PR #8, opened 2015-01-26
Tickets/dm 1218

jdswinbank:


src/table/slots.cc
```diff
@@ -10,31 +10,53 @@ bool startsWith(std::string const & a, std::string const b) {
     return a.compare(0, b.size(), b) == 0;
 }
 
+// helper class that resolves aliases for a slot's main measurement field, while
+// distiguishing between two cases where the field can't be found:
+//  - if the alias points to an invalid field, we should throw an exception.
+//  - if the alias simply isn't defined, we should just reset the slot with invalid keys
+class MeasFieldNameGetter {
```
jdswinbank: It might help readability to be more specific about exactly where the throwing, etc, is supposed to happen. Being unfamiliar with the code, I was confused to note that you seemed to be catching and handling exceptions below rather than raising them. Took me a moment to realise where the exception you're referring to here was coming from.

src/table/slots.cc
```diff
@@ -10,31 +10,53 @@ bool startsWith(std::string const & a, std::string const b) {
...
+public:
+
+    MeasFieldNameGetter(SubSchema const & s, Schema const & schema) :
+        replaced(schema[schema.getAliasMap()->apply(s.getPrefix())]),
+        defined(replaced.getPrefix() != s.getPrefix()) // slot is defined if applying alias wasn't a no-op
+    {}
+
+    SubSchema replaced; // a SubSchema that includes alias replacement
+    bool defined;     // whether the slot is defined at all
```
jdswinbank: Being picky, you the alignment of those comments looks weird.

src/table/slots.cc
```diff
@@ -10,31 +10,53 @@ bool startsWith(std::string const & a, std::string const b) {
...
-            } catch (pex::exceptions::NotFoundError &) {}
-            try {
-                _flagKey = s["flag"];
-            } catch (pex::exceptions::NotFoundError &) {}
+    _measKey = MeasKey();
+    _errKey = ErrKey();
+    _flagKey = Key<Flag>();
+    if (schema.getVersion() == 0) {
+        MeasFieldNameGetter helper(s, schema);
+        if (!helper.defined) return;
```
jdswinbank: Looks like this check should be outside the ``if (schema.getVersion()...`` block rather than duplicated.

tests/testSourceTable.py
```diff
@@ -407,6 +407,19 @@ def testDM1083(self):
...
+    def testSlotUndefine(self):
+        """Test that we can correctly define and undefine a slot after a SourceTable has been created"""
+        schema = lsst.afw.table.SourceTable.makeMinimalSchema()
+        schema.setVersion(1)
+        key = schema.addField("a_flux", type=float, doc="flux field")
+        table = lsst.afw.table.SourceTable.make(schema)
+        table.definePsfFlux("a")
+        self.assertEqual(table.getPsfFluxKey(), key)
+        table.schema.getAliasMap().erase("slot_PsfFlux")
+        print dict(table.schema.getAliasMap())
```
jdswinbank: I'm not a huge fan of print statements in unit tests -- they tend to disrupt whatever pretty test runner is in use. Maybe only print when in verbose mode, or use the logging framework?

src/table/slots.cc
TallJimbo: Will do.

src/table/slots.cc
TallJimbo: Fixed.

src/table/slots.cc
TallJimbo: It can't in this case, because it depends on the previous line, and that's not the same in both of the `if` clauses.

tests/testSourceTable.py
TallJimbo: Will remove it.  I never intended for it to remain there after I got the test working.

PR #9, opened 2015-01-30
U/krughoff/dm 1296

SimonKrughoff:


jchiang87:
Hi Simon,

I'm putting my code review comments here...not sure what, if anything, you want me to do at the JIRA issue.

### fitsUtils.py
These comments are based on commit 30eef22e
* lines 7, 14:
  * Apply convention for in-place object modification with that object as first argument (like a method).
  * Are these intended for use outside of module? if not, prepend "_" to function name (It seems these are intended for use outside of the module, since these tested are in unit tests, in which case these functions should have docstrings.)
* line 25: fix typos
* line 53: Add Exception.message to warning output.
* lines 117-122: variables are named "dmt*" whereas the header keywords are "DTM*" (the comment also uses "DMT").  Make these consistent.
* lines 125-126: Some comment on why these are the right values for FLIPX and FLIPY would be helpful here (I'm not sure that they are).
* lines 143-147: 'BIASSEC' is read from the FITS file hdu.  How are keyword arrays specified in FITS headers? Is an array-like BIASSEC an actual use case that will be encountered in practice?
* line 223: fix typos (incomplete comment?).

Ok to merge after fixing typos and adding comments where requested.  Other changes are up to you if you want to do them.

SimonKrughoff:
@jchiang87 thanks for doing this.  I fixed everything you asked for.  In response to the question about array BIASSEC values, this is part of the NOAO mosaic definition (as I understand it).  I think we will pick a saner default when we go back through and choose new defaults.

PR #11, opened 2015-02-17
Tickets/dm 2033

jchiang87:


PaulPrice:
I would rather use something based on the existing `lsst::afw::math::Function` instead of the new `lsst::afw::geom::Functor`, and I would suggest that adding a feature, exposing the feature and testing the feature are all part of the same functionality so they should be in the same commit.

TallJimbo:
@r-owen's recent message to dm-devel questioning the need for analytic spatially-varying kernels is relevant to the question of whether this should be based on `afw::math::Function`.

At this point, I'd lean towards just using the `Functor` that has been added here for now, because:
 - `afw::math::Function` has features we don't need here (get and set parameters), and we don't want to force users to implement those in order to use `SeparableXYTransform`.
 - Currently, `afw::geom` doesn't depend on `afw::math` at all, and that's a circular dependency I'd really like to avoid.  If it ever does turn out we can use the same functions in `afw::math::Kernel` that we use in `afw::geom::XYTransform`, then I think we either need to put those objects in `afw::geom` or consider splitting either `afw::geom` or `afw::math` or both into multiple packages.

include/lsst/afw/geom/EdgeRolloffFunctor.h
```diff
@@ -0,0 +1,67 @@
...
+#ifndef LSST_AFW_GEOM_EDGEROLLOFFFUNCTOR_H
+#define LSST_AFW_GEOM_EDGEROLLOFFFUNCTOR_H
+
+#include "lsst/afw/geom/Functor.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+/// @brief Stubbs parameterization of edge rolloff effect in LSST CCDs.
```
r-owen: Please expand this doc string, including a description of each parameter and some idea of the functional shape of the result. Also, my understanding is that the function is valid within a specific range of x; please make that range clear somehow, e.g. in the description of xmax or in the general comments. (I'm guessing the range is [0, xmax] but that is just a guess).

Also, what happens if x is outside the range? If this functor raises any exceptions then please list them here using @throw

include/lsst/afw/geom/Functor.h
```diff
@@ -0,0 +1,99 @@
...
+
+/// @brief Abstract base class for function objects.
+class Functor : public daf::base::Citizen {
+
+public:
+
+   /// @param owner The name of the class that owns a concrete
+   ///              instance of this class.  Used in the exception
+   ///              message from the inverse(...) member function.
+   explicit Functor(std::string const & owner);
```
PaulPrice: There's no need for the `Functor` to know the `owner` --- it can be deduced from the exceptions if you use the `LSST_EXCEPT_ADD` macro in the owner.

r-owen:
I can see why Jim Chiang chose to create a new functor base class for SeparableXYTransform: he wanted the derivative to compute the inverse transform. That said, I agree with Paul that it would be nice if we could at use existing afw::math::Function1D functions in SeparableXYTransform

Here is a proposed solution, which also addresses Jim Bosch's points:
- Move SeparableXYTransform and EdgeRolloffFunctor to afw::math:: namespace. I think that's a natural place for them. I hope someday XYTransform also ends up in afw::math.
- Get rid of the existing IntegerDelta functions. We don't use it and it complicates the rest of the proposal.
- Add a "derivative" method to Function1D and implement it for the remaining functions in FunctionLibrary.h, all of which have simple analytic derivatives.
- Consider adding a "derivative" method to Function2D and implementing it for existing functions, but it will be a bit more work and we don't have an immediate need.
- Change SeparableXYTransform to take a Function1D.
- Make EdgeRolloffFunctor into a Function1D (and rename it slightly). I agree that set/getParams is an annoyance, but it is easy to implement for this function.

One question is whether we want to continue to support subclasses of Function1D that cannot compute an analytic derivative. We could certainly compute a numeric estimate (offering a default implementation in the base class) but I worry about numeric solutions that have no parameters to control desired accuracy (and the existing unused delta function is a warning). Another option is offer two base classes for Function1D: one for functions with an analytic derivative and one for those without.

Even if people like this proposal, the question arises as to when it should be implemented and who will do the work. I suspect it will not be very hard, but I don't know how much time and patience Jim Chiang has for this stuff and I certainly don't want to discourage him from making future contributions by dramatically increasing the scope of this one. I am happy to do much of the work, but I'm not sure how it would fit into the schedule.

include/lsst/afw/geom/Functor.h
```diff
@@ -0,0 +1,99 @@
...
+#include <string>
+#include "boost/shared_ptr.hpp"
+#include "lsst/base.h"
+#include "lsst/daf/base.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+/// @brief Abstract base class for function objects.
```
r-owen: Please give more information here. For instance:
- The functor must be invertible
- Subclasses must override deriv but need not override inverse

include/lsst/afw/geom/Functor.h
r-owen: Why use _owner instead of _name? I'm not sure I understand the concept of "the class that owns a concrete instance of this subclass". Would it not suffice to have the name of the subclass?

include/lsst/afw/geom/SeparableXYTransform.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include "lsst/afw/geom/XYTransform.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+class Functor;
+
+class SeparableXYTransform : public XYTransform {
```
r-owen: Please add a document string describing the point of this class. It is fairly self-documenting, but still...

include/lsst/afw/geom/SeparableXYTransform.h
```diff
@@ -0,0 +1,86 @@
...
+   virtual Point2D forwardTransform(Point2D const & point) const;
+
+   /// @return The un-transformed Point2D in sensor coordinates.
+   /// @param point The Point2D location in sensor coordinates.  This
+   ///        corresponds to the actual location of charge deposition,
+   ///        i.e., with the pixel distortions applied.
+   virtual Point2D reverseTransform(Point2D const & point) const;
+
+   /// @return Const reference to the xfunctor.
+   Functor const & xfunctor() const;
```
r-owen: Unfortunately our standards mandate that getters be named with a leading "get", e.g. "getXFunctor" so I will reluctantly ask you to change the name of this and the next method.

include/lsst/afw/geom/EdgeRolloffFunctor.h
```diff
@@ -0,0 +1,67 @@
...
+   EdgeRolloffFunctor(std::string const & owner,
+                      double amplitude, double xscale, double xmax);
+
+   ~EdgeRolloffFunctor() {}
+
+   virtual PTR(Functor) clone() const;
+
+   virtual double operator()(double x) const;
+
+   virtual double deriv(double x) const;
```
r-owen: Should this be a public method (is it likely to be useful by others) or can it be protected (in which case it should have a leading underscore in the name).

In any case I suggest expanding the name to derivative, since our naming standard frowns on abbreviations.

include/lsst/afw/geom/Functor.h
```diff
@@ -0,0 +1,99 @@
...
+   virtual ~Functor() {}
+
+   virtual PTR(Functor) clone() const = 0;
+
+   /// @return y = f(x)
+   virtual double operator()(double x) const = 0;
+
+   /// @return df(x)/dx evaluated at x.  This is used in the
+   /// inverse(...) member function.
+   virtual double deriv(double x) const = 0;
```
r-owen: Should this be a public method (is it likely to be useful by others) or can it be protected (in which case it should have a leading underscore in the name).

In any case I suggest expanding the name to derivative, since our naming standard frowns on abbreviations.

include/lsst/afw/geom/Functor.h
```diff
@@ -0,0 +1,99 @@
...
+
+   /// @return The x value such that y = f(x).  Uses Newton-Raphson.
+   virtual double inverse(double y, double tol=1e-10,
+                          unsigned int maxiter=1000) const;
+protected:
+
+   std::string _owner;
```
r-owen: Can this be private? If so, I suggest making that change.

src/geom/Functor.cc
```diff
@@ -0,0 +1,56 @@
...
+namespace lsst {
+namespace afw {
+namespace geom {
+
+Functor::Functor(std::string const & owner) 
+   : daf::base::Citizen(typeid(this)), _owner(owner) {
+}
+
+double Functor::inverse(double y, double tol, unsigned int maxiter) const {
+   /// Use Newton-Raphson method to find the inverse.
```
r-owen: Should you clip x to the valid range for operator()? I worry that this may unexpected fail otherwise.

Also I suggest you sanity-check tol>0 and maxiter>0 before starting the iteration; that may catch errors of mis-ordering arguments.

src/geom/Functor.cc
```diff
@@ -0,0 +1,56 @@
...
+   double x(y);
```
r-owen: double x(y) looks like a function declaration. I think your code would be clearer if you used {} (c++ uniform initialization) to initialize variables, e.g. double x{y} here and similarly in other places.

tests/testSeparableXYTransform.py
```diff
@@ -0,0 +1,94 @@
...
+import lsst.utils.tests
+import lsst.afw.geom as afwGeom
+
+def num_deriv(func, x, eps=1e-7):
+    h = eps*max(abs(x), 1)
+    xp = x + h
+    dx = xp - x
+    return (func(x + dx) - func(x))/dx
+
+class FunctorTestCase(unittest.TestCase):
```
r-owen: Thanks for the unit tests! However, I suggest making this a separate unit test (e.g. testFunctor.py) so it's easier to find.

tests/testSeparableXYTransform.py
```diff
@@ -0,0 +1,94 @@
...
+    def setUp(self):
+        self.pars = [('', 5, 30), ('', 2, 30, 4000)]
+        self.funcs = [afwGeom.LinearFunctor,
+                      afwGeom.EdgeRolloffFunctor]
+        self.xvals = np.logspace(-8, 1, 10)
+    def tearDown(self):
+        pass
```
r-owen: Shouldn't tearDown delete the instance variables set in setUp (e.g. del self.pars, etc.)?

tests/testSeparableXYTransform.py
```diff
@@ -0,0 +1,94 @@
...
+        npts = 10
+        self.xvals = np.concatenate((random.uniform(0, dxy, size=npts),
+                                     random.uniform(self.xpars[-1] - dxy,
+                                                    self.xpars[-1], size=npts)))
+        self.yvals = np.concatenate((random.uniform(0, dxy, size=npts),
+                                     random.uniform(self.ypars[-1] - dxy,
+                                                    self.ypars[-1], size=npts)))
+        self.functorClass = afwGeom.EdgeRolloffFunctor
+    def tearDown(self):
+        pass
```
r-owen: Again, shouldn't tearDown delte the instance variables set in setUp?

tests/testSeparableXYTransform.py
RobertLuptonTheGood: There's no need to explicitly tear things down as they'll be deleted when the test goes out of scope.  The exception is things for which we're tracking memory, as testing for leaks occurs earlier.

tests/testSeparableXYTransform.py
PaulPrice: That's not necessary unless the objects in `setUp` are `Citizen`s, in which case it looks like there's a memory leak.

include/lsst/afw/geom/Functor.h
jchiang87: My motivation was to provide additional provenance of the error in the exception message and wasn't happy either with my solution.  Paul's suggestion of `LSST_EXCEPT_ADD` looks like what I wanted.  I'll also do the `_name` thing as you suggest (with _name becoming a private data member).

src/geom/Functor.cc
jchiang87: I didn't know about the {} syntax for uniform initialization.  Is this new in C++11?  I assume this is supported by all compilers we use....

r-owen: It is new in C++ and it is supported by gcc 4.4, which is the oldest compiler we support. (You can have an = before the {} or not; my recollection is that it makes no difference to the generated code).

If you are nervous about using {} then I suggest using = as being clearer than varname(value), since that often looks like a function declaration.

include/lsst/afw/geom/Functor.h
jchiang87: I'm going to keep derivative(...) public since it does need to be tested in the unit tests.

include/lsst/afw/geom/EdgeRolloffFunctor.h
jchiang87: Specifying a valid range for x is problematic.  It's true that for operator()(x), it wouldn't make sense to provide an x outside of [0, xmax], but when computing the inverse via Newton-Raphson, the initial guess is the provided y value, which often lies outside of [0, xmax], and the range of that value depends on amplitude, xmax, and xscale. Since inverse calls operator()(...) as well as derivative(...), it can potentially violate that range constraint.   One could restrict the initial estimate to lie within the [0, xmax] range, but then that would involve re-implementing inverse(...) and also wouldn't guarantee that the subsequent trial values would stay in range.  Unless there is a divide by zero issue or some other floating point problem, I don't think it is desirable to enforce a validity range on x.

src/geom/LinearFunctor.cc
```diff
@@ -0,0 +1,50 @@
...
+
+PTR(Functor) LinearFunctor::clone() const {
+   return boost::make_shared<LinearFunctor>(_slope, _intercept);
+}
+
+double LinearFunctor::operator()(double x) const {
+   return _slope*x + _intercept;
+}
+
+double LinearFunctor::derivative(double x) const {
```
ktlim: Sorry I'm kind of late to the party here, but I thought that LinearFunctor is supposed to set a good example, so I would have thought it should have an `inverse()` method since it's definitely analytically invertible.

jchiang87: You're right of course.  I added LinearFunctor to provide a concrete class to test the implementation of the Newton-Rafeson method in the default inverse().  I guess I should have picked something not analytically invertible.

PR #12, opened 2015-03-02
DM-2193: Add assertWcssAlmostEqual to testUtils

PaulPrice:


python/lsst/afw/image/basicUtils.py
```diff
@@ -0,0 +1,106 @@
...
+    @param[in] nx  number of points in x for the grid of pixel positions
+    @param[in] ny  number of points in y for the grid of pixel positions
+    @param[in] msg  exception message prefix; details of the error are appended after ": "
+
+    @throw AssertionError if the two WCSs do not match sufficiently closely
+    """
+    bboxd = afwGeom.Box2D(bbox)
+    xList = numpy.linspace(bboxd.getMinX(), bboxd.getMaxX(), nx)
+    yList = numpy.linspace(bboxd.getMinY(), bboxd.getMaxY(), ny)
+    measDiffSky = (afwGeom.Angle(0), -1, -1) # (sky diff, pix pos)
```
ktlim: You never refer to `measDiffSky[2]`, so I think you want either the tuple `(-1, -1)` or a `Point2D` as the second element of this tuple or perhaps just `None`, since it should not usually ever be shown (only if someone set `maxDiffSky` to a negative number).

python/lsst/afw/image/basicUtils.py
```diff
@@ -0,0 +1,106 @@
...
+    measDiffPix = (0, -1, -1) # (pix diff, sky pos)
+    for x in xList:
+        for y in yList:
+            fromPixPos = afwGeom.Point2D(x, y)
+            sky0 = wcs0.pixelToSky(fromPixPos)
+            sky1 = wcs1.pixelToSky(fromPixPos)
+            diffSky = sky0.angularSeparation(sky1)
+            if diffSky > measDiffSky[0]:
```
ktlim: It might be a little ugly, but couldn't you combine the assertion and the comparison function by adding cases to this and the next condition to exit early if `not raiseException`?

tests/testTestMethods.py
```diff
@@ -0,0 +1,176 @@
...
+                        box0, box2, maxDiff=radDiff*0.999999)
+
+                    box3 = afwGeom.Box2D(box0.getMin(), box0.getMax() + deltaExtent)
+                    self.assertBoxesNearlyEqual(box0, box3, maxDiff=radDiff*1.00001)
+                    self.assertRaises(AssertionError, self.assertBoxesNearlyEqual,
+                        box0, box3, maxDiff=radDiff*0.999999)
+
+    def testAssertCoordsNearlyEqual(self):
+        """Test assertCoordsNearlyEqual"""
+        for raDecDeg in ((45, 45), (-70, 89), (130, -89.5)):
```
ktlim: Does it make sense to test that (0.01, 20) is nearly equal to (359.99, 20)?

python/lsst/afw/image/basicUtils.py
```diff
@@ -0,0 +1,106 @@
...
+    return cls(array)
+
+def makeMaskedImageFromArrays(image, mask=None, variance=None):
+    """Construct a MaskedImage from three NumPy arrays, inferring the MaskedImage types from the NumPy types.
+    """
+    cls = globals()["imageLib.MaskedImage%s" % suffixes[str(image.dtype.type)]]
+    return cls(makeImageFromArray(image), makeMaskFromArray(mask), makeImageFromArray(variance))
+
+@lsst.utils.tests.inTestCase
+def assertWcssNearlyEqualOverBBox(testCase, wcs0, wcs1, bbox, maxDiffSky=0.01*afwGeom.arcseconds, maxDiffPix=0.01,
```
ktlim: The doubled "s" bothers me, but I can live with it if you really want to try to indicate the plural.

tests/testTestMethods.py
timj: Yes. That's a very common mistake.

python/lsst/afw/image/basicUtils.py
timj: How about `assertWcsAreNearlyEqualOverBBox`?

python/lsst/afw/coord/utils.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import absolute_import, division
+# 
+# LSST Data Management System
+# Copyright 2008, 2009, 2010 LSST Corporation.
```
timj: This year can't be right.

python/lsst/afw/image/basicUtils.py
r-owen: Good catch. Thanks. I had originally been saving x,y and RA/Dec separately and didn't fix it everywhere. I switched to "?" for initial value, because the value is only ever printed anyway.

I also added a test for nx, ny > 0

python/lsst/afw/image/basicUtils.py
r-owen: Yes, and I did write that code but I felt it was ugly and decided not to bother. I've put my reasoning in RFC-28. If we do go with functions that measure difference and can short-circuit then that code would certainly go right here.

tests/testTestMethods.py
r-owen: I don't understand your question. assertAnglesNearlyEqual wraps the difference, by design and as documented. If you don't want to wrap the difference then you'd just use assertLess(abs(ang1-ang0), maxAng).

python/lsst/afw/image/basicUtils.py
r-owen: The double s bothers me, too. I wanted predictability, but I guess it'll be predictable enough without the double s. I don't think WcsAre helps, since it would be AnglesAre, not AngleAre.

python/lsst/afw/coord/utils.py
r-owen: OK. I'll clean up the new files. Grr. I hate this boilerplate.

tests/testTestMethods.py
timj: I thought he just wanted an extra wraparound test.

python/lsst/afw/image/basicUtils.py
timj: What makes you think WCS is singular?

python/lsst/afw/coord/utils.py
timj: RFC-45 needs a kick.

tests/testTestMethods.py
r-owen: Ah. That makes sense. I'll add that.

PR #14, opened 2015-03-11
Review for DM-1576

TallJimbo:


include/lsst/afw/image/Calib.h
```diff
@@ -51,6 +52,29 @@ namespace cameraGeom {
     class Detector;
 }
 namespace image {
+
+static double const JanskysPerABFlux = 3631.0;
```
TallJimbo: I believe this should be `JANSKYS_PER_AB_FLUX`.

include/lsst/afw/image/Calib.h
```diff
@@ -51,6 +52,29 @@ namespace cameraGeom {
...
+
+/// Compute AB magnitude from flux in Janskys
+inline double abMagFromFlux(double flux) {
+    return -2.5*std::log10(flux/JanskysPerABFlux);
+}
+
+/// Compute AB magnitude error from flux and flux error in Janskys
+inline double abMagErrFromFluxErr(double fluxErr, double flux) {
+    return std::abs(fluxErr/(-0.4*flux*std::log(10)));
+}
```
TallJimbo: I'd prefer these to follow the same call signatures as the methods in `Calib`, that is, have a method that returns a `std::pair` of `flux` and `fluxErr` instead of one that returns just `fluxErr` (and the same for the inverse operations).

TallJimbo: I don't personally care about this, but I still feel obligated to point out that our name standards would likely prefer these to start with a verb (probably `compute`, unless we want to go with `get` for consistency with `Calib`, except that `Calib`'s should probably be `compute` as well, according to the standards)...

Also, to be especially troublesome, I'll point out that our standards also *require* the name of the methods below to use "Ab" instead of "AB", which I find just horrible (but then, I'm not a big fan of Psf, Wcs, or Lsst either).

PaulPrice: On the subject of names, "flux" doesn't necessarily imply "Janskys".

TallJimbo: Of course, "flux" doesn't imply "DN" either, and we've been quite happy to ignore that elsewhere in our code so far (well, perhaps because it *could* imply DN until now).

This is something we've been struggling with for the past week or so.  I'm not bothered by this as it is mostly because I don't expect it to stay that way forever.  But I also don't have a good long-term solution (I'd rather not take the step of embedding units as types in *all* of our interfaces).

PR #15, opened 2015-03-18
DM-2343: Move afw_extensions_rgb functionality into afw proper

PaulPrice:
https://jira.lsstcorp.org/browse/DM-2343

### DIFF
--- a/python/lsst/afw/display/saturated.cc
+++ b/python/lsst/afw/display/saturated.cc
@@ -1,0 +1,175 @@
+/**
+ * \file
+ *
+ * Handle saturated pixels when making colour images
+ */
+#include "boost/format.hpp"
+#include "lsst/utils/ieee.h"
+#include "lsst/afw/detection.h"
+#include "lsst/afw/image/MaskedImage.h"
+#include "Rgb.h"
+
+namespace lsst { namespace afw { namespace display {
+
+namespace {
+    template <typename ImageT>
+    class SetPixels : public detection::FootprintFunctor<ImageT> {
+    public:
+        explicit SetPixels(ImageT const& img     // The image the source lives in
+                          ) : detection::FootprintFunctor<ImageT>(img), _value(0) {}
+
+        void setValue(float value) { _value = value; }
+
+        // method called for each pixel by apply()
+        void operator()(typename ImageT::xy_locator loc,        // locator pointing at the pixel
+                        int,                                    // column-position of pixel
+                        int                                     // row-position of pixel
+                       ) {
+            *loc = _value;
+        }
+    private:
+        float _value;
+    };
+}
+
+template<typename ImageT>
+void
+replaceSaturatedPixels(ImageT & rim,    // R image (e.g. i)
+                       ImageT & gim,    // G image (e.g. r)
+                       ImageT & bim,    // B image (e.g. g)
+                       int borderWidth,	// width of border used to estimate colour of saturated regions
+                       float saturatedPixelValue // the brightness of a saturated pixel, once fixed
+                      )
+{
+    int const width = rim.getWidth(), height = rim.getHeight();
+    int const x0 = rim.getX0(), y0 = rim.getY0();
+
+    if (width != gim.getWidth() || height != gim.getHeight() || x0 != gim.getX0() || y0 != gim.getY0()) {
+        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError,
+                          str(boost::format("R image has different size/origin from G image "
+                                            "(%dx%d+%d+%d v. %dx%d+%d+%d") %
+                              width % height % x0 % y0 %
+                              gim.getWidth() % gim.getHeight() % gim.getX0() % gim.getY0()));
+
+    }
+    if (width != bim.getWidth() || height != bim.getHeight() || x0 != bim.getX0() || y0 != bim.getY0()) {
+        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError,
+                          str(boost::format("R image has different size/origin from B image "
+                                            "(%dx%d+%d+%d v. %dx%d+%d+%d") %
+                              width % height % x0 % y0 %
+                              bim.getWidth() % bim.getHeight() % bim.getX0() % bim.getY0()));
+
+    }
+
+    bool const useMaxPixel = !utils::isfinite(saturatedPixelValue);
+
+    SetPixels<typename ImageT::Image>
+        setR(*rim.getImage()),
+        setG(*gim.getImage()),
+        setB(*bim.getImage()); // functors used to set pixel values
+
+    // Find all the saturated pixels in any of the three image
+    int const npixMin = 1;              // minimum number of pixels in an object
+    afw::image::MaskPixel const SAT = rim.getMask()->getPlaneBitMask("SAT");
+    detection::Threshold const satThresh(SAT, detection::Threshold::BITMASK);
+
+    detection::FootprintSet       sat(*rim.getMask(), satThresh, npixMin);
+    sat.merge(detection::FootprintSet(*gim.getMask(), satThresh, npixMin));
+    sat.merge(detection::FootprintSet(*bim.getMask(), satThresh, npixMin));
+    // go through the list of saturated regions, determining the mean colour of the surrounding pixels
+    typedef detection::FootprintSet::FootprintList FootprintList;
+    PTR(FootprintList) feet = sat.getFootprints();
+    for (FootprintList::const_iterator ptr = feet->begin(), end = feet->end(); ptr != end; ++ptr) {
+        PTR(detection::Footprint) const foot = *ptr;
+        PTR(detection::Footprint) const bigFoot = growFootprint(*foot, borderWidth);
+
+        double sumR = 0, sumG = 0, sumB = 0; // sum of all non-saturated adjoining pixels
+        double maxR = 0, maxG = 0, maxB = 0; // maximum of non-saturated adjoining pixels
+
+        for (detection::Footprint::SpanList::const_iterator sptr = bigFoot->getSpans().begin(),
+                 send = bigFoot->getSpans().end(); sptr != send; ++sptr) {
+            PTR(detection::Span) const span = *sptr;
+
+            int const y = span->getY() - y0;
+            if (y < 0 || y >= height) {
+                continue;
+            }
+            int sx0 = span->getX0() - x0;
+            if (sx0 < 0) {
+                sx0 = 0;
+            }
+            int sx1 = span->getX1() - x0;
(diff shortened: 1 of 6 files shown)


python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+#include "lsst/afw/image/MaskedImage.h"
+#include "Rgb.h"
+
+namespace lsst { namespace afw { namespace display {
+
+namespace {
+    template <typename ImageT>
+    class SetPixels : public detection::FootprintFunctor<ImageT> {
+    public:
+        SetPixels(ImageT const& img     // The image the source lives in
```
PaulPrice: `explicit`?

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+
+        // method called for each pixel by apply()
+        void operator()(typename ImageT::xy_locator loc,        // locator pointing at the pixel
+                        int,                                    // column-position of pixel
+                        int                                     // row-position of pixel
+                       ) {
+            *loc = _value;
+        }
+    private:
+        float _value;
```
PaulPrice: You've got `float` here, and below you only instantiate for floating-point images, so why not drop the pretense (`template`) of supporting other types?

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+                       int borderWidth,	// width of border used to estimate colour of saturated regions
+                       float saturatedPixelValue // the brightness of a saturated pixel, once fixed
+                      )
+{
+    int const width = rim.getWidth(), height = rim.getHeight();
+    int const x0 = rim.getX0(), y0 = rim.getY0();
+
+    if (width != gim.getWidth() || height != gim.getHeight() || x0 != gim.getX0() || y0 != gim.getY0()) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::InvalidParameterError,
+                          str(boost::format("R image has different size/origin from G image "
```
PaulPrice: Can drop the leading `lsst::`.

Here and throughout.

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+                                            "(%dx%d+%d+%d v. %dx%d+%d+%d")
+                              % width % height % x0 % y0
```
PaulPrice: Break after an operator.

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+    bool const useMaxPixel = !lsst::utils::isfinite(saturatedPixelValue);
+
+    SetPixels<typename ImageT::Image>
+        setR(*rim.getImage()),
+        setG(*gim.getImage()),
+        setB(*bim.getImage()); // functors used to set pixel values
+
+    // Find all the saturated pixels in any of the three image
+    int const npixMin = 1;              // minimum number of pixels in an object
+    int const SAT = rim.getMask()->getPlaneBitMask("SAT");
```
PaulPrice: `int` should be `afw::image::MaskPixel` --- we're changing the type soon….

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+
+    detection::FootprintSet       sat(*rim.getMask(), satThresh, npixMin);
+    sat.merge(detection::FootprintSet(*gim.getMask(), satThresh, npixMin));
+    sat.merge(detection::FootprintSet(*bim.getMask(), satThresh, npixMin));
+    // go through the list of saturated regions, determining the mean colour of the surrounding pixels
+    typedef detection::FootprintSet::FootprintList FootprintList;
+    PTR(FootprintList) feet = sat.getFootprints();
+    for (FootprintList::iterator ptr = feet->begin(), end = feet->end(); ptr != end; ++ptr) {
+        PTR(detection::Footprint) foot = *ptr;
+        PTR(detection::Footprint) bigFoot = growFootprint(*foot, borderWidth);
```
PaulPrice: Shouldn't all these (`ptr`, `foot`, `bigFoot`) be `const`?

PaulPrice: I'm surprised `growFootprint` doesn't have a namespace (`detection::`) prepended, yet it compiles.  What am I missing?

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+
+        float sumR = 0, sumG = 0, sumB = 0; // sum of all non-saturated adjoining pixels
+        float maxR = 0, maxG = 0, maxB = 0; // maximum of non-saturated adjoining pixels
```
PaulPrice: Should accumulate with `double`?

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+
+        for (detection::Footprint::SpanList::const_iterator sptr = bigFoot->getSpans().begin(),
+                 send = bigFoot->getSpans().end(); sptr != send; ++sptr) {
```
PaulPrice: What's the advantage to defining `send`?
You might use `auto` and `cbegin` and `cend` here?

PaulPrice: Might this block be more clear and concise if you used a `FootprintFunctor`?

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+                    if (val > maxB) {
+                        maxB = val;
+                    }
+                }
+            }
+        }
+        // OK, we have the mean fluxes for the pixels surrounding this set of saturated pixels
+        // so we can figure out the proper values to use for the saturated ones
+        float R = 0, G = 0, B = 0;      // mean intensities
+        if(sumR + sumB + sumG > 0) {
```
PaulPrice: Space after reserved words.

Here and below.

python/lsst/afw/display/saturated.cc
```diff
@@ -0,0 +1,176 @@
...
+            float scale = 1.0;          // how much to scale up max values
```
PaulPrice: What's the purpose of this?  It should be `const`, if not removed.

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env python
```
PaulPrice: pyflakes reports:
    tests/rgb.py:12: 'sys' imported but unused
    tests/rgb.py:17: 'pexExceptions' imported but unused

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
...
+
+#-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+#
+# A context object to clean up temporary files
+#
+from contextlib import contextmanager
+
+@contextmanager
+def Tempfile(fileName, remove=True):
+    pass
```
PaulPrice: Why the `pass`?

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
...
+
+R, G, B = 2, 1, 0
+
+class RgbTestCase(unittest.TestCase):
+    """A test case for Rgb"""
+    def setUp(self):
+        if True:
+            self.min, self.range, self.Q = 0, 5, 20 # asinh
+        else:
+            self.min, self.range, self.Q = 0, 5, 0 # linear
```
PaulPrice: Get rid of the latter option?  Or convert the test to test both?

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
...
+                elif i == R:
+                    amp = A*math.pow(10, 0.4*r_i)
+
+                self.images[i].set(x, y, amp)
+
+        psf = afwMath.AnalyticKernel(15, 15, afwMath.GaussianFunction2D(2.5, 1.5, 0.5))
+
+        convolvedImage = type(self.images[0])(self.images[0].getDimensions())
+        randomImage = type(self.images[0])(self.images[0].getDimensions())
+        rand = afwMath.Random()
```
PaulPrice: Need a seed.

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
...
+        asinhMap = rgb.AsinhMapping(self.min, self.range, self.Q)
+        rgbImage = asinhMap.makeRgbImage(self.images[R], self.images[G], self.images[B])
+
+        for ext in ("jpeg", "jpg", "png", "tif", "tiff"):
+            fileName = "rgb.%s" % ext
+            with Tempfile(fileName, remove=True):
+                rgb.writeRGB(fileName, rgbImage)
+
+                if False:               # you'll also want to set remove=False in Tempfile manager
+                    os.system("open %s > /dev/null 2>&1" % fileName)
```
PaulPrice: Nice touch.

tests/rgb.py
```diff
@@ -0,0 +1,199 @@
...
+            feet[f] = ds.getFootprints()
+
+            arr = self.images[f].getImage().getArray()
+            arr[np.where(arr >= 1000)] = np.nan
+
+        rgb.replaceSaturatedPixels(self.images[R], self.images[G], self.images[B], 1, 2000)
+        #
+        # Check that we replaced those NaNs with some reasonable value
+        #
+        f0 = [k for k, v in feet.items() if v][0] # find a filter with a saturated region
```
PaulPrice: Shouldn't you check all the saturated footprints, instead of just one?

python/lsst/afw/display/saturated.cc
RobertLuptonTheGood: You're missing ADL (also known as Koenig Lookup).  growFootprint looks in detection:: because that's where *foot is found

python/lsst/afw/display/saturated.cc
RobertLuptonTheGood: An experiment to make the cores of brighter, but it's not configurable and I've removed it.

tests/rgb.py
RobertLuptonTheGood: Because the pattern is  "initial; yield; final".  I'll add a comment

python/lsst/afw/display/saturated.cc
RobertLuptonTheGood: I don't see a reason to remove the template.  We might want to support int types in the future.

python/lsst/afw/display/saturated.cc
ktlim: Potentially interesting point: ADL is also why str(boost::format()) works.

PR #16, opened 2015-03-25
Review for DM-1680: Change Background to use the xy0 of the image bbox

jdswinbank:
Background discarded the xy0 of the image when it was constructed from an image
and calling getImage(bbox) required that bbox be local, as well. This fixes that problem.

tests/background.py includes two new tests for these changes, plus some minor tweaks
to make the pyflakes linter happier.

Improve a change to a test to make the linter happier

### DIFF
--- a/tests/background.py
+++ b/tests/background.py
@@ -35,13 +35,13 @@ or
 
 import math
 import os
-import sys
 import unittest
 import numpy as np
 import pickle
 
 import lsst.utils.tests as utilsTests
 import lsst.pex.exceptions
+from lsst.daf.base import PropertySet
 import lsst.afw.image.imageLib as afwImage
 import lsst.afw.math as afwMath
 import lsst.afw.geom as afwGeom
@@ -53,6 +53,7 @@ try:
 except NameError:
     display = False
 
+AfwdataDir = eups.productDir("afwdata")
 
 # ==== summary to currently implemented tests ====
 # getPixel: tests basic functionality of getPixel() method (floats)
@@ -67,7 +68,7 @@ class BackgroundTestCase(unittest.TestCase):
     """A test case for Background"""
     def setUp(self):
         self.val = 10
-        self.image = afwImage.ImageF(afwGeom.Extent2I(100, 200))
+        self.image = afwImage.ImageF(afwGeom.Box2I(afwGeom.Point2I(1000, 500), afwGeom.Extent2I(100, 200)))
         self.image.set(self.val)
 
     def tearDown(self):
@@ -141,6 +142,7 @@ class BackgroundTestCase(unittest.TestCase):
         self.assertEqual(afwMath.cast_BackgroundMI(back).getPixel(xcen, ycen), self.val)
 
 
+    @unittest.skipIf(AfwdataDir is None, "afwdata not setup")
     def testBackgroundTestImages(self):
 
         imginfolist = []
@@ -158,16 +160,11 @@ class BackgroundTestCase(unittest.TestCase):
         #imgfiles.append("v2_i2_p_m9_f.fits")
         #imgfiles.append("v2_i2_p_m9_u16.fits")
         
-        afwdataDir = eups.productDir("afwdata")
-        if not afwdataDir:
-            print >> sys.stderr, "Skipping testBackgroundTestImages as afwdata is not setup"
-            return
-        
         for imginfo in imginfolist:
 
             imgfile, centerValue = imginfo
 
-            imgPath = afwdataDir + "/Statistics/" + imgfile
+            imgPath = os.path.join(AfwdataDir, "Statistics", imgfile)
 
             # get the image and header
             dimg = afwImage.DecoratedImageD(imgPath)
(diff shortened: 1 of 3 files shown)


tests/background.py
```diff
@@ -292,10 +292,10 @@ def testTicket1781(self):
         bctrl.getStatisticsControl().setNumSigmaClip(10.0)  
         bctrl.getStatisticsControl().setNumIter(1)
         backobj = afwMath.makeBackground(parabimg, bctrl)
-
-        #parabimg.writeFits('in.fits')
-        #backobj.getImageF().writeFits('out.fits')
-
+
```
jdswinbank: A trivial point, but: you don't need this white space.

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+        The statsImage and background image should not vary with xy0
+        """
+        bgImageList = [] # list of background images, one per xy0
+        statsImageList = [] # list of stats images, one per xy0
+        for xy0 in (afwGeom.Point2I(0, 0), afwGeom.Point2I(1000, 500)):
+            afwdataDir = eups.productDir("afwdata")
+            if not afwdataDir:
+                print >> sys.stderr, "Skipping testCFHT as afwdata is not setup"
+                return
```
jdswinbank: Shouldn't we use `unittest.skipIf` or similar here?

I notice that this pattern is reproduced through this file, both in the other test you've introduced (`testSubImage`) and elsewhere. It would be far nicer to just run the check for `afwdata` once and then skip the tests that require it, rather than repeating it in every test.

Also, this is not `testCFHT`.

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+
+            mi = afwImage.MaskedImageF(os.path.join(afwdataDir,
```
jdswinbank: This and the next few lines are near-duplicates with `testSubImage`. If there was much more duplication, I'd strongly urge you to consider refactoring them into a function which both these tests can call. Since it's only a few lines, I will just suggest that as something you might like to consider.

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+                                                    "CFHT", "D4", "cal-53535-i-797722_1.fits"))
+            bbox = afwGeom.Box2I(afwGeom.Point2I(500, 2000), afwGeom.Point2I(2079, 4609))
+            mi = mi.Factory(mi, bbox, afwImage.LOCAL)
```
jdswinbank: What is happening here? I'm not really familiar with this part of the codebase, so apologies if I'm missing the obvious, but: your test says it is testing that things do not vary with `xy0`. That test is (presumably) accomplished by the loop and repeated calls to `setXY0()`. What is this additional application of a bounding box achieving?

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+            mi.setXY0(xy0)
+            self.assertEqual(mi.getBBox(), afwGeom.Box2I(xy0, bbox.getDimensions()))
+
+            bctrl = afwMath.BackgroundControl(mi.getWidth()//128, mi.getHeight()//128)
+            bctrl.getStatisticsControl().setNumSigmaClip(3.0)  
+            bctrl.getStatisticsControl().setNumIter(2)
```
jdswinbank: "Magic numbers" that appear in the code make me nervous -- are they just there because you need to provide some sort of argument to make things run, or do they have some significance? In this case, I think it's fairly clear that the answer is the former, but it would still be nice to mention that and/or explain why they were chosen.

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+            bgImage = backobj.getImageF()
+            self.assertEqual(bgImage.getBBox(), mi.getBBox())
+            bgImageList.append(bgImage)
+
+            statsImage = afwMath.cast_BackgroundMI(backobj).getStatsImage()
+            statsImageList.append(statsImage)
+
+        # changing the bounding box should make no difference to the pixel values,
+        # so compare pixels using exact equality
+        for bgImage in bgImageList[1:]:
```
jdswinbank: There are only two entries in `bgImageList`. Do we need a list at all? It seems like you could simplify the structure of this test by removing the looping and the list, and just comparing the background/stats images generated from your two `xy0` values directly.

tests/background.py
```diff
@@ -361,7 +361,76 @@ def testCFHT_oldAPI(self):
...
+            self.assertTrue(np.all(bgImage.getArray() == bgImageList[0].getArray()))
+        for statsImage in statsImageList[1:]:
+            for i in range(3):
+                self.assertTrue(np.all(statsImage.getArrays()[i] == statsImageList[0].getArrays()[i]))
+
+    def testSubImage(self):
```
jdswinbank: Pretty much the same comments -- except for that about looping -- apply here as to the `testXY0()` method above.

jdswinbank:
The last line of the commit message seems a little stilted and rather a non sequitur with what has gone before -- "Improve a change to a test"? Which change? To which test? I'm guessing this got squashed in from another commit and could maybe use some clarifying.

tests/background.py
jdswinbank: To clarify, since it might not be obvious in the diff -- you do want a blank line here, but you currently have a line containing 8 space characters. It's them that I suggest you remove, not the line itself.

tests/background.py
r-owen: Good idea. I didn't know about skipIf. Doing the test inside the loop was a mistake in any case.

tests/background.py
r-owen: If you are asking about line 380 "mi = mi.Factory(mi, bbox, afwImage.LOCAL)" all this is doing is extracting a smallish piece of the overall image. It's not the best way to do the job; I'll change the code to specify the bbox as part of the constructor. The non-zero XY0 is desirable for making sure Background honors xy0, but setting the xy0 directly does that, so it's slightly redundant in this case.

tests/background.py
r-owen: I thought about that when I wrote it. It allows me to specify additional xy0 if desired and I didn't feel it made the code that much harder to read. Once you have more than one item a list is easier than keeping two sets of variable names.

tests/background.py
r-owen: The sigmaclip and numiter aren't needed for these tests so I'll change to use the defaults (it was just copy/paste from the CFHT test). The details of the points and extents aren't interesting, but probably don't need explanation.

PR #17, opened 2015-03-26
Design Review for Footprints Overhaul (DM-1126)

TallJimbo:
This PR isn't really a PR; it's a forum for a design review of the new Footprint interface.  The commits on this branch are intended to demonstrate how the Footprint library will appear after the overhaul is complete (not just the Footprint class, but several of its friends, including a major new one SpanRegion).  They don't contain working code, and some important details - private members, Citizen inheritance, persistence framework support - have been left out.  The documentation comments are also aimed more at a design reviewer than a future user.

While I've only included C++ files here, I've added remarks about the Python interface in the comments where expect it to differ (there are some significant differences), and I've also used `#ifndef SWIG` blocks to indicate methods I don't expect to expose to Python at all.

One overall feature is that while many of Footprint's operations have been shifted to other classes, others have been removed entirely, as I don't think they're actually needed (in many cases, we had multiple functions or methods that did very similar things, and only really needed one of them).  If it turns out we do need some of them, I'm happy to add them back in, but I'd like to see how far we get without them.

There are two major components of the new Footprint interface that aren't included here, and will be a part of a future design review, as I think they're largely orthogonal:
- The handling of Peaks: this is covered partially by RFC-30, but I plan to do some more work on this in the future.
- We'll need a new interface to handle merging FootprintSets from different detection images while recording the provenance in the peaks.  This is new functionality that we'll be bringing over from the HSC side shortly, but the interface will need to be cleaned up in the process.

I recommend looking at the new interface in the same order as the commits below - start with `SpanRegion`, then move on to `Footprint`, `FootprintSet`, and `HeavyFootprint`.  In some cases, it may be easiest to look at the new files on their own, rather than via the diff here; while I've tried to keep operations mostly in the same order, that hasn't always been possible, and the diff frequently connects the wrong lines anyway.


jdswinbank:
It would be a helpful if you could expand a little on the overall motivations behind the redesign. For example:

* What are you main aims here -- A simpler, more convenient API? More straightforward & reliable implementation? Performance? Enabling new functionality? All of the above?

* In your response to Paul above, I notice you're already considering the "next RFC on this general topic". Could you comment a bit on your larger scale plans here -- how does this redesign fit into the likely longer-term evolution of the codebase?

TallJimbo:
In terms of motivation, it's essentially "all of the above", but here are a few of the things that I particularly wanted to change:
 - In representing regions of pixels to include in measurement algorithms, I had a need for a lighter-weight `Footprint` class that didn't care about Peaks and had better support topological operations (hence refactoring `SpanRegion`'s functionality into a new class).
 - From the efficiency standpoint, the old classes had a few design choices that were mistakes, at least in retrospect - some of these may have made sense originally, either for expediency or because we didn't know what we'd use:
   * We used a container of `shared_ptr`s to `Spans`, rather than a direct container of `Spans`, even though the size of a `Span` is probably less (certainly no more) than the total size of a `shared_ptr`, and the latter always involves more memory allocations.
   * The `FootprintFunctor` class, which is used to implement many of our measurement algorithms, involves a virtual function call on every pixel, making code within it impossible to inline.
   * The limited topological operations (and, until recently, the mathematical morphology operations) we had were implemented very inefficiently, by converting to a mask and back.  These operations frequently dominated the performance of galaxy modeling algorithms, where they are only being used to determine the region of pixels to include in the fit.
   * The `HeavyFootprint` was carrying around mask and variance pixel values that were, to my knowledge, never used.
 - From an ease-of-use standpoint:
    * The old `Footprint` had many free functions that should have been methods, and many free functions and methods that did sufficiently similar things that they could be consolidated into a single more general interface.
    * We frequently have multiple objects sharing the same `Footprint` via `shared_ptr`, with some objects requiring that the `Footprint` remain unmodified during their lifetime, but no way of guaranteeing that other objects will respect that requirement (this is even more true of `Wcs`, which I hope to make immutable in the future, and was true of `Psf` before I made it immutable).
    * The low-level detection algorithm code was hidden in a multitude of `FootprintSet` constructors.  While running detection does generate a `FootprintSet`, one doesn't usually expect to find heavy-duty algorithmic code in a constructor, and having it there made it much harder to see what was going on when reading code that used it.
    * Now that we can use C++11 lambdas, we can make some of the most common operations - applying a functor to a pixel region - much more concise and readable.

In terms of timing, I felt it was important to overhaul these classes before putting significant effort into deblender improvements.

TallJimbo:
As for future plans for Footprint design changes, all I think remains are:
 - Making the use of `PeakRecord` in `Footprint` more natural, and ensuring that it maintains `Footprint`'s immutability.  This will probably just involve a new container class to augment or replace `Catalog` in this context, while still using the rest of `afw::table` under the hood (in fact, `Catalog` was originally intended to be only one of several types of containers for `afw::table` `Records`, but instead of having different containers for different purposes we've instead just kept adding more features to `Catalog`).
 - We need an interface for merging `FootprintSets`.  We have code to do this on the HSC side, but it was put together in a hurry, and I'd like to think through the design more carefully either just before or just after we pull it over.
 - Paul's request to change how `Threshold` works.
I do think it's appropriate to include all of those in the current Footprint overhaul, but I don't think they have to be done at the same time, so if we run out of time, they'd be good candidates for rescheduling.

I do also have long-term plans for similar overhauls to other afw components, including `afw::image`, `Wcs`/`Coord`, `Psf`, and of course `afw::table`, and the ideas I have for those (and the reasons I want to overhaul them) are similar to the ideas and motivation here - but those ideas are much less well-developed.

jdswinbank:
This is long obsolete; closing the PR.

PR #20, opened 2015-04-09
Review of afw DM-2307: Extend QuadrupoleKey to celestial coords.

TallJimbo:


include/lsst/afw/table/misc.h
```diff
@@ -20,6 +20,11 @@ namespace lsst { namespace afw { namespace table {
  */
 typedef boost::int64_t RecordId;
 
+enum CoordinateType {
+    PIXEL,
+    CELESTIAL
+};
+
```
TallJimbo: Could this be an `enum class` instead?  It looks like gcc 4.4 supports those, though I'm not sure if Swig does.  Looks like it's worth just trying it; your unit tests should be sufficient to see if it works.

TallJimbo:
I think this approach of modifying `QuadrupoleKey` to support celestial coordinates is the right idea in general (and sorry I didn't see your request for comments on this on HipChat earlier).  But I think that even in celestial coordinates we want to just continue calling the fields "xx, yy, xy", and instead just change the units and note that x=ra, y=dec in the documentation in `addFields`.  That may be inconsistent with how we've switched to "ra, dec" for centroids, but the reality is that no one uses "RaRa" or "DecDec", and I think most users will be quite comfortable just using xx, yy, xy here.  I think that should cut down slightly on the changes you need to make here - the constructor from `SubSchema` will no longer need an extra argument.

PR #21, opened 2015-04-13
Review comments for DM-1197 (Support some mixed-type operations for Point and Extent)

jdswinbank:


include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+/**
+ *  Return the component-wise floor (round towards more negative).
+ *
+ *  In Python, this is available as both a free function and a method on ExtentD.
+ */
+template <int N>
+Extent<int,N> floor(Extent<double,N> const & input);
+
+/**
+ *  Return the component-wise floor (round towards more positive).
```
jdswinbank: "ceil", not "floor".

include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+}
+
+template <int N>
+Extent<double,N> operator*(ExtentBase<int,N> const & lhs, double rhs) {
+    return Extent<double,N>(static_cast<Extent<int,N> const &>(lhs)) * rhs;
+}
+
+template <int N>
+void operator*=(ExtentBase<int,N> & lhs, double rhs) {
+    static_assert(N < 0, "In-place multiplication of Extent<int,N> by double will truncate.");
```
jdswinbank: I'm not sure I understand what's going on here. What's the significance check for `N <  0` -- isn't that always false? According to RFC-41 (and, indeed, the message above) this ought to truncate, but it seems like it will always trip this assertion.

include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+}
+
+template <int N>
+Extent<double,N> operator/(ExtentBase<int,N> const & lhs, double rhs) {
+    return Extent<double,N>(static_cast<Extent<int,N> const &>(lhs)) * rhs;
+}
+
+template <int N>
+void operator/=(ExtentBase<int,N> & lhs, double rhs) {
+    static_assert(N < 0, "In-place division of Extent<int,N> by double will truncate.");
```
jdswinbank: Same comment as for `operator*=` above.

include/lsst/afw/geom/Extent.h
jdswinbank: Ah, reading further into RFC-41, I realise that the behaviour might be designed to match some of the comments on the RFC, rather than the table of expected results. Is that correct?

I'm still not sure what the significance of `N < 0` is though.

include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+// functions on ExtentBase).
+
+template <typename T, int N>
+Extent<T,N> operator*(T scalar, ExtentBase<T,N> const & rhs) {
+    return rhs * scalar;
+}
+
+template <int N>
+Extent<double,N> operator*(ExtentBase<int,N> const & lhs, double rhs) {
+    return Extent<double,N>(static_cast<Extent<int,N> const &>(lhs)) * rhs;
```
jdswinbank: According to RFC-41, in C++ `ExtentI * double` ought to produce a "truncated `ExtentI`". In fact, I think this will produce an `ExtentD`.

jdswinbank: (...or perhaps not -- see my comment below re the table in RFC-41 vs the commentary.)

include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+}
+
+template <int N>
+void operator*=(ExtentBase<int,N> & lhs, double rhs) {
+    static_assert(N < 0, "In-place multiplication of Extent<int,N> by double will truncate.");
+}
+
+template <int N>
+Extent<double,N> operator/(ExtentBase<int,N> const & lhs, double rhs) {
+    return Extent<double,N>(static_cast<Extent<int,N> const &>(lhs)) * rhs;
```
jdswinbank: That looks like a multiplication, when it should be a division.

On experimentation, I note that `/` does multiply in C++, but that division works as expected in Python. I assume there's some magic happening in Swig that makes that happen (I've not attempted to figure that out yet). However, that makes me worry a bit about the test suite: so far as I can see, the tests for these operations are pure Python. Does that mean the C++ code is never being exercised?

include/lsst/afw/geom/Extent.h
```diff
@@ -353,6 +356,98 @@ typedef Extent<double,2> ExtentD;
...
+template <int N>
+Extent<int,N> floor(Extent<double,N> const & input);
+
+/**
+ *  Return the component-wise floor (round towards more positive).
+ *
+ *  In Python, this is available as both a free function and a method on ExtentD.
+ */
+template <int N>
+Extent<int,N> ceil(Extent<double,N> const & input);
```
jdswinbank: This should probably have an accompanying test case.

I don't see any tests for `truncate()` and `floor()` either, but they are used in the definitions of operators which are tested, so we can be reasonably confident they work. I don't think `ceil()` is used anywhere.

include/lsst/afw/geom/Extent.h
TallJimbo: Will fix.

include/lsst/afw/geom/Extent.h
TallJimbo: The `N < 0` bit is a workaround for a quirk of `static_assert`: I *want* the assert always trigger when this function is invoked, but if the assert condition isn't dependent on the template parameters, the compiler will assert even when the template isn't instantiated (i.e. all the time).

include/lsst/afw/geom/Extent.h
TallJimbo: Yup, sorry, I should have updated the table to reflect the comments.  This behavior is correct.

include/lsst/afw/geom/Extent.h
jdswinbank: Gotcha. I was wondering why we couldn't just assert on `false`, but this makes sense.

include/lsst/afw/geom/Extent.h
TallJimbo: Yup, that's definitely a bug.  I'll have to investigate how it slipped through.

The tests are all pure Python, but the vast majority of them delegate to the corresponding C++ operator.  That's apparently not the case for this one, or it got left out of the tests somehow.  I'll do a sweep and make sure any other C++ operators that aren't used in the Python implementation are tested elsewhere.  (I'd do it now, but I'd like to get to a better stopping point on the other afw ticket I'm working on before checking out this branch again).

include/lsst/afw/geom/Extent.h
TallJimbo: Agreed.  I'd intended to add tests for those and just forgot.

include/lsst/afw/geom/Extent.h
jdswinbank: Although maybe a comment to that effect would be useful? Also, I think "would truncate" would be better, since the truncation isn't actually happening.

python/lsst/afw/geom/Extent.i
```diff
@@ -47,12 +46,138 @@
...
+        Py_INCREF(*PYTHON_SELF);
+        return *PYTHON_SELF;
+    }
     %pythoncode %{
 # support "__from__ future import division" in Python 2; not needed for Python 3
 # warning: this indentation level is required for SWIG 3.0.2, at least
-__truediv__ = __div__
-__itruediv__ = __idiv__
+__div__ = __truediv__
+__idiv__ = __itruediv__
```
jdswinbank: I think the comment here is misleading -- post changes, this is defining `__div__`, not `__truediv__`, so it's setting normal (rather than future) division.

Plus given that this is division for an `ExtentD`, I'm not sure why future division comes into it at all -- you're always expecting to get doubles back, right?

include/lsst/afw/geom/Extent.h
TallJimbo: This one was a case where I was not delegating to the C++ operator, but should have been (historical reasons).  I've fixed that, and I'll now go add some pure C++ tests.

python/lsst/afw/geom/Extent.i
TallJimbo: I fixed the comment.  Without future division, I think `__div__` still needs to be defined (as opposed to just having `__truediv__`).

PR #23, opened 2015-04-28
Tickets/dm 2557

danielsf:
This ticket adds methods to perform findDetectors on numpy arrays of points.  It is necessary to speed up the CatSim methods that read it catalogs of objects and assign them to individual detectors.

### DIFF
--- a/python/lsst/afw/cameraGeom/camera.py
+++ b/python/lsst/afw/cameraGeom/camera.py
@@ -129,31 +129,31 @@ class Camera(DetectorCollection):
                 detectorList.append(detector)
         return detectorList
 
-    def findDetectorsArray(self, xCoord, yCoord, coordSys):
+    def findDetectorsList(self, xCoordList, yCoordList, coordSys):
         """!Find the detectors that cover a list of points specified by x and y coordinates in any system
 
-        @param[in] xCoord is an array of x coordinates
-        @param[in] yCoord is an array of y coordinates
-        @param[in] coordSys is the coordinate system in which xCoord and yCoord are defined
+        @param[in] xCoordList  an array of x coordinates
+        @param[in] yCoordList  an array of y coordinates
+        @param[in] coordSys  the coordinate system in which xCoord and yCoord are defined
         @return a list of lists; each list contains the names of all detectors which contain the
         corresponding point
         """
 
-        cameraPoints = [afwGeom.Point2D(x,y) for x,y in zip(xCoord, yCoord)]
+        cameraPointList = [afwGeom.Point2D(x,y) for x,y in zip(xCoordList, yCoordList)]
 
         #transform the points to the native coordinate system
-        nativePoints = self._transformSingleSysArray(cameraPoints, coordSys, self._nativeCameraSys)
+        nativePointList = self._transformSingleSysArray(cameraPointList, coordSys, self._nativeCameraSys)
 
         detectorList = []
-        for i in range(len(xCoord)):
+        for i in range(len(xCoordList)):
             detectorList.append([])
 
         for detector in self:
             coordMap = detector.getTransformMap()
             cameraSys = detector.makeCameraSys(PIXELS)
-            detectorPoints = coordMap.transform(nativePoints, self._nativeCameraSys, cameraSys)
+            detectorPointList = coordMap.transform(nativePointList, self._nativeCameraSys, cameraSys)
             box = afwGeom.Box2D(detector.getBBox())
-            for i, pt in enumerate(detectorPoints):
+            for i, pt in enumerate(detectorPointList):
                 if box.contains(pt):
                     detectorList[i].append(detector)
 
--- a/python/lsst/afw/cameraGeom/camera.py
+++ b/python/lsst/afw/cameraGeom/camera.py
@@ -139,7 +139,7 @@ class Camera(DetectorCollection):
         corresponding point
         """
 
-        cameraPoints = [afwGeom.Point2d(x,y) for x,y in zip(xCoord, yCoord)]
+        cameraPoints = [afwGeom.Point2D(x,y) for x,y in zip(xCoord, yCoord)]
 
         #transform the points to the native coordinate system
         nativePoints = self._transformSingleSysArray(cameraPoints, coordSys, self._nativeCameraSys)
--- a/python/lsst/afw/cameraGeom/camera.py
+++ b/python/lsst/afw/cameraGeom/camera.py
@@ -129,23 +129,20 @@ class Camera(DetectorCollection):
                 detectorList.append(detector)
         return detectorList
 
-    def findDetectorsList(self, xCoordList, yCoordList, coordSys):
+    def findDetectorsList(self, cameraPointList, coordSys):
         """!Find the detectors that cover a list of points specified by x and y coordinates in any system
 
-        @param[in] xCoordList  an array of x coordinates
-        @param[in] yCoordList  an array of y coordinates
-        @param[in] coordSys  the coordinate system in which xCoord and yCoord are defined
+        @param[in] cameraPointList  a list of cameraPoints
+        @param[in] coordSys  the camera coordinate system in which cameraPointList is defined
         @return a list of lists; each list contains the names of all detectors which contain the
         corresponding point
         """
-
-        cameraPointList = [afwGeom.Point2D(x,y) for x,y in zip(xCoordList, yCoordList)]
 
         #transform the points to the native coordinate system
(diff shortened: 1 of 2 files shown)


python/lsst/afw/cameraGeom/camera.py
```diff
@@ -106,8 +129,38 @@ def findDetectors(self, cameraPoint):
...
+    def findDetectorsArray(self, xCoord, yCoord, coordSys):
+        """!Find the detectors that cover a list of points specified by x and y coordinates in any system
+
+        @param[in] xCoord is an array of x coordinates
+        @param[in] yCoord is an array of y coordinates
+        @param[in] coordSys is the coordinate system in which xCoord and yCoord are defined
+        @return a list of lists; each list contains the names of all detectors which contain the
+        corresponding point
+        """
+
```
r-owen: We must be very careful in afw to disambiguate between afw.coord and afw.cameraGeom. Thus in cameraGeom we are careful to use the terms "camera coordinate system" (the word "camera" is always necessary) "point" or "camera point" instead of "coord" or "coordinate". We reserve the plain terms "coord", "coordinate", "coordinate system" for the coord package.

From this perspective this API (and perhaps others you added) need some work. Here are some suggestions:
- Please don't pass x and y as separate arrays. It would be much safer and cleaner to pass them as a single array of Point2D or some other object. This saves any worry that they may not be the same length or may not be in the same order.
- Change the name "coord" to pointList (or perhaps "cameraPointList", but does "camera" add useful information?). Note that our coding standards discourage simple plurals for collections, so please also change the code accordingly.
- For coordSys: in the description change "coordinate system" to "camera coordinate system"
- Please remove the word "is" from your @param commands; just use a pair of spaces (read the generated documentation to see if you agree)

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -106,8 +129,38 @@ def findDetectors(self, cameraPoint):
...
+            box = afwGeom.Box2D(detector.getBBox())
+            for i, pt in enumerate(detectorPoints):
+                if box.contains(pt):
+                    detectorList[i].append(detector)
+
+        return detectorList
+
     def getTransformMap(self):
-        """!Obtain a pointer to the transform registry.  
+        """!Obtain a pointer to the transform registry.
```
r-owen: While you are fixing things, could you please get rid of "a pointer to"? Pointers are a C++ concept.

tests/testCameraGeom.py
```diff
@@ -218,6 +218,8 @@ def testTransformDet(self):
 
     def testFindDetectors(self):
         for cw in self.cameraList:
+            detXPoints = []
+            detYPoints = []
```
r-owen: Use detPointList and don't bother to split into x and y. If you improve findDetectorsArray as suggested earlier then you won't need the split.

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -65,8 +65,31 @@ def _tranformFromNativeSys(self, nativePoint, toSys):
         else:
             return self._transformSingleSys(nativePoint, toSys)
 
+    def _transformSingleSysArray(self, positionArray, fromSys, toSys):
+        """!Transform an array of points from once CameraSys to another CameraSys
+        @warning This method only handles a single jump, not a transform linked by a common native sys.
+        
+        @param[in] positionArray Array of Point2D objects, one per position
+        @param[in] fromSys  Initial coordinate system
+        @param[in] toSys  Destination coordinate system
```
r-owen: On the whole this method looks fine. Two requested tweaks:
- Rename variable positionArray to positionList, our generic name for an ordered collection
- For the parameter documentation change "coordinate system" to "camera coordinate system"

python/lsst/afw/cameraGeom/camera.py
TallJimbo: I think we're going to have to start living with some interfaces that accept separate arrays of x and y instead of arrays of points, because arrays of points just aren't NumPy-friendly.

That said, this method apparently uses a vector of Points under the hood (which it populates in a loop in Python), so I can't see how it'd be any more efficient than the interfaces we have now.  If you want to speed it up, I'd think you'd need a C++ interface that takes an `ndarray::Array` for each of x and y, and then loops over them in C++.  Otherwise you're just substituting one loop over points in Python for another loop over the same points in Python.

danielsf: It's less that this makes camerGeom any faster than that it facilitates a faster interface with what is going on the catalog simulations code.

The catalog simulations code already reads object positions in from a databases and stores them as numpy arrays of Ra, Dec, which it converts quickly into numpy arrays of pupil coordinates (or whatever other coordinates we need at the time).  In order to speed up the conversion from Ra, Dec, we needed an interface that would allow us to pass in those numpy arrays of coordinates and get out a numpy array of corresponding chipNames.

So: I'm sure what Jim is saying is correct, but the existing catalog simulations code had forced us into writing something like this.

danielsf: For a similar reason, I'm going to push back on Russell's comment, since, in CatSim, we already have numpy arrays of x and y.  Converting them into Point2D in CatSim (which is all python), is what was slowing us down before.

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -106,8 +129,38 @@ def findDetectors(self, cameraPoint):
...
+        """
+
+        cameraPoints = [afwGeom.Point2D(x,y) for x,y in zip(xCoord, yCoord)]
+
+        #transform the points to the native coordinate system
+        nativePoints = self._transformSingleSysArray(cameraPoints, coordSys, self._nativeCameraSys)
+
+        detectorList = []
+        for i in range(len(xCoord)):
+            detectorList.append([])
```
PaulPrice: Better to use list comprehension:
    detectorList = [[] for _ in range(len(xCoord))]

There's no check that xCoord and yCoord have the same length.

python/lsst/afw/cameraGeom/camera.py
PaulPrice: If this is just an impedance match, why does it have to be on the afw side rather than the sims side?

r-owen: The first thing this routine does is [afwGeom.Point2D(x,y) for x,y in zip(xCoord, yCoord)]. All I am asking you to do is move that out of this method and accept a collection of Point2D. It cannot affect speed and it makes for a much cleaner interface.

As to Jim's comment: I would be much happier with a numpy structured array with x and y fields than two separate arrays. Perhaps we can find some way to efficiently convert between this and the bits of C++ that presently use vectors of Point2D. But I think that is for the future.

danielsf: The real answer to Jim's question may be: I'm not 100% sure why it's faster, but it is.  Running on a simulated catalog 5000 objects, using findDetectors for each takes a total of about 46 seconds.  Using findDetectorsList takes 7 seconds.

I think it is related to the frequency of calls to coordinate transformations, either to the camera system or the detector systems.

RobertLuptonTheGood: Please open an issue to sort this out later.

r-owen: I think there are three different issues in this discussion and I fear they are getting conflated:

1) The speedup. This is caused by using findDetectorsArray instead of findDetectors, and it makes sense because you are vectorizing coordinate transformations. That is perfectly understandable and does not require a ticket.

2) findDetectorsArray x,y arrays vs a list of Point2D. The first thing findDetectorsArray does is turn the x and y array into a list of findDetectorsArray (in python). All I'm asking is that the caller do this, instead, so findDetectorsArray can have a cleaner and API that looks more like the API of TransformMap.

3) Eventually we will want a way of passing around lists of points and possibly coordinates (e.g. ICRS) as numpy arrays. One obvious representation of a list of points in numpy is a structured array with x and y fields. But how do we get that into C++ (e.g. for the vector version of TransformMap.transform)? Coordinates add additional complications. This issue could probably use a separate ticket; it certainly should not be solved on this one.

TallJimbo: Thanks you all for the explanation on (1); just wanted to make sure we weren't just guessing this would go faster.

As for (2) and (3), I think it's at least a little bit relevant here because looking towards the future I don't think it's entirely obvious that we want to discourage interfaces that accept separate arrays of x and y.  I agree we don't want to try to come up with a general solution here, but if that interface was provided in addition to one that takes a vector of Points, I don't think it's a problem.

PR #25, opened 2015-05-27
afw tickets/DM-2533: remove "version" from Schemas

TallJimbo:
Going forward, old-version Schemas will only be supported with backwards compatibility in the FITS reading code: instead of loading older schemas directly (which required lots of special case code all over the codebase) we'll convert them to the new conventions as we read them.

That conversion isn't actually in place yet; it will be added in other issues in DM-1766.  This PR just includes the removal of the old backwards compatibility code.


laurenam:
Some of the deleted functions (e.g. getModelFluxKey) show up in the following repos:
afw-test
ap
analysis

I don't think these are actually currently in use, so I assume it's a non-issue?

TallJimbo:
`afw-test` is definitely not a real repository. @frossie, is that a temporary thing that should have already been removed?

`ap` is sort of in use, and I've created RFC-57 to try to sort that out.

`analysis` is in use by Robert, but I'm not sure all parts of it are expected to be in working condition at any given time, and currently he handles all the fixes there.

TallJimbo:

PR #26, opened 2015-05-27
afw tickets/DM-2534

TallJimbo:
This is a major refactoring of the afw::table FITS read code, along with a much smaller change to the FITS write code and an unrelated small fix to a warning in TanWcs.

The read code refactoring was necessary to support some new backwards compatibility features, which will be added in DM-2535 and DM-2536.


include/lsst/afw/table/Exposure.h
```diff
@@ -1,8 +1,8 @@
 // -*- lsst-c++ -*-
-/* 
+/*
```
timj: I do sometimes wonder if we shouldn't just do a single clean up of all whitespace issues (tied with people making sure that trailing whitespace is visible in their editors and possibly a CI check).

include/lsst/afw/table/io/FitsReader.h
```diff
@@ -15,135 +17,85 @@ namespace lsst { namespace afw { namespace table { namespace io {
...
-        return apply<ContainerT>(fits, flags);
-    }
-
-    /// @brief Low-level entry point for reading FITS files into arbitrary containers.
     template <typename ContainerT>
-    static ContainerT apply(Fits & fits, PTR(io::InputArchive) archive=PTR(io::InputArchive)(), int flags=0) {
-        PTR(FitsReader) reader = make(&fits, archive, flags);
-        return reader->template read<ContainerT>();
+    static ContainerT apply(afw::fits::Fits & fits, int ioFlags, PTR(InputArchive) archive=nullptr) {
+        PTR(daf::base::PropertyList) metadata = boost::make_shared<daf::base::PropertyList>();
```
timj: Are we using `boost::make_shared` in new code? Shouldn't we be switching to `std::make_shared`? I see that qserv have been migrating away from boost for things like this so the stack build system already has to handle it.

src/image/TanWcs.cc
```diff
@@ -324,8 +324,8 @@ geom::Point2D TanWcs::undistortPixel(geom::Point2D const & pix) const {
 
     // Polynomial orders for U = f(u,v) and V = g(u,v)
     // Note these may be different in general, but not usually in practise.
-    size_t const fOrder = _sipA.rows();
-    size_t const gOrder = _sipB.rows();
+    int const fOrder = _sipA.rows();
+    int const gOrder = _sipB.rows();
```
timj: why were these changed to `int` but the others changed to `std::size_t`?

include/lsst/afw/table/io/FitsReader.h
TallJimbo: At some point, yes, but there will be a lot of code we have to convert at once, and I haven't had the energy to RFC it and get it done.

include/lsst/afw/table/Exposure.h
TallJimbo: Ah, I meant to revert the whitespace changes before I pushed.  I might still do that with a force push, just to keep the changes distinct.  I've been trying out Atom as my text editor, and it actually just goes ahead and *deletes* all trailing whitespace automatically, which is a bit overzealous, especially when making small changes to old files.

src/image/TanWcs.cc
TallJimbo: These ones are used in a loop where they have to be able to go negative.

jdswinbank: From glancing at the code it's not obvious why these don't follow the same pattern as elsewhere -- both Tim & I did a double take. I think at least a comment pointing out what's happening would be useful.

Actually, though, these don't go negative (they're ``const``). I think the point is that you want to get a negative when you evaluate ``fOrder - i`` a few lines below. I wonder whether a cast there would be clearer than using ``int``s in the above... but I don't think it's a big deal either way.

include/lsst/afw/table/io/FitsReader.h
```diff
@@ -15,135 +17,85 @@ namespace lsst { namespace afw { namespace table { namespace io {
  *  FitsReader itself provides the implementation for reading standard FITS binary tables
```
jdswinbank: GitHub won't let me comment on line 15 above since it's not part of these changes, but note that this is no longer "A Reader subclass", so the documentation needs tweaking.

include/lsst/afw/table/io/FitsReader.h
```diff
@@ -15,135 +17,85 @@ namespace lsst { namespace afw { namespace table { namespace io {
...
-        fits.setHdu(hdu);
-        return apply<ContainerT>(fits, flags);
-    }
-
-    /// @brief Low-level entry point for reading FITS files into arbitrary containers.
     template <typename ContainerT>
-    static ContainerT apply(Fits & fits, PTR(io::InputArchive) archive=PTR(io::InputArchive)(), int flags=0) {
-        PTR(FitsReader) reader = make(&fits, archive, flags);
-        return reader->template read<ContainerT>();
+    static ContainerT apply(afw::fits::Fits & fits, int ioFlags, PTR(InputArchive) archive=nullptr) {
```
jdswinbank: I think the documentation for the ``apply()`` functions was useful.

In fact, I think it would be useful it could be expanded upon, rather than removed. Coming at this code cold, I'm not sure what ``InputArchive`` is for (and glancing at its documentation doesn't help much). I see that it's used when reading ``Exposure`` and ``Source``, so it'll probably become clear later, but a bit more up-front explanation of what's happening would help.

src/table/io/FitsReader.cc
```diff
@@ -1,547 +1,39 @@
...
+    int ioFlags,
+    bool stripMetadata
+) const {
+    PTR(BaseTable) result = BaseTable::make(mapper.finalize());
+    result->setMetadata(metadata);
+    return result;
+}
+
+FitsReader::FitsReader(std::string const & name) {
     getRegistry()[name] = this;
```
jdswinbank: So if you construct multiple ``FitsReader``s with the same name, the last one will take precedence. Is that desirable? Would it be better to throw an error if the name is already taken?

include/lsst/afw/table/io/FitsSchemaInputMapper.h
```diff
@@ -0,0 +1,160 @@
...
+    std::string tccls;   // which field class to use (from our own TCCLS keys)
+    std::string tunit;   // field units (from TUNIT keys)
+    std::string doc;     // field docs (from comments on TTYPE keys)
+
+    explicit FitsSchemaItem(int column_, int bit_) : column(column_), bit(bit_) {}
+};
+
+
+/**
+ *  A class that describes a mapping from a FITS binary table and an afw::table Schema.
```
jdswinbank: "between a... and an... ", or "from a... to an...".

(I've not yet looked to see if the mapping is bidrectional.)

include/lsst/afw/table/io/FitsSchemaInputMapper.h
```diff
@@ -0,0 +1,160 @@
...
+ *  repeatedly to read FITS rows into record objects according to the mapping that has been
+ *  defined.
+ */
+class FitsSchemaInputMapper {
+public:
+
+    typedef FitsSchemaItem Item;
+
+    /// Construct a mapper from a PropertyList of FITS header values, stripping recognized keys if desired.
+    FitsSchemaInputMapper(daf::base::PropertyList & metadata, bool stripMetadata);
```
jdswinbank: Why would stripping recognized keys be desirable? Or, perhaps more to the point: why would it not be?

I think that the only time this is called it has ``stripMetadata=true`` -- does the option need to exist at all?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -0,0 +1,619 @@
...
+    _impl(boost::make_shared<Impl>())
+{
+    // Set the table version.  If AFW_TABLE_VERSION tag exists, use that
+    // If not, set to 0 if it has an AFW_TYPE, Schema default otherwise (DM-590)
+    if (!metadata.exists("AFW_TYPE")) {
+        _impl->version = lsst::afw::table::Schema::VERSION;
+    }
+    _impl->version = metadata.get("AFW_TABLE_VERSION", _impl->version);
+    if (stripMetadata) {
+        metadata.remove("AFW_TABLE_VERSION");
```
jdswinbank: Do you also want to strip ``AFW_TYPE`` if it exists?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -0,0 +1,619 @@
...
+    }
+
+    // Find a key that indicates an Archive stored on other HDUs
+    _impl->archiveHdu = metadata.get("AR_HDU", -1);
+    if (stripMetadata && _impl->archiveHdu > 0) {
+        metadata.remove("AR_HDU");
+    }
+
+    // read aliases stored in the new, expected way
```
jdswinbank: Should we define what the "new, expected way" is? And what if there are aliases stored in an old way (is there an old way)?

More generally, I'm a bit conflicted as to what extent the on-disk structure of the FITS files should be documented. Certainly, if we ever expected anybody to understand the FITS files without reading them through the LSST code, we need to document things like how aliases are stored. I'm not sure if that's necessary though; maybe it's legit to say that this is basically an opaque blob which you need to read with the LSST stack. Even in that latter case, it might also be convenient for our own purposes to document how we're storing things in the files.

src/table/io/FitsSchemaInputMapper.cc
jdswinbank: Oh you do, but a couple of hundred lines below. Maybe do it here for consistency?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -0,0 +1,619 @@
...
+        }
+        --_impl->flagColumn; // switch from 1-indexed to 0-indexed
+        auto iter = _impl->byColumn().find(_impl->flagColumn);
+        if (iter == _impl->byColumn().end()) {
+            throw LSST_EXCEPT(
+                afw::fits::FitsError,
+                (boost::format("Column for flag data not found; FLAGCOL=%d") % _impl->flagColumn).str()
+            );
+        }
+        static boost::regex const regex("(\\d+)?X\\(?(\\d)*\\)?", boost::regex::perl);
```
jdswinbank: It would save the reader a few minutes to be explicit about what you're checking for here.

jdswinbank: (Although I do note that this is not new code but rather being moved in from elsewhere, so it's fair enough to leave it as it is.)

include/lsst/afw/table/io/FitsReader.h
```diff
@@ -15,135 +17,85 @@ namespace lsst { namespace afw { namespace table { namespace io {
...
-    struct ProcessRecords;
+    virtual ~FitsReader() {}
 
-    Fits * _fits;         // cfitsio pointer in a conveniencer wrapper
-    int _flags;           // subclass-defined flags to control FITS reading
-    std::size_t _row;     // which row we're currently reading
 private:
 
-    friend class afw::table::Schema;
+    friend class Schemaa;
```
jdswinbank: ``Schemaa``?

src/table/Exposure.cc
```diff
@@ -186,61 +186,72 @@ void ExposureFitsWriter::_writeRecord(BaseRecord const & r) {
...
-// used whenever we read a table with AFW_TYPE set to that value.
-
-namespace {
-
-class ExposureFitsReader : public io::FitsReader {
+// FitsColumnReader that reads a Persistable subclass T (Wcs, Psf, or Calib here) by using an int
+// column to retrieve the object from an InputArchive and attach it to an ExposureRecord via
+// the Setter member function pointer.
+template <typename T, void (ExposureRecord::*Setter)(PTR(T const))>
+class PersistableColumnReader : public io::FitsColumnReader {
```
jdswinbank: Naming things is hard, but:

At first glance, this sounds like a ``FitsColumnReader`` which can be persisted. In fact, I think it's a ``FitsColumnReader`` which is used to read things which are persistable. Not a big deal (and maybe this is actually following naming conventions elsewhere), but I wonder if the name could be more explicit?

src/table/Exposure.cc
```diff
@@ -186,61 +186,72 @@ void ExposureFitsWriter::_writeRecord(BaseRecord const & r) {
 //----- ExposureFitsReader ---------------------------------------------------------------------------------
```
jdswinbank: This name (``ExposureFitsReader``) is obsolete.

src/table/Source.cc
```diff
@@ -407,38 +316,127 @@ PTR(BaseRecord) SourceFitsReader::_readRecord(PTR(BaseTable) const & table) {
...
+private:
+    int _spanCol;
+    int _peakCol;
+    int _heavyPixCol;
+    int _heavyMaskCol;
+    int _heavyVarCol;
+};
+
+// FitsColumnReader for new-style Footprint persistence using archives.
+class SourceFootprintReader : public io::FitsColumnReader {
```
jdswinbank: This looks like it ought to be derived from ``PersistableColumnReader``.

src/table/Exposure.cc
```diff
@@ -186,61 +186,72 @@ void ExposureFitsWriter::_writeRecord(BaseRecord const & r) {
...
+        fits.readTableScalar<int>(row, _column, id);
+        PTR(T) value = archive->get<T>(id);
+        (static_cast<ExposureRecord&>(record).*(Setter))(value);
+    }
 
-    PTR(BaseTable) _inTable;
-    PTR(io::InputArchive) _archive;
-    SchemaMapper _mapper;
+private:
+    bool _noHeavy;
```
jdswinbank: ``_noHeavy`` isn't used, so far as I can see. (It looks like it belongs in ``SourceFootprintReader``.)

src/table/Source.cc
```diff
@@ -175,12 +175,13 @@ void SourceFitsWriter::_writeRecord(BaseRecord const & r) {
 //----- SourceFitsReader ------------------------------------------------------------------------------------
 //-----------------------------------------------------------------------------------------------------------
 
-// A custom FitsReader for Sources - this reads footprints as variable-length arrays, and adds header
-// keys that define the slots.  It gets registered with name SOURCE, so it should get used whenever
-// we read a table with AFW_TYPE set to that value.
+// A custom FitsReader for Sources - this reads footprints in addition to the regular fields.  It
```
jdswinbank: There are actually a couple of moderately weighty classes defined between here and ``SourceFitsReader``. They're necessary to make things work, but maybe this comment could be a bit closer to the class it actually describes?

include/lsst/afw/table/io/FitsSchemaInputMapper.h
```diff
@@ -0,0 +1,160 @@
...
+    /// Return true if the mapper has an InputArchive.
+    bool hasArchive() const;
+
+    /**
+     *  Return a mutable reference to the Schema object that is being built.
+     *
+     *  This allows new fields to be added by the user (usually those required by custom FitsColumnReaders).
+     *  It should not be called after finalize() has been called.
+     */
+    Schema & editSchema();
```
jdswinbank: I don't think this is actually used anywhere at present. Is it necessary?

Would it be worth storing a flag to indicate whether the mapper has been finalized and throwing if it is set when ``editSchema()`` is called?

include/lsst/afw/table/Exposure.h
TallJimbo: Fixed (on DM-1766; all my changes will go there to avoid disrupting the diffs for the ongoing reviews of other DM-1766 subtasks).

src/image/TanWcs.cc
TallJimbo: I added a comment.  I couldn't find a version using a cast in the loop that didn't disrupt the readability of the loop more than I was comfortable with.

include/lsst/afw/table/io/FitsReader.h
TallJimbo: Fixed.  Also fixed `read()` vs. `apply()` later in the text.

include/lsst/afw/table/io/FitsReader.h
TallJimbo: Sorry - I removed the documentation during the refactoring, as I could see it was getting inconsistent with the code, and I'd always meant to rewrite it.  Of course, I forgot.

Now fixed on tickets/DM-1766.

src/table/io/FitsReader.cc
TallJimbo: Maybe, but I'm a little worried about throwing exceptions during static initialization (note that `FitsReader`s will always be static variables).  Mostly, I just don't think there's any real danger of that happening.

include/lsst/afw/table/io/FitsSchemaInputMapper.h
TallJimbo: Fixed (it's unidirectional).

include/lsst/afw/table/io/FitsSchemaInputMapper.h
TallJimbo: The only code that could have passed `stripMetadata=false` was removed rather late in the refactor.  I'm not convinced the option needs to exist in the low-level code at all, but it is an option in all of our other FITS header parsing code.

I'm fairly ambivalent as to whether it's better to remove it to simplify the code a bit or leave it in for consistency with the other FITS header parsing code.  If you have a strong opinion, I'm happy to go with it.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: Good point.  I had it in my head that I couldn't remove it here because I needed it to still be present in some downstream processing, but looking again I must have reordered things so that's no longer the case.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: I'll clarify the code comment.  There is no "old way" for aliases, just an old way for slots, which predate aliases.

Your concern about how official the FITS format is is a timely one; I recently had a discussion with Gregory about how the SUI team can access some of the metadata in our FITS files that goes beyond what's standardized by FITS itself, and since they're using entirely different languages they'd like to avoid using afw to do the reading.  I definitely don't want to start formalizing our on-disk formats beyond what's in the FITS standard to the extent that we'll be unable to change those formats in the future, but I do anticipate we'll want to start documenting things like aliases for external users in the future - with a large "use at your own risk" disclaimer.

I'm going to treat that as a separate issue, though.

include/lsst/afw/table/io/FitsSchemaInputMapper.h
jdswinbank: My inclination would be to drop any code that we're not using and don't have a clear use case for in future, but, particularly given your point re consistency with other code, I don't think that amounts to a particularly strong opinion. Happy to go with your preference.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: I've added some comments to that effect.

include/lsst/afw/table/io/FitsReader.h
TallJimbo: Interesting.  Looks like that friend declaration isn't needed anymore, and I forgot to remove it.

src/table/Exposure.cc
TallJimbo: Renamed to `PersistableObjectColumnReader`.

src/table/Exposure.cc
TallJimbo: That's actually intentional - I'm trying to emphasize that the following class (`PersistableObjectColumnReader`) is just a helper class for (and implementation detail of) `ExposureFitsReader`.

src/table/Source.cc
TallJimbo: Yeah, I thought about trying to share some code here, but there simply wasn't any real code to share - it's just boilerplate that has to be repeated anyway.

src/table/Source.cc
TallJimbo: Instead I've mentioned in that comment that the reader class itself is after some helper classes.  I'd like to keep the description up here so someone reading down from the top of the file understands the context for the helper classes.

include/lsst/afw/table/io/FitsSchemaInputMapper.h
TallJimbo: Good catch.  It's not necessary.  I've removed it.

include/lsst/afw/table/io/FitsSchemaInputMapper.h
TallJimbo: I think I'll just leave it as it is.

TallJimbo:

PR #28, opened 2015-05-28
afw tickets/DM-2536: backwards-compatible reading of version 0 slots

TallJimbo:


TallJimbo:

PR #29, opened 2015-06-10
Tickets/dm 2599 Do something sensible when building images and exposures from MEFs.

SimonKrughoff:


timj:
Nicely comprehensive tests.

PR #30, opened 2015-06-23
tickets/DM-2931: truncation in Wcs persistence

TallJimbo:


src/image/ExposureInfo.cc
```diff
@@ -254,7 +254,10 @@ void ExposureInfo::_readFits(
         }
         int wcsId = popInt(*metadata, "WCS_ID");
         try {
-            _wcs = archive.get<Wcs>(wcsId);
+            auto archiveWcs = archive.get<Wcs>(wcsId);
+            if (archiveWcs) {
+                _wcs = archiveWcs;
+            }
```
TallJimbo: It looks like the warning message below this block is incorrect: we should say we're just using the Wcs we got from the header.

By the way, I just refreshed my memory on when that warning is supposed to appear: it's when we couldn't load the object because it's defined in an extension package that's not setup.  If the WCS wasn't saved in the tables at all, we'll just get a null pointer back, and the block above is in play.  If the WCS was saved but it's been corrupted, we'll get an exception that propagates all the way up.

ktlim:
Is it possible to write test cases for this code?

src/image/Wcs.cc
```diff
@@ -1102,13 +1102,22 @@ void Wcs::write(OutputArchiveHandle & handle) const {
...
 }
 
+bool Wcs::isPersistable() const {
+    if (!_mayBePersistable()) {
+        return false;
+    }
+    // The current table persistence only works for TAN and TAN-SIP projections
+    return false;
+}
+
```
TallJimbo: I don't think we need to call `_mayBePersistable()` at all here, since we're going to return false regardless.

RobertLuptonTheGood: I wrote it that way as a template for derived classes.  It doesn't cost enough to matter.

src/image/ExposureInfo.cc
RobertLuptonTheGood: I fixed the comment

I also note that we could check for wcsId == 0, but doing that would break the symmetry with e.g. Psf handing and I didn't want to fix all of them on this ticket.

PR #31, opened 2015-06-23
Review of DM-2967 (Be a bit pushier about deleting PVi_j cards)

jdswinbank:


src/image/Wcs.cc
```diff
@@ -217,21 +217,6 @@ void Wcs::initWcsLibFromFits(CONST_PTR(lsst::daf::base::PropertySet) const& head
...
-            std::string key = (boost::format("PV%i_%i") % j % i).str();
-            if (!access.toRead()->exists(key)) {
-                break;
-            }
-            double val = access.toRead()->getAsDouble(key);
-            access.toWrite()->remove(key);
-            access.toWrite()->add("X_"+key, val);
-        }
-    }
-
```
jdswinbank: A minor formatting point, but it would be nice to reinstate one of the blank lines you've removed for the sake of the rest of the code.

src/image/makeWcs.cc
```diff
@@ -95,13 +95,19 @@ afwImg::Wcs::Ptr afwImg::makeWcs(
         metadata->set<std::string>("CTYPE1", "RA---TAN");
         metadata->set<std::string>("CTYPE2", "DEC--TAN");
         metadata->set<bool>("TPV_WCS", true);
-
-        for (int i = 1; i <= 3; ++i) {
-            for (int j = (i == 1 ? 5 : 1); j <= 16; ++j) { // note that PV1_{1..4} are legal
+        // PV1_[1-4] are in principle legal, but Swarp reuses them as part of the TPV parameterisation
+        // As a compromise, we remove them iff there is at least one PV2_? card
+        bool sawPV2 = false;               // did we see a PV2_? card?
+        for (int i : { 2, 1 }) {
```
jdswinbank: Range-based for isn't supported by GCC 4.4, so this will cause Buildbot to choke.

src/image/makeWcs.cc
```diff
@@ -95,13 +95,19 @@ afwImg::Wcs::Ptr afwImg::makeWcs(
...
+            for (int j = (i == 1 && sawPV2 ? 5 : 1); j <= 99; ++j) {
```
jdswinbank: TPV is only defined up to `j=39` (according to http://fits.gsfc.nasa.gov/registry/tpvwcs/tpv.html). I suppose there's no harm in checking up to 99 though.

PR #33, opened 2015-07-30
tickets/DM-3265

danielsf:
I split up assertWcsNearlyEqualOverBBox into two methods:

wcsNearlyEqualOverBBox, which does the work and then returns a boolean (True if the WCSs are nearly equal to within the specified tolerance) and a list of generated messages

assertWcsNearlyEqualOverBBox which behaves as before, but calls wcsNearlyEqualOverBBox

This allows me to use wcsNearlyEqualOverBBox to validate WCS approximations outside of the context of unit tests.


danielsf:
Russell Owen suggested a better way to achieve what I needed in meas_astrom.  Abandoning this pull request.

PR #34, opened 2015-08-05
tickets/DM-3347 wcsNearlyEqual

ktlim:


python/lsst/afw/image/basicUtils.py
```diff
@@ -92,12 +94,16 @@ def assertWcsNearlyEqualOverBBox(testCase, wcs0, wcs1, bbox, maxDiffSky=0.01*afw
             diffSky = sky0.angularSeparation(sky1)
             if diffSky > measDiffSky[0]:
                 measDiffSky = (diffSky, fromPixPos)
+                if doShortCircuit:
+                    break
```
ktlim: These breaks only exit one level.  You need to break out of both levels.

python/lsst/afw/image/basicUtils.py
```diff
@@ -106,5 +112,70 @@ def assertWcsNearlyEqualOverBBox(testCase, wcs0, wcs1, bbox, maxDiffSky=0.01*afw
...
+    @param[in] bbox  boundaries of pixel grid over which to compare the WCSs (an lsst.afw.geom.Box2I or Box2D)
+    @param[in] maxDiffSky  maximum separation between sky positions computed using Wcs.pixelToSky
+        (an lsst.afw.geom.Angle)
+    @param[in] maxDiffPix  maximum separation between pixel positions computed using Wcs.skyToPixel
+    @param[in] nx  number of points in x for the grid of pixel positions
+    @param[in] ny  number of points in y for the grid of pixel positions
+    @param[in] doShortCircuit  if True then stop at the first error, else test all values in the grid
+        and return information about the worst violations found
+
+    @return None if the WCS are nearly equal, else return a string describing the largest error measured
```
ktlim: Using `not` on the (string) result of `_compareWcsOverBBox` does what the brief description on line 120 says but not what this comment says.

python/lsst/afw/image/basicUtils.py
r-owen: Good catch. I'll use itertools.product to make the for loop one level deep.

python/lsst/afw/image/basicUtils.py
r-owen: Thank you for catching this. I removed the `@return` documentation because it duplicated the summary. I also switched added an explicit bool cast to the value returned by _compareWcsOverBBox in hopes of making the intent a bit more obvious.

PR #36, opened 2015-09-28
Tickets/dm 3939

danielsf:
This is a much scaled-back version of what we talked about Friday.

After looking more closely at makeLsstRepository.py (in obs_lsstSim), I became less convinced that it could be made LSST-agnostic.  Instead, I have created a class BasicCameraFactory that can be used to create an afwCameraGeom.camera object that only contains the coordinate transformations and chip bounding boxes associated with a camera (i.e. there is no useful AmpTableDict associated with the camera).  This is sufficient for the sims needs I envisioned.  In the future, it should be able to expand this class to include AmpTableDict information (provided we come up with a suitably generic way of parametrizing amp information; the AmpTableDict methods are where most of the LSST-specific code in obs_lsstSim lives).

If you would rather I move this class to one of the sims packages, that will be very easy to do.

Thanks for looking at this.


tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+                                         detTypeMap = {'science':SCIENCE,
+                                                       'focus':FOCUS,
+                                                       'guider':GUIDER,
+                                                       'wave':WAVEFRONT})
+
+
+            camera = factory.makeCamera()
+
+            for detName, yaw, xCenter, yCenter, nx, ny, mmPerPixel in \
+            zip(detNameList, yawList, xCenterList, yCenterList, \
```
timj: can you please indent this so that it doesn't look like a completely new statement lined up with the `for`. `flake8` will complain.

tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+                                           'tests', 'data')
+
+    def testException(self):
+        """
+        Test that an exception is raised if you try to create a BasicCameraFactory
+        without specifying detectorIdFromAbbrevName
+        """
+        layoutFile = os.path.join(self.cameraDataDir, 'testFocalPlaneLayout.txt')
+
+        self.assertRaises(RuntimeError, BasicCameraFactory, layoutFile, \
```
timj: Remove the backslash. You don't need it if you are inside parens.

tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+        detNameList = ['Det00', 'Det01', 'Det02']
+        yawList = np.radians([20.0, 10.0, 30.0])
+        xCenterList = [0.0, 0.0, -9.0]
+        yCenterList = [0.0, 9.0, 0.0]
+        nxList = [400, 300, 300]
+        nyList = [400, 200, 300]
+        mmPerPixelList = [2.0e-3, 1.0e-3, 3.0e-3]
+
+        for detName, yaw, xCenter, yCenter, nx, ny, mmPerPixel in \
+        zip(detNameList, yawList, xCenterList, yCenterList, \
```
timj: Remove backslash.

tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+            nxList, nyList, mmPerPixelList):
+
+            pixelSystem = camera[detName].makeCameraSys(PIXELS)
+            focalSystem = camera[detName].makeCameraSys(FOCAL_PLANE)
```
timj: Does this variable get used?

tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+                                                       'wave':WAVEFRONT})
+
+
+            camera = factory.makeCamera()
+
+            for detName, yaw, xCenter, yCenter, nx, ny, mmPerPixel in \
+            zip(detNameList, yawList, xCenterList, yCenterList, \
+                nxList, nyList, mmPerPixelList):
+
+                pupilSystem = camera[detName].makeCameraSys(PUPIL)
```
timj: I don't think this variable is used.

tests/testBasicCameraFactory.py
```diff
@@ -0,0 +1,278 @@
...
+                # approximate radians per pixel
+                scale_approx = mmPerPixel/radial_coeffs[1]
+
+                for theta in np.arange(0.0, 2.0*np.pi, 0.5*np.pi):
+                    for rr in np.arange(10.0*scale_approx, 200.0*scale_approx, 50.0*scale_approx):
+
+                        xp = rr*np.cos(theta)
+                        yp = rr*np.sin(theta)
+
+                        rr_scale = np.array([radial_coeffs[ii]*rr**ii \
```
timj: remove backslash.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,233 @@
...
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division
+import argparse
+import os
+import re
+import shutil
```
timj: Check how many of these imports are used. I can't find `os` being used here.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,233 @@
...
+
+import lsst.utils
+import lsst.afw.geom as afwGeom
+import lsst.afw.table as afwTable
+from lsst.afw.cameraGeom import makeCameraFromCatalogs
+from lsst.afw.cameraGeom import (DetectorConfig, CameraConfig, PUPIL, FOCAL_PLANE, PIXELS,
+                                 SCIENCE, FOCUS, GUIDER, WAVEFRONT)
```
timj: Are all these used? GUIDER doesn't seem to be mentioned anywhere other than the documentation.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,233 @@
...
+
+        self._detectorLayoutFile = detectorLayoutFile
+        self._shortNameFromLongName = {}
+
+        self._cameraName = cameraName
+        self._radialTransform = radialTransform #[1] should be 1/rad per mm
+
+        if detectorIdFromAbbrevName is None:
+            raise RuntimeError("You must specify detectorIdFromAbbrevName when " \
+                               + "instantiating BasicCameraFactory")
```
timj: Remove backslash and the `+` operator.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,233 @@
...
+        camConfig.plateScale = 1.0/afwGeom.radToArcsec(self._radialTransform[1])
+
+        tConfig = afwGeom.TransformConfig()
+        tConfig.transform.name = 'inverted'
+        radialClass = afwGeom.xyTransformRegistry['radial']
+        tConfig.transform.active.transform.retarget(radialClass)
+        tConfig.transform.active.transform.coeffs = self._radialTransform
+        tmc = afwGeom.TransformMapConfig()
+        tmc.nativeSys = FOCAL_PLANE.getSysName()
+        tmc.transforms = {PUPIL.getSysName():tConfig}
```
timj: This looks a bit cramped. Please add a space after the colon (which is also the PEP8 style).

danielsf:
made all of the style changes requested by Tim Jenness

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,225 @@
...
+        from pupil coordinates (x, y, radians) to focal plane coordinates (x, y, mm)
+        according to the convention
+
+        (x, y)_focal = (x, y)_pupil * sum_i radialTransform[i] * r^i/r
+
+        where r is the magnitude of the point in pupil coordinates.  Note that the [1]
+        element of this list should be 1/plateScale where plateScale is in radians per mm.
+
+        @param [in] cameraName is a string referring to the name of the camera
+        (default 'LSST')
```
SimonKrughoff: Why default this to LSST?  You could make this a positional argument and make people name their cameras.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,225 @@
...
+                 cameraName='LSST'
+                 ):
+
+        """
+        @param [in] detectorLayoutFile is the absolute path to the file
+        listing the layout of all of the chips in the camera's focal plane
+        (see this class' docstring for detailed information on the contents
+        of that file).
+
+        @param [in] detectorIdFromAbbrevName is a (required) method mapping the
```
SimonKrughoff: If this is required, why default it?  Just make it a positional argument.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,225 @@
...
+        @param[in] detTypeMap is a dict mapping the group name of the sensors in
+        detectorLayoutFile to the integers:
+
+        0 = SCIENCE
+        1 = FOCUS
+        2 = GUIDER
+        3 = WAVEFRONT
+
+        if left as None, this will just try to cast the contents of detectorLayoutFile
+        as an int
```
SimonKrughoff: The other obvious option here is to assume all sensors are SCIENCE.  I'm not sure which is better.

python/lsst/afw/cameraGeom/basicCameraFactory.py
```diff
@@ -0,0 +1,225 @@
...
+        else:
+            return self._detTypeMap[typeName]
+
+    def makeDetectorConfigs(self):
+        """
+        Create the detector configs to use in building the Camera
+        """
+        detectorConfigs = []
+        #We know we need to rotate 3 times and also apply the yaw perturbation
+        nQuarter = 1
```
SimonKrughoff: I feel like if this is being motivated by the LSST chips, can we make this a passed parameter, so it doesn't need to be hard coded.

danielsf: I'll be honest: I'm not sure what the purpose of this parameter is (a lot of what I did here was cutting and pasting with minimal hacking to get something whose output I understood).

danielsf: Is it that the LSST configuration assumes a baseline 90 degree rotation in addition to the yaw specified in the configuration file?

Should we just get rid of nQuarter altogether and have the full yaw rotation be specified in the configuration file?

RobertLuptonTheGood: I think I'd rather keep nQuarter as it provides a way for code that crudely assembles focal planes to use a simple rotation not a "proper" one.  We could use `nQuarter = (yaw + 360 + 45)//90`  (+ 360 in case it's negative), but this seems simpler, and then yaw is always a small correction.  But the general yaw solution would work too.

PaulPrice: @SimonKrughoff's original point was about assuming a geometry (`nQuarter = 1`) that is not appropriate for all cameras.

jdswinbank:
Has been Won't Fix-ed.

PR #37, opened 2015-10-07
Tickets/dm 4021

SimonKrughoff:


PaulPrice:
For the first commit, please use the summary line to state what has *changed*, e.g., "remove const keys for unordered_map".  The explanation about `std::unordered_map` forbidding them can go underneath.

PR #40, opened 2015-11-09
Tickets/dm 3105

timj:


python/lsst/afw/image/__init__.py
```diff
@@ -25,3 +25,4 @@
 import numpy
 from .imageLib import *
 from .basicUtils import *
+from .testUtils import *
```
timj: new line needed here.

python/lsst/afw/image/testUtils.py
```diff
@@ -36,125 +41,108 @@ def makeGaussianNoiseMaskedImage(dimensions, sigma, variance=1.0):
...
+    @param[in] msg  exception message prefix; details of the error are appended after ": "
+
+    The mask planes must match exactly. The image and variance planes are nearly equal if all pixels obey:
+        |val1 - val0| <= rtol*|val1| + atol
+    or, for float types, if nan/inf/-inf pixels match.
+
+    @warning the comparison equation is not symmetric, so in rare cases the assertion
+    may give different results depending on which masked image comes first.
     
     Return a string describing the error if the images differ significantly, an empty string otherwise
```
timj: This is now an `assert` so it doesn't return anything any more. I think that's a left over comment from the previous implementation.

python/lsst/afw/image/testUtils.py
```diff
@@ -36,125 +41,108 @@ def makeGaussianNoiseMaskedImage(dimensions, sigma, variance=1.0):
...
     
-    rtol and atol are positive, typically very small numbers.
-    The relative difference (rtol * abs(b)) and the absolute difference "atol" are added together
-    to compare against the absolute difference between "a" and "b".
+    @param[in] testCase  unittest.TestCase instance the test is part of;
+                        an object supporting one method: fail(self, msgStr)
+    @param[in] maskedImage0  masked image 0 (an lsst.afw.image.MaskedImage or
+        collection of three transposed numpy arrays: image, mask, variance)
+    @param[in] maskedImage1  masked image 1 (an lsst.afw.image.MaskedImage or
+        collection of three transposed numpy arrays: image, mask, variance)
```
timj: Does the transpose requirement matter when we are comparing pixel for pixel? It seems that what matters is that `maskedImage0` and `maskedImage1` have the same order.

python/lsst/afw/image/testUtils.py
```diff
@@ -36,125 +41,108 @@ def makeGaussianNoiseMaskedImage(dimensions, sigma, variance=1.0):
...
+    may give different results depending on which masked image comes first.
     
     Return a string describing the error if the images differ significantly, an empty string otherwise
     """
-    retStrs = []
+    maskedImageArrSet0 = maskedImage0.getArrays() if hasattr(maskedImage0, "getArrays") else maskedImage0
+    maskedImageArrSet1 = maskedImage1.getArrays() if hasattr(maskedImage1, "getArrays") else maskedImage1
+
+    try:
+        assert len(maskedImageArrSet0) == 3
```
timj: These sanity checks don't seem to be very bullet proof. If I pass in any sequence with 3 elements in it (including a 3 character long string) I will get through all this code.

If we really want to check there are 2 x 3 transposed numpy arrays that are the same shape then we should check for that. If the intention is purely to ensure we have three whatevers to pass to `imagesDiffer` then we may want to note that here and also modify the error message. You aren't checking whether it's a supported type, only that you have 3 of them.

python/lsst/afw/image/testUtils.py
```diff
@@ -36,125 +41,108 @@ def makeGaussianNoiseMaskedImage(dimensions, sigma, variance=1.0):
...
+        |val1 - val0| <= rtol*|val1| + atol
+    or, for float types, if nan/inf/-inf pixels match.
+
+    @warning the comparison equation is not symmetric, so in rare cases the assertion
+    may give different results depending on which masked image comes first.
     
     Return a string describing the error if the images differ significantly, an empty string otherwise
     """
-    retStrs = []
+    maskedImageArrSet0 = maskedImage0.getArrays() if hasattr(maskedImage0, "getArrays") else maskedImage0
```
timj: Using "set" in the variable name might result in someone expecting these to be Python sets rather than tuples.

python/lsst/afw/image/testUtils.py
```diff
@@ -162,12 +150,115 @@ def maskedImagesDiffer(maskedImageArrSet1, maskedImageArrSet2,
...
+    @return a string which is non-empty if the images differ
+    """
+    errStrList = []
+    imageArr0   = image0.getArray()   if hasattr(image0,   "getArray") else image0
+    imageArr1   = image1.getArray()   if hasattr(image1,   "getArray") else image1
+    skipMaskArr = skipMask.getArray() if hasattr(skipMask, "getArray") else skipMask
+
+    # sanity check; this is useful because the error message from np.allclose
+    # is opaque for object arrays
+    try:
```
timj: Consider a for loop here to help with code repetition.

python/lsst/afw/image/testUtils.py
```diff
@@ -162,12 +150,115 @@ def maskedImagesDiffer(maskedImageArrSet1, maskedImageArrSet2,
...
+        assert imageArr0.dtype.kind != "O"
```
timj: Are we saying the "O" is the only type we can't use? Are "c", "S", "U" and "V" all okay?

python/lsst/afw/image/testUtils.py
```diff
@@ -162,12 +150,115 @@ def maskedImagesDiffer(maskedImageArrSet1, maskedImageArrSet2,
...
+    if skipMaskArr is not None:
+        try:
+            assert skipMaskArr.dtype.kind != "O"
+        except Exception:
+            raise TypeError("skipMask=%r is not a supported type" % (skipMask,))
+
+    if skipMaskArr is not None:
+        skipMaskArr = np.array(skipMaskArr, dtype=bool)
+        maskedArr0 = np.ma.array(imageArr0, copy=False, mask = skipMaskArr)
+        maskedArr1 = np.ma.array(imageArr1, copy=False, mask = skipMaskArr)
```
timj: spaces around `=` should be removed here.

python/lsst/afw/image/testUtils.py
```diff
@@ -162,12 +150,115 @@ def maskedImagesDiffer(maskedImageArrSet1, maskedImageArrSet2,
...
+        if np.any(neginf0 != neginf1):
+            errStrList.append("-infs differ")
+
+        valSkipMaskArr = nan0 | nan1 | posinf0 | posinf1 | neginf0 | neginf1
+        if skipMaskArr is not None:
+            valSkipMaskArr |= skipMaskArr
+
+    # compare values that should be comparable (are finite and not masked)
+    valMaskedArr1 = np.ma.array(imageArr0, copy=False, mask = valSkipMaskArr)
+    valMaskedArr2 = np.ma.array(imageArr1, copy=False, mask = valSkipMaskArr)
```
timj: Space around `=`

tests/testTestMethods.py
```diff
@@ -211,6 +213,218 @@ def testAssertWcssNearlyEqualOverBBox(self):
...
+            with self.assertRaises(TypeError):
+                self.assertMaskedImagesNearlyEqual(invalidType, mi)
+            with self.assertRaises(TypeError):
+                self.assertMaskedImagesNearlyEqual(mi, mi, skipMask=invalidType)
+
+            with self.assertRaises(TypeError):
+                self.assertMaskedImagesNearlyEqual(mi.getImage(), mi.getImage())
+
+
+def makeMaskedImage(width, height, imgClass=afwImage.MaskedImageF):
```
timj: This name is somewhat confusing given that it is making a ramp image, much like `makeRampMaskedImage`, but with additional nan/inf/-inf values. Would be more obvious if it was called something like `makeRampMaskedImageWithNans` :smile:

python/lsst/afw/image/__init__.py
r-owen: OK

python/lsst/afw/image/testUtils.py
r-owen: Good catch!

python/lsst/afw/image/testUtils.py
r-owen: I'm afraid I don't understand the question. For borderline cases swapping image0 and image1 may make the difference between the test passing and failing. This is an unfortunate side effect of the way numpy.allclose works. Fixing it is possible, but I don't think it is worth the work and it the current behavior is predictable.

timj: I'm trying to ask whether it really matters whether the arrays are in an explicit row ordering scheme (ie transposed) or whether it would work if neither of the numpy arrays were transposed (because if they are in the same shape and in C order or the same shape and Fortran order, then `numpy.allclose` should still work fine on them). It's a minor pedantic point from reading the doc string: is transpose important or is all that matters is that both of the input arrays/images agree.

python/lsst/afw/image/testUtils.py
r-owen: I agree. I'll beef it up and improve the tests.

python/lsst/afw/image/testUtils.py
r-owen: Good point. I'll switch to "List".

python/lsst/afw/image/testUtils.py
r-owen: I am creating new variables and I don't know a really clear way to do that programmatically. I fear the result would be less obvious than what I have now. But I am open to suggestions.

python/lsst/afw/image/testUtils.py
r-owen: Good point. As part of cleaning up the tests I changed it to allow any scalar type. There may be a cleaner way to handle this, but I am just testing for the valid scalar types: ("b", "i", "u", "f", "c")

python/lsst/afw/image/testUtils.py
r-owen: I see. The point is that numpy array axes are transposed relative to data in Image-like classes. Thus if one is testing a pair of 2x3 Images then one must provide a 3x2 skipMask array.

Suggestions on clarifying the doc string would be most welcome.

timj: Yes, and if the two images and the mask are all un-transposed and the same size then it will work. In the special case of 3 numpy arrays we only care that they are the same shape.

timj: I should add that I'm not going to hold up the merge on this documentation technicality.

r-owen: I am adding this sort of warning to each assert method, in hopes that it will clarify:

    @warning the axes of numpy arrays are transposed with respect to Image and Mask data.
    Thus for example if image0 and image1 are both lsst.afw.image.ImageD with dimensions (2, 3)
    and skipMask is a numpy array, then skipMask must have shape (3, 2).

And here is an example string for one of the inputs:

    @param[in] image0  image 0, an lsst.afw.image.Image, lsst.afw.image.Mask,
        or transposed numpy array (see warning)

It's not perfect (in particular there are two warnings), but I hope it will be sufficiently clear.

PR #41, opened 2015-11-13
FootprintMerge: fix bug when identifying existing peaks in a merge.

rearmstr:
When merging two separate footprints from the same catalog because they
both overlap an existing merged footprint, we were not propagating the flags of
which peaks were detected in which filters.


rearmstr:
The code looks fine, but the commit message I wrote is slightly confusing.  I would reword like this:

When merging a catalog with two separate footprints that both overlap an existing merged footprint, we were not correctly propagating the filter peak flags.

PR #42, opened 2015-11-13
SourceCatalog: test assumption on sort order for getChildren

rearmstr:
SourceCatalog.getChildren assumes that the catalog is sorted by
parent, but did not check this assumption, which can lead to
misleading results if the assumption is not true.  Test the
assumption and raise an AssertionError if it fails.


jdswinbank:
Long merged.

PR #43, opened 2015-12-12
DM-4162: XYTransform interface for warping

TallJimbo:


include/lsst/afw/coord/Coord.h
```diff
@@ -131,7 +131,7 @@ class Coord {
     void rotate(Coord const &axis, lsst::afw::geom::Angle const theta);
     lsst::afw::geom::Angle offset(lsst::afw::geom::Angle const phi, lsst::afw::geom::Angle const arcLen);
     
-    Coord::Ptr convert(CoordSystem system) const;
+    Coord::Ptr convert(CoordSystem system, double epoch=2000) const;
```
TallJimbo: Since you're touching this line, might as well switch from `Coord::Ptr` to `PTR(Coord)` at the same time.

TallJimbo:
Given that swapped sky axes and warping without interpolation are rare, it's not obvious to me that a 10% speedup under these conditions is worth an extra bool member in every `Wcs` instance.  Neither is a terribly big deal, though.

src/math/warpExposure.cc
```diff
@@ -595,7 +596,8 @@ int afwMath::warpImage(
     typename DestImageT::SinglePixel padValue
 ) {
     afwGeom::Point2D const destXY0(destImage.getXY0());
-    afwMath::detail::WcsPositionFunctor const computeSrcPos(destXY0, destWcs, srcWcs);
+    afwImage::XYTransformFromWcsPair xyTransform{destWcs.clone(), srcWcs.clone()};
```
TallJimbo: It's disappointing that we have to clone the `Wcs`s here; `Wcs`s are sufficiently complex that I don't think we want to assume this is an efficient operation.  I think I'd prefer to change the arguments to `PTR(Wcs const)` so we can pass them through directly.  I'm ambivalent as to whether we should add a new signature and leave this deprecated for a while before removing it, or just change it now and deal with code downstream immediately.

include/lsst/afw/coord/Coord.h
r-owen: This is a good point, but I would prefer consistency. I can change all instances now or fix it as part of adding getCoordSystem to Coord. I confess I'm inclined to do it as part of the latter work, since it involves more work on Coord anyway.

src/math/warpExposure.cc
r-owen: I agree that it is disappointing, but I want to push back here. Warping is already slow, and I doubt that the expense of cloning the WCS is a significant factor. I propose we leave it until we get warping fast enough that the slowdown is a significant concern.

include/lsst/afw/coord/Coord.h
r-owen: I don't see a way to comment directly after your remark about the 10% speedup below, so I'll do it here. I am not sure what you mean by "swapped sky axes"; there is no swapping involved in this speedup; it simply handles the overwhelmingly most common case that both WCS will have the same sky system. (It is fortunate it is so common since it was the ONLY case that was warped correctly!). I agree it may not be common to warp without interpolation, but the speedup makes interpolation a bit less crucial. Since you are willing to leave it in, I'm going to do so.

PR #45, opened 2015-12-18
DM-4133  Change type of LTV1/2 from int to float

hchiang2:


tests/test1079.py
```diff
@@ -244,11 +245,15 @@ def testFitsHeader(self):
             subImg.writeFits(outFile)
             hdr = afwImage.readMetadata(outFile)
             
-            self.assertTrue( hdr.exists("LTV1"), "LTV1 not saved to fits header")
-            self.assertTrue( hdr.exists("LTV2"), "LTV2 not saved to fits header")
-            self.assertEqual(hdr.get("LTV1"), -1*x0, "LTV1 has wrong value")
-            self.assertEqual(hdr.get("LTV2"), -1*y0, "LTV1 has wrong value")
-
+            def checkLtvHeader(hdr, name, value):
```
hchiang2: I was wondering why define checkLtvHeader inside testFitsHeader? Would it be more clear to have a _checkLtvHeader outside?

jdswinbank: I thought this was clearer, since it keeps the function definition close to the point of use, and it doesn't pollute the class namespace. I'm happy to move it if you prefer, though.

hchiang2: I guess it's just a style preference.  I don't see our coding standard suggests one way or another (or does it?) So don't worry then!

jdswinbank:
This has been merged; closing PR.

PR #46, opened 2015-12-21
Improve mask tests with full array comparisons

parejkoj:
Improvements in tests, so they actually compare full arrays instead of single values, discovered while digging into cfitsio upgrade issues.

My one question is the "should only change dmask" on line 186: I think that was an error in the original (certainly, the original test suggests as such), but someone else should check my logic.

### DIFF
--- a/tests/mask.py
+++ b/tests/mask.py
@@ -37,7 +37,7 @@ import os.path
 
 import sys
 import unittest
-import numpy
+import numpy as np
 
 import lsst.utils
 import lsst.utils.tests as utilsTests
@@ -47,7 +47,7 @@ import lsst.afw.image as afwImage
 import lsst.afw.geom as afwGeom
 import lsst.afw.display.ds9 as ds9
 
-numpy.random.seed(1)
+np.random.seed(1)
 
 try:
     type(display)
@@ -99,10 +99,7 @@ class MaskTestCase(unittest.TestCase):
         except Exception:
             self.maskFile = None
         else:
-            if True:
-                self.maskFile = os.path.join(dataDir, "small_MI.fits")
-            else:
-                self.maskFile = os.path.join(dataDir, "871034p_1_MI.fits")
+            self.maskFile = os.path.join(dataDir, "small_MI.fits")
 
     def tearDown(self):
         del self.mask1
@@ -120,11 +117,11 @@ class MaskTestCase(unittest.TestCase):
         self.assertEqual(array1.shape[0], image2.getHeight())
         self.assertEqual(array1.shape[1], image2.getWidth())
         self.assertEqual(type(image3), afwImage.MaskU)
-        array1[:,:] = numpy.random.uniform(low=0, high=10, size=array1.shape)
         for j in range(image1.getHeight()):
             for i in range(image1.getWidth()):
                 self.assertEqual(image1.get(i, j), array1[j, i])
                 self.assertEqual(image2.get(i, j), array1[j, i])
+        array1[:,:] = np.random.uniform(low=0, high=10, size=array1.shape)
 
     def testInitializeMasks(self):
         val = 0x1234
@@ -198,7 +195,7 @@ class MaskTestCase(unittest.TestCase):
         
         del smask
         del mask2
-        
+
         self.assertEqual(self.mask1.get(0, 0), self.val1)
         self.assertEqual(self.mask1.get(1, 1), 666)
         self.assertEqual(self.mask1.get(4, 1), self.val1)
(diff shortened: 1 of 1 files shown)


timj:
Can the commit message also indicate that you had removed the "dead" code?

tests/mask.py
```diff
@@ -99,10 +99,10 @@ def setUp(self):
         except Exception:
             self.maskFile = None
         else:
-            if True:
-                self.maskFile = os.path.join(dataDir, "small_MI.fits")
-            else:
-                self.maskFile = os.path.join(dataDir, "871034p_1_MI.fits")
+            self.maskFile = os.path.join(dataDir, "small_MI.fits")
+            # Above mask looks like this, possibly left-shifted by some number of mask planes.
```
r-owen: I don't understand this comment (what does "looks like this" mean?)

More generally I'm not thrilled with silently changing behavior based on whether afwdata is setup. In my opinion a test should either need it or not. If a test needs afwdata then it should be explicitly skipped if afwdata is missing, so it's obvious to users that it was skipped.

tests/mask.py
```diff
@@ -37,7 +37,7 @@
 
 import sys
 import unittest
-import numpy
+import numpy as np
```
r-owen: At risk of sounding too pedantic, I suggest making no-op cleanups like this (and cleaning up whitespace and indentation) as a separate commit from changes that change the behavior. In this case it may not be worth going back (though it's certainly doable).

tests/mask.py
```diff
@@ -138,23 +135,28 @@ def testOrMasks(self):
         self.mask2 |= self.mask1
         self.mask1 |= self.val2
         
-        self.assertEqual(self.mask1.get(0, 0), self.val1 | self.val2)
-        self.assertEqual(self.mask2.get(0, 0), self.val1 | self.val2)
-    
+        self.assertMasksEqual(self.mask1, self.mask2)
+        expect = np.zeros_like(self.mask1.getArray())
+        expect |= self.val2 | self.val1
```
r-owen: At risk of being too picky, I think the following would be slightly easier to follow because the user doesn't have to know the initial value (which is zero, because I looked, but still...)
expect[:] = self.val2 | self.val1

tests/mask.py
```diff
@@ -226,13 +228,7 @@ def testReadFitsConform(self):
...
         mask = afwImage.MaskU(self.maskFile, hdu, None, afwGeom.Box2I(), afwImage.LOCAL, True)
 
-        if False:
-            import lsst.afw.display.ds9 as ds9
-            ds9.mtv(mask)
-
-        self.assertEqual(mask.get(32, 1), 0)
-        self.assertEqual(mask.get(50, 50), 0)
-        self.assertEqual(mask.get(0, 0), 1)
+        self.assertMasksEqual(mask,self.expect)
```
r-owen: Please put a space after comma (here and several other places)

tests/mask.py
```diff
@@ -183,9 +185,11 @@ def testCopyConstructors(self):
         dmask = afwImage.MaskU(self.mask1, True) # deep copy
         smask = afwImage.MaskU(self.mask1) # shallow copy
         
-        self.mask1 |= 32767             # should only change dmask
-        self.assertEqual(dmask.get(0, 0), self.val1)
-        self.assertEqual(smask.get(0, 0), self.val1 | 32767)
+        self.mask1 |= 32767             # should only change smask
+        temp = np.zeros_like(self.mask1.getArray()) | self.val1
```
r-owen: Good catch!

tests/mask.py
timj: I think there is a ticket already for fixing the skipping in afw tests. We need it for the switch to `py.test`.

parejkoj: I'll not change to not use proper skipping, but I can put in a "#skipif" note or somesuch, if there's a common way to refer to them.

tests/mask.py
parejkoj: That's fair. It was a small enough change that I figured it wouldn't matter.

timj: They always matter :wink: More seriously, cleanup patches should always be distinct from functionality patches.

RobertLuptonTheGood: There was once a discussion about consistently importing numpy as np (and matplotlib.pyplot at plt).  Should we restart it on c.l.o?

tests/mask.py
timj: Ticket was https://jira.lsstcorp.org/browse/DM-609 and there are some duplicates as people keep wanting the tests to run and skip rather than skipping the entire test suite.

tests/mask.py
timj: https://jira.lsstcorp.org/browse/RFC-34

parejkoj:
Closing, as the changes were rebased after review and will be merged separately.

PR #47, opened 2015-12-23
Review for DM-4729: Add functions to generate 'unpacked matches' in a Catalog

jdswinbank:


python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
+# 
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
```
jdswinbank: 2016 now! (My fault for the slow review)

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+# You should have received a copy of the LSST License Statement and 
+# the GNU General Public License along with this program.  If not, 
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import
+
+import os.path
+
+from .tableLib import BaseCatalog, SimpleCatalog, SourceCatalog, SimpleTable, SourceTable, \
+                      Schema, ReferenceMatch
```
jdswinbank: The coding standards tell us we should prefer parens to backslashes for continuations.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+from lsst.utils import getPackageDir
+
+__all__ = ["copySchema", "copyCatalog", "matchesToCatalog", "matchesFromCatalog"]
+
+def copySchema(schema, target, targetPrefix=None, sourcePrefix=None):
+    """Return a deep copy the provided schema
```
jdswinbank: "... deep copy *of* the ..."

Also, could you please clarify in the docstring what the relationship between `target` and the return value is? Can the caller ignore the return value and assume that whatever they passed as `target` is modified by this function?

Also also: would it be possible to make this function and `copyCatalog`, below, considerably simpler & less verbose by using a `SchemaMapper`? I spent a few minutes playing around and came up with https://gist.github.com/jdswinbank/a2418d7c8a00c9c1f6ea, which provides replacements for these functions which are quite a lot shorter. They pass the `SourceMatchTestCase`, below, but haven't otherwise been sanity checked, so I might've missed some subtlety and could probably simplify further with a bit more effort.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+        if typeStr in ("ArrayF", "ArrayD", "ArrayI", "CovF", "CovD"):
+            fieldSize = field.getSize()
+        else:
+            fieldSize = None
+
+        target.addField((targetPrefix if targetPrefix is not None else "") + keyNameFix,
+                        type=typeStr, doc=fieldDoc, units=fieldUnits, size=fieldSize)
+    return target
+
+def copyCatalog(catalog, target, sourceSchema=None, targetPrefix=None, sourcePrefix=None):
```
jdswinbank: Similar comments apply here as to `copySchema()`, above.

* Please clarify whether callers can expect the `target` they supply to be modified or if they need to catch the return value;
* This could could be simplified quite a bit by using a `SchemaMapper`;
* Even if you don't want to rewrite to use the `SchemaMapper` approach, there's some duplication between these functions in the handling of `sourcePrefix` that you could consider trying to eliminate.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+                rTo.set(kTo, rFrom.get(kFrom))
+            except:
+                print "Error setting: %s %s %s %s" % (type(rFrom), type(rTo), type(kFrom), type(kTo))
+
+    return target
+
+def matchesToCatalog(matches, matchMeta):
+    """Denormalise matches into a Catalog of "unpacked matches"
+
+    \param[in] matches    unpacked matches
```
jdswinbank: Might be helpful to expand the documentation a little to describe exactly what's meant by "unpacked matches" and what types you expect the input to have. This is jargon we use elsewhere, but I'm not aware of a clear exposition of it; the newcomer (by which I mean me!) is left poking around at fragmentary text in Doxygen to figure out what it means.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+    \param[in] catalog           catalog of matches.  Must have schema where reference entries are
+                                 prefixed with "ref_" and source entries are prefixed with "src_"
+    \param[in] sourceSlotConfig  configuration for source slots (optional)
+    \param[in] prefix            prefix for slot config setup (optional)
+
+    \returns   lsst.afw.table.ReferenceMatch of matches
+    """
+    if catalog is None:
+        # There are none
+        return []
```
jdswinbank: Is this check necessary? Since `catalog` isn't an optional argument, the caller would have to explicitly call `matchesFromCatalog(None)` -- surely the responsibility for not doing that lies with the caller rather than the callee?

python/lsst/afw/table/catalogMatches.py
```diff
@@ -0,0 +1,164 @@
...
+    refSchema = copySchema(catalog.schema, SimpleTable.makeMinimalSchema(), sourcePrefix="ref_")
+    refCatalog = SimpleCatalog(refSchema)
+    copyCatalog(catalog, refCatalog, sourcePrefix="ref_")
+
+    srcSchema = copySchema(catalog.schema, SourceTable.makeMinimalSchema(), sourcePrefix="src_")
+    srcCatalog = SourceCatalog(srcSchema)
+    copyCatalog(catalog, srcCatalog, sourcePrefix="src_")
+
+    if sourceSlotConfig is not None:
+        sourceSlotConfig.setupTable(srcCatalog.table, prefix=prefix)
```
jdswinbank: What is `sourcesSlotConfig`? I searched both [Doxygen](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/search.php?query=setupTable) and [GitHub](https://github.com/search?utf8=✓&q=user%3Alsst+setupTable) for a `setupTable()` method, and came up blank.

python/lsst/afw/table/catalogMatches.py
laurenam: Done

python/lsst/afw/table/catalogMatches.py
laurenam: Done

python/lsst/afw/table/catalogMatches.py
laurenam: Good points.  I've indicated that the target is modified in place in the docstring and removed the return statement.  On that topic, do we have a preference on how to "return" from functions with no return value?  I've seen no `return` statements, naked `return`, explicit `return None`.  Is there any reason to prefer one over the others?

Also, thanks for pointing out the benefits of using the `SchemaMapper` here (and for writing the code to do it!  This will serve as a nice guideline to clean up the mappings I'm working on to map between HSC vs. LSST schemas for run comparisons).  The only real difference I see between your versions and the old ones is the handling of "existing" keys in the target schema.  In the latter, if a key was already present in the target schema it was silently skipped over.  Now, I believe a name collision will cause the SchemaMapper to throw (since we are using the default `doReplace=false`).  I actually think this is preferred as the user should probably not be trying to copy between two schemas that have some names in common (without having accommodated that with the optional prefix parameters).

python/lsst/afw/table/catalogMatches.py
laurenam: Adopted your `SchemaMapper` solution.

python/lsst/afw/table/catalogMatches.py
laurenam: I expanded the docstrings a bit.  Let me know if you think it's sufficient.

python/lsst/afw/table/catalogMatches.py
laurenam: Removed

python/lsst/afw/table/catalogMatches.py
laurenam: Good catch!  This is defined in HSC at: https://github.com/HyperSuprime-Cam/meas_algorithms/blob/master/python/lsst/meas/algorithms/measurement.py#L36

The equivalent in LSST is (where `setupTable()` --> `setupSchema()`): https://github.com/lsst/meas_base/blob/master/python/lsst/meas/base/baseMeasurement.py#L145

I've updated to the latter.

jdswinbank:
Has been merged.

PR #48, opened 2016-01-05
Add utility to display Exposure + source matches.

jdswinbank:


python/lsst/afw/display/utils.py
```diff
@@ -370,3 +370,31 @@ def drawCoaddInputs(exposure, frame=None, ctype=None, bin=1, display="deferToFra
...
+        for source in sources:
+            x, y = source.getX() - x0, source.getY() - y0
+            disp.dot('o', x, y, ctype="green", size=4)
+
+        for first, second, _ in matches:
+            catPos = wcs.skyToPixel(first.getCoord())
+            x1, y1 = catPos.getX() - x0, catPos.getY() - y0
+            disp.dot("+", x1, y1, ctype="yellow", size=8)
+            x2, y2 = second.getX() - x0, second.getY() - y0
+            disp.dot("x", x2, y2, ctype="red", size=8)
```
RobertLuptonTheGood: There's an awful lot of hard-coded decisions in here.  Do we need this routine in afw?  If we do, it should be more configurable (colours, symbols, sizes)

python/lsst/afw/display/utils.py
```diff
@@ -370,3 +370,31 @@ def drawCoaddInputs(exposure, frame=None, ctype=None, bin=1, display="deferToFra
...
+    """
+    disp = _getDisplayFromDisplayOrFrame(None, frame)
+    disp.mtv(exposure)
+    x0, y0 = exposure.getMaskedImage().getX0(), exposure.getMaskedImage().getY0()
+    wcs = exposure.getWcs()
+
+    with disp.Buffering():
+        for source in sources:
+            x, y = source.getX() - x0, source.getY() - y0
+            disp.dot('o', x, y, ctype="green", size=4)
```
RobertLuptonTheGood: You shouldn't need to deal with XY0 yourself (it's handled by the origin=XXX keyword).   Did you check that you needed to?

python/lsst/afw/display/utils.py
RobertLuptonTheGood: I think that this is replacing pipe_base's display code, but I'm not sure that pushing this routine down to afw is the correct solution

jdswinbank: I put this in `afw` since it is called by tasks from a variety of packages and it matches the contents of the rest of this file reasonably well. I'm happy to move it elsewhere if you have a preferred location.

No existing code would make use of configuration options. I'd prefer not to add them, as it's complexity with no tangible benefit. They could easily be added later if a requirement arose.

RobertLuptonTheGood: Why not pipe_base?   It's very much an inflexible pipeline-debugging utility as written.

jdswinbank: You're correct that this is lifted with minimal tweaks from `pipe_base`, where it's slated for removal. I have no objection with leaving it as a free function in that package, except, as I say, it seems broadly comparable to the other functionality in this file.

jdswinbank: I've no objection to it going in `pipe_base`. I'll shift it there.

python/lsst/afw/display/utils.py
jdswinbank: You're right; thanks.

python/lsst/afw/display/utils.py
TallJimbo: Sorry I'm coming late to this, but `pipe_tasks` (or meas_algorithms or even meas_astrom, if this is focused on mathces) seems much more relevant than `pipe_base`.  `pipe_base` is just plumbing for `Task` and `CmdLineTask` with very little that's astronomy specific, and this has nothing to do with `Task`.

jdswinbank: It can't go into `pipe_tasks`, because it's used by `meas_astrom` and the former depends on the latter.

I'm not clear what the qualifications for going into `meas_algorithms` are -- my impression is that repository is basically a grab back of assorted bits & pieces -- but this isn't an "astronomical measurement algorithm" so I don't think it fits there.

I'm agnostic as to whether it goes in `pipe_base` or `meas_astrom`. I will choose one based on whoever shouts loudest, but then specifically ask the code reviewer to weigh in with their opinion before it gets merged.

ktlim: It seems that this code is having a hard time fitting into an existing package.  It also seems like it may hopefully be the vanguard of an increasing number of debugging utilities and tools.  If so, perhaps it really belongs in a new package?

RobertLuptonTheGood: That was my thought; this is the sort of thing that my de-facto private analysis package supported.  But we need it a long way down the dependencies, and debugging utilities naturally end up at the top of the stack.

jdswinbank: It looks as though `meas_astrom`'s existing `displayAstrometry()` function produces results which are broadly equivalent to the functionality I was preserving from `pipe_base`. I think the solution here is to simply drop this code altogether and use that instead, so this problem vanishes.

That doesn't avoid the larger issue of where to put debugging utilities in future, of course, but it means we don't need to wait for a solution to that to proceed here.

PR #50, opened 2016-01-21
Review for DM-4881 (Statistics: assert that images have the same size)

jdswinbank:
If the image and mask don't have the same size, we can get a NAN
result if all pixels in the lower-left are masked. This patch adds
a check on the size of all images to catch accidental user errors.

### DIFF
--- a/include/lsst/afw/math/Statistics.h
+++ b/include/lsst/afw/math/Statistics.h
@@ -416,6 +416,7 @@ public:
     x_iterator row_end(int) const { return _v.end(); }
     int getWidth() const { return _v.size(); }
     int getHeight() const { return 1; }
+    afw::geom::Extent2I getDimensions() const { return afw::geom::Extent2I(getWidth(), getHeight()); }
     
     bool empty() const { return _v.empty(); }
 private:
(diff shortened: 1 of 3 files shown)


rearmstr:
We are trying to keep the HSC commit unchanged correct?  I'm not sure what the "lower-left" is referring to in the commit message.  The NAN is coming from the fact that all the corresponding pixels in the mask are set to be ignored, correct?

jdswinbank:
Thanks Bob! I don't think keeping the HSC commit unchanged is an intrinsic good -- if we can improve it by changing it, we should do so. Obviously, we don't want to waste time changing it unless it really is an improvement though.

In this case, I agree that a little rewording could make it clearer -- I'll take care of it.

laurenam:
This is my bad as I told Bob we generally try to keep the commit history the same.  Perhaps we should quickly go over our HSC port "best practices" at today's standup to make sure we're all on the same page?

timj:
My naive assumption would be to cherry-pick HSC commits as is and then commit LSST local fixes as separate commits.

jdswinbank:
The model we've adopted is broadly comparable to our code review process: when modest changes or cleaning up is needed, they are squashed into the original commit; when new features or significant re-writing is merited, that's a separate commit. This has been working well.

PR #57, opened 2016-04-25
Skip tests if afwdata is missing

timj:


tests/background.py
```diff
@@ -884,8 +888,8 @@ def suite():
 def run(shouldExit = False):
     """Run the tests
     """
-    if AfwdataDir is None:
-        print("Warning: afwdata is not setup so some tests will be skipped")
+    #if AfwdataDir is None:
+    #    print("Warning: afwdata is not setup so some tests will be skipped")
```
timj: Remove this commented out code.

tests/convolve.py
```diff
@@ -172,6 +177,7 @@ def sameMaskPlaneDicts(maskedImageA, maskedImageB):
     return True
 
 class ConvolveTestCase(utilsTests.TestCase):
+    @unittest.skipIf(dataDir is None, "afwdata not setup")
```
timj: I would expect that is setUp is skipped that all the tests in the class would be skipped so there would be no need to try to skip other functions explicitly.

tests/exposure.py
```diff
@@ -56,20 +56,24 @@
...
-InputMaskedImageName = "871034p_1_MI.fits"
-InputMaskedImageNameSmall = "small_MI.fits"
-InputImageNameSmall = "small"
-OutputMaskedImageName = "871034p_1_MInew.fits"
-
-currDir = os.path.abspath(os.path.dirname(__file__))
-inFilePath = os.path.join(dataDir, InputMaskedImageName)
-inFilePathSmall = os.path.join(dataDir, InputMaskedImageNameSmall)
-inFilePathSmallImage = os.path.join(dataDir, InputImageNameSmall)
+try:
```
timj: I would prefer a smaller `try` block with an `else` clause. If you do that (and elsewhere) you could probably avoid use `pex_exceptions` as you'd only be wanting to catch any exception associated with trying to get the `dataDir` rather than the possibility of other parts of the `try` block throwing exceptions.

tests/image.py
```diff
@@ -545,7 +549,7 @@ def setUp(self):
 
         try:
             dataDir = lsst.utils.getPackageDir("afwdata")
-        except Exception:
+        except pexExcept.NotFoundError:
```
timj: it's probably okay to leave that as `Exception`.

tests/image.py
```diff
@@ -48,12 +48,16 @@
...
 import lsst.afw.geom as afwGeom
 import lsst.afw.display.ds9 as ds9
-
-import numpy
+import lsst.pex.exceptions as pexExcept
 
 numpy.random.seed(1)
 
 try:
+    afwdataDir = lsst.utils.getPackageDir("afwdata")
```
timj: why does this test file have `afwdataDir` and `dataDir`?

tests/imageIo1.py
```diff
@@ -27,6 +27,7 @@
 Test cases to test image I/O
 """
 import os.path
+import pdb
```
timj: why add this? Note that you can run tests with `py.test --pdb` if you want to go into the debugger if a test fails.

tests/mask.py
```diff
@@ -54,6 +55,11 @@
 except NameError:
     display = False
 
+try:
+    afwdataDir = lsst.utils.getPackageDir("afwdata")
+except pexExcept.NotFoundError:
+    afwdataDir = None
```
timj: This test file has `afwdatadir` and `datadir` doing the same job.

tests/maskedImage1.cc
```diff
@@ -37,9 +38,26 @@ namespace image = lsst::afw::image;
...
 int test(int argc, char**argv) {
-    if (argc < 5) {
-       cerr << "Usage: inputBaseName1 inputBaseName2 outputBaseName1  outputBaseName2" << endl;
-       return EXIT_FAILURE;
+    
+    string dataDir, inImagePath1, inImagePath2, outImagePath1, outImagePath2;
+    if (argc < 2) {
+        try {
+            dataDir = lsst::utils::getPackageDir("afwdata");
+            inImagePath1 = dataDir + "/data/871034p_1_MI.fits";
```
timj: in theory you could use `boost::filesystem` for this but probably not worth it.

tests/statistics.cc
```diff
@@ -245,7 +246,7 @@ BOOST_AUTO_TEST_CASE(StatisticsTestAllNanButOne) { /* parasoft-suppress  LsstDm-
 
 }
 
-BOOST_AUTO_TEST_CASE(StatisticsTestImages) { /* parasoft-suppress  LsstDm-3-2a LsstDm-3-4a LsstDm-4-6 LsstDm-5-25 "Boost non-Std" */
+BOOST_AUTO_TEST_CASE(StatisticsTestImages, * utf::disabled()) { /* parasoft-suppress  LsstDm-3-2a LsstDm-3-4a LsstDm-4-6 LsstDm-5-25 "Boost non-Std" */
```
timj: It's not entirely obvious to me why this test has been disabled.

tests/warpExposure.py
```diff
@@ -57,15 +57,18 @@
...
-afwDataDir = lsst.utils.getPackageDir("afwdata")
-dataDir = os.path.join(afwDataDir, "data")
-
-originalExposureName = "medexp.fits"
-originalExposurePath = os.path.join(dataDir, originalExposureName)
-subExposureName = "medsub.fits"
-subExposurePath = os.path.join(dataDir, originalExposureName)
-originalFullExposureName = os.path.join("CFHT", "D4", "cal-53535-i-797722_1.fits")
-originalFullExposurePath = os.path.join(dataDir, originalFullExposureName)
+try:
```
timj: try/except/else preferred.

tests/wcs1.py
```diff
@@ -479,13 +491,15 @@ def tearDown(self):
         del self.wcs
         del self.metadata
 
+    @unittest.skipIf(afwdataDir is None, "afwdata not setup")
```
timj: Again, is this needed if the `setUp` is already being skipped?

tests/maskedImage1.cc
```diff
@@ -37,9 +38,26 @@ namespace image = lsst::afw::image;
...
+            inImagePath2 = dataDir + "/data/871034p_1_MI.fits"; // afw/tests/SConscript passes the same file twice in the previous avatar.
+            outImagePath1 = "tests/file:maskedImage1_output_1.fits";
+            outImagePath2 = "tests/file:maskedImage1_output_2.fits";
+        } catch (lsst::pex::exceptions::NotFoundError) {
+            cerr << "Usage: convolveGPU [fitsFile]" << endl;
```
timj: This is the wrong message.

tests/background.cc
```diff
@@ -80,7 +81,7 @@ BOOST_AUTO_TEST_CASE(BackgroundBasic) { /* parasoft-suppress  LsstDm-3-2a LsstDm
 
 }
 
-BOOST_AUTO_TEST_CASE(BackgroundTestImages) { /* parasoft-suppress  LsstDm-3-2a LsstDm-3-4a LsstDm-4-6 LsstDm-5-25 "Boost non-Std" */
+BOOST_AUTO_TEST_CASE(BackgroundTestImages, * utf::disabled()) { /* parasoft-suppress  LsstDm-3-2a LsstDm-3-4a LsstDm-4-6 LsstDm-5-25 "Boost non-Std" */
```
timj: Why is this test disabled? It does test for AFWDATA inside it? Is that meant to be a conditional skipper?

tests/statistics.cc
AstroVPK: This test is disabled by default because we should assume that afwdata is not setup (in case we drop afwdata from the stack in the future etc...). If it is setup (checked in the tests/Sconscript file), we pass a command line arg to the test - '--run_test=+StatisticsTestImages' to run this test.

TallJimbo: I'm not sure that this is the wrong sense for the default, but I don't like the argument that this is a guard against dropping afwdata in the future; if we were to drop afwdata in the future, we'd certainly want to rewrite this test, and not have silently dead code.

timj: I don't think this is doing what you think then. If I run this test file without {{afwdata}} setup it runs the 3 enabled tests. If I setup {{afwdata}} it still only runs the 3 tests.

tests/convolve.py
```diff
@@ -173,26 +179,28 @@ def sameMaskPlaneDicts(maskedImageA, maskedImageB):
...
+            # provide destinations for the convolved MaskedImage and Image that contain junk
+            # to verify that convolve overwrites all pixels;
+            # make them deep copies so we can mess with them without affecting self.inImage
+            self.cnvMaskedImage = afwImage.MaskedImageF(FullMaskedImage, ShiftedBBox, afwImage.LOCAL, True)
+            self.cnvImage = afwImage.ImageF(FullMaskedImage.getImage(), ShiftedBBox, afwImage.LOCAL, True)
+    
+            self.width = self.maskedImage.getWidth()
+            self.height = self.maskedImage.getHeight()
+    #         smask = afwImage.MaskU(self.maskedImage.getMask(), afwGeom.Box2I(afwGeom.Point2I(15, 17), afwGeom.Extent2I(10, 5)))
+    #         smask.set(0x8)
```
timj: I know this was commented out before but why is it there?

tests/warper.py
```diff
@@ -35,18 +35,19 @@
...
 pexLog.Debug("lsst.afw.math", VERBOSITY)
 
 try:
-    afwDataDir = lsst.utils.getPackageDir("afwdata")
-except Exception:
-    warnings.warn("skipping all tests because afwdata is not setup")
+    afwdataDir = lsst.utils.getPackageDir("afwdata")
+except pexExcept.NotFoundError:
+    afwdataDir = None
     dataDir = None
```
timj: `dataDir`?

tests/wcs1.py
```diff
@@ -298,17 +302,20 @@ class WCSTestCaseSDSS(unittest.TestCase):
     """A test case for WCS using a small (SDSS) image with a slightly weird WCS"""
 
     def setUp(self):
-        self.im = afwImage.DecoratedImageD(InputSmallImagePath)
-
-        self.wcs = afwImage.makeWcs(self.im.getMetadata())
-
-        if False:
-            ds9.mtv(self.im, wcs=self.wcs)
+        if afwdataDir is not None:
```
timj: I think this test should really be checking `InputSmallImagePath` is not None rather than `afwdataDir`.

tests/wcs1.py
```diff
@@ -467,17 +481,20 @@ def makeWcs(crPixPos, crValDeg, projection):
...
 
+@unittest.skipIf(afwdataDir is None, "afwdata not setup")
 class WCSTestCaseCFHT(unittest.TestCase):
     """A test case for WCS"""
 
     def setUp(self):
-        path = InputImagePath + ".fits"
-        self.metadata = afwImage.readMetadata(path)
-        self.wcs = afwImage.makeWcs(self.metadata)
+        if afwdataDir is not None:
```
timj: Again, it's `InputImagePath` that matters here.

tests/wcs1.py
```diff
@@ -467,17 +481,20 @@ def makeWcs(crPixPos, crValDeg, projection):
...
+            path = InputImagePath + ".fits"
+            self.metadata = afwImage.readMetadata(path)
+            self.wcs = afwImage.makeWcs(self.metadata)
 
     def tearDown(self):
-        del self.wcs
-        del self.metadata
+        if afwdataDir is not None:
+            del self.wcs
```
timj: I'd be inclined to be explicit about calling del on `self.wcs` if `self.wcs` is an attribute and not None.

tests/convolve.py
AstroVPK: ConvolveTestCase includes a number of tests that don't require afwdata (eg. runBasicTest, runStdTest etc...). If I used skipIf on the setUp or on the whole class, they would also gety skipped. This way, only parts of ConvolveTestCase get skipped.

timj: I'm talking about the commented out `smask` code.

tests/warper.py
AstroVPK: It get's used later in getSwarpedImage(). I wanted to keep as much as in to avoid introducing new bugs.

timj: So just skip on `dataDir` then. Use `afwdataDir` as a stepping stone to `dataDir`.

tests/convolve.py
AstroVPK: Oh! No idea to be honest.

PR #59, opened 2016-05-05
Use and enforce Astropy compliant units

pschella:


src/table/AmpInfo.cc
```diff
@@ -144,23 +144,23 @@ AmpInfoTable::MinimalSchema::MinimalSchema() {
...
-        "amplifier gain in e-/ADU",
-        "e-/ADU");
+        "amplifier gain",
+        "electron adu^-1");
     saturation = schema.addField<int>(
         "saturation",
-        "saturation value, in ADU",
-        "ADU");
+        "saturation value",
+        "adu");
```
TallJimbo: As I just posted on the RFC, I don't think we want to use `adu` in some places and `count` in others, since they mean the same thing.  But I'm not sure which is preferred.

(I'm aware this was a preexisting inconsistency, but I'd like to address it on this issue).

tests/testSchema.py
```diff
@@ -106,6 +106,52 @@ def testKey(name, key):
...
+        except ValueError:
+            self.fail("addField raised ValueError unexpectedly!")
+        try:
+            schema.addField("g", type="Angle", units="rad")
+        except ValueError:
+            self.fail("addField raised ValueError unexpectedly!")
+        try:
+            schema.checkUnits()
+        except ValueError:
+            self.fail("checkUnits raised ValueError unexpectedly!")
```
TallJimbo: I don't think all the try/except blocks here are actually helpful - if we just let the exception propagate up, the test will still fail, and the traceback in that case might give whoever is debugging the failure a bit more useful information.

tests/testSchema.py
```diff
@@ -132,11 +178,11 @@ def testKeyAccessors(self):
...
         schema1 = lsst.afw.table.Schema()
-        schema1.addField("a", type=float, doc="doc for a", units="units for a")
-        schema1.addField("b", type=int, doc="doc for b", units="units for b")
+        schema1.addField("a", type=float, doc="doc for a", units="")
+        schema1.addField("b", type=int, doc="doc for b", units="")
         schema2 = lsst.afw.table.Schema()
-        schema2.addField("a", type=int, doc="doc for a", units="units for a")
-        schema2.addField("b", type=float, doc="doc for b", units="units for b")
+        schema2.addField("a", type=int, doc="doc for a", units="")
+        schema2.addField("b", type=float, doc="doc for b", units="")
```
TallJimbo: To maintain full test coverage here, I think we want to continue to pass real units for all of these calls (it doesn't matter what they are , but it matters that "a" and "b" have different units).

PR #64, opened 2016-05-13
Add Astropy view support to afw.table (DM-5641)

TallJimbo:
This adds an `asAstropy` method to get a `astropy.table.Table` or `astropy.table.QTable` view directly, as well as an `__astropy_table__` that will allow the Astropy table constructors to accept LSST `Catalog` objects (once astropy/astropy#4885 is complete).

The astropy PR has not quite been finalized, but I'm confident that any further changes will be minor, and hence it should be fine to review this now.  I will delay the merge of this ticket until the astropy PR is finalized (they also have a feature freeze this weekend, so I don't think we'll have to wait long).

### DIFF
--- a/python/lsst/afw/table/Base.i
+++ b/python/lsst/afw/table/Base.i
@@ -63,6 +63,8 @@ template <> struct NumpyTraits<lsst::afw::geom::Angle> : public NumpyTraits<doub
 %}
 
 %declareNumPyConverters(ndarray::Array<bool const,1>);
+%declareNumPyConverters(ndarray::Array<bool const,1,1>);
+%declareNumPyConverters(ndarray::Array<bool,1,1>);
 %declareNumPyConverters(ndarray::Array<lsst::afw::table::RecordId const,1>);
 %declareNumPyConverters(ndarray::Array<boost::uint16_t const,1>);
 %declareNumPyConverters(ndarray::Array<boost::int32_t const,1>);
--- a/python/lsst/afw/table/Base.i
+++ b/python/lsst/afw/table/Base.i
@@ -686,4 +686,16 @@ namespace lsst { namespace afw { namespace table {
 
 %declareCatalog(CatalogT, Base)
 
+// This needs to be here, not Catalog.i, to prevent it from being picked up in afw.detection, where _syntax.py is not available.
+%extend CatalogT<BaseRecord> {
+    %pythoncode %{
+    asAstropy = _syntax.BaseCatalog_asAstropy
+
+    def __astropy_table__(self, cls, copy, **kwds):
+        """Implement interface called by Astropy table constructors to construct a view or copy.
+        """
+        return _syntax.BaseCatalog_asAstropy(cls=cls, copy=copy)
+    %}
+}
+
 }}} // namespace lsst::afw::table
--- a/python/lsst/afw/table/_syntax.py
+++ b/python/lsst/afw/table/_syntax.py
@@ -202,3 +202,70 @@ def BaseColumnView_extract(self, *patterns, **kwds):
         else:
             d[name] = processArray(self.get(schemaItem.key))
     return d
+
+def BaseCatalog_asAstropy(self, cls=None, copy=False, unviewable="copy"):
+    """!
+    Return an astropy.table.Table (or subclass thereof) view into this catalog.
+
+    @param[in]   cls        Table subclass to use; None implies astropy.table.Table itself.
+                            Use astropy.table.QTable to get Quantity columns.
+
+    @param[in]  copy        Whether to copy data from the LSST catalog to the astropy table.
+                            Not copying is usually faster, but can keep memory from being
+                            freed if columns are later removed from the Astropy view.
+
+    @param[in]  unviewable  One of the following options, indicating how to handle field types
+                            (string and Flag) for which views cannot be constructed:
+                              - 'copy' (default): copy only the unviewable fields.
+                              - 'raise': raise ValueError if unviewable fields are present.
+                              - 'skip': do not include unviewable fields in the Astropy Table.
+                            This option is ignored if copy=True.
+    """
+    import astropy.table
+    if cls is None:
+        cls = astropy.table.Table
+    if unviewable not in ("copy", "raise", "skip"):
+        raise ValueError("'unviewable' must be one of 'copy', 'raise', or 'skip'")
+    ps = self.getMetadata()
+    meta = ps.toOrderedDict() if ps is not None else None
+    columns = []
+    items = self.schema.extract("*", ordered=True)
+    for name, item in items.iteritems():
+        key = item.key
+        unit = item.field.getUnits() or None  # use None instead of "" when empty
+        if key.getTypeString() == "String":
+            if not copy:
+                if unviewable == "raise":
+                    raise ValueError("Cannot extract string unless copy=True or unviewable='copy' or 'skip'.")
+                elif unviewable == "skip":
+                    continue
+            data = numpy.zeros(len(self), dtype=numpy.dtype((str, key.getSize())))
+            for i, record in enumerate(self):
+                data[i] = record.get(key)
+        elif key.getTypeString() == "Flag":
+            if not copy:
+                if unviewable == "raise":
+                    raise ValueError(
+                        "Cannot extract packed bit columns unless copy=True or unviewable='copy' or 'skip'."
+                    )
+                elif unviewable == "skip":
+                    continue
+            data = self.columns.get_bool_array(key)
+        elif key.getTypeString() == "Angle":
+            data = self.columns.get(key)
+            unit = "radian"
+            if copy:
+                data = data.copy()
+        else:
+            data = self.columns.get(key)
+            if copy:
+                data = data.copy()
+        columns.append(
+            astropy.table.Column(
+                data,
+                name=item.field.getName(),
+                unit=unit,
+                description=item.field.getDoc()
+            )
+        )
+    return cls(columns, meta=meta, copy=False)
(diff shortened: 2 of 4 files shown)


tests/testAstropyTableViews.py
```diff
@@ -0,0 +1,187 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+"""
+Tests for Astropy views into afw.table Catalogs
```
timj: Am I correct in thinking that these tests are not testing the new functionality in Astropy but are testing that the `asAstropy` method does the right thing. ie we are not (yet) depending on Astropy v1.2 for this to work. When v1.2 is released shouldn't we be able to instantiate an astropy.table directly with `asAstropy` being a private method? Or did I misunderstand?

python/lsst/afw/table/_syntax.py
```diff
@@ -202,3 +202,70 @@ def processArray(a):
         else:
             d[name] = processArray(self.get(schemaItem.key))
     return d
+
+def BaseCatalog_asAstropy(self, cls=None, copy=False, unviewable="copy"):
```
timj: I don't really like `asAstropy`. I'd prefer `as_astropy_table` but I understand that camel case is king. I don't really care if it's an internal method.

tests/testAstropyTableViews.py
TallJimbo: Basically correct.  The tests do not exercise any new Astropy functionality.  We do have a method that will hopefully work with v1.2, but that's untested.

But even once v1.2 is out, I'd like to keep the `asAstropy` (or whatever we call it) method public, as it provides useful customization options that will likely never be available through the Astropy table constructors (since they'd require API changes to those constructors).

python/lsst/afw/table/_syntax.py
```diff
@@ -202,3 +202,70 @@ def processArray(a):
...
+    Return an astropy.table.Table (or subclass thereof) view into this catalog.
+
+    @param[in]   cls        Table subclass to use; None implies astropy.table.Table itself.
+                            Use astropy.table.QTable to get Quantity columns.
+
+    @param[in]  copy        Whether to copy data from the LSST catalog to the astropy table.
+                            Not copying is usually faster, but can keep memory from being
+                            freed if columns are later removed from the Astropy view.
+
+    @param[in]  unviewable  One of the following options, indicating how to handle field types
```
timj: Does this method exist because you can't get the `unviewable` keyword argument into the `__astropy_table__` call? Could they forward some keyword arguments through the astropy.table constructor to this method? Then you could get rid of the entire trampoline.

python/lsst/afw/table/_syntax.py
TallJimbo: I'm fairly ambivalent about underscores vs. CamelCase here; overall I have a slight preference for switching everything to underscores, but I don't want to switch just some interfaces because then it becomes hard to remember which convention to use where.  But I also don't want people to be stuck trying to remember whether it's `asAstropy` or `asAstroPy`.  Maybe let's see if we get any more opinions?

In any case, I'd like to avoid appending "table"; I think that's pretty strongly implied by the fact that this is a Catalog object, and I do think this is something people may end up typing a fair amount in interactive use, so fewer characters matters.

RobertLuptonTheGood: It's "import astropy" so I don't think that people will think "asAstroPy".   I vote for asAstropy for consistency with the rest of our camelCaseCode.

tests/testAstropyTableViews.py
timj: As I say above, I would prefer it if we just said "you pass this afw table to the astropy table constructor" rather than "you call this special exporter routine because it has an extra feature". Getting some extra `**kwargs` passed in to `__astropy_table__` would seem feasible if we asked about it quickly. We'd then have a transitional APIs that would allow people to use table views before 1.2 astropy is released but with the expectation that in the future we'd use the astropy constructor.

timj: Oh, and my main comment here was that I'd like a comment in this test explaining that it's not actually testing the Astropy native constructor.

python/lsst/afw/table/_syntax.py
TallJimbo: Yes and no.  This argument is indeed the only one we can't get through the `__astropy_table__` call.  But I also think it's not obvious that the astropy table constructor interface is the more natural one even if it did kwarg forwarding.  We had some discussion of this on the current Astropy PR's predecessor (astropy/astropy/pull/4864, but it's a long thread); it's sort of unfortunate that Python users expect overloaded constructors, even though the language doesn't actually support overloading.

timj: Ok. So we have both.

tests/testAstropyTableViews.py
TallJimbo: Comment added.

python/lsst/afw/table/_syntax.py
TallJimbo: I think my worry about confusion comes from the fact that I [write `NumPy` in CamelCase code](https://github.com/lsst/ndarray/blob/master/python/ndarray.i#L41) even though it's `import numpy`.

But the NumPy and Astropy documentation and web pages seem to be quite internally consistent on the capitalization of the "p" while being unfortunately inconsistent with each other, so I maintain it's not my fault.

python/lsst/afw/table/_syntax.py
pschella: If `copy=False` actually might leak shouldn't `copy=True` be the default?

TallJimbo: It's not actually a leak; it's just that you're keeping a large array alive via a view into a subset of it.  It's exactly the same as what happens here:
```
a = numpy.zeros(100, dtype=float)
b = a[0:5]
del a
```

taldcroft: Since this is the first case of using `__astropy_table__` and it looks like passing through `**kwargs` would be useful, this might indicate it will be useful in other cases.  This is already officially part of the `__astropy_table__` API (except there is nothing to populate `**kwargs`).  We would need to have some discussion, but it isn't entirely unreasonable to have `**kwargs` in the main `Table` API for this purpose.  I would be much happier to have `Table(lsst_table, unviewable='raise')` be the one true way to convert, rather than one mostly-works method and another fully-works method.

taldcroft: On the philosophical side, when I see `Table(lsst_table)` I almost view as a casting operation and it seems obvious to me that this should work.  So one just needs to provide the machinery to cast from one kind of table to other. 

Right now `Table` has methods `to_pandas()` and `from_pandas()` to convert back and forth to pandas DataFrames.  I think this is really inelegant compared to `DataFrame(astropy_table)` and `Table(dataframe)`.  Part of the reason is that `Table(dataframe)` can actually take all the other constructor args, for instance I could change the column names or types at that time.  With `from_pandas()` currently you don't have the full constructor API.  It is certainly possible to put the full API into `from_pandas()` but that just gets ugly and repetitious for each new compatible table type.

TallJimbo: If there was no precedent here, I think I'd feel pretty strongly that `to_X()` methods are the better interface, because it puts the responsibility in the right place (if object A is viewable as a B, it's A that actually has to know how to do that), and it lets constructors stay simple.  And there's nothing stopping you from adding pandas constructor args to `to_pandas()`, and in fact you can limit it to just the pandas constructor arguments that make sense for this conversion. That's potentially an advantage over the constructor, which has to have all possible arguments for all possible construction patterns and thus pretty complex documentation to specify how those options interact.  You can use `**kwargs` to get around some of that, but I think it's always preferable to avoid that when you can.

All that said, using constructors for type casting is very much an established pattern in Python, even way down in the standard library.  I think that may have started largely because it's familiar to people coming from C++, Java, or other strongly-typed languages with overloading (where this pattern very much makes sense), but it's here now, and it's not something I can wish away.

I don't have a well-formed opinion about `from_X` vs. a constructor; generally I'd prefer the former to reduce overloading, but your point about replicating the full constructor API is a good one.

Anyhow, I think the current interface should accept `**kwargs` correctly in `__astropy_table__` already, so if `Table`'s constructor does allow those to be forwarded from the user, the type-casting constructor will indeed be as fully functional as the `asAstropy` method, and I'd be happy to eventually deprecate the latter if it ultimately proves unpopular in comparison.

taldcroft:
OK astropy/astropy#4885 is merged with `**kwarg` keyword passthrough supported and will be in 1.2.

python/lsst/afw/table/Base.i
```diff
@@ -684,4 +686,16 @@ namespace lsst { namespace afw { namespace table {
...
 
+// This needs to be here, not Catalog.i, to prevent it from being picked up in afw.detection, where _syntax.py is not available.
+%extend CatalogT<BaseRecord> {
+    %pythoncode %{
+    asAstropy = _syntax.BaseCatalog_asAstropy
+
+    def __astropy_table__(self, cls, copy, **kwds):
+        """Implement interface called by Astropy table constructors to construct a view or copy.
+        """
+        return _syntax.BaseCatalog_asAstropy(cls=cls, copy=copy)
```
timj: Shouldn't the kwargs have been forwarded on here?

TallJimbo: Thought I did that, but apparently not.  I'll create a mini-ticket to fix.

PR #65, opened 2016-05-19
DM-6147

r-owen:


parejkoj:
These look good. The only thing I might add would be a clarification on what "suspect" means in the help description. Is that a term of art in LSST land, or is it supposed to be a general catch-all for "some undefined thing may be wrong here"?

r-owen:
That is an excellent point. I will expand the explanation. For the record: it means a pixel whose value is high enough that the non-linearity correction may not work very well, and so very picky algorithms may wish to avoid using this pixel. I worry it's too granular, but HSC apparently finds it useful.

parejkoj:
Ah, then can we call it something other than "suspect"? Because that sounds like a more generic thing. Maybe "non-linear suspect" or something?

r-owen:
Ack!!!!!

parejkoj:
I'm sorry about the naming complaint, but I was definitely confused by it when I saw the name, and I could see wanting to introduce a "suspect" mask in the future as a catch-all.

r-owen:
I agree it's not a great name, but the mask plane is called SUSPECT so I think it's the best choice unless we rename the mask plane (and I don't want to do that unless it's really important)

parejkoj:
Ok. Just document the name well.

parejkoj:
The new description is definitely better. Thanks.

RobertLuptonTheGood:
I like SUSPECT, as we may set it for other sorts of problems too.

parejkoj:
If we're defining "suspect" as "pixels with values >= suspectLevel may be poorly linearized", then we shouldn't use "suspect" for other problems. If we want to use it for other problems, we should define it as "pixels whose post-ISR values are suspect for various reasons." And if we do that, the definition of suspectLevel will be meaningless, since it's not a `value >= suspectLevel` test anymore.

RobertLuptonTheGood:
I think we're using suspect as "pixels that we don't trust".   One type of untrusted pixel is those above a certain level, so suspectLevel is OK.  I _think_ I'm agreeing with John, but I'm not quite sure

r-owen:
Robert: could you please provide a definition of suspect/suspect level?

parejkoj:
That was what I initially suspected (see what I did there?) `suspect` meant, but in this context it appears that it has specific meaning. The documentation should be clear about this: possibly isrTask says what it currently says (because that's how `suspect` is used in that context) while afw says the more generic thing (suspect means we don't trust it).

RobertLuptonTheGood:
Does this help?

> Pixels whose values may be affected by unknown systematics are labelled SUSPECT.  An example is that the NAOJ instrument scientist for SuprimeCam recommends not trusting the values of pixels above a certain level (this could happen if the non-linearity corrections above this level are unstable).  We accordingly set the SUSPECT mask bit for all pixels above a level, suspectLevel -- which may be infinite for the LSST electronics.   In the future we may identify other sources of suspect pixels.

r-owen:
Thank you Robert. That is very helpful.

r-owen:
Here is the description I came up with for the amp info catalog field (pushed to the ticket):

    level above which pixels are considered suspicious, meaning they may be affected by unknown
    systematics; for example if non-linearity corrections above a certain level are unstable
    then that would be a useful value for suspectLevel; use `nan` if no such level applies

parejkoj:
I'm not sure about the "level above which" part. Robert's description above suggests that there could be other reasons for a pixel to be suspicious (I think? @RobertLuptonTheGood ).

r-owen:
In support of "level above": if pixels with low signal are hard to correct accurately then we are in deep trouble, since much of our signal is faint. I suggest we keep "level above" for now. It's probably sufficient, and if not, I suspect we will need to do much more work.

PR #72, opened 2016-05-28
DM-6095, afw: use forwarding references instead of boost::ref

TallJimbo:


include/lsst/afw/table/detail/SchemaMapperImpl.h
```diff
@@ -78,7 +78,8 @@ class SchemaMapperImpl {
         }
 
         /// @brief Construct the wrappper.
-        explicit VisitorWrapper(F func) : _func(func) {}
+        template <typename T>
+        explicit VisitorWrapper(T func) : _func(std::forward<T>(func)) {}
```
TallJimbo: I'm a bit surprised the signature here doesn't need to be `T&& func`.  If it's just because this is only invoked by another template function that does use a unversal reference, it seems like it'd be safer to do it here too, in case there is another caller in the future.

pschella: Yes, this should be `T&&` I just overlooked it. Of course the internal forwarding still works because it isn't used after but that is not really the point. Will fix it.

PR #74, opened 2016-05-29
Review for DM-6094: Use type traits from type_traits over boost equivalents where possible

jdswinbank:


src/geom/Extent.cc
```diff
@@ -19,12 +19,12 @@
  * the GNU General Public License along with this program.  If not, 
  * see <http://www.lsstcorp.org/LegalNotices/>.
  */
+
+#include <type_traits>
```
jdswinbank: Don't think this is used in this file.

src/math/FunctionLibrary.cc
```diff
@@ -1,7 +1,7 @@
 // -*- lsst-c++ -*-
 
 #include <memory>
-#include "boost/type_traits/is_same.hpp"
+#include <type_traits>
```
jdswinbank: Unused?

PR #79, opened 2016-06-22
Add updateRefCentroids and updateSourceCoords

r-owen:
Add two utility functions to lsst.afw.table:
updateRefCentroids and updateSourceCoords.
Add a unit test for these functions testTableUtils.py
(since the functions are defined in new module lsst.afw.table.utils)

### DIFF
--- a/python/lsst/afw/table/utils.py
+++ b/python/lsst/afw/table/utils.py
@@ -1,0 +1,69 @@
+#
+# LSST Data Management System
+# Copyright 2016 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
+from .tableLib import CoordKey, Point2DKey
+
+__all__ = ["updateRefCentroids", "updateSourceCoords"]
+
+
+def updateRefCentroids(wcs, refList):
+    """Update centroids in a collection of reference objects
+
+    This code supports any kind of collection, instead of requiring a catalog,
+    to make it easy to use with match lists. For example:
+
+        updateRefCentroids(wcs, refList=[match.first for match in matches])
+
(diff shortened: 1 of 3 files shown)


tests/testTableUtils.py
```diff
@@ -0,0 +1,181 @@
...
+        for val in nanRefCentroid:
+            self.assertTrue(math.isnan(val))
+
+        # computed centroid should be crpix
+        afwTable.updateRefCentroids(self.wcs, self.refCat)
+        refCentroid = self.refCat[0].get(self.refCentroidKey)
+        self.assertPairsNearlyEqual(refCentroid, self.crpix)
+
+        # coord should not be changed
+        self.assertPairsNearlyEqual(self.refCat[0].get(self.refCoordKey), self.crval, maxDiff=0)
```
parejkoj: Shouldn't you be able to use assertEqual here, since coord really shouldn't be touched at all?

tests/testTableUtils.py
```diff
@@ -0,0 +1,181 @@
...
+        for val in nanSourceCoord:
+            self.assertTrue(math.isnan(val))
+
+        # compute coord should be crval
+        afwTable.updateSourceCoords(self.wcs, self.sourceCat)
+        srcCoord = self.sourceCat[0].get(self.srcCoordKey)
+        self.assertPairsNearlyEqual(srcCoord, self.crval)
+
+        # centroid should not be changed; also make sure that getCentroid words
+        self.assertPairsNearlyEqual(self.sourceCat[0].getCentroid(), self.crpix, maxDiff=0)
```
parejkoj: Same as above: assertEqual should be ok.

tests/testTableUtils.py
r-owen: maxDiff=0 has the same effect, and I don't know if Coord supports operator==

python/lsst/afw/table/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    """Update coords in a collection of sources
+
+    This code supports any kind of collection, instead of requiring a catalog,
+    to make it easy to use with match lists. For example:
+
+        updateSourceCoords(wcs, sourceList=[match.second for match in matches])
+
+    @param[in] wcs  WCS to map from pixels to sky; an lsst.afw.image.Wcs
+    @param[in,out] sourceList    collection of sources (lsst.afw.table.SourceRecords) with fields:
+                                - "slot_Centroid" (as returned by getCentroid) is read
```
parejkoj: I'd rather this be worded `read from "slot_Centroid" (as returned by getCentroid)` and `write to "coord" (a field...)`.

tests/testTableUtils.py
parejkoj: Ah, right, it's an afw object. Can you try it with equal and see what happens? I see "nearlyEqual" and think "something could have gotten jittered a bit", even with the maxDiff.

r-owen: Will do. I just checked and Coord does support operator== so this ought to work.

python/lsst/afw/table/utils.py
r-owen: Good suggestion. Will do.

PR #80, opened 2016-07-13
Tickets/dm 2280

r-owen:


rearmstr:
Do you still need the plateScale argument in makePixelToTanPixel.py?

tests/testMakePixelToTanPixel.py
```diff
@@ -26,14 +26,73 @@
...
+import math
 
 import lsst.utils.tests
 import lsst.afw.geom as afwGeom
 import lsst.afw.cameraGeom as cameraGeom
 from lsst.afw.cameraGeom import makePixelToTanPixel
 
-class MakePixelToTanPixelTestCaseCase(unittest.TestCase):
+class MakePixelToTanPixelTestCaseCase(lsst.utils.tests.TestCase):
+    def testSimpleCurvedFocalPlane(self):
```
rearmstr: How useful is it to have a test of this simple curved focal plane?  Isn't the same functionality being tested in the curved focal plane test below.  I think it was helpful for me trying to figure what is going on, but I'm not sure it adds anything new to the test suite.

r-owen:
Good catch. plateScale is ignored. Removing it means that makeDetector no longer needs that argument, either. I'll change the code accordingly.

tests/testMakePixelToTanPixel.py
r-owen: I feel it is important because it allows some low-level sanity checks that are difficult with a more complex geometry (e.g. a detector with rectangular pixels and offset and rotated with respect to the focal plane, as is used in testCurvedFocalPlane). In particular I can check that correction is purely radial. I agree that the comparison of actual pixel positions is redundant, but I felt it best to leave it in as an extra check.

PR #82, opened 2016-08-04
DM-6613: Typo in afw/display/interface.py

fred3m:
_callbacks property was missing an underscore.

### DIFF
--- a/python/lsst/afw/display/interface.py
+++ b/python/lsst/afw/display/interface.py
@@ -537,7 +537,7 @@ class Display(object):
             k, x, y = ev.k, ev.x, ev.y      # for now
 
             try:
-                if self.callbacks[k](k, x, y):
+                if self._callbacks[k](k, x, y):
                     break
             except KeyError:
                 print >> sys.stderr, "No callback is registered for %s" % k
--- a/python/lsst/afw/display/interface.py
+++ b/python/lsst/afw/display/interface.py
@@ -29,6 +29,9 @@ import sys
 import importlib
 import lsst.afw.geom  as afwGeom
 import lsst.afw.image as afwImage
+import lsst.log
+
+logger = lsst.log.Log.getLogger("afw.display.interface")
 
 __all__ = (
     "WHITE", "BLACK", "RED", "GREEN", "BLUE", "CYAN", "MAGENTA", "YELLOW", "ORANGE",
@@ -526,24 +529,23 @@ class Display(object):
 
     def interact(self):
         """!Enter an interactive loop, listening for key presses in display and firing callbacks.
-
-        Exit with q, \c CR, or \c ESC
-    """
-
-        while True:
+            Exit with q, \c CR, \c ESC, or any other callback function that returns a ``True`` value.
+        """
+        interactFinished = False
+        
+        while not interactFinished:
             ev = self._impl._getEvent()
             if not ev:
                 continue
             k, x, y = ev.k, ev.x, ev.y      # for now
-
-            try:
-                if self._callbacks[k](k, x, y):
-                    break
-            except KeyError:
-                print >> sys.stderr, "No callback is registered for %s" % k
-            except Exception, e:
-                print >> sys.stderr, "Display.callbacks[%s](%s, %s, %s) failed: %s" % \
-                    (k, k, x, y, e)
+            
+            if k not in self._callbacks:
+                logger.warn("No callback registered for {0}".format(k))
+            else:
+                try:
+                    interactFinished = self._callbacks[k](k, x, y)
+                except Exception as e:
+                    logger.error("Display._callbacks[{0}]({0},{1},{2}) failed: {3}".format(k, x, y, e))
 
     def setCallback(self, k, func=None, noRaise=False):
         """!Set the callback for key k to be func, returning the old callback
--- a/tests/display.py
+++ b/tests/display.py
@@ -159,6 +159,14 @@ class DisplayTestCase(unittest.TestCase):
 
             im = afwImage.MaskU(self.fileName, 3)
             dummy.mtv(im)
+    
+    def testInteract(self):
+        """Check that interact exits when a q, \c CR, or \c ESC is pressed, or if a callback function
+        returns a ``True`` value.
+        If this is run using the virtualDevice a "q" is automatically triggered.
+        If running the tests using ds9 you will be expected to do this manually.
+        """
+        self.display0.interact()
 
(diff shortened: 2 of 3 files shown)


python/lsst/afw/display/interface.py
```diff
@@ -523,27 +526,26 @@ def pan(self, colc=None, rowc=None, origin=afwImage.PARENT):
...
-            except Exception, e:
-                print >> sys.stderr, "Display.callbacks[%s](%s, %s, %s) failed: %s" % \
-                    (k, k, x, y, e)
+            
+            if k not in self._callbacks:
+                logger.warn("No callback registered for {0}".format(k))
+            else:
+                try:
+                    interact_finished = self._callbacks[k](k, x, y)
+                except Exception, e:
```
timj: If these lines are changes can you please upgrade them to modern Exceptions?
```python
except Exception as e:
```

tests/display.py
```diff
@@ -159,6 +159,13 @@ def testImageTypes(self):
 
             im = afwImage.MaskU(self.fileName, 3)
             dummy.mtv(im)
+    
+    def testInteract(self, timeout=1):
+        """Check that interact exists if a q, \c CR, or \c ESC is pressed.
+        If this is run using the virtualDevice a "q" is automatically triggered.
```
jdswinbank: "exists" or "exits"?

tests/display.py
```diff
@@ -159,6 +159,13 @@ def testImageTypes(self):
 
             im = afwImage.MaskU(self.fileName, 3)
             dummy.mtv(im)
+    
+    def testInteract(self, timeout=1):
```
jdswinbank: `timeout` seems to be unused.

python/lsst/afw/display/interface.py
```diff
@@ -523,27 +526,26 @@ def pan(self, colc=None, rowc=None, origin=afwImage.PARENT):
...
     def interact(self):
         """!Enter an interactive loop, listening for key presses in display and firing callbacks.
-
-        Exit with q, \c CR, or \c ESC
-    """
-
-        while True:
+            Exit with q, \c CR, or \c ESC
+        """
+        interact_finished = False
```
jdswinbank: [Coding standard](https://developer.lsst.io/coding/python_style_guide.html#naming-conventions) prefers `interactFinished`, ugly though it is.

python/lsst/afw/display/interface.py
```diff
@@ -523,27 +526,26 @@ def pan(self, colc=None, rowc=None, origin=afwImage.PARENT):
...
-
+    
     def interact(self):
         """!Enter an interactive loop, listening for key presses in display and firing callbacks.
-
-        Exit with q, \c CR, or \c ESC
-    """
-
-        while True:
+            Exit with q, \c CR, or \c ESC
```
jdswinbank: In fact, we exit when any callback returns `True`, right? (I realise this isn't a new addition in your changes, though).

python/lsst/afw/display/interface.py
```diff
@@ -523,27 +526,26 @@ def pan(self, colc=None, rowc=None, origin=afwImage.PARENT):
         """!Pan to (rowc, colc); see also zoom"""
 
         self.zoom(None, colc, rowc, origin)
-
+
```
jdswinbank: Please don't add unnecessary whitespace.

python/lsst/afw/display/interface.py
```diff
@@ -29,6 +29,9 @@
 import importlib
 import lsst.afw.geom  as afwGeom
 import lsst.afw.image as afwImage
+import lsst.log
+
+logger = lsst.log.Log.getLogger("afw.display.interact")
```
jdswinbank: I actually don't know the answer to this: do we have a convention for choosing logger names?

Regardless of a convention, the above looks a bit suspicious because it's in scope for the whole file but calls out `interact`, which is one specific method in one class contained in the file, by name.

tests/display.py
```diff
@@ -159,6 +159,13 @@ def testImageTypes(self):
...
+        """Check that interact exists if a q, \c CR, or \c ESC is pressed.
```
jdswinbank: "exists" or "exits"?

python/lsst/afw/display/interface.py
hchiang2: While doing the logging framework transition I have the exactly same question about the convention of logger names. If the log belongs to a Task the logger name is its `_DefaultName`.  If it's not a Task, the name seems a bit arbitrary and depends on the author's intent from I've seen so far.

python/lsst/afw/display/interface.py
```diff
@@ -523,27 +526,26 @@ def pan(self, colc=None, rowc=None, origin=afwImage.PARENT):
...
-                print >> sys.stderr, "Display.callbacks[%s](%s, %s, %s) failed: %s" % \
-                    (k, k, x, y, e)
+            
+            if k not in self._callbacks:
+                logger.warn("No callback registered for {0}".format(k))
+            else:
+                try:
+                    interact_finished = self._callbacks[k](k, x, y)
+                except Exception, e:
+                    logger.error("Display._callbacks[{0}]({0},{1},{2}) failed: {3}".format(k,x,y,e))
```
jdswinbank: Need spaces after commas in the argument list — `k, x, y, e`.

python/lsst/afw/display/interface.py
fred3m: This is a typo. It should have been 
`logger = lsst.log.Log.getogger("afw.display.interface")`.

If lsst.log is like the logging module, this can matter a great deal. For example, if the "afw" logger is set to a given level (logging.INFO for example), then "afw.display.interface" will inherit that setting. For this reason other projects that I've been involved in use this convention, although I have also seen it become more specific ("afw.display.interface.interact") in a function if there is a reason that the function might want to be singled out explicitly. I don't know enough about the C++ logger and what effect this might have, but if it has similar functionality I would recommend that we adopt a similar convention. Even if it isn't, it might still be useful for the python side of the code.

python/lsst/afw/display/interface.py
fred3m: Sorry I didn't know that this is the convention. My text editor, Textmate, adds whitespace by default that matches the current indentation level (for the class in this case). Personally I like this style *much* better but I'll be more careful in the future.

timj: Yes, we try to remove trailing whitespace. We are also actively updating the style guide to be able to express it in terms of PEP-8.

tests/display.py
fred3m: This was corrected in a later commit. I rebased before pushing my commits but it didn't seem to fix this (but the current file [https://github.com/lsst/afw/blob/master/python/lsst/afw/display/interface.py](url) shows timeout removed).

PR #83, opened 2016-08-07
DM-7152: Port to python 3

timj:
Includes some code cleanup. Not working yet.


examples/timeWarpExposure.py
```diff
@@ -176,10 +179,10 @@ def run():
                     )
                     destExposure.setWcs(destWcs)
                     dTime, nIter, goodPix = timeWarp(destExposure, srcExposure, warpingControl)
-                    print "%4d  %5d  %8.1f  %6.1f, %6.1f  %7.1f %10s %8d %6.2f" % (
+                    print("%4d  %5d  %8.1f  %6.1f, %6.1f  %7.1f %10s %8d %6.2f" % (
```
pschella: We are not using new style print formatting?

timj: Switching to `format` is not required by the Python 3 migration. In fact, use of `format` at all is highly contentious. I always use `format` in new code as I grew up with Python 3.

python/lsst/afw/display/interface.py
```diff
@@ -532,13 +533,13 @@ def interact(self):
             Exit with q, \c CR, \c ESC, or any other callback function that returns a ``True`` value.
         """
         interactFinished = False
-
```
pschella: Why were these not part of the previous whitespace fix?

timj: Not entirely sure.

python/lsst/gdb/afw/printers.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
+from __future__ import absolute_import
```
pschella: Doesn't seem to be used.

python/lsst/gdb/afw/printers.py
```diff
@@ -884,6 +886,6 @@ def build_daf_base_dictionary():
         return printer
 
     printers.append(build_daf_base_dictionary())
-except ImportError, e:
-    print "RHL", e
-    from printers_oldgdb import *
+except ImportError as e:
+    print("RHL", e)
+    from .printers_oldgdb import *
```
pschella: Ok, ignore my previous comment.

tests/coordptr.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
 from __future__ import absolute_import, division
```
pschella: Separate or combine all?

tests/display.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
 from __future__ import absolute_import, division
```
pschella: Separate or combine everywhere?

python/lsst/gdb/afw/printers.py
timj: It was added because the code was originally `from printers_oldgdb import *`.

tests/exposure.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
 from __future__ import absolute_import, division
```
pschella: Same question (not repeated from now on).

tests/coordptr.py
timj: Yes. I know. `futurize` does not combine automatically and I wonder whether it's better for them to be together or separate (and I think I prefer separate). But I have not gone through the old code to split older items.

tests/exposure.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
```
timj: I obviously need to fix these shebangs...

examples/catTables.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from __future__ import print_function
+from past.builtins import basestring
```
pschella: Isn't the "from builtins import str" alternative better here?

examples/spatialCellExample.py
```diff
@@ -32,6 +32,8 @@
    >>> import SpatialCellExample; SpatialCellExample.run()
 """
 from __future__ import print_function
+from __future__ import division
+from builtins import range
 
 import os
```
pschella: Import ordering and spacing doesn't seem very consistent. Not a big problem. But it might be nice to cleanup some day.

examples/catTables.py
timj: On python3 they are synonyms. In Python2 I think the reason I do this is because string literals are a different class to the override `str` and I needed `basestring` to cover all possible variants.

python/lsst/afw/cameraGeom/cameraFactory.py
```diff
@@ -87,7 +87,7 @@ def makeCameraFromPath(cameraConfig, ampInfoPath, shortNameFunc):
     @return camera (an lsst.afw.cameraGeom.Camera)
     """
     ampInfoCatDict = dict()
-    for detectorConfig in cameraConfig.detectorList.itervalues():
+    for detectorConfig in cameraConfig.detectorList.values():
```
pschella: Why is something called {{dectectorList}} a dict? And don't we need {{from builtins import dict}} here as well to make it efficient?

python/lsst/afw/cameraGeom/cameraGeomEnumDicts.py
```diff
@@ -28,4 +28,4 @@
     cameraGeomLib.GUIDER:    "GUIDER",
     cameraGeomLib.WAVEFRONT: "WAVEFRONT",
 }
-DetectorTypeNameValDict = dict((val, key) for key, val in DetectorTypeValNameDict.iteritems())
+DetectorTypeNameValDict = dict((val, key) for key, val in DetectorTypeValNameDict.items())
```
pschella: Add `from builtins import dict` for efficiency?

python/lsst/afw/cameraGeom/detectorCollection.py
```diff
@@ -20,6 +20,8 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
 from __future__ import absolute_import, division
+from past.builtins import basestring
```
pschella: Use `from builtins import str` instead?

python/lsst/afw/cameraGeom/cameraFactory.py
timj: I didn't name these attributes :-)  Regarding dict efficiency, you may be right. `futurize` does not do that automatically.

python/lsst/afw/display/rgb.py
```diff
@@ -257,12 +259,12 @@ def __init__(self, minimum, dataRange, Q=8):
                 Q = Qmax
 
         if False:
-            self._slope = self._uint8Max/Q # gradient at origin is self._slope
+            self._slope = self._uint8Max/Q  # gradient at origin is self._slope
```
pschella: Would be nicer as different commit. But not really critical :)

python/lsst/afw/cameraGeom/detectorCollection.py
timj: No. Because of this:
```
>>> from builtins import str
>>> from past.builtins import basestring
>>> isinstance("xx", str)
False
>>> isinstance("xx", basestring)
True
>>> isinstance(str("xx"), basestring)
True
```

python/lsst/afw/table/_syntax.py
```diff
@@ -123,7 +124,7 @@ def BaseRecord_extract(self, *patterns, **kwds):
         d = self.schema.extract(*patterns, **kwds).copy()
     elif kwds:
         raise ValueError("Unrecognized keyword arguments for extract: %s" % ", ".join(kwds.keys()))
-    for name, schemaItem in d.items():  # can't use iteritems because we might be adding/deleting elements
+    for name, schemaItem in list(d.items()):  # can't use iteritems because we might be adding/deleting elements
```
pschella: take out obsolete comment?

python/lsst/afw/table/_syntax.py
```diff
@@ -188,14 +189,14 @@ def BaseColumnView_extract(self, *patterns, **kwds):
...
-        raise ValueError("Unrecognized keyword arguments for extract: %s" % ", ".join(kwds.keys()))
+        raise ValueError("Unrecognized keyword arguments for extract: %s" % ", ".list(kwds.keys()))
     def processArray(a):
         if where is not None:
             a = a[where]
         if copy:
             a = numpy.ascontiguousarray(a)
         return a
-    for name, schemaItem in d.items(): # can't use iteritems because we might be adding/deleting elements
+    for name, schemaItem in list(d.items()): # can't use iteritems because we might be adding/deleting elements
```
pschella: take out obsolete comment?

python/lsst/afw/display/rgb.py
timj: Yes, this was caused by an interactive rebase where I backed out an `old_div` change but ended up retaining the whitespace fix by mistake.

python/lsst/afw/table/tableEnumDicts.py
```diff
@@ -28,4 +28,4 @@
     tableLib.UR: "UR",
     tableLib.UL: "UL",
 }
-ReadoutCornerNameValDict = dict((val, key) for key, val in ReadoutCornerValNameDict.iteritems())
+ReadoutCornerNameValDict = dict((val, key) for key, val in ReadoutCornerValNameDict.items())
```
pschella: Use dict comprehension?

python/lsst/gdb/afw/printers.py
```diff
@@ -1,5 +1,10 @@
 from __future__ import print_function
 from __future__ import absolute_import
+from __future__ import division
+from builtins import str
+from builtins import range
+from builtins import object
+from past.builtins import long
```
pschella: Why not "from builtins import int"?

timj: I wanted to retain the distinction on the python2 version as I had assumed it was relatively important. That's how I did previous python3 migrations (e.g. in lsst@daf_base#8).

python/lsst/gdb/afw/printers.py
```diff
@@ -687,7 +692,7 @@ def invoke (self, args, fromTty):
                 print("%*d" % (opts.formatWidth, x), end=' ')
             print("")
 
-            for y in reversed(range(y0, y0 + ny)):
+            for y in reversed(list(range(y0, y0 + ny))):
```
pschella: Why?

python/lsst/gdb/afw/printers_oldgdb.py
```diff
@@ -205,7 +209,7 @@ def invoke (self, args, fromTty):
             print("%8d" % x, end=' ')
         print("")
 
-        for y in reversed(range(y0, y0 + ny)):
+        for y in reversed(list(range(y0, y0 + ny))):
```
pschella: Why?

tests/convolve.py
```diff
@@ -168,11 +169,11 @@ def sameMaskPlaneDicts(maskedImageA, maskedImageB):
     """
     mpDictA = maskedImageA.getMask().getMaskPlaneDict()
     mpDictB = maskedImageB.getMask().getMaskPlaneDict()
-    if mpDictA.keys() != mpDictB.keys():
+    if list(mpDictA.keys()) != list(mpDictB.keys()):
```
pschella: Seems inefficient. Twice iterating over the list.

tests/convolve.py
```diff
@@ -168,11 +169,11 @@ def sameMaskPlaneDicts(maskedImageA, maskedImageB):
...
         print("mpDictA.keys()  ", mpDictA.keys())
         print("mpDictB.keys()  ", mpDictB.keys())
         return False
-    if mpDictA.values() != mpDictB.values():
+    if list(mpDictA.values()) != list(mpDictB.values()):
```
pschella: Same

python/lsst/gdb/afw/printers.py
timj: No idea. I had a failing test here without the `list`. Will take another look.

tests/imagePca.py
```diff
@@ -179,7 +180,7 @@ def testPca(self):
         self.assertEqual(len(eImages), numInputs)
 
         # Test for orthogonality
-        for i1, i2 in itertools.combinations(range(len(eImages)), 2):
+        for i1, i2 in itertools.combinations(list(range(len(eImages))), 2):
```
pschella: Are you sure this is needed?

tests/convolve.py
timj: I'll run the test without the `list`.

tests/imagePca.py
timj: Will check. It might be one that I missed.

tests/testSchema.py
```diff
@@ -103,7 +104,7 @@ def testKey(name, key):
         self.assertEqual(keys[3].getBit(), 1)
         for n1, n2 in zip(schema4.getOrderedNames(), "abcd"):
             self.assertEqual(n1, n2)
-        keys2 = map(lambda x: x.key, schema4.asList())
+        keys2 = [x.key for x in schema4.asList()]
```
pschella: Not possible to iterate over schema directly?

python/lsst/afw/table/Base.i
```diff
@@ -122,6 +122,7 @@ template <> struct NumpyTraits<lsst::afw::geom::Angle> : public NumpyTraits<doub
 %pythoncode %{
 from . import _syntax
 import astropy.units
+from past.builtins import basestring
```
pschella: Can't use `from builtins import str` instead?

tests/testSchema.py
timj: I think this is a question for @TallJimbo. I am just modernizing the usage.

python/lsst/afw/image/wcs.i
```diff
@@ -48,39 +48,20 @@
...
-        ar >> *header;
-        return header;
-    }
-%}
-
 %pythoncode %{
     def unpickleWcs(pick):
-        header = unpickleMetadata(pick)
-        return makeWcs(header)
+        import pickle
```
pschella: If we use protocol v2 (and cpickle) it works nicer with pybind11 in the future.

python/lsst/afw/table/Base.i
timj: In this particular case I don't want the futurized `str` to contaminate the `aliases` table.

python/lsst/afw/image/wcs.i
timj: `cPickle` does not exist on python 3. Are you asking for us to explicitly attempt to load `cPickle` for python2 and `pickle` on python3? What's the default protocol. I don't think @PaulPrice cares.

tests/footprint2.py
```diff
@@ -455,9 +449,11 @@ def doTestPeaks(self, dwidth=0, dheight=0, x0=0, y0=0, threshold=10, callback=No
         #
         # Sort self.peaks in decreasing peak height to match Footprint.getPeaks()
         #
+        def peakKey(p):
+            return p[2] * -1.0
```
pschella: Slightly confusing names `keyPeaks` and `peakKey`. Not a big issue though.

tests/footprint2.py
```diff
@@ -575,7 +571,9 @@ def callback():
             self.im.getImage().set(10, 4, 20)
             self.peaks[-2].append((10, 4, ))
 
-            self.peaks[0] = sorted(sum(self.peaks, []), lambda x, y: cmpPeaks(self.im, x, y))
+            def peaksKey(p):
```
pschella: Ok, now it is getting confusing.

tests/footprint2.py
timj: Yes. That was grating on my mind as I iterated on solving the problems in this file...

python/lsst/afw/display/interface.py
```diff
@@ -80,7 +80,7 @@ def _makeDisplayImpl(display, backend, *args, **kwargs):
         try:
             _disp = importlib.import_module(dt, package="lsst.display")
             break
-        except ImportError as e:
+        except (ImportError, SystemError) as e:
```
pschella: What is the SystemError for?

python/lsst/afw/geom/ellipses/PixelRegion.i
```diff
@@ -59,12 +59,15 @@
...
 %pythoncode %{
-def next(self):
+def __next__(self):
     if self.atEnd():
         raise StopIteration()
     current = self.get()
     self.increment()
     return current
+
+next = __next__
```
pschella: What is this for? Why not use `from builtins import object`? Per http://python-future.org/compatible_idioms.html#custom-iterators

tests/testSchema.py
TallJimbo: `__iter__` actually delegates to `asList`, due to Swig-related implementation details.

python/lsst/afw/display/ds9Regions.py
```diff
@@ -46,7 +46,7 @@ def dot(symb, c, r, size, ctype=None, fontFamily="helvetica", textAngle=None):
 
 N.b. objects derived from BaseCore include Axes and Quadrupole.
 """
-    if ctype == None:
+    if ctype is None:
```
pschella: This is the only case? I'm impressed :)

python/lsst/afw/table/Base.i
```diff
@@ -609,6 +610,7 @@ aliases = {
     int: "I",
     float: "D",
     str: "String",
+    futurestr: "String",
```
pschella: I'm confused why this is needed. Can't we just use `from builtins import str` everywhere and forget about the old `str`?

tests/random1.py
```diff
@@ -109,6 +110,7 @@ def setUp(self):
     def tearDown(self):
         del self.image
 
+    @unittest.skipIf(sys.version_info.major >= 3, "setState segv on Py3k. Bytes vs str issue.")
```
pschella: I assume this is fixed now?

python/lsst/afw/display/interface.py
timj: That is what seems to be triggered from `import_module` if `lsst.display` does not exist.

python/lsst/afw/geom/ellipses/PixelRegion.i
timj: I had not spotted that inheriting from `object` automatically added the old `next` method as a clone for `__next__`.

python/lsst/afw/table/Base.i
timj: No, because there is code above AFW that has not been migrated, and also string literals in Python 2 will not be compatible with future `str` so we we would have to enable `unicode_literals` in all the caller code. This approach lets me support both during transition to python 3. When we drop python2 this line can go away.

tests/random1.py
timj: Not on this branch. I will cherry-pick the fix from @TallJimbo tonight. We then have to decide how much we care about the exception problem.

python/lsst/afw/display/interface.py
```diff
@@ -76,15 +76,26 @@ def _makeDisplayImpl(display, backend, *args, **kwargs):
...
+            # Copy the exception into outer scope
+            exc = e
 
     if not _disp:
-        raise e
+        if exc is not None:
+            # re-raise the final exception
+            raise exc
+        else:
+            raise ImportError("Could not load the requested backend: {}".format(backend))
```
pschella: Wow, this is convoluted. Why not simply test (in the handler) if there are more list entries to consider and re-raise if there aren't any? Perhaps using the input list as a stack?

python/lsst/afw/display/virtualDevice.py
```diff
@@ -41,7 +41,7 @@ def __del__(self):
 
     def _close(self):
         """!Close the display, cleaning up any allocated resources"""
-        if self.verbose:
+        if hasattr(self, "verbose") and self.verbose and hasattr(self, "display"):
```
pschella: :( that this is needed.

python/lsst/afw/display/interface.py
timj: I was trying to keep with the original spirit of the code. The original problem was the `e` not being accessible outside the scope. I also then learned I shouldn't be asking for a relative import if I'm doing an absolute import because that causes Python3 to trigger an error when it can't find `lsst.display` package anywhere.

python/lsst/afw/display/virtualDevice.py
timj: It's because the attributes are not created if the object `__init__` has not been run.

python/lsst/afw/geom/ellipses/PixelRegion.i
timj: The problem is this is a swig object so it doesn't seem to be easy to make it inherit from something else. The `.next()` method is their for old code that has not been migrated to use the `next()` function. We can remove it at some point. Maybe we should add a deprecation decorator at some point.

python/lsst/afw/display/interface.py
timj: For example:
```
Importing '.virtualDevice' with args {'package': 'lsst.display'}
Traceback (most recent call last):
  File "tests/display.py", line 202, in <module>
    with afwDisplay.Display("test", backend=args.backend) as disp:
  File "/Volumes/G-RAID with Thunderbolt/transient/lsstsw3/build/afw/python/lsst/afw/display/interface.py", line 143, in __init__
    self._impl = _makeDisplayImpl(self, backend, *args, **kwargs)
  File "/Volumes/G-RAID with Thunderbolt/transient/lsstsw3/build/afw/python/lsst/afw/display/interface.py", line 88, in _makeDisplayImpl
    _disp = importlib.import_module(dt, **impargs)
  File "/Users/timj/work/lsstsw3/miniconda/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 981, in _gcd_import
  File "<frozen importlib._bootstrap>", line 931, in _sanity_check
SystemError: Parent module 'lsst.display' not loaded, cannot perform relative import
```
so an error triggers before it can even attempt to do the import when `lsst.display` does not exist.

tests/testSchema.py
timj: Yes, I removed the `asList` and everything all worked.

python/lsst/afw/table/tableEnumDicts.py
timj: ok

python/lsst/afw/cameraGeom/cameraGeomEnumDicts.py
timj: I don't think it's needed for 4 elements.

PR #84, opened 2016-08-17
DM-7207: Pytest updates

timj:


parejkoj:
I just want to register my agreement that random seeding should go in setUp: you don't want any dependence on other tests being run, so you want to reset the seed every time.

tests/testAmpInfoTable.py
```diff
@@ -150,20 +150,12 @@ def testBasics(self):
...
-
-
-def run(shouldExit=False):
-    """Run the tests"""
-    lsst.utils.tests.run(suite(), shouldExit)
-
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: Needs newline

tests/testAngle.py
```diff
@@ -226,21 +220,21 @@ def testWrap(self):
                             self.assertAlmostEqual(math.cos(nearAngRad), cosAng)
 
 
-#-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+class DemoTestCase(lsst.utils.tests.TestCase):
```
timj: I don't think you wanted to copy this example in from the documentation :smile:

tests/testApCorrMap.py
```diff
@@ -87,10 +83,10 @@ def testAccessors(self):
...
         self.assertEqual(list(collections.OrderedDict(self.map).keys()), list(self.map))
-        self.assertTrue("b" in self.map)
-        self.assertTrue("d" not in self.map)
+        self.assertIn("b", self.map)
+        self.assertNotIn("d", self.map)
         self.map["d"] = lsst.afw.math.ChebyshevBoundedField(self.bbox, numpy.random.randn(2, 2))
-        self.assertTrue("d" in self.map)
+        self.assertIn("d", self.map)
         self.assertIsNone(self.map.get("e"))
         self.assertRaisesLsstCpp(lsst.pex.exceptions.NotFoundError, self.map.__getitem__, "e")
```
timj: This assert is deprecated. Please replace with `with self.assertRaises...`

timj:
There are a few bare `assertRaises` in the code (some of which have `lambda`). I imagine @jdswinbank doesn't want us to spend more time on fixing them so can you please file a ticket so we can remember to clean them up.

timj:
`testConvolve.py` still fails for me and I'm not entirely clear how it's working for anyone else. This line:
```python
from kernel import makeDeltaFunctionKernelList, makeGaussianKernelList
```
can not possibly work because there is no `kernel.py` anywhere. I think it is loading some helper functions from `testKernel.py` (renamed). Do you have an old `kernel.py` somewhere? How does this pass on Jenkins? Did anyone run this branch on Jenkins (py2 and py3)?

tests/testApproximate.py
```diff
@@ -172,29 +139,19 @@ def testLinearRampAsBackground(self):
                         rampCoeffs[0] + rampCoeffs[1]*x + rampCoeffs[1]*y
 
                     self.assertEqual(aim.get(x, y), val)
-        #
         # Check that we can't "truncate" the expansion to a higher order than we requested
-        #
         self.assertRaises(pexExcept.InvalidParameterError,
                           lambda: approx.getImage(orderMax + 1, orderMax + 1))
```
timj: This is one of those `assertRaises` + `lambda` that need to go at some point.

tests/testLeastSquares.py
```diff
@@ -42,36 +42,34 @@
...
 
-    def assertClose(self, a, b, rtol=1E-5, atol=1E-8):
-        self.assert_(numpy.allclose(a, b, rtol=rtol, atol=atol), "\n%s\n!=\n%s" % (a, b))
+    def _assertNotClose(self, a, b, rtol=1E-5, atol=1E-8):
+        self.assertFloatsNotEqual(a, b, rtol=rtol, atol=atol, msg="\n%s\n==\n%s" % (a, b))
 
-    def assertNotClose(self, a, b, rtol=1E-5, atol=1E-8):
-        self.assert_(not numpy.allclose(a, b, rtol=rtol, atol=atol), "\n%s\n==\n%s" % (a, b))
+    def setUp(self):
+        numpy.random.seed(500)
```
r-owen: I think it's too late for this ticket, but for future test improvements please also replace `import numpy` with `import numpy as np`

tests/testAstropyTableViews.py
```diff
@@ -179,20 +179,12 @@ def testUnviewableSkip(self):
...
-
-
-def run(shouldExit=False):
-    """Run the tests"""
-    lsst.utils.tests.run(suite(), shouldExit)
-
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testMakeWcs.py
```diff
@@ -86,8 +86,8 @@ def testNoCreateTanWcs(self):
         """Test than an exception is thrown if we try to upcast to a TanWcs inappropriately"""
         wcs = afwImage.makeWcs(self.metadata)
 
-        excpt = lsst.pex.exceptions.Exception
-        self.assertRaises(excpt, afwImage.cast_TanWcs, wcs)
+        with self.assertRaises(lsst.pex.exceptions.Exception):
```
r-owen: Thank you for ditching the confusing variable `excpt`

tests/testMaskedImage.py
```diff
@@ -120,9 +120,10 @@ def tearDown(self):
 
     def testArrays(self):
         image, mask, variance = self.mimage.getArrays()
-        self.assert_((self.mimage.getImage().getArray() == image).all())
-        self.assert_((self.mimage.getMask().getArray() == mask).all())
-        self.assert_((self.mimage.getVariance().getArray() == variance).all())
+        self.assertListEqual(self.mimage.getImage().getArray().flatten().tolist(), image.flatten().tolist())
```
r-owen: Is `assertListEqual` necessary? If the items are lists then`assertEqual` is preferred (and it will call assertListsEqual internally)

tests/testBackground.py
```diff
@@ -52,32 +42,18 @@
 import lsst.afw.display.ds9 as ds9
 import lsst.pex.exceptions as pexExcept
 
-try:
-    type(display)
-except NameError:
-    display = False
+# Set to True to display things in ds9.
+display = False
```
timj: I'm not sure whether we should keep this boilerplate or not. Given that it turned up 6 years ago in ff0c8fc4d3fb8d3e18c90ccf5d5efac743619b4a without explanation and has no description indicating why it exists, I'm happy to let it disappear.

tests/testMaskedImageIO.py
```diff
@@ -74,9 +74,15 @@ def setUp(self):
...
+        mask = self.Mask()
 
-        mask.clearMaskPlaneDict()
+        # Store the default mask planes for later use
+        maskPlaneDict = self.Mask().getMaskPlaneDict()
+        self.defaultMaskPlanes = sorted(maskPlaneDict, key=maskPlaneDict.__getitem__)
+
+        # reset so tests will be deterministic
+        self.Mask.clearMaskPlaneDict()
         for p in ("ZERO", "BAD", "SAT", "INTRP", "CR", "EDGE"):
```
r-owen: Is this the standard list of mask planes? If so, you could use defaultMaskPlanes. (I don't recognize "ZERO" so maybe it is not). It's no big deal, since it's just for these tests.

tests/testMaskedImage.py
timj: Hmm. You are right. It's not obvious from the `unittest` documentation but the code does try to compare standard types using specialist methods so `assertEqual` would work here.

tests/testBackground.py
```diff
@@ -113,9 +89,9 @@ def testOddSize(self):
 
         for iy, by in zip([image.get(0, y) for y in range(H)],
                           [back.get(0, y) for y in range(H)]):
-            self.assertTrue(abs(iy - by) < 5)
+            self.assertLess(abs(iy - by), 5)
 
-        if False:
+        if display:
```
timj: Interesting. The line was added in commit f31e7b32ba20a8b3b3a531361fdeabc24a7245d1 -- does the code work if a display is actually enabled? I'm not sure that this display code is compatible with the intent of the `display` variable.

tests/testBackground.py
```diff
@@ -146,20 +121,9 @@ def testgetPixel(self):
 
     @unittest.skipIf(AfwdataDir is None, "afwdata not setup")
     def testBackgroundTestImages(self):
+        """tests Laher's afwdata/Statistics/*.fits images (doubles)"""
```
timj: `Tests` vs `tests` at start of docstring.

tests/testBackground.py
```diff
@@ -218,7 +183,8 @@ def testRamp(self):
         bctrl.getStatisticsControl().setNumIter(1)
         backobj = afwMath.cast_BackgroundMI(afwMath.makeBackground(rampimg, bctrl))
 
-        print(rampimg.getArray())
+        if display:
+            print(rampimg.getArray())
```
timj: Did this break before? Is it garbage if the `display` variable is not set?

tests/testBackground.py
```diff
@@ -416,12 +374,11 @@ def testSubImage(self):
         subArr = bgSubImage.getArray()
 
         # the pixels happen to be identical but it is safer not to rely on that; close is good enough
-        self.assertTrue(np.allclose(subArr, subFullArr))
+        self.assertClose(subArr, subFullArr)
```
timj: `assertClose` is now `assertFloatsEqual` or `assertFloatAlmostEqual`.

tests/testBackground.py
```diff
@@ -458,15 +413,6 @@ def testUndersample(self):
         bctrl.setNxSample(3)
         bctrl.setNySample(3)
 
-        if False:                       # INCREASE_NXNYSAMPLE is no longer supported post #2074
```
timj: 👍 for removing code that is explicitly stated to no longer work.

tests/testBackground.py
```diff
@@ -513,22 +458,8 @@ def testOnlyOneGridCell(self):
                 testval = afwMath.cast_BackgroundMI(backobj).getPixel(bctrl.getInterpStyle(), xpix, ypix)
                 self.assertAlmostEqual(testval/mean, 1)
 
-    def testTicket1681OffByOne(self):
```
timj: @RobertLuptonTheGood disabled the code in fbbd59207be0c0181213e72ca6baa901f183d3a3 but I'm not sure why he left it around.

tests/testMaskedImage.py
r-owen: Here are the docs I'm using:

    assertListEqual(list1, list2, msg=None)
    assertTupleEqual(tuple1, tuple2, msg=None)
        Tests that two lists or tuples are equal. If not, an error message is constructed that shows only the differences between the two. An error is also raised if either of the parameters are of the wrong type. These methods are used by default when comparing lists or tuples with assertEqual().

Based on this it is clear that `assertEqual` calls these specialized versions if appropriate, but less clear if that means `assertEqual` is usually preferred. I tend to prefer `assertEqual`, but I can see situations where the longer name might make it clearer what is going on.

timj: I think the key point here is that if it is important to the test that you are comparing lists then you use the specialist routine. If you only care that the entities you are comparing are equal in the `==` sense then you use the generic version.

tests/testCamGeomFitsUtils.py
```diff
@@ -125,20 +125,12 @@ def testDetectorBuilder(self):
...
-
-
-def run(shouldExit=False):
-    """Run the tests"""
-    lsst.utils.tests.run(suite(), shouldExit)
-
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testMaskedImagePersistence1.py
```diff
@@ -106,20 +106,12 @@ def testBoostPersistence(self):
```
r-owen: This test needs fixing. The two testX method call self.checkImages(...), which does nothing (the body is `pass` despite the comforting documentation). I suggest changing it to call assertMaskedImagesEqual, but that will require having the test case inherit from `lsst.utils.tests.TestCase`

tests/testCameraGeom.py
```diff
@@ -95,8 +98,7 @@ def testMakeCameraPoint(self):
 
                     # test == and !=
                     cp2 = camera.makeCameraPoint(pt, coordSys)
-                    self.assertTrue(cp == cp2)
-                    self.assertFalse(cp != cp2)
+                    self.assertEqual(cp, cp2)
```
timj: I think we still want that `assertFalse` line to test the `!=` operator.

tests/testSpatialCell.py
```diff
@@ -333,22 +333,12 @@ def testInsertCandidate(self):
 #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
```
r-owen: Please change `self.assertRaises(pexExcept.NotFoundError, tst)` at line 120 to `with self.assertRaises(...)` and delete the definition of `tst` (just move the body of that function to the body of the with statement)

tests/testCameraGeom.py
```diff
@@ -342,30 +345,12 @@ def assertCamPointAlmostEquals(self, cp1, cp2, ndig=6):
...
-    if display:
-        ds9.setDefaultFrame(0)
-    utilsTests.run(suite(), exit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testCameraSys.py
```diff
@@ -38,47 +38,39 @@ def testBasics(self):
...
-                    self.assertTrue(cameraSys == noDetSys)
-                    self.assertFalse(cameraSys != noDetSys)
+                    self.assertEqual(cameraSys, noDetSys)
 
-                self.assertTrue(cameraSys != camSysPrefix)
-                self.assertTrue(noDetSys != camSysPrefix)
-                self.assertFalse(cameraSys == camSysPrefix)
-                self.assertFalse(noDetSys == camSysPrefix)
+                self.assertNotEqual(cameraSys, camSysPrefix)
+                self.assertNotEqual(noDetSys, camSysPrefix)
```
timj: It looks like @r-owen deliberately wanted explicit tests of `!=` and `==` here. See commit 03b91f2271b901654ac340375900e33df79d669c

tests/testMinimize.py
```diff
@@ -72,22 +72,19 @@ def testMinimize2(self):
 
         print("modelParams=", modelParams)
         print("fitParams  =", fitResults.parameterList)
-        self.assert_(fitResults.isValid, "fit failed")
+        self.assertTrue(fitResults.isValid, "fit failed")
         if not numpy.allclose(modelParams, fitResults.parameterList):
```
r-owen: I suggest using assertFloatsNotEqual (from lsst.utils.tests.TestCase) or the numpy.testing equivalent

tests/testColor.py
```diff
@@ -183,15 +165,13 @@ def testCtorFromMetadata(self):
...
-        #
         # Check that we can clean up metadata
-        #
         afwImage.stripCalibKeywords(metadata)
         self.assertEqual(len(metadata.names()), 0)
 
     def testCalibEquality(self):
         self.assertEqual(self.calib, self.calib)
-        self.assertFalse(self.calib != self.calib)
+        self.assertFalse(self.calib != self.calib)  # using assertFalse to directly test =! operator
```
timj: typo in comment `=!` but I like the comment there to stop people trying to delete the line.

tests/testColor.py
```diff
@@ -218,42 +198,38 @@ def testCalibFromCalibs(self):
...
         self.calib.setFluxMag0(1e12)
 
-        funcs = [lambda: self.calib.getMagnitude(-10), lambda: self.calib.getMagnitude(-10, 1)]
-
-        for func in funcs:
-            self.assertRaises(pexExcept.DomainError, func)
+        with self.assertRaises(pexExcept.DomainError):
+            self.calib.getMagnitude(-10)
+        with self.assertRaises(pexExcept.DomainError):
+            self.calib.getMagnitude(-10, 1)
```
timj: Much clearer code. Thanks.

tests/testColor.py
```diff
@@ -308,11 +275,6 @@ def defineFilterProperty(self, name, lambdaEff, force=False):
     def testListFilters(self):
         self.assertEqual(afwImage.Filter_getNames(), self.filters)
 
-    def testCtor(self):
```
timj: Why has this test been deleted?

tests/testConvolve.py
```diff
@@ -39,7 +39,7 @@
 import numpy
 
 import lsst.utils
-import lsst.utils.tests as utilsTests
+import lsst.utils.tests
 import lsst.pex.logging as pexLog
 import lsst.afw.geom as afwGeom
 import lsst.afw.image as afwImage
```
timj: The kernel import on line 50 seems to be completely broken.

tests/testApCorrMap.py
timj: Also, if it is rewritten as:
```python
with self.assertRaises(lsst.pex.exceptions.NotFoundError):
    self.map["e"]
```
I think it's much clearer what is happening (assuming you don't really need that `__getitem__` call to be explicit).

tests/testCoord.py
```diff
@@ -689,19 +696,20 @@ def testTicket3093(self):
 
     def testEquality(self):
         # (In)equality is determined by value, not identity. See DM-2347, -2465.
+        # These asserts are testing the functionality of `==` and `!=` and should not be changed
         c1 = afwCoord.IcrsCoord(self.ra, self.dec)
-        self.assertTrue(c1 == c1)
-        self.assertFalse(c1 != c1)
+        self.assertTrue(c1==c1)
+        self.assertFalse(c1!=c1)
```
timj: Put the spaces back.

tests/testCoord.py
```diff
@@ -689,19 +696,20 @@ def testTicket3093(self):
...
 
         c2 = afwCoord.IcrsCoord(self.ra.replace('1', '2'), self.dec)
-        self.assertFalse(c2 == c1)
-        self.assertTrue(c2 != c1)
+        self.assertTrue(c2!=c1)
+        self.assertFalse(c2==c1)
```
timj: Spaces around the operator.

tests/testCoordinates.py
```diff
@@ -40,21 +40,25 @@
...
 #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 
 
-class CoordinateTestCase(unittest.TestCase):
+class CoordinateTestCase(object):
+    """Mixin for some of the tests below.
+    """
 
-    def assertClose(self, a, b, msg=None):
+    def _assertClose(self, a, b, msg=None):
```
timj: @parejkoj would you like to comment on why this variant is required?

tests/testCoordinates.py
```diff
@@ -40,21 +40,25 @@
...
+        """Note that this will not be used by all tests, only by those
+           that derive from this class. The rest will use assertClose
```
timj: Replace mention of `assertClose`.

tests/testCoordinates.py
timj: It seems to me that this function is always comparing whether the contents of tuples are close so I think we could do with a better name.

tests/testCameraSys.py
r-owen: Good catch! Testing the operator== and operator!= is a bit tricky and best done explicitly rather than hoping that `assertEqual` and `assertNotEqual` do the right thing.

tests/testDisplay.py
```diff
@@ -172,35 +172,15 @@ def testInteract(self):
...
 
-def run(shouldExit=False):
-    """Run the tests"""
+def setup_module(module):
+    lsst.utils.tests.init()
 
-    tests.run(suite(), shouldExit)
 
 if __name__ == "__main__":
-    import argparse
```
timj: Please add this argument parsing code back in. It doesn't hurt anything and we have no reason to display the ability to specify a display backend on the command line.

tests/testDs9.py
```diff
@@ -140,20 +140,12 @@ def testStretch(self):
...
-    """Run the tests"""
-
-    tests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testExposure.py
```diff
@@ -443,13 +443,13 @@ def testDeepCopyData(self):
         miCopy.getMask().set(2)
         miCopy.getVariance().set(175)
 
-        self.assertTrue(numpy.allclose(miCopy.getImage().getArray(), -50))
+        self.assertClose(miCopy.getImage().getArray(), -50)
```
timj: Deprecated. Use `assertFloatsAlmostEqual`.

tests/testFootprint1.py
```diff
@@ -115,6 +115,8 @@ def testLessThan(self):
         span7 = afwDetect.Span(42, 0, 200)
         span8 = afwDetect.Span(42, 0, 100)
 
+        # Cannot use assertLess and friends here
+        # because Span only has operator <
```
timj: Can someone file a ticket to add the missing operator to Span?

tests/testConvolve.py
r-owen: I suspect tests/kernel.py was renamed to tests/testKernel.py, which would have broken this. It would be safer to move the shared code to afw test utils, but that may be out of scope for this ticket.

tests/testFootprintMergeCatalog.py
```diff
@@ -122,23 +140,22 @@ def testMerge1(self):
         pixArea.fill(69)
         pixArea[1] = 135
         measArea = [i.getFootprint().getArea() for i in merge]
-        self.assert_(np.all(pixArea == measArea))
+        np.testing.assert_array_equal(pixArea, measArea)
```
timj: @parejkoj can we replace this with something from `utils`?

tests/testHeader.py
```diff
@@ -58,28 +58,20 @@ def testHeaders(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testHeavyFootprint.py
```diff
@@ -317,19 +317,12 @@ def testDot(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    tests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: new line.

tests/testImageIo1.py
```diff
@@ -176,20 +176,12 @@ def testLongStrings(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testImageIo2.py
```diff
@@ -68,20 +68,12 @@ def testIo(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testCameraSys.py
r-owen: Please restore `assertNotEqual(foo != bar)`, preferably with a comment such as `# use this instead of assertNotEqual to test operator!=`.

You can probably retain `assertEqual(foo, bar)` instead of `assertTrue(foo == bar) as that is surely a safe substitution.

tests/testImagePersistence1.py
```diff
@@ -143,20 +143,12 @@ def testBoostPersistenceU16(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: new line.

tests/testImagePickle.py
```diff
@@ -106,20 +106,12 @@ def testMaskedImage(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testInterpolate.py
```diff
@@ -143,20 +143,12 @@ def testInvalidInputs(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line

tests/testKernel.py
```diff
@@ -316,7 +312,7 @@ def testLinearCombinationKernelAnalytic(self):
 
         kParams = [0.0]*len(basisKernelList)
         kernel = afwMath.LinearCombinationKernel(basisKernelList, kParams)
-        self.assert_(not kernel.isDeltaFunctionBasis())
+        self.assertTrue(not kernel.isDeltaFunctionBasis())
```
timj: Use `assertFalse` and remove the `not`.

tests/testKernel.py
```diff
@@ -283,7 +279,7 @@ def testComputeImageRaise(self):
             kernelList = afwMath.KernelList()
             kernelList.append(analKernel)
             lcKernel = afwMath.LinearCombinationKernel(kernelList, [1])
-            self.assert_(not lcKernel.isDeltaFunctionBasis())
+            self.assertTrue(not lcKernel.isDeltaFunctionBasis())
```
timj: Use `assertFalse` and remove the `not`.

tests/testKernel.py
```diff
@@ -538,8 +533,8 @@ def testSVLinearCombinationKernelFixed(self):
         )
 
         kernel = afwMath.LinearCombinationKernel(basisKernelList, spFunc)
-        self.assert_(not kernel.isDeltaFunctionBasis())
-        self.basicTests(kernel, 2, 3)
+        self.assertTrue(not kernel.isDeltaFunctionBasis())
```
timj: `assertFalse` and remove `not`.

tests/testKernel.py
```diff
@@ -695,37 +690,45 @@ def testRefactorGaussianLinearCombinationKernel(self):
...
 
     def basicTests(self, kernel, nKernelParams, nSpatialParams=0, dimMustMatch=True):
         """Basic tests of a kernel"""
-        self.assert_(kernel.getNSpatialParameters() == nSpatialParams)
-        self.assert_(kernel.getNKernelParameters() == nKernelParams)
+        self.assertEqual(kernel.getNSpatialParameters(), nSpatialParams)
+        self.assertEqual(kernel.getNKernelParameters(), nKernelParams)
         if nSpatialParams == 0:
-            self.assert_(not kernel.isSpatiallyVarying())
+            self.assertTrue(not kernel.isSpatiallyVarying())
```
timj: `assertFalse` and remove `not`

tests/testKernelImagesForRegion.py
```diff
@@ -153,9 +153,9 @@ def testComputeNextRow(self):
         nx = 6
         ny = 5
         regionRow = mathDetail.RowOfKernelImagesForRegion(nx, ny)
-        self.assert_(not regionRow.hasData())
-        self.assert_(not regionRow.isLastRow())
-        self.assert_(regionRow.getYInd() == -1)
+        self.assertTrue(not regionRow.hasData())
```
timj: These are negating to use `assertFalse`.

tests/testKernelIo1.py
```diff
@@ -405,20 +408,12 @@ def testSetCtr(self):
...
-def run(shouldExit=False):
-    """Run the tests"""
-    utilsTests.run(suite(), shouldExit)
+def setup_module(module):
+    lsst.utils.tests.init()
 
 if __name__ == "__main__":
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
```
timj: New line.

tests/testBackground.py
parejkoj: I'm happy to just delete this matplotlib block entirely. but I figured that it should be wrapped in something accessible from the outside.

tests/testBackground.py
parejkoj: We can delete this: print statements in tests should not be run by default, which is why I wrapped it in the "debugging" variable here.

tests/testBackground.py
parejkoj: It doesn't actually have an assert, so it's pointless anyway.

tests/testColor.py
parejkoj: There are no asserts and there are many other tests that make exactly that function call. It's not clear what the failure condition for this test would be, anyway.

tests/testCoordinates.py
parejkoj: I have no idea why it is required. @pschella made the changes here. I would try replacing all of the calls with assertNearlyEqual and kill this method.

tests/testFootprintMergeCatalog.py
parejkoj: utils.assert* don't work on lists or tuples (it's documented as "array-like"). The numpy methods do work on other generic sequences, so this seems fine to me here. The other option would be to do `np.array(pixArea)` inside one of our tests.

tests/testConvolve.py
parejkoj: Oh, fun. Tests that directly import other tests. Yuck. I'd suggest moving all of those functions that are wanted by multiple tests here into a kernelTestHelpers.py or something in this directory (with a docstring saying what it's for), and then importing that.

parejkoj: Oh, the reason it probably kept working is various kernel.pyc lying around.

tests/testBackground.py
timj: Depends if it is clear that `display` is a debugging variable.

tests/testBackground.py
timj: That's not correct. You don't need an assert to have a test. If the test code crashes then the test fails. Some tests really do exist purely to ensure that nothing breaks even if nothing is tested explicitly.

tests/testColor.py
timj: As I say above, `asserts` are not required in test code. It may be that the test code used to throw an exception but then something was fixed. Of course, if everything that is in this test is found in other tests then it is no longer needed.

tests/testConvolve.py
timj: Yes, although that doesn't explain how Jenkins worked.

tests/testBackground.py
parejkoj: If the test is explicitly checking whether something won't raise an exception, it should be wrapped like so:

```python
try:
    do_something()
except SomeError:
    self.fail("do_something() should not raise SomeError")
```

Otherwise, it's not clear what is being tested, if it fails it will be an "error" instead of a test "failure", and we can be clear about what exception we're checking for.

Since this test was disabled anyway, in an un-documented `if False`, we might as well delete it, since the test method was doing nothing.

parejkoj: No, it would be reported as an "Error", not a test "Failure". Those are explicitly different things, and it's easier to debug if we're explicit about Failing tests so that real errors (e.g. someone using the wrong call signature) are clearly distinct from failed tests.

If you really do want to allow tests without asserts or explicit failure conditions (e.g. ones that "Error" when something goes wrong), then they need to be clearly documented as such, not just a bunch of random function calls.

tests/testAngle.py
fred3m: :)

tests/testApCorrMap.py
fred3m: 👍

tests/testBackground.py
timj: That's no different to what would happen in some other test if something broke prior to the assert being called so I don't see what the problem is.

tests/testMaskedImage.py
fred3m: I think I did this before I knew about the assertFloatsEqual method, so I added the .flatten().tolist(). I'll update this to assertFloatsEqual, which will be much more clear.

tests/testMaskedImageIO.py
fred3m: No, these mask planes are different from the default and defined purely for this test, which is the reason this was causing the other tests to fail using pytest.

tests/testBackground.py
fred3m: No @timj, this block doesn't work with display==True! I agree with @parejkoj, it should just be deleted.

tests/testBackground.py
fred3m: Should I change `display` to `debugMode` or something similar? 

Alternatively, I could create a ticket to remove print functions in tests and replace them with lsst.log (once logging has been completed). This would eliminate the need for a debug variable and is the proper way to show debugging info.

For debug code that uses ds9, the module could check whether or not the logging level is <= DEBUG and only use ds9 if this is True.

tests/testMaskedImagePersistence1.py
fred3m: Good point, the class should inherit from lsst.utils.tests.TestCase anyway.

tests/testCameraGeom.py
fred3m: You're right, made this change before the discussion on the other ticket.

tests/testMaskedImage.py
parejkoj: In this case, they're images, so you should call `self.assertMaskedImagesEqual` on the images themselves, skipping the `.getArray()` and everything beyond. That also checks the mask and variance planes.

tests/testMinimize.py
fred3m: I think that fitsResults.isValid is a boolean that gives that state of the fit. The "fit failed" was a message used by assert_ to give the reason for the failure. I think this is ok as is, do you agree or am I missing something?

tests/testColor.py
fred3m: You can thank @parejkoj for these changes!

tests/testColor.py
fred3m: It looks like this functionality is already in several other tests, including the an exact replica in `testFilterEquality`, so I think it's ok to leave this deleted.

tests/testConvolve.py
fred3m: I've been wondering about this too. I know that Jenkins caches a ticket branch for future tests, so is it possible that someone ran a test before the filename was changed and that Jenkins didn't bother to remove the old kernel.pyc file either? I'm really not sure but this seems like a question we might want to have answered for the future.

tests/testMinimize.py
timj: I think @r-owen is talking about the `numpy.allclose` not the `assert_`.

fred3m: Of course... I wasn't paying close enough attention. Thanks for the clarification.

tests/testMaskedImage.py
fred3m: I think this function is explicitly checking that `getArrays` returns a proper image, mask, and variance.

parejkoj: OH! I get it. Please add a docstring to the method explaining that. It makes sense now, but took some thinking to understand, so it should be documented better.

fred3m:
Yes @timj, I'm not sure what happened there.

PR #85, opened 2016-08-23
Tickets/dm 7221

parejkoj:


timj:
Is this going to cause difficulties for #84's merge?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -416,6 +393,49 @@ class StandardReader : public FitsColumnReader {
...
+    ) :
+        _column(item.column),
+        _key(schema.addField<afw::geom::Angle>(item.ttype, item.doc, "", base))
+    {
+        // We require an LSST-specific key in the headers before parsing a column
+        // as Angle at all, so we don't need to worry about other units or other
+        // spellings of radians.  We do continue to support no units for backwards
+        // compatibility.
+        if (!item.tunit.empty() && item.tunit != "rad") {
+            throw LSST_EXCEPT(afw::fits::FitsError, "Angle fields must be persisted in radians.");
```
parejkoj: Maybe add "(tunit=rad)" to the exception message to specify how.

parejkoj: I guess supporting us reading in non-radian fields (e.g. from other data sets) is work for an other time?

tests/testTableIO.py
```diff
@@ -0,0 +1,62 @@
...
+import lsst.utils.tests
+import lsst.afw.geom
+import lsst.afw.table
+
+
+class TableIoTestCase(lsst.utils.tests.TestCase):
+
+    def testAngleUnitWriting(self):
+        """Test that Angle columns have TUNIT set appropriately,
+        as per DM-7221.
```
parejkoj: Yay helpful test docstrings!

tests/testTableArchives.cc
```diff
@@ -642,3 +644,31 @@ BOOST_AUTO_TEST_CASE(ArchiveImporter) {
...
+    MemFileManager manager;
+    Fits outFits(manager, "w", Fits::AUTO_CHECK);
+    outArchive.writeFits(outFits);
+    outFits.closeFile();
+    Fits inFits(manager, "r", Fits::AUTO_CHECK);
+    inFits.setHdu(3);
+    lsst::daf::base::PropertyList metadata;
+    inFits.readMetadata(metadata);
+    inFits.closeFile();
+    auto names = metadata.getArray<std::string>("AR_NAME");
```
parejkoj: Maybe just a short comment here saying that this is checking that there's exactly one of each? It's mostly clear from the code and function docstring, but it took me a bit to realize that's what this meant.

parejkoj: Oh, and would this code have failed before you made the fix?

tests/testTableIO.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        schema = lsst.afw.table.Schema()
+        key = schema.addField("a", type="Angle", doc="angle field")
+        outCat = lsst.afw.table.BaseCatalog(schema)
+        outCat.addNew().set(key, 1.0*lsst.afw.geom.degrees)
```
parejkoj: Nice: make it degrees, but ensure it gets output as radians. Good test.

TallJimbo:
> Is this going to cause difficulties for #84's merge?

I don't think so; the only Python test code here is in a new file that uses the new boilerplate.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: > Maybe add "(tunit=rad)" to the exception message to specify how.

Done

> I guess supporting us reading in non-radian fields (e.g. from other data sets) is work for another time?

Yes.  I actually originally tried to do that here, but hard to roll it back - it turns out we're using non-Angle fields with non-radian units in `Wcs` persistence, and those changes broke that.

tests/testTableArchives.cc
TallJimbo: I'll add the comment.  I'm pretty certain this test would have failed before the fix, but I didn't actually do it in that order.

tests/testTableIO.py
TallJimbo: Thanks!  I'd like to think mine are usually okay, but I have to admit that I was thinking of the conversation we had at the hack session last week when I included the ticket number.

PR #86, opened 2016-08-24
Add temporary workaround for DM-7377

timj:
Introduce a workaround to prevent infinite recursion in the swig
catalog bindings

### DIFF
--- a/python/lsst/afw/table/Catalog.i
+++ b/python/lsst/afw/table/Catalog.i
@@ -255,7 +255,8 @@ public:
             # self._columns is created the when self.columns is accessed -
             # looking for it in self.columns below would trigger infinite
             # recursion.
-            if name == "_columns":
+            # Test for "this" as a temporary workaround for ticket 7377
+            if name in ("_columns", "this"):
                 raise
         try:
             return getattr(self.table, name)


python/lsst/afw/table/Catalog.i
```diff
@@ -255,7 +255,8 @@ public:
             # self._columns is created the when self.columns is accessed -
             # looking for it in self.columns below would trigger infinite
             # recursion.
-            if name == "_columns":
+            # Test for "this" as a temporary workaround for ticktet 7377
```
timj: Completely unimportant spelling mistake in this comment...

timj: Do we know where "this" is coming from and why trapping it is a good idea?

TallJimbo: I'm fairly certain it's coming from Swig: I think it must be trying to get it as an attribute before it exists, while wrapping that in a `try/except` block.  But that only works if `AttributeError` is actually raised when it doesn't exist (as opposed to infinite recursion).

It might be a good idea to look at the Swig generated code (if we can get enough of a traceback to find it) to confirm that this is the case, but we could defer that to DM-7377.

timj: Highly likely given this code in the `.py` file:
```python
def _swig_setattr_nondynamic(self, class_type, name, value, static=1):
    if (name == "thisown"):
        return self.this.own(value)
    if (name == "this"):
        if type(value).__name__ == 'SwigPyObject':
            self.__dict__[name] = value
            return
```
so SWIG is using a "this" attribute.

PR #88, opened 2016-08-25
DM-6296: Wrap geom and coord with pybind11

pschella:


python/lsst/afw/geom/angle.cc
```diff
@@ -0,0 +1,125 @@
...
+
+    /* Non members */
+    mod.attr("PI") = py::float_(PI);
+    mod.attr("TWOPI") = py::float_(TWOPI);
+    mod.attr("HALFPI") = py::float_(HALFPI);
+    mod.attr("ONE_OVER_PI") = py::float_(ONE_OVER_PI);
+    mod.attr("SQRTPI") = py::float_(SQRTPI);
+    mod.attr("INVSQRTPI") = py::float_(INVSQRTPI);
+    mod.attr("ROOT2") = py::float_(ROOT2);
+    mod.attr("NullAngle") = py::cast(new Angle(-1000000., degrees));
```
timj: Why can't this use the definition from the `Angle.h` header file? It seems dangerous to define it in C++ and in Python.

python/lsst/afw/geom/extent.cc
```diff
@@ -0,0 +1,228 @@
...
+
+    // And then the types
+    auto clsExtent2I = declareExtent2<int>(mod, "I");
+    auto clsExtent3I = declareExtent3<int>(mod, "I");
+    auto clsExtent2D = declareExtent2<double>(mod, "D");
+    auto clsExtent3D = declareExtent3<double>(mod, "D");
+
+    /* Operators */
+
+    // Python's integer division works differently that C++'s for negative numbers - Python
```
timj: s/that/than/

python/lsst/afw/geom/extent.cc
```diff
@@ -0,0 +1,228 @@
...
+    // uses floor (rounds towards more negative), while C++ truncates (rounds towards zero).
+    // Therefore one needs to be carefull in the definition of division operators.
+    // Also note that pybind11 tries operators (like any overload) `in order'. So int has
+    // to come before double if specialization is needed.
+    clsExtent2I.def("__floordiv__", [](Extent<int,2> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent3I.def("__floordiv__", [](Extent<int,3> &e, int val) { return floor(e / static_cast<double>(val)); });
```
timj: are these meant to return doubles or ints?

python/lsst/afw/geom/extent.cc
```diff
@@ -0,0 +1,228 @@
...
+
+    clsExtent2I.def("__div__", [](Extent<int,2> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent3I.def("__div__", [](Extent<int,3> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent2I.def("__div__", [](Extent<int,2> &e, double val) { return e / val; });
+    clsExtent3I.def("__div__", [](Extent<int,3> &e, double val) { return e / val; });
+    clsExtent2D.def("__div__", [](Extent<double,2> &e, double val) { return e / val; });
+    clsExtent3D.def("__div__", [](Extent<double,3> &e, double val) { return e / val; });
```
timj: It's a bit annoying that we need to define all the `__div__` items her as clones of trudiv and floordiv.

python/lsst/afw/geom/extent.cc
```diff
@@ -0,0 +1,228 @@
...
+    // Also note that pybind11 tries operators (like any overload) `in order'. So int has
+    // to come before double if specialization is needed.
+    clsExtent2I.def("__floordiv__", [](Extent<int,2> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent3I.def("__floordiv__", [](Extent<int,3> &e, int val) { return floor(e / static_cast<double>(val)); });
+
+    clsExtent2I.def("__div__", [](Extent<int,2> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent3I.def("__div__", [](Extent<int,3> &e, int val) { return floor(e / static_cast<double>(val)); });
+    clsExtent2I.def("__div__", [](Extent<int,2> &e, double val) { return e / val; });
+    clsExtent3I.def("__div__", [](Extent<int,3> &e, double val) { return e / val; });
+    clsExtent2D.def("__div__", [](Extent<double,2> &e, double val) { return e / val; });
```
timj: So `ExtentD/int` automatically ends up in the `ExtentD/double` option?

python/lsst/afw/geom/functor.cc
```diff
@@ -0,0 +1,58 @@
...
+    py::module mod("_functor", "Python wrapper for afw _functor library");
+
+    py::class_<Functor> clsFunctor(mod, "Functor");
+
+    /* Operators */
+    clsFunctor.def("__call__", &Functor::operator());
+
+    /* Members */
+    clsFunctor.def("inverse", &Functor::inverse,
+        py::arg("y"), py::arg("tol")=1e-10, py::arg("maxiter")=1000);
```
timj: I don't like how these defaults can get out of sync with the defaults on the C++ side.

python/lsst/afw/geom/span.cc
```diff
@@ -0,0 +1,95 @@
...
+#include "lsst/afw/geom/Span.h"
+#include "lsst/afw/geom/SpanPixelIterator.h"
+
+namespace py = pybind11;
+
+using namespace lsst::afw::geom;
+
+// A thin wrapper around SpanPixelIterator.
+// Unfortunately we cannot use py::make_iterator here, as we normally
+// should, because for some reason the return values then all refer
```
timj: a bug in pybind11 or bug in SpanPixelIterator?

python/lsst/afw/geom/xyTransformFactory.py
```diff
@@ -53,6 +53,7 @@ class OneXYTransformConfig(Config):
 def makeInvertedTransform(config):
     """Make an InvertedXYTransform
     """
+
```
timj: whitespace change should be separate.

python/lsst/afw/geom/xyTransformFactory.py
```diff
@@ -100,7 +101,7 @@ def validate(self):
 def makeRadialXYTransform(config):
     """Make a RadialXYTransform
     """
-    return RadialXYTransform(config.coeffs)
+    return RadialXYTransform(config.coeffs._list)
```
timj: The `_list` method :smile:

python/lsst/afw/geom/xyTransformFactory.py
```diff
@@ -119,4 +120,3 @@ def makeMultiTransform(config):
     return MultiXYTransform(transformList)
 makeMultiTransform.ConfigClass = MultiXYTransformConfig
 xyTransformRegistry.register("multi", makeMultiTransform)
-
```
timj: whitespace change in separate commit.

tests/box.py
```diff
@@ -171,6 +171,12 @@ def testAccessors(self):
         self.assertEqual(box.getBeginY(), ymin)
         self.assertEqual(box.getEndX(), xmax + 1)
         self.assertEqual(box.getEndY(), ymax + 1)
+        a = box.getDimensions()
+        b = (pmax - pmin + geom.Extent2I(1))
+
+        print a, b
```
timj: if you are adding prints please make them functions.

tests/box.py
```diff
@@ -186,7 +192,7 @@ def testAccessors(self):
         
     def testRelations(self):
         box = geom.Box2I(geom.Point2I(-2,-3), geom.Point2I(2,1), True)
-        self.assertNotEqual(box, (3,4,5)) # should not throw
+#        self.assertNotEqual(box, (3,4,5)) # should not throw
```
timj: does this no longer work?

tests/temporary_disabled_tests/coord.py
```diff
@@ -1,788 +0,0 @@
-#!/usr/bin/env python2
```
timj: This deletion is presumably in the wrong commit as it's not related to geom.

tests/testXYTransform.py
```diff
@@ -220,7 +220,7 @@ def testRadial(self):
         """
         radialClass = xyTransformRegistry["radial"]
         radialConfig = radialClass.ConfigClass()
-        radialConfig.coeffs = (0, 1.05, 0.1)
+        radialConfig.coeffs = [0, 1.05, 0.1]
```
timj: the standard list versus tuple problem :smile:

python/lsst/afw/coord/coord.cc
```diff
@@ -0,0 +1,300 @@
...
+
+    mod.def("makeCoordEnum", makeCoordEnum);
+
+    py::class_<Coord, std::shared_ptr<Coord>> clsCoord(mod, "Coord");
+
+    /* Constructors */
+    clsCoord.def(py::init<lsst::afw::geom::Point2D const, lsst::afw::geom::AngleUnit, double const>(),
+        py::arg("p2d"), py::arg("unit") = lsst::afw::geom::degrees, py::arg("epoch") = 2000.0);
+    clsCoord.def(py::init<lsst::afw::geom::Point3D const, double const, bool, lsst::afw::geom::Angle const>(),
+        py::arg("p2d"), py::arg("epoch") = 2000.0, py::arg("normalize") = true, py::arg("defaultLongitude") = lsst::afw::geom::Angle(0.));
```
timj: I will grumble some more about duplicated default argument values but I guess we can't do anything about it.

python/lsst/afw/coord/coord.cc
```diff
@@ -0,0 +1,300 @@
...
+    clsFk5Coord.def(py::init<lsst::afw::geom::Point3D const &, double const, bool, lsst::afw::geom::Angle const>(),
+        py::arg("p3d"), py::arg("epoch") = 2000.0, py::arg("normalize") = true, py::arg("defaultLongitude") = lsst::afw::geom::Angle(0.));
+    clsFk5Coord.def(py::init<lsst::afw::geom::Angle const, lsst::afw::geom::Angle const, double const>(),
+        py::arg("ra"), py::arg("dec"), py::arg("epoch") = 2000.0);
+    clsFk5Coord.def(py::init<std::string const, std::string const, double const>(),
+        py::arg("ra"), py::arg("dec"), py::arg("epoch") = 2000.0);
+    clsFk5Coord.def(py::init<>());
+
+    /* Operators */
+    clsFk5Coord.def("__str__", [](Fk5Coord &c) {
```
timj: maybe we should add a `toString` method to Coord like most other afw classes have and then this would be less repetitive.

python/lsst/afw/coord/observatory.cc
```diff
@@ -0,0 +1,58 @@
...
+    clsObservatory.def(py::init<lsst::afw::geom::Angle const, lsst::afw::geom::Angle const,  double const>());
+    clsObservatory.def(py::init<std::string const, std::string const, double const>());
+
+    /* Operators */
+    clsObservatory.def("__str__", [](lsst::afw::coord::Observatory &o) {
+        std::ostringstream os;
+        os << o;
+        return os.str();
+    });
+    clsObservatory.def("__repr__", [](lsst::afw::coord::Observatory &o) {
```
timj: if there was a `toString` in the C++ this code duplication would be reduced and you could just assign `__repr__` to `__str__` (ignoring the fact that `repr()` should give you a string that you can use to reconstruct the object).

tests/coord.py
```diff
@@ -0,0 +1,791 @@
+#!/usr/bin/env python2
```
timj: It's not possible to work out what had to change in the tests as the file is entirely new.

python/lsst/afw/geom/span.cc
TallJimbo: A different discussion with @pschella led me to look up the iterator category for `SpanPixelIterator`, and it is indeed incorrectly set to `random_access_iterator` rather than either `input_iterator` or `forward_iterator` (I haven't thought hard enough about this particular case to figure out which it ought to be yet).  I'm not certain fixing that will fix the problem; pybind11 would have to have different implementations for different iterator categories (presumably as an optimization) for this to matter, but it could be what's going wrong.  In any case, our code should be fixed, and it'd be worth seeing if that allows `py::make_iterator` to work properly.

python/lsst/afw/coord/coord.cc
TallJimbo: I can think of two possible ways around this:
 - Instead of specifying the default arguments in this way, we wrap each different version of the function (corresponding to different numbers of arguments provided) as a different overload.  That would be *very* verbose in our wrapper definition code, and depending on how pybind11 does its overload resolution, it could degrade the performance of the wrappers slightly.
 - We could (as a matter of convention) define static data members for all of our default values (or just any that are conceivably subject to change), and use those in both the wrapper definitions and the original function signatures.  I'm slightly worried about [static initialization order problems](https://isocpp.org/wiki/faq/ctors#static-init-order), but we can probably come up with a safe approach.

python/lsst/afw/geom/angle.cc
pschella: It can indeed.

python/lsst/afw/geom/extent.cc
pschella: No, they return ints. Making that more clear with lambda return value now.

python/lsst/afw/geom/extent.cc
pschella: Yes it is :(

python/lsst/afw/geom/extent.cc
pschella: yes

python/lsst/afw/geom/functor.cc
pschella: I don't like it either. But there is no way to fix this without introducing some variable to hold the defaults on the C++ side. Unfortunately default arguments are not part of the function type information so pybind11 cannot get at them.

timj: I guess the important part is that `pydoc` returns these defaults and not the defaults found in the C++ header file.

python/lsst/afw/geom/span.cc
pschella: I don't think this is the problem. At least pybind11 doesn't seem to know about iterator categories. It must somehow be a problem with our boost-iterator-facade using implementation and pybind11 expecting a standard STL iterator. It must behave differently somehow (although I also think it really shouldn't). But I'm not sure. Also not sure if we really want to dig in deep here (now at least). I suggest we reevaluate when we have stripped out Boost.

python/lsst/afw/geom/xyTransformFactory.py
pschella: Yes, unfortunately exposing the internal `_list` is needed (at least temporarily) because pybind11 doesn't accept all sequence types and just knows about actual Python lists.

python/lsst/afw/geom/span.cc
timj: I'll leave the decision on when/where to pursue this to @TallJimbo.

tests/testXYTransform.py
pschella: Yes :(

python/lsst/afw/coord/coord.cc
pschella: I will keep it as is now. But will consider RFC the proposed fix in the future.

python/lsst/afw/coord/coord.cc
pschella: Standardising on a `__str__` (and `__repr__`) equivalent in C++ would be nice.

python/lsst/afw/coord/observatory.cc
pschella: Indeed.

pschella: Not sure if adding it is in-scope for this ticket though.

timj: I would argue "yes" on the basis that it simplifies the pybind11 binding. I'm happy for @jdswinbank to say to punt it.

python/lsst/afw/geom/functor.cc
pschella: Yes it does.

PR #89, opened 2016-09-01
DM-5503: Implement single interface to sanitized exposure metadata

timj:


python/lsst/afw/image/makeVisitInfo.py
```diff
@@ -0,0 +1,93 @@
...
+    boresightAzAlt=Coord(nanAngle, nanAngle),
+    boresightAirmass=nanFloat,
+    boresightRotAngle=nanAngle,
+    rotType=RotType_UNKNOWN,
+    observatory=Observatory(nanAngle, nanAngle, nanFloat),
+    weather=Weather(nanFloat, nanFloat, nanFloat),
+):
+    """Make a VisitInfo from keyword arguments
+
+    This function will be replaced by a VisitInfo constructor once we switch to pybind11
```
PaulPrice: I'm surprised pybind11 will make this any easier.

src/image/VisitInfo.cc
```diff
@@ -0,0 +1,440 @@
...
+    auto weather = visitInfo.getWeather();
+    setDouble(metadata, "AIRTEMP", weather.getAirTemperature(), "Outside air temperature (C)");
+    setDouble(metadata, "AIRPRESS", weather.getAirPressure() , "Outdoor air pressure (P)");
+    setDouble(metadata, "HUMID", weather.getHumidity(), "Relative humidity (fraction)");
+}
+
+}  // lsst::afw::image::detail
+
+
+VisitInfo::VisitInfo(daf::base::PropertySet const & metadata) :
```
PaulPrice: I see that the makeVisitInfo factory function will do the job, so ignore this.

src/coord/Weather.cc
```diff
@@ -0,0 +1,72 @@
...
+}
+
+bool Weather::operator==(Weather const & other) const {
+    return (_airTemperature == other.getAirTemperature()
+            && _airPressure == other.getAirPressure()
+            && _humidity == other.getHumidity());
+}
+
+void Weather::validate() const {
+    if ((_humidity < 0.0) || (_humidity > 1.0)) {
```
r-owen: I'll hope we can live with it. But the main goal is to avoid folks trying to set % instead of fraction, and I suppose the upper limit could be increased a bit of necessary.

tests/testWeather.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python2
```
r-owen: I don't think we allow executable tests anymore anyway, alas. I'll remove it.

tests/testWeather.py
```diff
@@ -0,0 +1,74 @@
...
+                # test == (using a copy, to make sure the test is not based on identity) and !=
+                self.assertEqual(weather, weatherCopy)
+                if prevWeather is not None:
+                    self.assertNotEqual(weather, prevWeather)
+                prevWeather = weather
+
+    def testBadHumidity(self):
+        """Check bad humidity handling (humidity is the only value whose range is checked)"""
+        for humidity in (-1, -0.0001, 1.00001, 99):
+            with self.assertRaises(Exception):
```
r-owen: OK

tests/testWeather.py
```diff
@@ -0,0 +1,74 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+"""
+Tests for lsst.afw.cameraGeom.Detector
```
r-owen: Good

include/lsst/afw/coord/Observatory.h
```diff
@@ -47,19 +47,40 @@ namespace coord {
...
 
+    /*
+     * @brief Constructor for the observatory with lat/long as strings
+     *
+     * @param[in] longitude  observatory longitude (dd:mm:ss.s, positive values are E of Greenwich)
+     * @param[in] latitude  observatory latitude  (dd:mm:ss.s)
+     * @param[in] elevation  observatory elevation (meters above reference spheroid)
+     *
+     */
+    Observatory(std::string const longitude, std::string const latitude, double const elevation);
```
r-owen: Old code. I'm not keen on constructing these from strings anyway. But adding & seems logical and I'll do it.

include/lsst/afw/image/VisitInfo.h
```diff
@@ -0,0 +1,217 @@
...
+class VisitInfo : public table::io::PersistableFacade<VisitInfo>, public table::io::Persistable {
+public:
+
+    /**
+     * Construct a VisitInfo
+     *
+     * @param[in] exposureId  exposure ID
+     * @param[in] exposureTime  exposure duration (sec)
+     * @param[in] date  TAI (international atomic time) MJD date at middle of exposure
+     * @param[in] ut1  UT1 (universal time) MJD date at middle of exposure
```
r-owen: No. UT1 is not a uniform time and DateTime cannot handle it, which is why UT1 is a float instead of a DateTime. Computing UT1 requires earth orientation predictions for the UTC date in question, e.g. as obtained from IERS Bulletin B. This information is a pain to access after the fact, and is best provided at the time the observations are taken (since telescope control systems require the information in order to track accurately).

Note: it is possible to compute earth rotation angle from UT1, but not the other way around. Since a header may include one or the other (typically neither, alas) it's safer to just have slots and provide the function elsewhere, e.g. as a method on MakeRawVisitInfo (I probably should have added that, but haven't need it yet).

include/lsst/afw/image/VisitInfo.h
```diff
@@ -0,0 +1,217 @@
...
+        _boresightRaDec(boresightRaDec),
+        _boresightAzAlt(boresightAzAlt),
+        _boresightAirmass(boresightAirmass),
+        _boresightRotAngle(boresightRotAngle),
+        _rotType(rotType),
+        _observatory(observatory),
+        _weather(weather)
+    {};
+
+    explicit VisitInfo(daf::base::PropertySet const & metadata);
```
r-owen: I have a markdown document describing almost all our FITS keywords. I just need some place to serve it.

include/lsst/afw/image/VisitInfo.h
```diff
@@ -0,0 +1,217 @@
...
+
+namespace detail {
+
+/**
+ * Set FITS metadata from a VisitInfo
+ *
+ * @param[in,out] metadata  FITS keyword metadata to set
+ * @param[in] visitInfo  instance of VisitInfo from which to set metadata
+ */
+void setVisitInfoMetadata(daf::base::PropertyList &metadata, VisitInfo const &visitInfo);
```
r-owen: I copied how Calib did it. I don't have strong feelings, but prefer to keep the API simple and use external free functions when it makes sense to do so, and FITS arcana may be such a situation.

src/image/VisitInfo.cc
```diff
@@ -0,0 +1,440 @@
...
+    os << "Unknown RotType name: \"" << rotTypeName << "\"";
+    throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, os.str());
+}
+
+class VisitInfoSchema {
+public:
+    table::Schema schema;
+    table::Key<table::RecordId> exposureId;
+    table::Key<double> exposureTime;
+    table::Key<std::int64_t> tai;
```
r-owen: I wish. Unfortunately double precision isn't accurate enough to maintain full precision. That is why DateTime uses nsecs from the unix epoch internally (and one reason it has a rather narrow valid range). Furthermore, the handling of invalid DateTime works using a special value of nsecs and was designed with afw table persistence in mind.

python/lsst/afw/image/makeVisitInfo.py
r-owen: pybind11 has support for wrapping existing code while providing named arguments. SWIG can apparently do it, but I don't know how. I did spend some time trying to %extend the class in SWIG to provide a keyword-based constructor, but was not successful and it's probably just as well, since it would have wrapped the same constructor that SWIG was already exposing without keyword arguments.

src/image/VisitInfo.cc
```diff
@@ -0,0 +1,440 @@
...
+    os << "Unknown RotType enum: " << static_cast<int>(rotType);
+    throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, os.str());
+}
+
+/**
+ * Get rotation type as an enum, given the FITS keyword value string equivalent
+ *
+ * @throw lsst::pex::exceptions::RuntimeError if the string is not recognized
+ */
+RotType rotTypeEnumFromStr(std::string rotTypeName) {
```
r-owen: Good catch.

src/coord/Weather.cc
```diff
@@ -0,0 +1,72 @@
...
+    _airPressure(airPressure),
+    _humidity(humidity)
+{
+    validate();
+}
+
+bool Weather::operator==(Weather const & other) const {
+    return (_airTemperature == other.getAirTemperature()
+            && _airPressure == other.getAirPressure()
+            && _humidity == other.getHumidity());
```
r-owen: I prefer the format I used as more readable, though I definitely see your point (and I used to code that way for that reason).

src/coord/Weather.cc
r-owen: I don't think telescopes run in supersaturated conditions. But I also think I should switch to percent for relative humidity, as the unit of least surprise.

tests/testWeather.py
```diff
@@ -0,0 +1,74 @@
...
+            with self.assertRaises(Exception):
+                Weather(1.1, 2.2, humidity)
+
+
+class MemoryTester(lsst.utils.tests.MemoryTestCase):
+    pass
+
+
+if __name__ == "__main__":
+    lsst.utils.tests.init()
```
r-owen: Yes. I believe it is there now. However, I moved `setup_module` closer to the above to make it more obvious.

src/coord/Weather.cc
RobertLuptonTheGood: I thought that break after operator was in the coding standard.  It was certainly supposed to be!

Consistency is important, and the trailing operator makes it clear that the line is continued.

include/lsst/afw/image/VisitInfo.h
PaulPrice: Sounds like a job for Super Documentation Man.  Paging his mild-mannered alter ego, @jonathansick...

src/coord/Weather.cc
TallJimbo: I'm with @r-owen on this.  I've never understood the preference for break after operator.  Putting the operator on the next line makes it so much more like the way we write out mathematics.

As for safety, isn't that a moot point in a case where you've got parentheses on both sides?

All that said, the coding standards do *suggest* break after operator, though the implication is this is one of several considerations in making an expression easier to read (and I feel it almost never does).

timj: PEP8 agrees with @r-owen and @TallJimbo: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator -- the DM python style guide does not make a statement on this so PEP8 wins. Leading to inconsistency now between C++ and Python.

include/lsst/afw/image/VisitInfo.h
jonathansick: Ah yes, this follows from https://community.lsst.org/t/documenting-internal-designs-and-implementations-following-fits-data-model-discussion-from-hipchat/1160?u=jsick. I'm going to call the result of that the FITS keywords will be a page in the Science Pipelines guide (which in turn will have DOI'd and archived pages).

So my recommendation to @r-owen is:

- Commit the markdown doc to the `doc/` directory
- Make a ticket stating that the markdown document should be incorporated into the Science Pipelines documentation. Assign it to me, with a `Stack Documentation and UX` component. I'll be able to pick it up in a couple weeks when I start on a new stack documentation epic.

r-owen: Thanks, @jonathansick. I made it a .rst file because I wanted the table support. I hope that is OK. Please feel free to rework the document to improve clarity; the current table is wider than I would like. Also, I could not figure out how to get rst to make low-key headers for the two tables and the footnotes so I used bold formatting instead (markdown wins there; pity its table support is so lame).

Also, the document is by no means complete in that it punts on describing the binary tables that may be present. I intended it to be complete in in describing LSST-specific header keywords (I did not attempt to describe the standard keywords such as SIMPLE, nor the standard WCS keywords).

jonathansick: Thank you @r-owen for [DM-7790](https://jira.lsstcorp.org/browse/DM-7790). The rST is great. I'll figure out how to make reST tables great; the [linux kernel community](https://github.com/return42/linuxdoc/blob/master/linuxdoc/rstFlatTable.py) is doing some interesting stuff in this area.

r-owen:
@PaulPrice thank you for your many excellent suggestions. I have added darkTime to VisitInfo and applied all your other suggested fixes (except reformatting the lines with leading &&).

src/table/Exposure.cc
```diff
@@ -25,6 +29,14 @@ namespace lsst { namespace afw { namespace table {
 
 namespace {
 
+int const TableVersion = 2;  // current version of ExposureTable
+std::string const TableVersionKey = "TBL_VERS";  // FITS metadata key for ExposureTable version
```
r-owen: I'm going with @PaulPrice here because a short name that is unique will be illegible. There are standard ways to write long FITS header keys and I think we are simply going to have to figure out how to manage it. `VisitInfo` also saves information using some long key names.

src/image/Calib.cc
```diff
@@ -392,31 +393,41 @@ std::pair<ndarray::Array<double,1>, ndarray::Array<double,1> > Calib::getMagnitu
...
+        if (tableVersion == 1) {
+            // obsolete fields
+            midTime = schema.addField<std::int64_t>(
+                "midtime", "middle of the time of the exposure relative to Unix epoch", "ns"
+            );
+            expTime = schema.addField<double>(ExpTimeFieldName, "exposure time", "s");
+        }
+        fluxMag0 = schema.addField<double>("fluxmag0", "flux of a zero-magnitude object", "count");
+        fluxMag0Sigma = schema.addField<double>("fluxmag0.err", "1-sigma error on fluxmag0", "count");
         schema.getCitizen().markPersistent();
```
r-owen: Good catch!

src/image/Calib.cc
```diff
@@ -392,31 +393,41 @@ std::pair<ndarray::Array<double,1>, ndarray::Array<double,1> > Calib::getMagnitu
 
 namespace {
 
-class CalibSchema {
+int const TableVersion = 2;  // current version of ExposureTable
+std::string const ExpTimeFieldName = "exptime";  // name of exposure time field
```
r-owen: I can't see having constants that start with lowercase so I'll put up with UPPER_CASE

PR #94, opened 2016-09-13
DM-6985: Logging framework migration

hchiang2:


examples/warpExposure.py
```diff
@@ -28,7 +28,15 @@
...
-import lsst.pex.logging
+import lsst.log
+
+
+def traceSetAt(name, number):
+    """Display messages with trace number <= NUMBER"""
+    for i in range(1, 6):
+        level = lsst.log.INFO if i > number else lsst.log.DEBUG
+        lsst.log.Log.getLogger('TRACE%d.%s' % (i, name)).setLevel(level)
+
```
TallJimbo: This utility function seems pretty generic.  Maybe we should move it to the `log` package itself so it can be used elsewhere?

src/detection/PsfFormatter.cc
```diff
@@ -22,14 +22,13 @@
...
+#include "lsst/log/Log.h"
 #include "lsst/pex/policy/Policy.h"
 
 BOOST_CLASS_EXPORT(lsst::afw::detection::Psf)
 
-#define EXEC_TRACE  20
-static void execTrace(std::string s, int level = EXEC_TRACE) {
-    lsst::pex::logging::Trace("afw.detection.PsfFormatter", level, s);
+namespace {
+LOG_LOGGER _log = LOG_GET("afw.detection.PsfFormatter");
```
TallJimbo: Is it always safe to construct a logger at static initialization time?  I'm a bit worried about [static initialization order](https://isocpp.org/wiki/faq/ctors#static-init-order) here.

src/detection/Threshold.cc
```diff
@@ -115,7 +114,7 @@ double Threshold::getValue(ImageT const& image) const {
         math::Statistics stats = math::makeStatistics(image, math::STDEVCLIP);
         double const sd = stats.getValue(math::STDEVCLIP);
 
-        pexLogging::TTrace<3>("afw.detection", "St. Dev = %g", sd);
+        LOGLS_DEBUG("afw.detection.threshold", "St. Dev = " << sd);
```
TallJimbo: It looks like `LOGL_DEBUG("afw.detection.threshold", "St. Dev = %g", sd);` should also work, and that would be a smaller change from the original.

src/fits.cc
```diff
@@ -720,12 +720,11 @@ void writeKeyFromProperty(
...
         // FIXME: inherited this error handling from fitsIo.cc; need a better option.
-        pex::logging::Log::getDefaultLog().log(
-            pex::logging::Log::WARN,
+        LOGLS_WARN("afw.writeKeyFromProperty",
             makeErrorMessage(
                 fits.fptr, fits.status,
-                boost::format("In %s, unknown type '%s' for key '%s'.")
-                % BOOST_CURRENT_FUNCTION % valueType.name() % key
+                boost::format("Unknown type '%s' for key '%s'.")
+                % valueType.name() % key
```
TallJimbo: I assume you've dropped the `BOOST_CURRENT_FUNCTION` bit here because `LOGLS_WARN` automatically includes the function name?  If not, it seems like a regression.

src/detection/PsfFormatter.cc
TallJimbo: At first glance, it looks like this isn't safe: default-constructing a `Log` object invokes `lsst::log::log4cxxInit`, which checks the status of another global variable, `lsst::log::doBasicConfig`, which isn't guaranteed to have been initialized yet at that point.  I'm not *completely* positive that's a problem, because `doBasicConfig` is a bool, and `log4cxxInit` seems to do something sensible whether it's true or false (but its behavior does seem to depend on the undefined status of `doBasicConfig`, which is worrying).

You can fix this using the construct-on-first-use approach recommended in the link above, but I wonder if it would be possible to make it safe to construct `Log` instances at global initialization time, since a lot of people are probably going to try to do that.  Perhaps @andy-slac has an opinion on this?  It looks like there's already an attempt to make it safe to initialize loggers at global initialization time, and maybe I've just misread the code in thinking that it isn't.

src/image/Mask.cc
```diff
@@ -392,7 +391,7 @@ namespace {
  */
 template<typename MaskPixelT>
 void Mask<MaskPixelT>::_initializePlanes(MaskPlaneDict const& planeDefs) {
-    pexLog::Trace("afw.Mask", 5, boost::format("Number of mask planes: %d") % getNumPlanesMax());
+    LOGLS_DEBUG("afw.Mask", "Number of mask planes: " << getNumPlanesMax());
```
TallJimbo: Consider `LOGL_DEBUG` here too.

src/image/makeWcs.cc
```diff
@@ -62,12 +62,9 @@ afwImg::Wcs::Ptr afwImg::makeWcs(
     //
     // Follow Dave's AST and switch TAN to TPV
     //
-    using pex::logging::Log;
-    auto log = Log(Log::getDefaultLog(), "makeWcs");
-
     if (ctype1.substr(5, 3) == "TAN" &&
         (metadata->exists("PV1_5") || metadata->exists("PV2_1"))) {
-        log.log(Log::INFO, str(boost::format("Interpreting %s/%s + PVi_j as TPV") % ctype1 % ctype2));
+        LOGLS_INFO("makeWcs", "Interpreting " << ctype1 << "/" << ctype2 << " + PVi_j as TPV");
```
TallJimbo: I think `LOGL_DEBUG` with `sprintf`-style formatting would make this a bit more concise.

src/math/LeastSquares.cc
```diff
@@ -137,15 +134,16 @@ class EigensystemSolver : public LeastSquares::Impl {
         _eig.compute(fisher);
         if (_eig.info() == Eigen::Success) {
             setRank(_eig.eigenvalues().reverse());
-            log.debug<5>("SelfAdjointEigenSolver succeeded: dimension=%d, rank=%d", dimension, rank);
+            LOGL_DEBUG("TRACE4.afw.math.LeastSquares",
+                "SelfAdjointEigenSolver succeeded: dimension=%d, rank=%d", dimension, rank);
```
TallJimbo: I don't there's a need for the `TRACE4` prefix here; as long as there was only one level used in this file with this particular logging name, I don't think it matters if there were other levels used in `afw.math`.

src/math/SpatialCell.cc
```diff
@@ -86,9 +86,8 @@ SpatialCell::SpatialCell(std::string const& label, ///< string representing "nam
     _candidateList(candidateList),
     _ignoreBad(true)
 {
-    lsst::pex::logging::TTrace<3>("lsst.afw.math.SpatialCell",
-                                  "Cell %s : created with %d candidates",
-                                  this->_label.c_str(), this->_candidateList.size());
+    LOGL_DEBUG("TRACE2.afw.math.SpatialCell", "Cell %s : created with %d candidates",
+               this->_label.c_str(), this->_candidateList.size());
```
TallJimbo: I also don't think the `TRACE2` prefix is needed here.

I may find an exception, but in general I think `afw.math` is enough of a smorgasbord of different things that no one will ever want to enable logging on all `afw.math.*` at a particular level; they'll only care about enabling individual named sub-logs.

TallJimbo: I did find some exceptions: we should keep the `TRACE#` prefixes for `afw.math.convolve` and `afw.math.warp` loggers, since I think it's conceivable they could be directly enabled and they contain multiple levels.

tests/testLeastSquares.py
```diff
@@ -39,10 +39,11 @@
 
 import lsst.utils.tests as utilsTests
 import lsst.pex.exceptions
-import lsst.pex.logging
 from lsst.afw.math import LeastSquares
+from lsst.log import Log
+
+Log.getLogger("TRACE5.afw.math.LeastSquares").setLevel(Log.DEBUG)
```
TallJimbo: Just a note that `TRACE5` prefix can be removed here too.

src/image/Mask.cc
TallJimbo: Oh, and while this might be outside the scope of the original ticket, it seems like the log name ought to be `afw.image.Mask`, and this might be a good time to clean that up (both here and in the afw tests that enable this logger).

src/detection/PsfFormatter.cc
andy-slac: @TallJimbo, the initialization code as it is written now depends on `log` being built as a shared object, in a shared object you can only call methods/functions after all globals in that shared object have been (dynamically) initialized. You are right to doubt that approach, it's not very portable and it may break if someone decides to link that code statically. It can be made safer of course, either with function-static approach or depending only on static initialization. Feel free to open a ticket if you think it needs improvement.

TallJimbo: Ah, got it - I wasn't considering the guarantees about ordering that come from the linker.  I think the current approach is fine, then.  Thanks!

examples/warpExposure.py
hchiang2: [This utility function is moved to `log` in `utils.py`.](https://github.com/lsst/log/pull/15)

src/detection/Threshold.cc
hchiang2: Corrected.

src/fits.cc
hchiang2: log should include the function name. But I didn't mean to drop `BOOST_CURRENT_FUNCTION` here, so I'm adding it back. ( So no changes here)

src/image/Mask.cc
hchiang2: Used `LOGL_DEBUG` instead. And added another commit to rename this logger to `afw.image.Mask` along with the placeholders in tests.

src/image/makeWcs.cc
hchiang2: Corrected to use `LOGL_INFO` instead.

src/math/LeastSquares.cc
hchiang2: Dropped the `TRACE4` prefix for `afw.math.LeastSquares`.

src/math/SpatialCell.cc
hchiang2: Agree with what you said. Under `afw.math` loggers, these use one tracing/debugging level: 
`afw.math.KernelFormatter`
`afw.math.LeastSquares`
`afw.math.SpatialCell`
So no prefix is needed. 

These two use more than one level so the `TRACE#` prefixes are used: 
`afw.math.warp`
`afw.math.convolve`

tests/testLeastSquares.py
hchiang2: Corrected

PR #95, opened 2016-09-14
Use size_t and ptrdiff_t with ndarray

pschella:


timj:
My one comment is that I don't see any `ptrdiff_t` changes in the commits...

PR #96, opened 2016-09-16
DM-7549: GaussianProcess updates

timj:


tests/testGaussianProcess.py
```diff
@@ -35,8 +35,495 @@
...
+        """
+        This test will test the construction of KdTree in the pathological case
+        where many of the input data points are identical.
+        """
+        pp = 100
+        dd = 5
+        data = np.zeros((pp, dd), dtype=float)
+        tol = 1.0e-10
+
+        f = open(os.path.join(testPath, "data", "kd_test_data.sav"))
```
timj: Maybe use `with open` so you don't need the `close`?

tests/testGaussianProcess.py
```diff
@@ -35,8 +35,495 @@
...
+        ff = f.readlines()
+        f.close()
+        for i in range(len(ff)):
+            s = ff[i].split()
+            for j in range(dd):
+                data[i][j] = float(s[j])
+
+        kd = gp.KdTreeD()
+        try:
```
timj: What's the purpose of these try/catch blocks? Are you expecting the `pex.Exception` to happen, or are you trying to intercept an error that should not happen and print it out? If the former, why not use the more standard `with self.assertRaises`, if the latter, shouldn't this be a test failure? If it's something non-standard then there should be a comment explaining it before the first try.

danielsf: I think the latter.  I have removed the try/except blocks.

tests/testGaussianProcess.py
danielsf: done

danielsf:
This pull request adds tests to all of the user-facing methods in GaussianProcess.cc. These tests ensure (or ought to) that exceptions are thrown if users pass in invalid inputs.

I also took the opportunity to flesh-out testGaussianProcess.py so that the functionality of GaussianProcess.cc is tested more methodically.

I added some methods to the GaussianProcess class to allow users to inspect its contents (specifically methods to return the data points and function values underlying the GaussianProcess).

danielsf:
@jmeyers314 Just wanted to make sure you saw this

include/lsst/afw/math/GaussianProcess.h
```diff
@@ -603,6 +603,42 @@ class GaussianProcess {
                     std::shared_ptr< Covariogram<T> > const &covarIn);
 
     /**
+     * @brief return the number of data points stored in the GaussianProcess
+    */
+    int getPoints() const;
```
jmeyers314: Can we use getNPoints() or something like that perhaps?  getPoints sounds too much like getData to me.

I suppose this applies to the private variable `_pts` as well: -> `_npts` or something.

tests/testGaussianProcess.py
```diff
@@ -45,21 +739,21 @@ def testTooManyNeighbors(self):
...
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, 0, 2*nData)
+            gg.selfInterpolate(sigma, 0, 2*nData)
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, 0, -5)
+            gg.selfInterpolate(sigma, 0, -5)
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, -1, nData-1)
+            gg.selfInterpolate(sigma, -1, nData-1)
         # the following segfaults, for unknown reasons, so run directly instead
```
jmeyers314: Does this still segfault?

tests/testGaussianProcess.py
```diff
@@ -45,21 +739,21 @@ def testTooManyNeighbors(self):
...
+            gg.interpolate(sigma, test, -5)
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, 0, 2*nData)
+            gg.selfInterpolate(sigma, 0, 2*nData)
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, 0, -5)
+            gg.selfInterpolate(sigma, 0, -5)
         with self.assertRaises(pex.Exception):
-             gg.selfInterpolate(sigma, -1, nData-1)
+            gg.selfInterpolate(sigma, -1, nData-1)
```
jmeyers314: Part of what made the reported bug difficult to track before was that while most of the time this line did raise an exception, it was actually not the intended exception.  Is there anyway within pex.Exception to check that the right exception is being raised?  If so, no need to implement that everywhere, but probably a good idea to implement it here where we no there was a problem in the past.

danielsf: I can use `unittest`'s code to check the message that comes with the exception.  Probably a good idea.

danielsf:
Regarding the high-level comment: at the time I wrote this code, I think we had several kdTree implementations floating around, but hadn't picked one to be our "official" kdTree class (thus I was allowed to go ahead and implement one specifically for GaussianProcess).  I don't know if that is still the case, but you're right: picking a "blessed" kdTree implementation is probably something we should do.

tests/testGaussianProcess.py
danielsf: No.  I have restored the code block.

jmeyers314:
Sorry so slow, @danielsf .  Didn't realize you had completed these all 9 days ago.  I'm happy to merge now.

PR #97, opened 2016-10-03
afw/DM-7736: Avoid signed/unsigned integer comparison warnings.

TallJimbo:


include/lsst/afw/table/FieldBase.h
```diff
@@ -211,7 +211,7 @@ struct FieldBase< Array<U> > {
 
     template <typename Derived>
     void setValueDeep(Element * p, ndarray::ExpressionBase<Derived> const & value) const {
-        if (value.template getSize<0>() != _size) {
+        if (value.template getSize<0>() != static_cast<std::size_t>(_size)) {
```
kfindeisen: It might make more sense for `_size` to be `size_t`, rather than `int`. Or will this cause problems for `ndarray::makeVector`?

src/detection/FootprintSet.cc
```diff
@@ -720,7 +720,7 @@ static void findFootprints(
                     fp->addSpan(spans[i0]->y + row0, spans[i0]->x0 + col0, spans[i0]->x1 + col0);
                 }
 
-                if (good && !(fp->getNpix() < npixMin)) {
+                if (good && !(fp->getNpix() < static_cast<std::size_t>(npixMin))) {
```
kfindeisen: A bit illegible. Perhaps rephrase as `fp->getNpix() >= static_cast<std::size_t>(npixMin)` to at least get rid of one set of parentheses?

src/table/arrays.cc
```diff
@@ -133,7 +133,7 @@ ndarray::Array<T const,1,1> ArrayKey<T>::get(BaseRecord const & record) const {
 template <typename T>
 void ArrayKey<T>::set(BaseRecord & record, ndarray::Array<T const,1,1> const & value) const {
     LSST_THROW_IF_NE(
-        value.template getSize<0>(), _size,
+        value.template getSize<0>(), static_cast<std::size_t>(_size),
```
kfindeisen: Should `_size` be of type `size_t`?

src/detection/FootprintSet.cc
TallJimbo: Will do.

include/lsst/afw/table/FieldBase.h
TallJimbo: The signature of this `FieldBase` instantiation's constructor needs to match that of another instantiation for which negative values at schema-definition time are used to indicate a variable-size array field (i.e. a different size for each record, so this needs to stay as a signed `int`. Switching to 64-bits might be worthwhile, but I'd rather not do that on this ticket.

src/table/arrays.cc
TallJimbo: (same as above - it's probably a bit more important to have the same time as another field type for which negative values are meaningful)

include/lsst/afw/table/FieldBase.h
TallJimbo: Your instincts weren't at all wrong; it was actually a historical decision more than a deliberate one, and I liked your proposal to switch enough that I started looking into doing it here before I realized there was a closely-related type I couldn't easily change.  But it is more of a change than I'd like to do now.  I'm planning to do a bit of refactoring of afw::table in the not-too-distance future, and I'll include another look at all of these types in that.

PR #100, opened 2016-10-03
afw/DM-3549: transform features in support of new astrometry fitter

TallJimbo:


r-owen:
This looks like a very useful change

include/lsst/afw/geom/AffineTransform.h
```diff
@@ -109,7 +109,7 @@ class AffineTransform {
     ) : _linear(linear), _translation(translation) {}
 
     /** Construct an AffineTransform from a LinearTransform. */
-    explicit AffineTransform(LinearTransform const & linear)
```
TallJimbo: It allows you to pass a `LinearTransform` to functions that take an `AffineTransform` without an explicit conversion.  In particular, that includes overloaded operators, so it lets me do `linear1 + affine2` instead of `AffineTransform(linear1) + affine1`.

PR #102, opened 2016-10-03
Cleanup after rebase on Python 3 and pytest changes

pschella:


python/lsst/afw/coord/coord.cc
```diff
@@ -76,9 +76,9 @@ PYBIND11_PLUGIN(_coord) {
     });
 
     /* Members */
-    clsCoord.def("clone", &Coord::clone);
     clsCoord.def("reset", (void (Coord::*)(lsst::afw::geom::Angle const, lsst::afw::geom::Angle const)) &Coord::reset);
     clsCoord.def("reset", (void (Coord::*)(lsst::afw::geom::Angle const, lsst::afw::geom::Angle const, double const)) &Coord::reset);
+    clsCoord.def("clone", &Coord::clone);
```
timj: Did this need to move?

pschella: No, but I suggest to keep it moved now to not upset Fred's rebasing.

PR #104, opened 2016-10-14
DM-7976 fix reading old coadds (CoaddInputs/ExposureTable)

r-owen:


src/table/Exposure.cc
```diff
@@ -121,15 +122,23 @@ struct PersistenceHelper {
...
         output.setValidPolygon(archive.get<geom::polygon::Polygon>(input.get(validPolygon)));
-        output.setVisitInfo(archive.get<image::VisitInfo>(input.get(visitInfo)));
+        if (tableVersion > 1) {
+            output.setVisitInfo(archive.get<image::VisitInfo>(input.get(visitInfo)));
+        } else {
+            // VisitInfo not available; set to invalid values
+            daf::base::PropertySet md;
+            auto nullVisitInfoPtr = std::make_shared<image::VisitInfo>(md);
+            output.setVisitInfo(nullVisitInfoPtr);
+        }
```
TallJimbo: I'm a little surprised you consider it better to default-construct a `VisitInfo` here rather than leave it an empty pointer.  Is having a non-null `VisitInfo` required for all `Exposure` objects (even coadd?!) and hence something you want to enforce for `ExposureRecords` as well?

r-owen: I would prefer to leave it an empty pointer. How do I do that?

TallJimbo: If you simply do nothing, it should already be set to an empty pointer.  If you want to make it explicit (for readers of the code), you could do:
```
output.setVisitInfo(nullptr);
```
I'm ambivalent about which is better.

PR #105, opened 2016-10-14
DM-5529: Implement SphPoint

kfindeisen:
`afw::geom::SphPoint` has been implemented as discussed in [DM-5529](https://jira.lsstcorp.org/browse/DM-5529). The one substantial change is the replacement of `offsetTo(SphPoint) -> pair<Angle, Angle>` with `bearingTo(SphPoint) -> Angle`, as discussed previously with @r-owen.

Unit tests are a combination of existing tests for `afw::coord::Coord` and new tests.

Class documentation has been written in the format proposed in [RFC-225](https://jira.lsstcorp.org/browse/RFC-225) for forward-compatibility.

### DIFF
--- a/include/lsst/afw/geom/SpherePoint.h
+++ b/include/lsst/afw/geom/SpherePoint.h
@@ -1,0 +1,337 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * See COPYRIGHT file at the top of the source tree.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program. If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_AFW_GEOM_SPHEREPOINT_H_
+#define LSST_AFW_GEOM_SPHEREPOINT_H_
+
+#include <ostream>
+#include <utility>
+
+#include "lsst/afw/geom/Angle.h"
+#include "lsst/afw/geom/Point.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+    /**
+     * @brief Point in an unspecified spherical coordinate system.
+     *
+     * This class represents a point on a sphere in the mathematical rather
+     * than the astronomical sense. It does not refer to any astronomical
+     * reference frame, nor does it have any concept of (radial) distance.
+     *
+     * Points can be represented either as spherical coordinates or as a unit
+     * vector. The adopted convention for converting between these two systems
+     * is that (0, 0) corresponds to <1, 0, 0>, that (&pi;/2, 0) corresponds
+     * to <0, 1, 0>, and that (0, &pi;/2) corresponds to <0, 0, 1>.
+     *
+     * To keep usage simple, SpherePoint does not support modification of existing
+     * values; transformations of SpherePoints should be expressed as a new object.
+     * To support cases where a SpherePoint *must* be updated in-place, SpherePoint
+     * supports assignment to an existing object. One example is in containers
+     * of SpherePoints; no STL container has an atomic element-replacement method,
+     * so complicated constructions would need to be used if you couldn't
+     * overwrite an existing element.
+     *
+     * @see @ref coord::Coord
+     */
+    class SpherePoint
+#ifndef SWIG
+    final
+#endif
+    {
+    public:
+        /**
+         * @brief Construct a SpherePoint from a longitude and latitude.
+         *
+         * @param longitude The longitude of the point.
+         * @param latitude The latitude of the point. Must be in the
+         *                 interval [-&pi;/2, &pi;/2] radians.
+         *
+         * @throws OutOfRangeError Thrown if @c latitude is out of range.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         */
+        SpherePoint(Angle const & longitude, Angle const & latitude);
+
+        /**
+         * @brief Construct a SpherePoint from a vector representing a direction.
+         *
+         * @param vector A position whose direction will be stored as a SpherePoint.
+         *               Must not be the zero vector. Need not be normalized,
+         *               and the norm will not affect the value of the point.
+         *
+         * @throws InvalidParameterError Thrown if @c vector is the zero vector.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         */
+        explicit SpherePoint(Point3D const & vector);
+
+        /**
+         * @brief Create a copy of a SpherePoint.
+         *
+         * @param other The point to copy.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         */
+        SpherePoint(SpherePoint const & other) = default;
+
+        /**
+         * Overwrite this object with the value of another SpherePoint.
+         *
+         * This is the only method that can alter the state of a SpherePoint after
+         * its construction, and is provided to allow in-place replacement of
+         * SpherePoints where swapping is not possible.
+         *
+         * @param other The object with which to overwrite this one.
+         * @return a reference to this object.
+         *
+         * @exceptsafe This operator shall not throw exceptions.
+         */
+        SpherePoint& operator=(SpherePoint const & other) = default;
+
+        /*
+         * Accessors
+         */
+
+        /**
+         * @brief The longitude of this point.
+         *
+         * If this point is at a coordinate pole, the longitude is undefined, and
+         * this method may return any value. If the SpherePoint implementation
+         * allows multiple values of longitude from a pole, they shall all be
+         * treated as valid representations of the same point.
+         *
+         * @return the longitude, in the interval [0, 2&pi;) radians.
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        Angle getLongitude() const noexcept { return _longitude*radians; };
+
+        /**
+         * @brief The latitude of this point.
+         *
+         * @return the latitude, in the interval [-&pi;/2, &pi;/2] radians.
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        Angle getLatitude() const noexcept { return _latitude*radians; };
+
+        /**
+         * @brief A unit vector representation of this point.
+         *
+         * @return a unit vector whose direction corresponds to this point
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        Point3D getVector() const noexcept;
+
+        /**
+         * @brief Longitude and latitude by index.
+         *
+         * @param index the index of the spherical coordinate to return. Must
+         *              be either 0 or 1.
+         *
+         * @return @ref getLongitude() if @c index = 0, or @ref getLatitude()
+         *         if @c index = 1
+         *
+         * @throws OutOfRangeError Thrown if @c index is neither 0 nor 1.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         */
+        Angle operator[](size_t index) const;
+
+        /**
+         * @brief @c true if this point is either coordinate pole.
+         *
+         * @return @c true if this point is at the north or south pole,
+         *         @c false otherwise
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        bool atPole() const noexcept {
+            // Unit tests indicate we don't need to worry about epsilon-errors
+            // Objects constructed from lat=90*degrees, <0,0,1>, etc. all have
+            // atPole() = true. More complex operations like bearingTo have also
+            // been tested near the poles with no ill effects
+            return fabs(_latitude) >= HALFPI;
+        }
+
+
+        /**
+         * @brief @c true if this point is a well-defined position.
+         *
+         * @return @c true if @ref getLongitude(), @ref getLatitude(), and
+         *         @ref getVector() return finite floating-point values;
+         *         @c false if any return NaN or infinity.
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        bool isFinite() const noexcept;
+
+        /*
+         * Comparisons between points
+         */
+
+        /**
+         * @brief Return @c true if two points represent the same position.
+         *
+         * Points are considered equal if and only if they represent the same
+         * location, regardless of how they were constructed. In particular,
+         * representations of the north or south pole with different longitudes
+         * are considered equal.
+         *
+         * @param other the point to test for equality
+         * @return true if this point matches @c other exactly, false otherwise
+         *
+         * @exceptsafe This operator shall not throw exceptions.
+         *
+         * @warning Points whose @ref getLongitude(), @ref getLatitude(), or
+         *          @ref getVector() methods return @c NaN shall not compare
+         *          equal to any point, including themselves. This may break
+         *          algorithms that assume object equality is reflexive; use
+         *          @ref isFinite() to filter objects if necessary.
+         */
+        bool operator==(SpherePoint const & other) const noexcept;
+
+        /**
+         * @brief Return @c false if two points represent the same position.
+         *
+         * This operator shall always return the logical negation of @c ==; see
+         * its documentation for a detailed specification.
+         */
+        bool operator!=(SpherePoint const & other) const noexcept;
+
+        /**
+         * @brief Direction from one point to another.
+         *
+         * This method finds the shortest (great-circle) arc between two
+         * points, and characterizes its direction by the angle between
+         * it and a line of latitude passing through this point. 0 represents
+         * due east, &pi;/2 represents due north. If the points are on
+         * opposite sides of the sphere, the bearing may be any value.
+         *
+         * @param other the point to which to measure the bearing
+         * @return the direction, as defined above, in the interval [0, 2&pi;).
+         *
+         * @throws DomainError Thrown if <tt>this.atPole()</tt>.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         *
+         * @note For two points @c A and @c B, <tt>A.bearingTo(B)</tt> will in
+         *       general not be 180 degrees away from <tt>B.bearingTo(A)</tt>
+         */
+        Angle bearingTo(SpherePoint const & other) const;
+
+        /**
+         * @brief Angular distance between two points.
+         *
+         * @param other the point to which to measure the separation
+         * @return the length of the shortest (great circle) arc between the
+         *         two points. Shall not be negative.
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        Angle separation(SpherePoint const & other) const noexcept;
+
+        /*
+         * Transformations of points
+         */
+
+        /**
+         * @brief Return a point rotated from this one around an axis.
+         *
+         * @param axis a point defining the north pole of the rotation axis.
+         * @param amount the amount of rotation, where positive values
+         *               represent right-handed rotations around @c axis.
+         * @return a new point created by rotating this point
+         *
+         * @exceptsafe This method shall not throw exceptions.
+         */
+        SpherePoint rotated(SpherePoint const & axis, Angle const & amount) const noexcept;
+
+        /**
+         * @brief Return a point offset from this one along a great circle.
+         *
+         * For any point @c A not at a coordinate pole, and any two angles @c b
+         * and @c delta, <tt>A.bearingTo(A.offset(b, delta))</tt> = @c b and
+         * <tt>A.separationTo(A.offset(b, delta))</tt> = @c delta.
+         *
+         * @param bearing the direction in which to move this point, following
+         *                the conventions described in @ref bearingTo.
+         * @param amount the distance by which to move along the great
+         *               circle defined by @c bearing
+         * @return a new point created by shifting this point
+         *
+         * @throws DomainError Thrown if <tt>this.atPole()</tt>.
+         * @throws OutOfRangeError Thrown if @c amount is negative.
+         *
+         * @exceptsafe The program state shall be unchanged in the event of an
+         *             exception.
+         */
+        SpherePoint offset(Angle const & bearing, Angle const & amount) const;
+
+    private:
+        // For compatibility with Starlink AST, the implementation must be a
+        // pair of floating-point numbers, with no other data. Do not change
+        // the implementation without an RFC.
+        double _longitude;      // radians
+        double _latitude;       // radians
+    };
+
+    /*
+     * Object-level display
+     */
+
+    /**
+     * @brief Print the value of a point to a stream.
+     *
+     * The exact details of the string representation are unspecified and
+     * subject to change, but the following may be regarded as typical:
+     * <tt>"(10.543250, +32.830583)"</tt>.
+     *
+     * @param os the stream to which to print @c point
+     * @param point the point to print to the stream
+     * @return a reference to @c os
+     *
+     * @throws std::ostream::failure Thrown if an I/O state flag was set that
+     *      was registered with <tt>os.exceptions()</tt>. See the documentation
+     *      of std::ostream for more details.
+     *
+     * @exceptsafe All objects shall be left in valid states, with no resource
+     *             leaks, in the event of an exception.
+     *
+     * @relatesalso SpherePoint
+     */
+    std::ostream& operator<<(std::ostream& os, SpherePoint const & point);
+
+}}} /* namespace lsst::afw::geom */
+
+#endif /* LSST_AFW_GEOM_SPHEREPOINT_H_ */
--- a/tests/testSpherePoint.py
+++ b/tests/testSpherePoint.py
@@ -1,0 +1,801 @@
+#
+# LSST Data Management System
+# See COPYRIGHT file at the top of the source tree.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+# -*- python -*-
+"""
+Unit tests for SpherePoint
+
+Run with:
+   python testSpherePoint.py
+or
+   python
+   >>> import testSpherePoint
+   >>> testSpherePoint.run()
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import copy
+import math
+import re
+import unittest
+
+import numpy as np
+
+import lsst.utils.tests
+import lsst.afw.geom as afwGeom
+import lsst.pex.exceptions as pexEx
+
+from lsst.afw.geom import degrees, radians, SpherePoint
+from numpy import nan, inf
+
+
+class SpherePointTestSuite(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        self._dataset = SpherePointTestSuite.positions()
+        self._poleLatitudes = [
+            afwGeom.HALFPI*afwGeom.radians,
+            6.0*afwGeom.hours,
+            90.0*afwGeom.degrees,
+            5400.0*afwGeom.arcminutes,
+            324000.0*afwGeom.arcseconds,
+        ]
+
+    @staticmethod
+    def positions():
+        """Provide valid coordinates for nominal-case testing.
+
+        Returns
+        -------
+        positions : `iterable`
+            An iterable of pairs of Angles, each representing the
+            longitude and latitude (in that order) of a test point.
+        """
+        nValidPoints = 100
+        rng = np.random.RandomState(42)
+        ra = rng.uniform(0.0, 360.0, nValidPoints)
+        dec = rng.uniform(-90.0, 90.0, nValidPoints)
+
+        points = list(zip(ra*degrees, dec*degrees))
+        # Ensure corner cases are tested.
+        points += [
+            (0.0*degrees, 0.0*degrees),
+            (afwGeom.PI*radians, -6.0*degrees),
+            (42.0*degrees, -90.0*degrees),
+            (172.0*degrees, afwGeom.HALFPI*radians),
+            (360.0*degrees, 45.0*degrees),
+            (-278.0*degrees, -42.0*degrees),
+            (765.0*degrees, 0.25*afwGeom.PI*radians),
+            (180.0*degrees, nan*radians),
+            (inf*degrees, 45.0*degrees),
+            (nan*degrees, -8.3*degrees),
+        ]
+        return points
+
+    def testInit2ArgErrors(self):
+        """Test if 2-argument form of __init__ handles invalid input.
+        """
+        # Latitude should be checked for out-of-range.
+        for lat in self._poleLatitudes:
+            with self.assertRaises(pexEx.OutOfRangeError):
+                SpherePoint(0.0*degrees, self.nextUp(lat))
+            with self.assertRaises(pexEx.OutOfRangeError):
+                SpherePoint(0.0*degrees, self.nextDown(0.0*radians - lat))
+
+        # Longitude should not be checked for out of range.
+        SpherePoint(360.0*degrees, 45.0*degrees)
+        SpherePoint(-42.0*degrees, 45.0*degrees)
+        SpherePoint(391.0*degrees, 45.0*degrees)
+
+        # Infinite latitude is not allowed.
+        with self.assertRaises(pexEx.OutOfRangeError):
+            SpherePoint(-42.0*degrees, inf*degrees)
+        with self.assertRaises(pexEx.OutOfRangeError):
+            SpherePoint(-42.0*degrees, -inf*degrees)
+
+    def testInit1ArgErrors(self):
+        """Test if 1-argument form of __init__ handles invalid input.
+        """
+        # Only one singularity, at zero
+        with self.assertRaises(pexEx.InvalidParameterError):
+            SpherePoint(afwGeom.Point3D(0.0, 0.0, 0.0))
+        SpherePoint(afwGeom.Point3D(0.0, -0.2, 0.0))
+        SpherePoint(afwGeom.Point3D(0.0, 0.0, 1.0))
+        SpherePoint(afwGeom.Point3D(42.78, -46.29, 38.27))
+
+    def testInitNArgFail(self):
+        """Tests if only 1- or 2-argument initializers are allowed.
+        """
+        with self.assertRaises(NotImplementedError):
+            SpherePoint()
+        with self.assertRaises(NotImplementedError):
+            SpherePoint("Rotund", "Bovine")
+        with self.assertRaises(NotImplementedError):
+            SpherePoint(42)
+        with self.assertRaises(NotImplementedError):
+            SpherePoint("ICRS", 34.0, -56.0)
+
+    def testGetLongitudeValue(self):
+        """Test if getLongitude() returns the expected value.
+        """
+        for lon, lat in self._dataset:
+            point = SpherePoint(lon, lat)
+            self.assertIsInstance(point.getLongitude(), afwGeom.Angle)
+            # Behavior for non-finite points is undefined; depends on internal data representation
+            if point.isFinite():
+                self.assertGreaterEqual(point.getLongitude().asDegrees(), 0.0)
+                self.assertLess(point.getLongitude().asDegrees(), 360.0)
+
+                # Longitude not guaranteed to match input at pole
+                if not point.atPole():
+                    # assertAnglesNearlyEqual handles angle wrapping internally
+                    self.assertAnglesNearlyEqual(lon, point.getLongitude())
+
+        # Vector construction should return valid longitude even in edge cases.
+        point = SpherePoint(afwGeom.Point3D(0.0, 0.0, -1.0))
+        self.assertGreaterEqual(point.getLongitude().asDegrees(), 0.0)
+        self.assertLess(point.getLongitude().asDegrees(), 360.0)
+
+    def testTicket1394(self):
+        """Regression test for Ticket 1761.
+
+        Checks that negative longitudes within epsilon of lon=0 lead
+        are correctly bounded and rounded.
+        """
+        # The problem was that the coordinate is less than epsilon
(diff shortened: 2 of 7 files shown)


include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+        Angle getLatitude() const noexcept { return _latitude*radians; };
+
+        /**
+         * @brief A vector representation of this point.
+         *
+         * The adopted convention is that <1, 0, 0> corresponds to (0, 0), that
+         * <0, 1, 0> corresponds to (&pi;/2, 0), and that <0, 0, 1> corresponds
+         * to (0, &pi;/2).
+         *
+         * @return a unit vector whose direction corresponds to this point
```
r-owen: Please document that the norm of the returned vector is 1

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         *
+         * @par Exception Safety:
+         * This method shall not throw exceptions.
+         */
+        Point3D getVector() const noexcept;
+
+        /**
+         * @brief An array-like representation of this point.
```
r-owen: Please change the brief description to make it clear that it is indexing over longitude and latitude, rather than a 3d vector. Perhaps something like "Get longitude or latitude by index"

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         * @param other the point to which to measure the bearing
+         * @return the direction, as defined above, in the interval [0, 2&pi;).
+         *
+         * @throws DomainError Thrown if this point is either coordinate pole.
+         *
+         * @par Exception Safety:
+         * The program state shall be unchanged in the event of an exception.
+         *
+         * @note For two points @c A and @c B, <tt>A.bearingTo(B)</tt> will in
+         *       general not be 180 degrees away from <tt>B.bearingTo(A)</tt>
```
r-owen: Nice

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         */
+        Angle bearingTo(SphPoint const & other) const;
+
+        /**
+         * @brief Angular distance between two points.
+         *
+         * @param other the point to which to measure the separation
+         * @return the length of the shortest arc between the two points. Shall
```
r-owen: Please specify that this is a great circle arc

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         * Transformations of points
+         */
+
+        /**
+         * @brief A point defined by rotating this point around an arbitrary axis.
+         *
+         * @param axis the point around which to perform the rotation.
+         * @param amount the angle by which to rotate around @c axis.
+         *               Positive angles represent counterclockwise
+         *               rotations.
```
r-owen: I find the two parameter descriptions unclear. What does it mean to rotate around a point? Perhaps something like this might be clearer, though I hope it may give you an idea on how to do even better:
- @brief axis  A point defining the Z axis of rotation frame
- @brief amount  The amount of rotation, where positive values rotate from the X axis to Y axis of the rotation frame

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+
+    /*
+     * Object-level display
+     */
+
+    /**
+     * @brief Prints the value of a point to a stream.
+     *
+     * The point shall be printed as an ordered pair of longitude and (signed)
+     * latitude, in degrees.
```
r-owen: From this description it sounds like "10.5, 22.3" but I'm hoping it is more like "SphPoint(10.5, 22.3)", though presumably with higher precision. Perhaps an example?

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         *                               NaN or infinite, or if @c vector is
+         *                               the zero vector.
+         *
+         * @par Exception Safety:
+         * The program state shall be unchanged in the event of an exception.
+         */
+        explicit SphPoint(Point3D const & vector);
+
+        /**
+         * @brief Creates a copy of a SphPoint.
```
r-owen: I personally prefer `SphPoint` to go with the `sphgeom` package. I don't think `SpherePoint` helps much, and `SphericalPoint` (the fully written out version) seems far longer than necessary.

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * See COPYRIGHT file at the top of the source tree.
```
kfindeisen: The current documentation guidelines require it: https://developer.lsst.io/docs/cpp_docs.html#boilerplate-for-header-h-and-source-cpp-files

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+     * This class was not designed for inheritance, and must not be subclassed.
+     *
+     * @see @ref coord::Coord
+     */
+    // TODO: add final keyword after migrating to Pybind11
+    class SphPoint
+    {
+    public:
+        /**
+         * @brief Represent a specific pair of spherical coordinates.
```
kfindeisen: Ok, but I'd leave out the "whose length is ignored", since that's confusing for anybody who isn't familiar with the historical context (when I hear "length" I think "number of elements").

include/lsst/afw/geom/SphPoint.h
kfindeisen: Good point, I'll add that rule to DM-7891 as well.

Re: the names, I'm fine with changing it to `SpherePoint`. It doesn't sound to my ears like a contraction of `SphericalPoint`, but more like "Point on a Sphere".

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         *
+         * This is the only method that can alter the state of a SphPoint after
+         * its construction, and is provided to allow in-place replacement of
+         * SphPoints where swapping is not possible.
+         *
+         * @param other The object with which to overwrite this one.
+         * @return a reference to this object.
+         *
+         * @par Exception Safety:
+         * This operator shall not throw exceptions.
```
kfindeisen: There's no sensible reason for `operator=` to throw exceptions, but at the same time I can't take responsibility for what Compiler X decides to give as the default implementation. As far as I understand, compilers don't warn you even if the implementation of a `noexcept` method is provably throwing, and the runtime behavior in that case is CTD. So telling people "if this code throws exceptions, that's a bug" without specifying it at the language level seemed the safest approach.

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         */
+        SphPoint& operator=(SphPoint const & other) = default;
+
+        /*
+         * Accessors
+         */
+
+        /**
+         * @brief The azimuth angle of this point.
```
kfindeisen: I was trying to pick something [as coordinate-independent as possible](https://en.wikipedia.org/wiki/Spherical_coordinate_system), but I was afraid this would happen. I'll change the terminology to longitude/latitude.

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         *
+         * If this point is at a coordinate pole, this method may return any
+         * value.
```
kfindeisen: What do you mean by "can be trusted"? The longitude is valid, it's simply arbitrary.

include/lsst/afw/geom/SphPoint.h
kfindeisen: ... are there unit vectors whose norm is not 1?

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+            return _latitude == other._latitude;
+        }
+    }
+
+    bool SphPoint::operator!=(SphPoint const & other) const noexcept {
+        return !(*this == other);
+    }
+
+    Angle SphPoint::bearingTo(SphPoint const & other) const {
+        if (abs(_latitude) >= HALFPI) {
```
r-owen: Another instance of needing `std::abs` or `fabs` or `std::fabs`.

I am worried that your unit tests did not find this.

Also, I suggest you use the `atPole()` method here (should you agree to add it).

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+    SphPoint::SphPoint(Angle const & longitude, Angle const & latitude) :
+                    _longitude(wrap(longitude).asRadians()),
+                    _latitude(latitude.asRadians()) {
+        if (!isfinite(_latitude) || !isfinite(_longitude)) {
+            throw pexExcept::InvalidParameterError("NaN detected in constructing (" +
+                                                   to_string(longitude.asDegrees()) +
+                                                   ", " + to_string(latitude.asDegrees()) +
+                                                   ")");
+        }
+        // For some reason abs(double) is ignored
```
r-owen: http://stackoverflow.com/a/13792680 explains. You must use `std::abs` not `abs`. You can also use `fabs` or `std::fabs` if you prefer.

I personally suggest you always make `std::` explicit. I realize it is a matter of taste, and you are free to do as you prefer, but I personally find it helps clarify where symbols come from (a bit), and it avoids this very issue you ran into.

Please look through the rest of the code for `abs` and change it. I found at least two other instances (which I will note, but I may not find them all).

include/lsst/afw/geom/SphPoint.h
kfindeisen: The current implementation is (10.5, +22.3), with the precision controlled by the output stream.

I don't think it's a good idea to include the class name, because that can reduce human-readability. Suppose you have some client code:
> cerr << "Value " << pixel << " at " << point << " is out of range." << endl

Do you want it to print 
> Value -3e-6 at (10.543250, +32.830583) is out of range.

or 
> Value -3e-6 at SphPoint(10.543250, +32.830583) is out of range.

A similar question arises for printing containers of points.

This is also why I give `__str__` and `__repr__` different behavior in the Python API; as I understand it, `__str__` should be human-readable, while `__repr__` should be parseable as valid Python.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+        if (fabs(_latitude) > HALFPI) {
+            throw pexExcept::OutOfRangeError("Angle " + to_string(latitude.asDegrees()) +
+                " is not a valid latitude.");
+        }
+    }
+
+    SphPoint::SphPoint(Point3D const & vector) {
+        double const norm = sqrt(vector.distanceSquared(Point3D()));
+        if (!isfinite(norm)) {
```
r-owen: Again, please do not forbid NaNs

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+        double const cosDelta1 = cos(      getLatitude().asRadians());
+        double const cosDelta2 = cos(other.getLatitude().asRadians());
+
+        // Adapted from http://www.movable-type.co.uk/scripts/latlong.html
+        double const y = sin(deltaLon)*cosDelta2;
+        double const x = cosDelta1*sinDelta2 - sinDelta1*cosDelta2*cos(deltaLon);
+        return wrap(90.0*degrees - atan2(y, x)*radians);
+    }
+
+    Angle SphPoint::separationTo(SphPoint const & other) const noexcept {
```
r-owen: Is the "To" informative here? This is a symmetrical operation. I gently suggest `separation` or `angularSeparation`.

include/lsst/afw/geom/SphPoint.h
kfindeisen: We should add `atPole` anyway to let clients guard against exceptions from the offset methods, so this is mainly about whether it's important enough to be a warning.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+
+        // take r x v to get the axis
+        SphPoint axis = SphPoint(Point3D(r.cross(v)));
+
+        return rotated(axis, amount);
+    }
+
+    ostream& operator <<(ostream & os, SphPoint const & point) {
+        // Can't provide atomic guarantee anyway for I/O, so ok to be sloppy.
+        os << "(" << point.getLongitude().asDegrees() << ", ";
```
r-owen: Please set the precision here, rather than in `__str__` in the .i file.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+        auto const x = getVector().asEigen();
+        auto const xprime = rotation * x;
+        return SphPoint(Point3D(xprime));
+    }
+
+    SphPoint SphPoint::offset(Angle const & bearing, Angle const & amount) const {
+        double const phi = bearing.asRadians();
+        if (!isfinite(phi)) {
+            stringstream buffer;
+            buffer << "NaN offset bearing " << bearing << ".";
```
r-owen: I suggest you simply propagate NaNs so as to avoid unnecessarily raising exceptions.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+            throw pexExcept::InvalidParameterError(buffer.str());
+        }
+        if (!isfinite(amount.asRadians())) {
+            stringstream buffer;
+            buffer << "NaN offset amount " << amount << ".";
+            throw pexExcept::InvalidParameterError(buffer.str());
+        }
+        if (abs(_latitude) >= HALFPI) {
```
r-owen: Another bare `abs` and worry about why the existing tests didn't find it.

include/lsst/afw/geom/SphPoint.h
r-owen: You could use "norm", but as long as the documentation says it somewhere, that's what matters. You already made it clear elsewhere that `SphPoint` describes points on a unit sphere.

include/lsst/afw/geom/SphPoint.h
r-owen: Good. In thinking about it further I don't fully understand what you are warning about; I think it need some clarification. If I construct a SphPoint at the pole from a longitude and latitude then getLatitude reliably returns the original latitude. Furthermore, I think methods such as `rotated` will have a predictable result on the longitude (I'm not so sure about `offset`).

On the other hand, if a `SphPoint` is constructed from a Vector3D then `longitude` is not self-evident, though you can and probably ought to document what it will be. Similarly, if a method such as `offset` or `rotated` returns a point at the pole then longitude is probably not predictable.

So the situation is nuanced. It seems only a few methods can make a point with unpredictable longitude, and perhaps the warning about longitude belongs with those methods? If so, I'm not convinced anything need be said here (or can be said here without some additional explanation).

include/lsst/afw/geom/SphPoint.h
r-owen: I'd personally go for "SphPoint(10.543250, +32.830583)" to avoid confusion with a cartesian point, but it's your call. They both work. I'm not thrilled about using the default stream precision (and then using the stream operator to code `__str__`, as noted later), but I guess it makes sense.

include/lsst/afw/geom/SphPoint.h
kfindeisen: I saw the original comment as merely a matter of being as explicit as possible in the spec -- there is no such thing as "the" longitude of a polar point, so the implementation is free to choose any value that conforms to the [0, 2π) requirement. It could be whatever was passed to the constructor (the current implementation), it could always be zero, it could be some magic value that makes internal calculations more efficient, etc.

My question is, why should the user *care* about the longitude of a polar SphPoint? It won't affect the result of `separationTo` or `rotated` (or, at least, it shouldn't -- that would be a good test case), it won't affect equality comparisons, etc.. It doesn't even affect the singularity associated with `bearingTo`, because that's caused by the absence of well-defined directions at the pole, not the absence of well-defined longitude.

The only thing it does affect is the value of `getLongitude()` (and synonyms like `[0]`), and if you're using those methods because you want to do coordinate math yourself, then it's up to you to make sure your equations are valid in spherical geometry.

python/lsst/afw/geom/SphPoint.i
```diff
@@ -0,0 +1,71 @@
...
+
+%include "lsst/afw/geom/SphPoint.h"
+
+// add __str__ and __repr__ methods
+%extend lsst::afw::geom::SphPoint {
+    std::string __str__() const {
+        std::ostringstream os;
+        // Ensure string gives enough precision to enable same-field comparisons
+        os.precision(6);
+        os << std::fixed << (*self);
```
kfindeisen: I'm not thrilled about forcing a specific precision on the user, though; unless we pick an extreme number of decimal places, there will always be cases where the user wants more.

How about guaranteeing output to at least N decimal places (6?), while allowing the user to request more using the stream precision and/or `boost::format`? That would make the implementation of `<<` a bit more complex, but it would address both our concerns.

include/lsst/afw/geom/SphPoint.h
r-owen: Well, that explains the pole test for equality.

I think you what would be clearer is to point out that the longitude is not important at the poles; all points with a given latitude are identical.

src/geom/SphPoint.cc
kfindeisen: At the time that I ran into this problem, I *was* using `std::abs`.

python/lsst/afw/geom/SphPoint.i
r-owen: That would work, though I think it would also be acceptable and simpler to make the user do their own formatting when they want more precision.

tests/testSphPoint.cc
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program. If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include <iostream>
+#include <sstream>
+
+// FIXME: violates https://developer.lsst.io/coding/unit_test_policy.html#testing-frameworks
```
r-owen: In what way does this violate the standards, and what would be required to fix it (or should we broaden our standards?). A few more words here would be helpful. (And yes I followed the link but couldn't figure it out).

tests/testSphPoint.cc
```diff
@@ -0,0 +1,69 @@
...
+
+        original = SphPoint(-42*degrees, 45*degrees);
+        BOOST_TEST(original != copy);
+        BOOST_TEST(copy.getLongitude().asDegrees() == copyLon);
+        BOOST_TEST(copy.getLatitude() .asDegrees() == copyLat);
+    }
+
+    // TODO: add a test for propagation of ostream errors
+
+}}} /* namespace lsst::afw::geom */
```
r-owen: Excellent idea testing the C++ specific replacement operator

tests/testSphPoint.cc
```diff
@@ -0,0 +1,69 @@
...
+namespace afw {
+namespace geom {
+
+    /**
+     * @brief Tests whether the result of SphPoint::SphPoint(SphPoint const &)
+     * makes an identical but independent copy.
+     */
+    BOOST_AUTO_TEST_CASE(SphPointCopyResult, *boost::unit_test::tolerance(1e-10))
+    {
+        SphPoint original(Point3D(0.0, 0.0, 1.0));
```
r-owen: I'm surprised you picked a point at the pole, since the equality operator ignores longitude there. A different point would give slightly broader coverage.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+            return getLongitude();
+          case 1:
+            return getLatitude();
+          default:
+            throw pexExcept::OutOfRangeError("Index " + to_string(index) + " must be 0 or 1.");
+        }
+    }
+
+    bool SphPoint::operator==(SphPoint const & other) const noexcept {
+        if (abs(_latitude) < HALFPI) {
```
kfindeisen: I agree. `testEqualityAlias` should not be passing if this condition always evaluates to `true`. I'll look into it tomorrow.

include/lsst/afw/geom/SphPoint.h
jdswinbank: Please check the note at the top of that page:

> This is a preview documentation format specification. Software documentation should currently be written in the format described at https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards

kfindeisen: The confluence page seems to say the same thing:
> * The LSST Copyright statement appropriate for the file type. Refer to:
>     - git@git.lsstcorp.org:LSST/DMS/devenv/templates.git  {CopyrightHeader.py, CopyrightHeader.h, CopyrightHeader.cc}

I can't clone that URI (is it out of date?), but I assume it's the same code found at https://dev.lsstcorp.org/trac/browser/DMS/devenv/templates/trunk. See also https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards#DocumentationStandards-Example.hfile.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+        test will fail for some value of `delta`. Testing multiple
+        values is recommended.
+
+        Parameters
+        ----------
+        delta : `number`
+            The separation, in degrees, at which point equality may
+            become intransitive.
+        """
+        point1 = SphPoint((45.0 - delta)*degrees, -18.0*degrees);
```
r-owen: Why not test at all values in the data set? You suggest that testing multiple values is recommended, and I would think that applies equally to deltas and points on the sphere.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+        point3 = SphPoint((45.0 + delta)*degrees, -18.0*degrees);
+
+        self.assertTrue(point1 != point2 or point2 != point3 or point1 == point3)
+        self.assertTrue(point3 != point1 or point1 != point2 or point3 == point2)
+        self.assertTrue(point2 == point3 or point3 != point1 or point2 == point1)
+
+    def testEqualityAlias(self):
+        """Test if == handles coordinate degeneracies correctly.
+        """
+        # Coordinate wrapping
```
r-owen: I suggest `# Longitude wrapping`. Let's save "coord" for the `Coord` class.

Speaking of which, you use the variable "coord" elsewhere and should consider renaming it, e.g. to longLat

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+                            SphPoint( 83.0*degrees, -90.0*degrees))
+
+    def testInquality(self):
+        """Test if == and != give mutually consistent results.
+        """
+        for lon1, lat1 in self._dataset[0:20]:
+            point1 = SphPoint(lon1*degrees, lat1*degrees)
+            for lon2, lat2 in self._dataset[0:20]:
+                point2 = SphPoint(lon2*degrees, lat2*degrees)
+                self.assertTrue((point1 == point2) == (point2 == point1))
```
r-owen: This and the next two lines would be better expressed as using `assertEqual` and `assertNotEqual`, for example: `self.assertEqual((point1 == point2),  (point2 == point1))`. This gives more useful error messages.

Your use of `assertTrue(foo == bar)` is definitely appropriate when you are explicitly testing `operator==`, but otherwise `assertTrue` should be avoided if another assert is more expressive of what you are testing, both for readability and because the resulting error message is much more informative if the test fails.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+                self.assertTrue((point1 != point2) == (point2 != point1))
+                self.assertTrue((point1 == point2) != (point1 != point2))
+
+    def testBearingToError(self):
+        """Test if bearingTo() correctly handles invalid input.
+        """
+        destination = SphPoint(42.0*degrees, -90.0*degrees)
```
r-owen: Why is `destination` at a pole? I would expect you to assert that that "north pole to safe" and south pole to safe" raise exceptions. It seems less interesting to test "north pole to destination at a pole".

Please consider eliminating `destination` and just having `northPole`, `southPole` and `safe` for simplicity and clarity. Or if you want to keep all of them, then please explain what you are doing in a code comment.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+
+        with self.assertRaises(pexEx.DomainError):
+            northPole.bearingTo(destination)
+        with self.assertRaises(pexEx.DomainError):
+            southPole.bearingTo(destination)
+        with self.assertRaises(pexEx.DomainError):
+            destination.bearingTo(safe)
+
+        # Argument is allowed to be a pole.
+        safe.bearingTo(destination)
```
r-owen: I suggest ditching this because it duplicates `testBearingToValue`, which is more rigorous.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+                self.assertGreaterEqual(sep.asDegrees(), 0.0)
+                self.assertLessEqual(sep.asDegrees(), 180.0)
+
+                self.assertAlmostEqual(expected, sep.asRadians())
+                self.assertAnglesNearlyEqual(sep, point2.separationTo(point1))
+
+    def testSeparationToValueAbsolute(self):
+        """Test if separationTo() returns specific values.
+        """
+        # Test from Meeus, p. 110
```
r-owen: Please provide more information, e.g. if this is a book then provide the title, edition and perhaps even the publication date

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+        spica = SphPoint(201.2983*degrees, -11.1614*degrees)
+        arcturus = SphPoint(213.9154*degrees, 19.1825*degrees)
+
+        # Verify to precision of quoted distance and _positions.
```
r-owen: "positions" instead of "_positions"?

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+            pointRepr = repr(point)
+            className = type(point).__name__
+            self.assertIn("degrees", pointRepr)
+            self.assertEqual(2, len(pointRepr.split(",")))
+
+            copy = eval(pointRepr)
+            self.assertAnglesNearlyEqual(point.getLongitude(), copy.getLongitude())
+            self.assertAnglesNearlyEqual(point.getLatitude(),  copy.getLatitude())
+
+    def _assertPointsNearlyEqual(self, lhs, rhs, *args):
```
r-owen: I strongly suggest you test that the separation is less than a given value, instead of comparing longitude and latitude, because the measure is invariant over the sphere.

Please add this to `afw.geom.utils` (using separation as the test). The code will be essentially identical to `assertCoordsNearlyEqual` in lsst.coords.utils`. This will be a big help to anyone using `SphCoord` in tests, as it helps avoid naive errors, such as testing that longitude values are nearly equal when two points may be nearly the same but in different wraps.

I am less convinced `assertPointsNotNearlyEqual` is needed as a publicly available assertion, but it's up to you.

If you do, a test will be required in `tests/testMethods.py`.

src/geom/SphPoint.cc
kfindeisen: Ok, this is really weird: it looks like `g++`, at least, picks `abs(double)` as the overload in this expression.

include/lsst/afw/geom/SphPoint.h
```diff
@@ -0,0 +1,319 @@
...
+         * The program state shall be unchanged in the event of an exception.
+         */
+        SphPoint(SphPoint const & other) = default;
+
+        /**
+         * Overwrites this object with the value of another SphPoint.
+         *
+         * This is the only method that can alter the state of a SphPoint after
+         * its construction, and is provided to allow in-place replacement of
+         * SphPoints where swapping is not possible.
```
kfindeisen: Well, move semantics don't work that way. In your example, `operator[i]` evaluates to an object reference, and assignment to a reference is treated as assignment to the object. Now, if you had a container of _pointers_...

Note also that most STL containers don't have `operator[]`; you need to use iterators to access arbitrary elements.

TallJimbo: I'm okay with it.  I could imagine that we'll find that we do want to set longitude and latitude separately a lot, and hence want to revisit this decision, but for now I'm with you both on hoping we can get away with just an assignment operator.

include/lsst/afw/geom/SphPoint.h
r-owen: That sounds fine

tests/testSphPoint.cc
kfindeisen: The linked standards say we need to use the single-header variant of Boost.Test, but only the shared library variant works in afw.

I assume fixing this would require changing build settings for afw, but I don't know the details.

TallJimbo: I think it'd be better to just change the standards in this case.  I certainly didn't know this is what they said, and indeed all of the C++ tests in our codebase use the shared library version.

kfindeisen: Requested as [DM-8038](https://jira.lsstcorp.org/browse/DM-8038); I'll remove the comment in the meantime.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+            point = SphPoint(lon*degrees, lat*degrees)
+            self.assertIsInstance(point.getLongitude(), afwGeom.Angle)
+            self.assertGreaterEqual(point.getLongitude().asDegrees(), 0.0)
+            self.assertLess(point.getLongitude().asDegrees(), 360.0)
+            if lon >= 0.0 and lon < 360.0 and abs(lat) != 90.0:
+                self.assertAlmostEqual(lon, point.getLongitude().asDegrees())
+
+    def testGetLongitudeWrap(self):
+        """Test if getLongitude() returns properly wrapped numbers.
+        """
```
kfindeisen: When the wrapping happens is an implementation detail, and subject to change. The spec only says that getLongitude() must return wrapped values, and it's the externally visible behavior of the class that's being tested.

tests/testSphPoint.cc
r-owen: Thank you @kfindeisen. That sounds perfect.

tests/testSphPoint.py
```diff
@@ -0,0 +1,615 @@
...
+        longitudes = [(0.0, 0.0), (180.0, 180.0), (360.0, 0.0), (-270.0, 90.0), (765.0, 45.0)]
+        for rawLon, cleanLon in longitudes:
+            point = SphPoint(rawLon*degrees, -42.0*degrees)
+            self.assertGreaterEqual(point.getLongitude().asDegrees(), 0.0)
+            self.assertLess(point.getLongitude().asDegrees(), 360.0)
+            self.assertAlmostEqual(cleanLon, point.getLongitude().asDegrees())
```
kfindeisen: Thanks, I didn't realize `assertAnglesNearlyEqual` handled wrapping internally.

src/geom/SphPoint.cc
```diff
@@ -0,0 +1,244 @@
...
+        }
+        // For some reason abs(double) is ignored
+        if (fabs(_latitude) > HALFPI) {
+            throw pexExcept::OutOfRangeError("Angle " + to_string(latitude.asDegrees()) +
+                " is not a valid latitude.");
+        }
+    }
+
+    SphPoint::SphPoint(Point3D const & vector) {
+        double const norm = sqrt(vector.distanceSquared(Point3D()));
```
kfindeisen: @r-owen What's the preferred idiom for doing this in Python? It appears `Point::asEigen()` doesn't have a corresponding method.

tests/testSphPoint.py
kfindeisen: @r-owen This test was copied from `Coord`. I kept it mainly because it provides a human-validated oracle, but maybe it's not worth it?

src/geom/SphPoint.cc
r-owen: I would use numpy in Python. I'm a bit worried about the question, though, as this is C++ code and I don't know where you would have to do this in Python.

tests/testSphPoint.py
r-owen: Drat. It's probably this: http://www.willbell.com/math/mc1.htm but unless we know somebody with a copy that could be hard to confirm. Well...please note that the details are unknown.

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+            point = SpherePoint(vector)
+            newLon, newLat = point
+            self.assertAlmostEqual(lonLat[0], newLon.asDegrees())
+            self.assertAlmostEqual(lonLat[1], newLat.asDegrees())
+            self.assertAlmostEqual(1.0, point.getVector().distanceSquared(afwGeom.Point3D(0.0, 0.0, 0.0)))
+
+        # Ill-defined points should be all NaN after normalization
+        pointList = [
+            afwGeom.Point3D(0.5, nan, 0.0),
+            afwGeom.Point3D(0.5, 0.0, -inf),
```
r-owen: I suggest a full test of nan and +/- inf appearing once in each of x, y and z. I realize x and y are probably handled fairly symmetrically, but z is handled differently and might as well cover all bases. Personally I'd use a loop to replace each element in turn by (nan, inf or -inf) but I realize loops are not necessarily well liked by all test authors.

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+
+    def testAtPoleValue(self):
+        """Test if atPole() returns the expected value.
+        """
+        poleList = [
+            SpherePoint(0.0*degrees, -90.0*degrees),
+            SpherePoint(42.0*degrees, 324000.0*afwGeom.arcseconds),
+            SpherePoint(-386.0*degrees, afwGeom.HALFPI*radians),
+            SpherePoint(afwGeom.Point3D(0.0, 0.0, 1.0)),
+            SpherePoint(afwGeom.Point3D(0.0, 0.0, -1.0)),
```
r-owen: As mentioned earlier, please test both poles for all units. I don't care about longitude, but I care a lot about latitude.

Also, test just inside of the atPole limit for both poles, all units. Consider scaling epsilon by the max value (e.g. epsilon/90 for degrees, etc.) for unit-independent resolution.

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+            SpherePoint(afwGeom.Point3D(0.1, 0.2, 0.3)),
+        ]
+        nonFiniteList = [
+            SpherePoint(0.0*degrees, nan*degrees),
+            SpherePoint(nan*degrees, 0.0*degrees),
+            SpherePoint(inf*degrees, 0.0*degrees),
+            SpherePoint(-inf*degrees, 0.0*degrees),
+            SpherePoint(afwGeom.Point3D(nan, 0.2, 0.3)),
+            SpherePoint(afwGeom.Point3D(0.1, inf, 0.3)),
+            SpherePoint(afwGeom.Point3D(0.1, 0.2, -inf)),
```
r-owen: Nice coverage!

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+            self.assertFalse(nonFinite.isFinite())
+
+    def testGetItemError(self):
+        """Test if indexing correctly handles invalid input.
+        """
+        point = SpherePoint(afwGeom.Point3D(1.0, 1.0, 1.0))
+
+        with self.assertRaises(pexEx.OutOfRangeError):
+            point[2]
+        with self.assertRaises((pexEx.OutOfRangeError, OverflowError)):
```
r-owen: I am a bit disappointed this fails. In Python I would expect point[-1] to return latitude (last element) and point[-2] to return longitude (2nd-to-last element).

I encourage you to spend a little time trying to fix this. For instance it may suffice to hide the C++ operator[] from SWIG (using %ignore, for instance). The SWIG wrapper already has `__iter__` and `__len__` and that might suffice. If not, defining `__getitem__` should definitely do it.

If you do fix this, then indices in range [-2, 1] are all valid, and any other might throw a standard Python exception instead of an LSST C++ exception (but it depends how you implement it).

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+                self.assertTrue(point2 != point1)
+                self.assertTrue(point1 != point2)
+                self.assertFalse(point2 == point1)
+                self.assertFalse(point1 == point2)
+
+        # Test for transitivity (may be assumed by algorithms).
+        for delta in [10.0**(0.1*x) for x in range(-150, -49, 5)]:
+            self._checkTransitive(delta*radians)
+
+    def _checkTransitive(self, delta):
```
r-owen: Consider making thisi public by deleting the leading underscore. It is our usual pattern.

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+                         SpherePoint(270.0*degrees, -42.0*degrees))
+
+        # Polar degeneracy
+        self.assertEqual(SpherePoint(42.0*degrees, 90.0*degrees),
+                         SpherePoint(270.0*degrees, 90.0*degrees))
+        self.assertEqual(SpherePoint(-42.0*degrees, -90.0*degrees),
+                         SpherePoint(83.0*degrees, -90.0*degrees))
+
+        self.assertNotEqual(SpherePoint(-42.0*degrees, 90.0*degrees),
+                            SpherePoint(83.0*degrees, -90.0*degrees))
```
r-owen: I suggest you use the same longitude for these two, so that the non-equality is certain to be due to latitude, rather than longitude.

tests/testSpherePoint.py
```diff
@@ -0,0 +1,751 @@
...
+        """White-box test: bearingTo() may be unstable if points are near opposite poles.
+        """
+        southPole = SpherePoint(0.0*degrees, (-90.0 + 1e-14)*degrees)
+        northPoleSame = SpherePoint(0.0*degrees, (90.0 - 1e-14)*degrees)
+        # Don't let it be on exactly the opposite side.
+        northPoleOpposite = SpherePoint(180.0*degrees, (90.0 - 1e-13)*degrees)
+
+        self.assertAnglesNearlyEqual(southPole.bearingTo(northPoleSame), afwGeom.HALFPI*afwGeom.radians)
+        self.assertAnglesNearlyEqual(southPole.bearingTo(northPoleOpposite),
+                                     (afwGeom.PI + afwGeom.HALFPI)*afwGeom.radians)
```
r-owen: Please also test from near each pole to some point that is not near a pole (perhaps on the equator or in another spot there the expected answer is obvious).

Similarly, please make such tests for `offset` and `rotated`. I think those are the only methods sensitive to numerical errors near the pole, though I may have missed something.

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -0,0 +1,343 @@
...
+
+#include "lsst/afw/geom/Angle.h"
+#include "lsst/afw/geom/Point.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+    /**
+     * @brief Point in an unspecified spherical coordinate system.
```
kfindeisen: I find the "unit sphere" wording confusing (we have a representation as unit vectors, sure, but you can use the class exclusively through longitude/latitude if you want). I can add an explicit "does not have distance to an object" to the first paragraph, though.

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -0,0 +1,343 @@
...
+         * @param other the point to test for equality
+         * @return true if this point matches @c other exactly, false otherwise
+         *
+         * @par Exception Safety
+         * This operator shall not throw exceptions.
+         */
+        bool operator==(SpherePoint const & other) const noexcept;
+
+        /**
+         * @brief Test whether two points represent the same position.
```
kfindeisen: Unfortunately, there's a bug in Doxygen that prevents explicit links to operators from working (autoref works, but it's disabled in the Stack). I don't think it's a major problem here (the documentation is generated in the order !=, =, ==, and all three are short), but let me know if that changes your preference.

r-owen: I agree it's not optimal, but I still think it is better not to duplicate information. Also the doc for `operator==` should be right before the doc for `operator!=` so it will be easy to find (in fact the user will likely have already seen it)

tests/testSpherePoint.py
kfindeisen: Do you mean that making test helpers public is the usual pattern?

r-owen: Yes

PR #106, opened 2016-10-17
SpanSets

kfindeisen:
Pull request for simultaneous review of [DM-2427](https://jira.lsstcorp.org/browse/DM-2427), [DM-7170](https://jira.lsstcorp.org/browse/DM-7170), and others. (@rearmstr, can you add to this list?)

### DIFF
--- a/include/lsst/afw/geom/SpanSet.h
+++ b/include/lsst/afw/geom/SpanSet.h
@@ -1,0 +1,578 @@
+// -*- lsst-c++ -*-
+
+/*
+ * LSST Data Management System
+ * Copyright 2008-2016  AURA/LSST.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <https://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_AFW_GEOM_SPANSET_H
+#define LSST_AFW_GEOM_SPANSET_H
+/*
+  Implements a compact representation of a collection of pixels
+ */
+
+#include <vector>
+#include <algorithm>
+#include <functional>
+#include <memory>
+#include <utility>
+#include "lsst/pex/exceptions.h"
+#include "lsst/afw/geom/Span.h"
+#include "lsst/afw/geom/Box.h"
+#include "lsst/afw/geom/SpanSetFunctorGetters.h"
+
+namespace lsst { namespace afw { namespace geom {
+
+/** @brief An enumeration class which describes the shapes
+
+           A stencil is a shape used in creating SpanSets, erosion kernels,
+           or dilation kernels. CIRCLE creates a circle shape, BOX
+           creates a box shape, and MANHATTAN creates a diamond shape.
+ */
+enum class Stencil { CIRCLE, BOX, MANHATTAN };
+
+/**
+ * @brief A compact representation of a collection of pixels
+ *
+ * A SpanSet is a collection of Span classes. As each Span encodes a range of pixels
+ * on a given row, a SpanSet represents an arbitrary collection of pixels on an image.
+ * The SpanSet class also contains mathematical set style operators, for working with
+ * the collection of pixels, and helper functions which make use of the area defined
+ * to perform localized actions
+ */
+class SpanSet {
+ public:
+    typedef std::vector<Span>::const_iterator const_iterator;
+    typedef std::vector<Span>::size_type size_type;
+    typedef Span value_type;
+    typedef value_type const & const_reference;
+    const_iterator begin() const;
+    const_iterator end() const;
+    const_iterator cbegin() const;
+    const_iterator cend() const;
+    const_reference front() const;
+    const_reference back() const;
+    size_type size() const;
+    bool empty() const;
+
+    /** @brief Default constructor
+     *
+     * Construct a null SpanSet, with zero size. This is useful as a placeholder
+     */
+    SpanSet();
+
+    /** @brief Construct a SpanSet from a box
+     *
+     * @param box - A integer box that defines the shape for which a span set should be created
+     */
+    explicit SpanSet(Box2I const & box);
+
+    /** @brief Construct a SpanSet from an iterator
+     *
+     * This constructor accepts the begin and end points of an arbitrary iterator
+     * of a container which contains previously created Spans. These Spans are
+     * used to construct the SpanSet.
+     *
+     * @param begin - Beginning iterator of a container of Spans
+     * @param end - End iterator of a container of Spans
+     * @param normalize - Controls if the constructor attempts to merge connected or
+                          overlapping spans. Defaults to true. Set to false to save
+                          computational time when the container is sure to already
+                          be normalized.
+     */
+    template <typename iter>
+    SpanSet(iter begin, iter end, bool normalize = true):_spanVector(begin, end) {
+        // Return a null SpanSet if spanVector is 0
+        if (_spanVector.size() == 0) {
+            _bBox = geom::Box2I();
+            _area = 0;
+        } else {
+            if (normalize) {
+                _runNormalize();
+            }
+            _initialize();
+        }
+    }
+
+    /** @brief Construct a SpanSet from a std vector by copying
+     *
+     * This constructor accepts a standard vector which contains already created
+     * Spans. These Spans are copied into the internal container of spans.
+     *
+     * @param vec - Standard vector containing Spans
+     * @param normalize - Controls if the constructor attempts to merge connected or
+                          overlapping spans. Defaults to true. Set to false to save
+                          computational time when the container is sure to already
+                          be normalized.
+    */
+    explicit SpanSet(const std::vector<Span> & vec, bool normalize = true);
+
+
+    /** @brief Construct a SpanSet from a std vector from a move
+     *
+     * This constructor accepts a standard vector r-reference which contains already
+     * created Spans. These Spans become the internal container of spans.
+     *
+     * @param vec - Standard vector containing Spans
+     * @param normalize - Controls if the constructor attempts to merge connected or
+                          overlapping spans. Defaults to true. Set to false to save
+                          computational time when the container is sure to already
+                          be normalized.
+    */
+    explicit SpanSet(std::vector<Span> && vec, bool normalize = true);
+
+    // Explicitly delete copy and move constructors
+    SpanSet(const SpanSet & other) = delete;
+    SpanSet(SpanSet && other) = delete;
+
+    // Define class methods
+    /** @brief Return the number of pixels in the SpanSet
+     */
+    size_type getArea() const;
+
+    /** @brief Return a new integer box which is the minimum size to contain
+               the pixels
+     */
+    Box2I getBBox() const;
+
+    /** @brief Defines if the SpanSet is simply contiguous
+     *
+     * If the pixels can be traversed in such a way that every pixel can be reached
+     * without going over a pixel not contained in the SpanSet this method will return
+     * true. If the SpanSet is disjoint aka the above is not true and there is more
+     * than one region, returns false.
+     */
+    bool isContiguous() const;
+
+    /** @brief Return a new SpanSet shifted by specified amount
+     *
+     * @param x - number of pixels to shift in x dimension
+     * @param y - number of pixels to shift in y dimension
+     */
+    std::shared_ptr<SpanSet> shiftedBy(int x, int y) const;
+    /** @brief Return a new SpanSet shifted by specified amount
+     *
+     * @param offset - integer extent which specifies amount to offset in x and y
+     */
+    std::shared_ptr<SpanSet> shiftedBy(Extent2I const & offset) const;
+
+    /** @brief Return a new SpanSet which has all pixel values inside specified box
+     *
+     * @param box - Integer box specifying the bounds for which all pixels must be within
+     */
+    std::shared_ptr<SpanSet> clippedTo(Box2I const & box) const;
+
+    /** @brief Return a new SpanSet who's pixels are the product of applying the specified
+               transformation
+     *
+     * @param t - A linear transform object which will be used to map the pixels
+     */
+    std::shared_ptr<SpanSet> transformedBy(LinearTransform const & t) const;
+
+    /** @brief Return a new SpanSet who's pixels are the product of applying the specified
+               transformation
+     *
+     * @param t - An affine transform object which will be used to map the pixels
+     */
+    std::shared_ptr<SpanSet> transformedBy(AffineTransform const & t) const;
+
+    /** @brief Return a new SpanSet who's pixels are the product of applying the specified
+               transformation
+     *
+     * @param t - A XY transform object which will be used to map the pixels
+     */
+    std::shared_ptr<SpanSet> transformedBy(XYTransform const & t) const;
+
+    /** @brief Specifies if this SpanSet overlaps with another SpanSet
+     *
+     * @param other - A SpanSet for which overlapping comparison will be made
+     */
+    bool overlaps(SpanSet const & other) const;
+
+    /** @brief Check if a SpanSet instance entirely contains another SpanSet
+     *
+     * @param other - The SpanSet who's membership is to be tested for
+     */
+    bool contains(SpanSet const & other) const;
+
+    /** @brief Check if a point is contained within the SpanSet instance
+     *
+     * @param point - An integer point object for which membership is to be tested
+     */
+    bool contains(Point2I const & point) const;
+
+    /** @brief Compute the point about which the SpanSet is symmetrically distributed
+     */
+    Point2D computeCentroid() const;
+
+    /** @brief Compute the shape parameters for the distribution of points in the SpanSet
+     */
+    ellipses::Quadrupole computeShape() const;
+
+    /** @brief Perform a set dilation operation, and return a new object
+     *
+     * Dilate a SpanSet with a kernel specified with the stencil parameter
+     *
+     * @param r - radius of the stencil, the length is inclusive i.e. 3 ranges from -3 to 3
+     * @param s - must be an enumeration of type geom::Stencil. Specifies the shape of the
+                  dilation kernel. May be CIRCLE, MANHATTAN, or BOX
+     */
+    std::shared_ptr<SpanSet> dilate(int r, Stencil s = Stencil::CIRCLE) const;
+
+    /** @brief Perform a set dilation operation, and return a new object
+     *
+     * Dilate a SpanSet with a kernel specified by another SpanSet
+     *
+     * @param other - A SpanSet which specifies the kernel to use for dilation
+     */
+    std::shared_ptr<SpanSet> dilate(SpanSet const & other) const;
+
+    /** @brief Perform a set erosion, and return a new object
+     *
+     * Erode a SpanSet with a kernel specified with the stencil parameter
+     *
+     * @param r - radius of the stencil, the length is inclusive i.e. 3 ranges from -3 to 3
+     * @param s - must be an enumeration of type geom::Stencil. Specifies the shape of the
+                  erosion kernel. May be CIRCLE, MANHATTAN, or BOX
+     */
+    std::shared_ptr<SpanSet> erode(int r, Stencil s = Stencil::CIRCLE) const;
+
+    /** @brief Perform a set erosion operation, and return a new object
+     *
+     * Erode a SpanSet with a kernel specified by another SpanSet
+     *
+     * @param other - A SpanSet which specifies the kernel to use for erosion
+     */
+    std::shared_ptr<SpanSet> erode(SpanSet const & other) const;
+
+    /** @brief Create 1d array at points given by SpanSet
+     *
+     * Take values from an array at points defined by SpanSet and use them to populate a new 1D ndarray
+     *
+     * @param input - The ndarray from which the values will be taken
+     * @param xy0 - A point object with is used as the origin point for the SpanSet coordinate system
+     *
+     * @tparam Pixel - The datatype for the ndarray
+     * @tparam inC - Number of guaranteed row-major contiguous dimensions, starting from the end
+     */
+    template <typename Pixel, int inC>
+    ndarray::Array<Pixel, 1, 1> flatten(ndarray::Array<Pixel, 2, inC>  & input,
+                                        Point2I const & xy0 = Point2I()) const {
+        // Populate a one dimensional array with the values from input taken at the points of SpanSet
+        ndarray::Array<Pixel, 1, 1> outputArray = ndarray::allocate(ndarray::makeVector(getArea()));
+        outputArray.deep() = 0;
+        flatten(outputArray, input, xy0);
+        return outputArray;
+    }
+
+    /** @brief Populate 1d array at points given by SpanSet
+     *
+     * Take values from an array at points defined by SpanSet and use them to populate a new 1D ndarray
+     *
+     * @param output - The 1d ndarray which will be populated with output parameters, will happen in place
+     * @param input - The ndarray from which the values will be taken
+     * @param xy0 - A point object which is used as the origin point for the SpanSet coordinate system
+     *
+     * @tparam Pixel - The data-type for the ndarray
+     * @tparam outC - Number of guaranteed row-major contiguous dimensions in the output array,
+                      starting from the end
+     * @tparam inC - Number of guaranteed row-major contiguous dimensions in the input array,
+                     starting from the end
+     */
+    template <typename Pixel, int outC, int inC>
+    void flatten(
+            ndarray::Array<Pixel, 1, outC> const & output,
+            ndarray::Array<Pixel, 2, inC> const & input,
+            Point2I const & xy0 =  Point2I()) const {
+        auto ndAssigner = []
+                          (Point2I const & point,
+                           typename details::FlatNdGetter<Pixel>::Reference out,
+                           typename details::ImageNdGetter<Pixel, 2, inC>::Reference in)
+                          {out = in;};
+        // Populate array output with values from input at positions given by SpanSet
+        applyFunctor(ndAssigner, ndarray::ndFlat(output), ndarray::ndImage(input, xy0));
+    }
+
+    /** @brief Create 2d array at pionts given by SpanSet
+     *
+     * Take values from a 1d array and insert them in a new 2D array at points defined by the SpanSet,
+     * offset by the lower left hand corner of the bounding box of the SpanSet
+     *
+     * @param input - The ndarray from which the values will be taken
+     *
+     * @tparam Pixel - The data-type for the ndarray
+     * @tparam inC - Number of guaranteed row-major contiguous dimensions, starting from the end
+     */
+    template <typename Pixel, int inC>
+    ndarray::Array<Pixel, 2, 2> unflatten(ndarray::Array<Pixel, 1, inC> & input) const {
+        // Create a two dimensional array the size of the bounding box. Populate values from input, placed at
+        // locations corresponding to SpanSet, offset by the lower corner of the bounding box
+        ndarray::Array<Pixel, 2, 2> outputArray = ndarray::allocate(_bBox.getHeight(), _bBox.getWidth());
+        outputArray.deep() = 0;
+        unflatten(outputArray, input, Point2I(_bBox.getMinX(), _bBox.getMinY()));
+        return outputArray;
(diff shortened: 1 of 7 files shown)


src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+            bool operator()(int mval, PrimaryRun const& pr) {
+                return mval < pr.m;
+            }
+};
+}  // namespace
+
+namespace geom = lsst::afw::geom;
+
+// Expose properties of the underlying vector constaining spans such that the
+// SpanSet can be concidered a container
```
rearmstr: concidered -> "considered"

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+    return area;
+}
+
+// Getter for the bounding box of the SpanSet
+geom::Box2I geom::SpanSet::getBBox() const {
+    return bBox;
+}
+
+
+bool geom::SpanSet::isContiguous() const {
```
rearmstr: Are you assuming here that spans from a single y value can only have two disjoint spans? I'm not sure this algorithm works when you have three, but maybe I'm not understanding it correctly.

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+        if (counter == 0) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool geom::SpanSet::contains(geom::Point2I const & point) const {
+    // Check to see if a given point is found within any spans in this
+    for (auto & spn : spanVector) {
```
rearmstr: Should this also be const?

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    // Iterate over the primary runs in such a way that we consider all values of m
+    // for a given y, then all m for y+1 etc.
+    std::sort(primaryRuns.begin(), primaryRuns.end(), comparePrimaryRun);
+
+    for (int y = primaryRuns.front().y; y <= primaryRuns.back().y; ++y) {
+        auto yRange = std::equal_range(primaryRuns.begin(), primaryRuns.end(), y, ComparePrimaryRunY());
+
+        /* Discard runs for any balue of y for wich we find fewer groups than M,
```
rearmstr: "balue" -> "value"

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+         * the total Y range of the structuring element. THis is step 3.1 of the
```
rearmstr: "THis" -> "This"

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+        for (auto const & otherSpn : other) {
+            if (spansOverlap(spn, otherSpn)) {
+                auto newMin = std::max(spn.getMinX(), otherSpn.getMinX());
+                auto newMax = std::min(spn.getMaxX(), otherSpn.getMaxX());
+                auto newSpan = geom::Span(spn.getY(), newMin, newMax);
+                tempVec.push_back(newSpan);
+            }
+        }
+    }
+    return geom::SpanSet(std::move(tempVec), false);
```
rearmstr: Do you need to nomalize the SpanSet here in case one span from other overlaps two spans?

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+     */
+    std::vector<Span> tempVec;
+    int added;
+    for (auto const & spn : spanVector) {
+        added = 0;
+        for (auto const & otherSpn : other) {
+            if (spansOverlap(spn, otherSpn)) {
+                added = 1;
+                /* To handle one span containing the other, the spans will be added
+                 * peice wise, and let the SpanSet contructor normalize spans which
```
rearmstr: "peice wise" -> "piecewise"

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+                 */
+                if (spn.getMinX() < otherSpn.getMinX()) {
+                    tempVec.push_back(geom::Span(spn.getY(), spn.getMinX(), otherSpn.getMinX()-1));
+                }
+                if (spn.getMaxX() > otherSpn.getMaxX()) {
+                    tempVec.push_back(geom::Span(spn.getY(), otherSpn.getMaxX()+1, spn.getMaxX()));
+                }
+            }
+        }
+        /* If added is still zero, that means it did not verlap any of the spans in other
```
rearmstr: "verlap" -> "overlap"

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+    SpanSet union_(SpanSet const & other) const;
+
+    /// Compareison Operators
+    bool operator==(SpanSet const & other) const;
+    bool operator!=(SpanSet const & other) const;
+
+    static SpanSet spanSetFromShape(int r, Stencil s = Stencil::CIRCLE);
+
+ private:
+    void runNormalize();
```
rearmstr: According to the style guideline 3-4, private member functions must be preceded by an underscore.

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+#pragma clang diagnostic pop
+#include "boost/test/floating_point_comparison.hpp"
+
+#include "lsst/log/Log.h"
+#include "lsst/afw/geom/SpanSet.h"
+#include "lsst/afw/image.h"
+#include "ndarray.h"
+#include "Eigen/Core"
+
+namespace geom = lsst::afw::geom;
```
rearmstr: Change to afwGeom.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+    SpanSet dilate(int r, Stencil s = Stencil::CIRCLE) const;
+    SpanSet dilate(SpanSet const & other) const;
+
+    SpanSet erode(int r, Stencil s = Stencil::CIRCLE) const;
+    SpanSet erode(SpanSet const & other) const;
+
+    template <typename Pixel, int inC>
+    ndarray::Array<Pixel, 1, 1> flatten(ndarray::Array<Pixel, 2, inC>  & input,
+                                                       Point2I const xy0 = Point2I()) const {
+        // Populate a one dimensional array with the values from input at taken at the points of SpanSet
```
rearmstr: "at taken" -> "taken"

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+        return outputArray;
+    }
+
+    template <typename Pixel, int outC, int inC>
+    void flatten(
+            ndarray::Array<Pixel, 1, outC> & output,
+            ndarray::Array<Pixel, 2, inC> & input,
+            Point2I const xy0 =  Point2I()) const {
+        // Populate array output with values from input at positions given by SpanSet
+        applyFunctor([](Point2I point, int n, typename details::FlatNdGetter<Pixel>::Reference out,
```
RobertLuptonTheGood: I think that Point2I is just a pair of ints (sizeof(Point2I) = 2*sizeof(int)), so this may be small enough to count as POD and be passed by value.

kfindeisen: True, but passing by reference is still good practice:
> Even when small objects have inexpensive copy constructors, there can be performance issues. Some compilers treat built-in and user-defined types differently, even if they have the same underlying representation. For example, some compilers refuse to put objects consisting of only a `double` into a register, even though they happily place naked `doubles` there on a regular basis. When that kind of thing happens, you can be better off passing such objects by reference, because compilers will certainly put pointers (the implementation of references) into registers.
-- Effective C++, 3rd Ed., p. 89

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+#include "lsst/afw/geom/Box.h"
+#include "lsst/afw/geom/SpanSetFunctorGetters.h"
+
+namespace lsst { namespace afw { namespace geom {
+
+enum class Stencil { CIRCLE, BOX, MANHATTAN };
+
+class SpanSet {
+ public:
+    typedef std::vector<Span>::const_iterator const_iterator;
```
pschella: Hmm, but not for STL standard conventions I hope?

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+           where the values of the input arguments are intelegently generated at each point
+           in SpanSet, and passed to the functor object for evaluation. The functor object
+           must operate inplace. The first two arguments of the functor must be the point
+           in the SpanSet where the operation is occuring and an integer n representing
+           the number of points that have been operated on. i.e. if the SpanSet has an
+           area of 10 pixels, the functor will be passed a value of 9 on the last
+           invocation. All other arguments to the functor are generated from the getters
+           and will be supplied in the order in which they are passed to applyFunctor
+         */
+        applyFunctorImpl(func, details::makeGetter(x)...);
```
pschella: `std::forward` the arguments.

include/lsst/afw/geom/SpanSet.h
TallJimbo: We should use the STL lowercase form here.  I believe this exception the our guidelines is broadly recognized, but it may need to be made explicit if it isn't already.

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+            bool operator()(PrimaryRun const& pr, int mval) {
+                return pr.m < mval;
+            }
+            bool operator()(int mval, PrimaryRun const& pr) {
+                return mval < pr.m;
+            }
+};
+}  // namespace
+
+namespace geom = lsst::afw::geom;
```
TallJimbo: Actually, we shouldn't need this namespace alias here at all; names in `afw::geom` are already visible.

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+    initialize();
+}
+
+void geom::SpanSet::runNormalize() {
+    // Ensure the span set is sorted according to Span < operator
+    std::sort(spanVector.begin(), spanVector.end());
+
+    // Now that the span is sorted, overlapping spans should be combined
+    // Start from 1 as the comparison will always be with the previous element
+    for (auto iter = ++(spanVector.begin()); iter != spanVector.end(); iter++) {
```
TallJimbo: It probably isn't a context where it matters, but the `++` form has the potential to be more efficient.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,252 @@
...
+
+    template <typename Pixel, int outC, int inC>
+    void unflatten(ndarray::Array<Pixel, 2, outC> & output,
+                                  ndarray::Array<Pixel, 1, inC> & input,
+                                  Point2I const xy0 = Point2I()) const {
+        // Populate 2D ndarray output with values from input, at locations defined by SpanSet, optionally
+        // offset by xy0
+        applyFunctor([](Point2I point, int n, typename details::ImageNdGetter<Pixel, 2, outC>::Reference out,
+                        typename details::FlatNdGetter<Pixel>::Reference in) {out = in;},
+                        ndarray::ndImage(output, xy0), ndarray::ndFlat(input));
```
TallJimbo: Could you use more whitespace in multi-line lambdas like this one (there are a few more in this file) to make them easier to read?  At the very least, the lambda implementation should not be on the same line as part of the argument list if the argument list consumes more than one line.  For example, I'd reformat this as:

```
applyFunctor(
    [](
        Point2I point,
        int n,
        typename details::ImageNdGetter<Pixel, 2, outC>::Reference out,
        typename details::FlatNdGetter<Pixel>::Reference in
    ) {
        out = in;
    },
    ndarray::ndImage(output, xy0),
    ndarray::ndFlat(input)
);
```

(Also, you should pass the `Point` by const reference here as well).

src/geom/SpanSet.cc
natelust: I was just thinking it would avoid some constructor overhead

src/geom/SpanSet.cc
natelust: check

src/geom/SpanSet.cc
natelust: removed

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+
+geom::SpanSet::size_type geom::SpanSet::size() const {
+    return spanVector.size();
+}
+
+bool geom::SpanSet::empty() const {
+    return spanVector.empty();
+}
+
+// Default constructor, creates a null null spanset which may be useful for
```
natelust: check

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+
+
+bool geom::SpanSet::isContiguous() const {
+    /* A contiguous region means that each Span value should have some overlap in the x range with the
+     * previous span, if the current and previous span have different y values
+     */
+    for (auto iter = spanVector.begin() + 1 ; iter != spanVector.end() ; iter++) {
+        // This means they are on the same row, and are not contiguous in x
+        if ((*(iter -1)).getY() == (*iter).getY() && !spansContiguous(*(iter - 1), *iter, false)) {
+            // If two spans are not contiguous on the same row, they must bothe be contiguous with either
```
natelust: check

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+            // the row before or after
+            bool notBefore(false);
+            bool notAfter(false);
+            // ensure the first span is not the begining, if so cant be joined before
```
natelust: check

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+            if ((iter - 1) != spanVector.begin()) {
+                if ( !spansContiguous(*(iter - 2), *(iter - 1), false) ||
+                     !spansContiguous(*(iter - 2), *iter, false)) {
+                         // one of the spans was not contiguous, set notBeforeto true;
```
natelust: check

src/geom/SpanSet.cc
natelust: changed

src/geom/SpanSet.cc
natelust: check

src/geom/SpanSet.cc
natelust: check

src/geom/SpanSet.cc
natelust: Changed to normalize

src/geom/SpanSet.cc
natelust: check

tests/testSpanSets.cc
natelust: changed

include/lsst/afw/geom/SpanSetFunctorGetters.h
```diff
@@ -0,0 +1,261 @@
...
+};
+
+template <typename T>
+class FlatNdGetter {
+    // Getter class to manage iterating though an ndarray which is interpreted as a 2D image
+ public:
+    using Reference = typename ndarray::Array<T, 1, 1>::Reference;
+
+    explicit FlatNdGetter(ndarray::Array<T, 1, 1> const & array): _array(array) {
+        _iter = _array.begin();
```
natelust: changed

include/lsst/afw/geom/SpanSetFunctorGetters.h
```diff
@@ -0,0 +1,261 @@
...
+}
+
+template <typename T, int inA, int inB>
+ImageNdGetter<T, inA, inB> makeGetter(ImageNdGetter<T, inA, inB> & getter) {
+    // This function simply passes though any imageNdGetter passed to it
+    return getter;
+}
+
+template <typename T>
+ImageNdGetter<T, 2, 1> makeGetter(lsst::afw::image::Image<T> & image) {
```
natelust: This is not a redundant overload. They are here to catch the two cases of const vs non const

include/lsst/afw/geom/SpanSetFunctorGetters.h
```diff
@@ -0,0 +1,261 @@
...
+}
+
+template <typename T>
+ImageNdGetter<T, 2, 1> makeGetter(lsst::afw::image::Image<T> const & image) {
+    // Function to create a ndarray getter from an afw image
+    return ImageNdGetter<T, 2, 1>(image.getArray(), image.getXY0());
+}
+
+template <typename T>
+ImageNdGetter<T, 2, 1> makeGetter(lsst::afw::image::Mask<T> & image) {
```
natelust: see above

include/lsst/afw/geom/SpanSetFunctorGetters.h
```diff
@@ -0,0 +1,261 @@
...
+    return IterGetter<T>(iter);
+}
+}}}} // Close namespace lsst::afw::geom::details
+
+namespace ndarray {
+    // These function are placed in the ndarray namespace to enable function argument namespace lookup
+    // This means the function can be used on an ndarray without the need to specify the namespace of
+    // the function itself
+    namespace details = lsst::afw::geom::details;
+    namespace geom = lsst::afw::geom;
```
natelust: changed

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    // overload a few pixels with the value 1
+    input[1][1] = 1;
+    input[1][2] = 1;
+    input[2][1] = 1;
+    input[2][2] = 1;
+    // Create a spanSet for those points
+    std::vector<geom::Span> spanVector = {geom::Span(0, 0, 1), geom::Span(1, 0, 1)};
+    geom::SpanSet SpnSt(spanVector);
+    auto flatArr = SpnSt.flatten(input, geom::Point2I(-1, -1));
+    for (auto const &  val : flatArr) {
```
natelust: updated

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+        BOOST_CHECK(val == 1);
+    }
+}
+
+void testUnflatten() {
+    // Test version without output array, as this simply delegates
+    ndarray::Array<int, 1, 1> input = ndarray::allocate(6);
+    input.deep() = 9;
```
natelust: I have addressed many of these where I felt it appropriate

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+
+    // Create a SpanSet which has six points int it
+    std::vector<geom::Span> spanVector = {geom::Span(9, 2, 3), geom::Span(10, 3, 4), geom::Span(11, 2, 3)};
+    geom::SpanSet SpnSt(spanVector);
```
natelust: Those guidelines say that obscure abbreviations should be avoided to aid in clarity. As dropping vowels is a common practice in programming, and the meaning is not really in question here, I think this is an appropriate name

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    auto output = SpnSt.unflatten(input);
+
+    // Loop over each point in the array correspointing to the spanset, minux xy0 (the min of the SpanSet)
+    // verify the array contains the value 9
+    for (auto const & spn : SpnSt) {
```
natelust: added some checks on dimentionality

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+        for (auto const & pt : spn) {
+            BOOST_CHECK(output[pt.getY()-9][pt.getX()-2] == 9);
+        }
+    }
+    // Verify that a point outside the vector has the value zero
+    BOOST_CHECK(output[0][2] == 0);
```
natelust: Added additional points

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+}
+
+std::pair<lsst::afw::image::Mask<lsst::afw::image::MaskPixel>, geom::SpanSet> populateMask() {
+    // Create a mask and populate it with the value 2
+    lsst::afw::image::Mask<lsst::afw::image::MaskPixel> msk(10, 10);
+    auto spnSt = geom::SpanSet::spanSetFromShape(3).shiftedBy(5, 5);
+    spnSt.setMask(msk, static_cast<lsst::afw::image::MaskPixel>(2));
```
natelust: The reason this was in this function is simply to share code between multiple tests. This does not cause any loss of functionality or clarity, and so I would push back on the necessity of moving it

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+
+void testSetMask() {
+    auto ret = populateMask();
+    auto msk = ret.first;
+    auto spnSt = ret.second;
+    // Verify the mask was populated with the value 2
+    auto mskArray = msk.getArray();
+    for (auto const & spn : spnSt) {
+        for (auto const & pt : spn) {
+            BOOST_CHECK(mskArray[pt.getY()][pt.getX()] == static_cast<lsst::afw::image::MaskPixel>(2));
```
natelust: Added checks on additional points

include/lsst/afw/geom/SpanSetFunctorGetters.h
kfindeisen: It is redundant because your non-const overload doesn't do anything different -- it doesn't modify the image, and it doesn't return a getter with extra permissions. In fact, the call to the `ImageNdGetter` constructor const-qualifies the object returned by `image.getArray()` regardless of whether `image` is const or not, so you would have exactly the same effect if you just had the const method.

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    BOOST_CHECK((firstSS != secondSSShift) == false);
+}
+
+void testFunctor() {
+    // Test the remaining types of functors. Above code has tested ndarray functors
+    // need to test, constants, iterators, Images
+    lsst::afw::image::Image<int> imageObject(5, 5, 0);
+    std::vector<int> vecObject(25, 0);
+
+    auto SSShape = geom::SpanSet::spanSetFromShape(2, geom::Stencil::BOX).shiftedBy(2, 2);
```
natelust: included points outside SpanSet region

include/lsst/afw/geom/SpanSetFunctorGetters.h
kfindeisen: Yes, if the return type is supposed to be `ImageNdGetter<T const, 2, 1>` for the second function then this makes much more sense.

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    // use a constant as a test of constantGetter
+    namespace afwImage = lsst::afw::image;
```
natelust: done

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    std::cout << "before apply" << std::endl;
+    for (auto & val : SSShape) {
+        std::cout << "SS is " << val << std::endl;
+    }
+    SSShape.applyFunctor([](geom::Point2I pt, int n, afwImage::Image<int>::Pixel & out, int in){ out = in; },
+                            imageObject, 6);
```
natelust: done

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+    SSShape.applyFunctor([](geom::Point2I, int n, int & out, int in){out = in;}, vecObject.begin(), 6);
+    auto imageIter = imageObject.begin();
+    auto vectorIter = vecObject.begin();
+    for (int i = 0; i < 25; ++i, ++imageIter, ++vectorIter) {
```
natelust: changed algorithm all together

tests/testSpanSets.cc
```diff
@@ -0,0 +1,397 @@
...
+        BOOST_CHECK(*imageIter == 6);
+        BOOST_CHECK(*vectorIter == 6);
+    }
+}
+
+
+BOOST_AUTO_TEST_CASE(SpanSet_test) {
```
natelust: Split apart

src/geom/SpanSet.cc
```diff
@@ -0,0 +1,640 @@
...
+    newBBoxD.include(t.forwardTransform(geom::Point2D(bBox.getMaxX(), bBox.getMaxY())));
+
+    geom::Box2I newBBoxI(newBBoxD);
+
+    std::vector<Span> tempVec;
+    for (int y = newBBoxI.getBeginY(); y < newBBoxI.getEndY(); ++y) {
+        bool inSpan = false;    // Are we in a span?
+        int start = -1;         // Start of span
+
+        for (int x = newBBoxI.getBeginX(); x < newBBoxI.getEndY(); ++x) {
```
natelust: good catch

tests/testSpanSets.cc
```diff
@@ -0,0 +1,638 @@
...
+BOOST_AUTO_TEST_CASE(SpanSet_testClearMask) {
+    auto result = populateMask();
+    auto mask = result.first;
+    auto spnSt = result.second;
+    spnSt->clearMask(mask, static_cast<lsst::afw::image::MaskPixel>(2));
+    // Verify the mask is now zero
+    auto mskArray = mask.getArray();
+    for (auto const & spn : *spnSt) {
+        for (auto const & pt : spn) {
+            BOOST_CHECK(mskArray[pt.getY()][pt.getX()] == static_cast<lsst::afw::image::MaskPixel>(0));
```
kfindeisen: As above, there's no constraint on `clearMask` actually doing a bitwise operation. While the null case below does check that values are unchanged, it would also be helpful to test this for a nontrivial SpanSet.

tests/testSpanSets.cc
```diff
@@ -0,0 +1,638 @@
...
+                                                      afwGeom::Stencil::BOX)->shiftedBy(spanRadius,
+                                                                                       spanRadius);
+    // use a constant as a test of constantGetter
+    int dataValue = 6;
+    SSShape->applyFunctor([](afwGeom::Point2I pt, afwImage::Image<int>::Pixel & out, int in){ out = in; },
+                            imageObject, dataValue);
+    SSShape->applyFunctor([](afwGeom::Point2I, int & out, int in){out = in;},
+                             vecObject.begin(), dataValue);
+
+    // Check on the point input of a functor
```
kfindeisen: Consider moving this block to after the dataValue/initialValue checks, so that related code is together.

include/lsst/afw/geom/SpanSetFunctorGetters.h
natelust: @kfindeisen I plan on changing the output to T const. I was waiting to see if bob had any feedback before I made any further changes.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -0,0 +1,563 @@
...
+    void applyFunctor(Functor && func, Args && ...args) const {
+        /* Use a variadic template to take a functor object, and an arbitrary number
+           of parameters. For each of the arguments, construct a Getter class using
+           a function (makeGetter) which is overloaded to all the types applyFunctorImpl
+           supports: Images, MaskedImages, Exposures, ndarrays, numeric values, and
+           iterators. The functor and the getters are then passed to the implementation of
+           applyFunctor where the values of the input arguments are intelligently
+           generated at each point in SpanSet, and passed to the functor object for evaluation.
+         */
+        applyFunctorImpl(func, details::makeGetter(args)...);
```
natelust: No, that will be changed. Jim and I tired a few different things but could not get the syntax of forwarding to work quite right with the variadic arguments. I chose to not keep beating my head against it and just wait until Pim got back to work it out the forwarding. It is already fixed in my working copy.

src/geom/SpanSet.cc
```diff
@@ -63,217 +65,293 @@ class ComparePrimaryRunM{
...
+    return _area;
 }
 
 // Getter for the bounding box of the SpanSet
 geom::Box2I geom::SpanSet::getBBox() const {
-    return bBox;
+    return _bBox;
+}
+
+/* Here is a description of how the _makeLabels and _label function works. In the
```
rearmstr: "works" -> "work"

src/geom/SpanSet.cc
```diff
@@ -63,217 +65,293 @@ class ComparePrimaryRunM{
...
+   dimension and have not yet been labeled, and mark them with the current label,
+   and recursively call the _label function with the adjacent row as the new Span
+   under consideration. This results in all Spans which overlap each other in x
+   being labeled with the current label.
+
+   Once _label reaches the end of Spans  which meet it's criteria, control falls
+   back to the _makeLabels function and the current label is incremented. The
+   function then moves onto the next Span in the SpanSet which has not been
+   labeled. If all spans were labeled in the recursive call the loop falls to the
+   end, and the label vector and the last label number  (number of labels) are
```
rearmstr: Extra space after "label number"

include/lsst/afw/geom/SpanSet.h
```diff
@@ -51,152 +70,437 @@ class SpanSet {
...
+    /* A function to combine overlapping Spans in a SpanSet into a single Span
+     */
+    void _runNormalize();
+
+    /* Initializes the SpanSet class. Contains code that is common to multiple constructors
+     */
+    void _initialize();
+
+    /* Label Spans according to contiguous group. If the SpanSet is contiguous, all Spans will be labeled 1.
+     * If there are more than one groups each group will receive a label one higher than the previous.
```
rearmstr: "there are more than one groups" -> "there is more than one group"

PR #107, opened 2016-10-20
DM-7129: temporary coadd covariance experiments

TallJimbo:
This PR is for review comments only; it should not be merged to master.


src/math/warpExposure.cc
```diff
@@ -350,13 +352,326 @@ namespace {
...
+
+    template<typename DestImageT, typename SrcImageT>
+    int doWarpCovariance(
+      DestImageT &destImage,                      ///< remapped %image
+      SrcImageT const &srcImage,                  ///< source %image
+      afwMath::detail::PositionFunctor const &computeSrcPos,   ///< Functor to compute source position
+          ///< called with dest row, column; returns source position (as a Point2D)
+      afwMath::WarpingControl const &control,     ///< warping parameters
+      afwImage::Image<double> &covImage,           ///< covariance matrix
+      typename DestImageT::SinglePixel padValue   ///< value to use for undefined pixels
```
TallJimbo: It may not matter for this throw-away branch, but you might as well set you editor and your habits to 4-space indents rather than 2.

src/math/warpExposure.cc
```diff
@@ -350,13 +352,326 @@ namespace {
...
+              for (typename DestImageT::x_iterator destPtr = destImage.row_begin(y), end = destImage.row_end(y);
+                  destPtr != end; ++destPtr) {
+                  *destPtr = padValue;
+              }
+          }
+          return 0;
+      }
+      PTR(afwMath::SeparableKernel) warpingKernelPtr = control.getWarpingKernel();
+
+      std::shared_ptr<afwMath::LanczosWarpingKernel const> const lanczosKernelPtr = std::dynamic_pointer_cast<afwMath::LanczosWarpingKernel>(warpingKernelPtr);
```
TallJimbo: Too-long line.

src/math/warpExposure.cc
```diff
@@ -350,13 +352,326 @@ namespace {
...
+
+      std::vector<afwGeom::Point2D> _srcPosList(1 + destWidth);
+
+      afwMath::detail::WarpAtOnePoint<DestImageT, SrcImageT> warpAtOnePoint(srcImage, control, padValue);
+
+      std::vector<afwGeom::Point2D>::iterator srcPosView = _srcPosList.begin() + 1;
+      double kernelFactor = 0.0;
+      for (int col = -1; col < destWidth; ++col) {
+          srcPosView[col] = computeSrcPos(col, -1);
+      }
```
TallJimbo: Why (effectively) iterate over `(_srcPosList.begin() + 1 - 1)[col]` instead of just `_srcPostList[col]`?

src/math/warpExposure.cc
```diff
@@ -575,6 +857,9 @@ namespace {
                     if (warpAtOnePoint(destXIter, srcPos, relativeArea,
                         typename lsst::afw::image::detail::image_traits<DestImageT>::image_category())) {
                         ++numGoodPixels;
+                        /*if ((covImage.getWidth() != 0) and (covImage.getHeight() != 0)) { // Sometimes we don't care about computing the covariance matrix. Obviously there are cleaner ways of doing this that don't involve making a check at every loop iteration.
+                          doWarpCovariance(destImage, srcImage, computeSrcPos, control, covImage, padValue, warpAtOnePoint, row, col, kernelWidth, kernelHeight); // Compute covariance between destPixel[col, row] and neighbor destPixels.
+                        }*/
```
TallJimbo: Is this something you still plan to re-enable sometimes, or should it just be removed?

src/math/warpExposure.cc
```diff
@@ -657,7 +948,7 @@ int afwMath::warpCenteredImage(
     t += dt;
     std::cout <<srcImage.getWidth()<<"x"<<srcImage.getHeight()<<": "<< dt <<" "<< t <<std::endl;
 #else
-    int n = warpImage(destImage, srcImageCopy, affXYTransform, control, padValue);
+    int n = warpImage(destImage, srcImageCopy, affXYTransform, control, covImage, padValue);
```
TallJimbo: I don't understand why you've renamed some of these to `warpCovariance` and not others.  Do all of these variants compute both the image and the covariance now?

src/math/warpExposure.cc
AstroVPK: Ok, fixed!

src/math/warpExposure.cc
AstroVPK: Fixed!

src/math/warpExposure.cc
AstroVPK: This bit of code is copied over from the original. The original code had the comment
 // A cache of pixel positions on the source corresponding to the previous or current row
 // of the destination image.
 // The first value is for column -1 because the previous source position is used to compute relative area
 // To simplify the indexing, use an iterator that starts at begin+1, thus:
 // srcPosView = _srcPosList.begin() + 1
 // srcPosView[col-1] and lower indices are for this row
 // srcPosView[col] and higher indices are for the previous row

src/math/warpExposure.cc
AstroVPK: Removed...

src/math/warpExposure.cc
AstroVPK: The naming has been mangled a bit because of the route taken by me to hack at the old code and because of my inexperience with templated code going in. The first variant of dowarpCovariance warps both the image and the covariance matrices while the second one warps just the image. If we end up turning any of this into production code, I'll clean the naming up (amongst a couple of other things that could be cleaned).

PR #109, opened 2016-10-24
afw/DM-8043: getMaskTransparency() and show() need to return values

stargaser:
show() is used in jupiter-enabled devices which need a value to display.


stargaser:
Review complete. The changes look fine and pass PEP8.

PR #110, opened 2016-10-25
Wrap afw::table

fred3m:


python/lsst/afw/table/baseColumnView.py
```diff
@@ -0,0 +1,92 @@
...
+def _set(self, key, value):
+    """
+    Set the value of a column
+    """
+    self[key] = value
+
+def _extract(self, *patterns, **kwds):
+    """
+    Extract a dictionary of {<name>: <column-array>} in which the field names
+    match the given shell-style glob pattern(s).
```
r-owen: I'm very grateful for the docs. However, as a minor style point: the first sentence is a brief description (and usually starts right after the initial `"""` rather than on its own line). Then a blank line, then details (if any).

python/lsst/afw/table/baseColumnView.py
```diff
@@ -0,0 +1,92 @@
+from __future__ import absolute_import
+
+import numpy as np
+
+from ._baseColumnView import *
```
r-owen: Please fix

python/lsst/afw/table/baseColumnView.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _get(self, key):
```
r-owen: This is surprising code: defining a method of a class and then assigning it later. Why is it necessary to do it this way? If this is not normal pybind11 technique then a comment here would be much appreciated, since it's not obvious until much later in the file what's going on.

python/lsst/afw/table/baseColumnView.py
```diff
@@ -0,0 +1,92 @@
...
+      ordered----- If True, a collections.OrderedDict will be returned instead of a standard
+                   dict, with the order corresponding to the definition order of the Schema.
+                   Default is False.
+    """
+    copy = kwds.pop("copy", False)
+    where = kwds.pop("where", None)
+    d = kwds.pop("items", None)
+    if d is None:
+        d = self.schema.extract(*patterns, **kwds).copy()
+    elif kwds:
```
r-owen: A code comment here reminding the reader what's going on would be appreciated.

What I *think* is happening is that `items` is not compatible with positional keywords or with most keywords (excepting the few you already popped). However, if so:
- The error message is not very clear; I think it would help to be explicit about `items`
- You don't seem to test that `patterns` is empty

python/lsst/afw/table/baseRecord.cc
```diff
@@ -22,16 +22,81 @@
...
 using namespace lsst::afw::table;
 
+template <typename T>
+void declareBaseRecordOverloads(py::class_<BaseRecord, std::shared_ptr<BaseRecord>> clsBaseRecord,
+                                const std::string & suffix){
+    clsBaseRecord.def(("_get_"+suffix).c_str(),
+                      (typename Field<T>::Value (BaseRecord::*)(Key<T> const &) const) &BaseRecord::get);
+    clsBaseRecord.def("_getitem_", [](BaseRecord & baseRecord, Key<T> const & key)->typename Field<T>::Reference{
+        /*
+        Define the python __getitem__ method in python to return a baserecord for the requested key
```
r-owen: Please consider `baserecord` -> `BaseRecord` (and similarly below)

python/lsst/afw/table/baseRecord.py
```diff
@@ -0,0 +1,123 @@
+from __future__ import absolute_import
+
+from ._baseRecord import *
+
+def _baseRecordSet(self, key, value):
+    """
+    Given a string or `lsst.afw.table.Key`, find the `BaseRecord` for the given key and set its value.
+    """
+    if isinstance(key, basestring):
```
r-owen: `basestring` is not Python 3 compatible. However, the `future` module provides a version, so if you import that the code will be compatible with Python 2.7 and Python 3

python/lsst/afw/fits/fitsLib.py
```diff
@@ -1 +1,3 @@
 from __future__ import absolute_import
```
r-owen: Please update to `from __future__ import absolute_import, division, print_function`

both here and in all other .py files you modified. It is our standard, and division is helpful in avoiding subtle bugs that may only show up in Python 2.7 or Python 3.

python/lsst/afw/geom/ellipses/__init__.py
```diff
@@ -23,23 +23,20 @@
 """lsst.afw.geom.ellipses
 """
 from __future__ import absolute_import
-from builtins import object
 from .ellipsesLib import *
```
r-owen: We do not permit `from <foo> import *` (except in `__init__.py` files) because it makes the code hard to understand and prevents linters from checking symbols.

Please either import explicit symbols by name or use `import ellipsesLib` and `ellipsesLib.<foo>`. This shows up in many .py files in this commit -- some of which you added (I know this one was pre-existing). Please fix all of the .py files that were touched on this ticket.

TallJimbo: Isn't this *exactly* the case (an `__init__.py` file that brings in wrapped C++ classes) that the exception is for?

python/lsst/afw/geom/ellipses/ellipse.cc
```diff
@@ -22,24 +22,36 @@
...
 //#include <pybind11/operators.h>
-//#include <pybind11/stl.h>
+#include <pybind11/stl.h>
+
+#include "lsst/afw/geom/ellipses/BaseCore.h"
+#include "lsst/afw/geom/Point.h"
+#include "lsst/afw/geom/ellipses/Ellipse.h"
 
 namespace py = pybind11;
+using namespace pybind11::literals;
```
TallJimbo: This is for the `"name"_a` syntax used to define kwargs, so I think we need this using declaration.  Might be good to add a comment to that effect, though if we have it in every file that will quickly become common knowledge.

python/lsst/afw/table/arrays.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
+
+from ._arrays import *
```
TallJimbo: I think we're going to have to introduce a new exception in the conventions for this sort of case; it's filling much the same role as an `__init__.py` file.

python/lsst/afw/table/baseRecord.cc
TallJimbo: When using lambdas or free functions as shims around member functions, I'd recommend we get in the habit of calling the first argument `self`.

python/lsst/afw/geom/ellipses/__init__.py
r-owen: @TallJimbo you are absolutely right. Here. There are many other instances that are not in `__init__.py` files and it was my mistake moving this comment too far up to an `__init__.py` file. My apologies.

python/lsst/afw/fits/fitsLib.py
```diff
@@ -1 +1,3 @@
...
+
+from .._fits import *
```
r-owen: Please change not use `from <foo> import *`

python/lsst/afw/geom/ellipses/ellipsesLib.py
```diff
@@ -1 +1,5 @@
-from __future__ import absolute_import
\ No newline at end of file
+from __future__ import absolute_import
+
+from ._baseCore import *
+from ._quadrupole import *
+from ._ellipse import *
```
r-owen: Please change to not use `from <foo> import *` (and provide closing \n)

timj: I think these are right and the wildcards should be used. They are forwarding the API into `ellipsesLib.py` to retain compatibility with the approach used by SWIG.

python/lsst/afw/geom/ellipses/ellipse.cc
r-owen: Got it. That's certainly an excellent justification and I agree it should be kept as is, without explanatory comment.

python/lsst/afw/table/arrays.py
r-owen: I hope we can avoid it because we lose so much readability and linter functionality with this construct. How bad would it be to import symbols by name or use the library prefix? Obviously I won't fully appreciate the utility until I do some pybind11 wrapping myself.

timj: I think in theory we could get rid of the `*Lib.py` intermediary completely and just import all the symbols directly into the `__init__.py`. These files are so short and are acting like `__init__.py` files so I'm not entirely convinced they are anything to worry about.

python/lsst/afw/fits/SConscript
```diff
@@ -1,3 +1,3 @@
 ## -*- python -*-
 #from lsst.sconsUtils import scripts
-#scripts.BasicSConscript.python(['fitsLib'])
+#scripts.BasicSConscript.pybind11(['fits'])
```
fred3m: In general Pim and I have been leaving the original commented line in all of the Sconscript files and uncommenting and changing them as needed. I think at one point it looked like we would need to wrap the fits.h header but as of now it has not been necessary. However, if we need python to have access to any of those methods this line will need to be uncommented.

python/lsst/afw/SConscript
```diff
@@ -0,0 +1,3 @@
+## -*- python -*-
+from lsst.sconsUtils import scripts
+scripts.BasicSConscript.pybind11(['fits'])
```
fred3m: Ok

python/lsst/afw/fits.cc
```diff
@@ -22,20 +22,28 @@
 
 #include <pybind11/pybind11.h>
 //#include <pybind11/operators.h>
```
fred3m: In general the template file comments out the operators and stl include lines and we uncomment them as needed. Pim and I have been leaving these in because it is common to require new functionality (like operators) when wrapping packages downstream.

python/lsst/afw/fits/__init__.py
```diff
@@ -1,2 +1,2 @@
 from .fitsLib import FitsError, FitsTypeError, MemFileManager
-from .pickleFits import reduceToFits, unreduceFromFits
+#from .pickleFits import reduceToFits, unreduceFromFits
```
fred3m: pickleFits is a library that is not yet needed, and might not be needed at all. When we have finished wrapping afw, if this module (which currently breaks when we attempt to import it) is not needed, we can delete the line. Until then it should remain commented out.

python/lsst/afw/geom/ellipses/__init__.py
```diff
@@ -23,23 +23,20 @@
...
+#Separable = {
+#    (Distortion, DeterminantRadius):SeparableDistortionDeterminantRadius,
+#    (Distortion, TraceRadius):SeparableDistortionTraceRadius,
+#    (Distortion, LogDeterminantRadius):SeparableDistortionLogDeterminantRadius,
+#    (Distortion, LogTraceRadius):SeparableDistortionLogTraceRadius,
+#    (ConformalShear, DeterminantRadius):SeparableConformalShearDeterminantRadius,
+#    (ConformalShear, TraceRadius):SeparableConformalShearTraceRadius,
+#    (ConformalShear, LogDeterminantRadius):SeparableConformalShearLogDeterminantRadius,
+#    (ConformalShear, LogTraceRadius):SeparableConformalShearLogTraceRadius
+#}
```
fred3m: See above. It is better to comment out lines that do not currently work than to delete them and have to search for them later.

python/lsst/afw/geom/ellipses/ellipse.cc
```diff
@@ -22,24 +22,36 @@
...
 
-using namespace lsst::afw::geom::ellipses;
+using namespace lsst::afw::geom;
+using namespace ellipses;
```
fred3m: This is the format we have been using for the pybind11 .cc files. @pschella, do you have any comments about this?

python/lsst/afw/geom/ellipses/quadrupole.cc
```diff
@@ -24,22 +24,40 @@
...
     /* Member types and enums */
+    typedef Eigen::Matrix<double,2,2,Eigen::DontAlign> Matrix;
 
     /* Constructors */
+    clsQuadrupole.def(py::init<double, double, double, bool>(),
+                      "ixx"_a=1.0, "iyy"_a=1.0, "ixy"_a=0.0, "normalize"_a=false);
+    clsQuadrupole.def(py::init<BaseCore::ParameterVector const &, bool>(),
+                      "vector"_a, "normalize"_a=false);
+    clsQuadrupole.def(py::init<Matrix const &, bool>(),
+                               "matrix"_a, "normalize"_a=true);
```
fred3m: Yup, I'll fix that.

python/lsst/afw/table/aggregates.cc
```diff
@@ -21,25 +21,185 @@
...
+    typedef std::vector<std::string> NameArray;
+    
+    py::class_<CovarianceMatrixKey<T,N>> cls(mod, ("CovarianceMatrix"+suffix+"Key").c_str());
+    
+    cls.def(py::init<>());
+    cls.def(py::init<SigmaKeyArray const &, CovarianceKeyArray const &>(),
+            "sigma"_a, "cov"_a=CovarianceKeyArray());
+    cls.def(py::init<SubSchema const &, NameArray const &>());
+    
+    cls.def_static("addFields", (CovarianceMatrixKey<T,N> (*)(
```
fred3m: This is the standard way to wrap an overloaded function (see dmtn-026), which in this case is a static method.

python/lsst/afw/table/aggregates.cc
```diff
@@ -21,25 +21,185 @@
...
 
+    if (_import_array() < 0) {
+            PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+            return nullptr;
+    };
+
     /* Module level */
+    declarePointKey<double>(mod, "2D");
+    declarePointKey<int>(mod, "2I");
+    declareCovarianceMatrixKey<float,2>(mod, "2f");
```
fred3m: Jim is correct (see https://github.com/lsst/afw/blob/057223eaf06faccd9bb52423f8e0505a01a9cbc6/python/lsst/afw/table/aggregates.i)

python/lsst/afw/table/baseColumnView.cc
```diff
@@ -22,24 +22,89 @@
...
+#include "numpy/arrayobject.h"
+#include "ndarray/pybind11.h"
+#include "ndarray/converter.h"
+
+#include "lsst/afw/table/Source.h"
+#include "lsst/afw/table/Key.h"
+#include "lsst/afw/table/BaseColumnView.h"
+
+namespace ndarray { namespace detail {
+template <> struct NumpyTraits<lsst::afw::geom::Angle> : public NumpyTraits<double> {};
```
fred3m: I had never seen this before either. I learned a lot of new C++ in the process of wrapping this module!

python/lsst/afw/table/tableLib.py
```diff
@@ -1 +1,23 @@
...
+from ._schemaMapper import *
+from .baseRecord import *
+from ._baseTable import *
+from .baseColumnView import *
+from .catalog import *
+from ._simple import *
+from ._source import *
+from ._sortedCatalog import *
+
+import lsst.afw.geom.ellipses
```
r-owen: Why is this last import wanted? It is very surprising, and probably worth a code comment.

python/lsst/afw/table/baseColumnView.py
fred3m: see previous comment

tests/testSimpleTable.py
```diff
@@ -434,23 +434,23 @@ def testTicket2850(self):
...
+        cat3 = cat1[np.array([True, True, False], dtype=bool)]
+        self.assertFloatsEqual(cat3[key], np.array([1, 2], dtype=np.int32))
+        cat4 = cat1[np.array([True, False, True], dtype=bool)]
+        self.assertFloatsEqual(cat4.copy(deep=True)[key], np.array([1, 3], dtype=np.int32))
 
     def testTicket2938(self):
         """Test heterogenous catalogs that have records from multiple tables"""
         schema = lsst.afw.table.Schema()
-        schema.addField("i", type=int, doc="doc for i")
+        schema.addField("i", type=np.int32, doc="doc for i")
```
r-owen: Why the change from bare `int` to `np.int32` (and similarly for `float`)? Do we expect users to always use the numpy types now? I hope not, as it will be a lot of retraining.

Also, I would expect `int` to be equivalent to `np.int64` but based on this and other similar changes it is not. Quite surprising.

python/lsst/afw/fits/SConscript
r-owen: We have strict standards about commented-out lines. There should always be an explanation of why the code is being left in place, commented out, and what it will take to get rid of it. In this case some of this is boilerplate and probably best left alone until we can take a final cleanup pass and just throw it all out. But aside from the boilerplate I strongly feel a code comment is necessary if the code is to be left there. Otherwise it's just confusing and frustrating for the reader.

python/lsst/afw/table/baseColumnView.py
fred3m: There is no way to directly extend a C++ class in python, so we have to define a function and then assign it to the class. This is how we've been doing all the modules but if you know a better way I'd be happy to implement it.

python/lsst/afw/table/baseColumnView.py
fred3m: Sure, I can fix that

python/lsst/afw/table/baseColumnView.py
fred3m: This was directly copied from the swig file. In general we haven't been adding new comments unless we have deviated from the SWIG functionality, but I can add some clarifications in places like this where you think they are necessary.

python/lsst/afw/table/aggregates.cc
r-owen: Perhaps I was not being sufficiently clear. What does the code starting with `(CovarianceMatrixKey<T,N> (*)` do?

python/lsst/afw/table/baseRecord.cc
fred3m: Ok

python/lsst/afw/table/baseColumnView.py
r-owen: In my opinion this file is too complicated to justify the use of `from <foo> import *`. It is not just a case of "hoist a few names into an `<foo>Lib.py` file so the `__init__.py` file can hoist them yet another level up".

I am very reluctantly willing to accept trivial cases (though even those violate our standards and should trigger an RFC), but I'm really unhappy about it in any file that has nontrivial code.

tests/testSimpleTable.py
timj: See [RFC-227](https://jira.lsstcorp.org/browse/RFC-227).

python/lsst/afw/table/baseColumnView.py
TallJimbo: I don't think complexity matters here.  What matters is that this file is an implementation detail; it should not be imported anywhere else, and all of its contents should appear to outsiders as if they were defined in this module (or one that imports it).

python/lsst/afw/table/baseRecord.py
fred3m: Good catch, I forgot about that

python/lsst/afw/fits/SConscript
timj: I think the aim is to remove the commented out code when the pybind11 port is about to be merged to master itself. This commented out code will not be retained.

python/lsst/afw/fits/fitsLib.py
fred3m: see above comments

python/lsst/afw/fits/fitsLib.py
fred3m: Done

python/lsst/afw/table/functorKey.cc
```diff
@@ -22,16 +22,43 @@
...
+#include "lsst/afw/table/FunctorKey.h"
 
 namespace py = pybind11;
 
 using namespace lsst::afw::table;
 
+template <typename T>
+void declareFunctorKeys(py::module & mod, const std::string suffix){
+    py::class_<OutputFunctorKey<T>, std::shared_ptr<OutputFunctorKey<T>>>
+        clsOutputFunctorKey(mod, ("OutputFunctorKey"+suffix).c_str());
```
fred3m: Nope, this is what Pim and I have been doing

python/lsst/afw/table/key.py
```diff
@@ -0,0 +1,42 @@
...
+
+from .tableLib import *
+from ._key import *
+
+from .schema import *
+
+def key_eq(self, other):
+    """
+    Compare two keys. If the key types are the same they are checked for equality using the
+    C++ layer. Otherwise NotImplemented is returned.
```
fred3m: Fixed the comment. This is directly from the SWIG code and there is a test for this explicitly, so I recommend leaving it for now.

python/lsst/afw/table/key.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    lsst::afw::table.Key<T> templates are wrapped as different objects in python, for example
+    lsst::afw::table::Key<int> -> lsst.afw.table.Key_I, and
+    lsst::afw::table::Key<Float> -> lsst.afw.table.Key_F.
+    This function sets __eq__ and __ne__ for all of the key objects.
+    """
+    for k in Key:
+        Key[k].__eq__ = key_eq
+        Key[k].__ne__ = key_ne
+        if isinstance(k, basestring) and k.startswith('Array'):
```
fred3m: done

python/lsst/afw/table/schema.py
```diff
@@ -0,0 +1,324 @@
...
+        self.find(key)
+        return True
+    except:
+        return False
+
+def schemaComparisonAnd(self, other):
+    """
+    Enumerated types do not have __and__ and __rand__ operators defined, so for
+    SchemaComparisons we cast the Schema.ComparisonFlag enumerated types into integers
+    to do a bitwise comparison.
```
fred3m: Good point, but this is another place mimicking SWIG functionality. So as of now, no one has  needed any other functionality although this could be a new ticket...

python/lsst/afw/table/tableLib.py
fred3m: Add the comment:

# In order for some of the module to be imported, lsst.afw.geom.ellipses is required
# Even though the modules themselves are wrapped using ellipses, it must still be
# explicitly imported in python

python/lsst/afw/table/baseColumnView.py
TallJimbo: I'm dismayed that linter use seems to encourage otherwise-unnecessary repetition, but it sounds like we should at least have the RFC so we can get others to weigh in.  @fred3m, would you be up for putting that together?

python/lsst/afw/table/aggregates.cc
fred3m: ``CovarianceMatrixKey<T,N>`` is the return type, and the ``(*)`` is needed to cast the function. Typically this would be ``(CovarianceMatrixKey<T,N>::*)``, but since this is a static method we drop the ``CovarianceMatrixKey<T,N>::``. This is required to wrap overloaded functions.

python/lsst/afw/table/baseColumnView.py
fred3m: I can, but I'd like to give @pschella a chance to defend the current procedure to see if he can convince Russel before we go to the trouble of creating an RFC. I'm fairly certain that not only afw but all of the upstream packages that he's wrapped use the same syntax.

python/lsst/afw/table/baseColumnView.py
pschella: There is no way to directly extend a Python class in Python either. And this is the standard way to do it, although it can be hidden behind a class decorator. In fact Nate and I recently wrote a PEP specifically to fix this issue (which was unfortunately rejected by the Python developers). In the current Swig code this approach is used as well, although Swig itself has `%extends` with `%pythoncode` which is used in most places.

python/lsst/afw/table/baseColumnView.py
pschella: In general I would like to add that almost all Python code in the pybind11 wrappers (at least 99.9% of it) is copied verbatim from Swig. The intent has always been to first reproduce what Swig does, and then (possibly) clean it up later.

python/lsst/afw/table/baseRecord.cc
pschella: Using `self` sounds like a good suggestion. We should add that to DMTN-024.

python/lsst/afw/table/functorKey.cc
pschella: I was a bit lazy, not wanting to worry about the size of the buffer when concatenating the suffix.

python/lsst/afw/table/baseRecord.cc
r-owen: Just to be sure I understand: @TallJimbo is suggesting changing `Key<T> const & key)` to `Key<T> const & self)`? (C++ lambdas are still quite new to me).

TallJimbo: Not quite; I was proposing changing `BaseRecord & baseRecord` to `BaseRecord & self`.

python/lsst/afw/fits/SConscript
PaulPrice: I just added some functions in `lsst.afw.fits` recently, so this will go live.

python/lsst/afw/fits.cc
r-owen: Sounds reasonable (especially since we do plan to take a final cleanup pass when pybind11 porting is complete). Unlike some commented-out code it's pretty obvious what this does.

python/lsst/afw/fits/SConscript
r-owen: I still think this deserves a comment in the code describing why it's commented out (so a user has a clue what might be needed to uncomment it or delete it). It is more obscure than the boilerplate pybind11 includes that are commented out.

python/lsst/afw/fits/__init__.py
r-owen: Please add that info as a code comment.

python/lsst/afw/geom/ellipses/__init__.py
r-owen: Yes, that's fine, but they deserve a code comment. Please? Note that in some cases the justification has been "because we don't need it yet" but in others it's because "it doesn't work yet". That is important information that needs to be in the code as a comment, not just on github.

python/lsst/afw/geom/ellipses/ellipse.cc
r-owen: I'm still curious, but it's not worth doing here. I'll try it when it's my turn to do some pybind11 wrapping.

python/lsst/afw/geom/ellipses/ellipsesLib.py
r-owen: It avoids some clutter in the `__init__.py` file, so it's probably reasonable (though that clutter has to go somewhere, so it's not a big win). It's fine with me to leave it, but a "to do" item is to update of our coding standards once we figure out what we want. There are other instances that are more of an issue; @fred3m fixed them in a nice way for this ticket but @pschella has said that technique may not work for other pybind11 code.

python/lsst/afw/table/baseColumnView.py
r-owen: When we find clear disagreements between documentation and code I suggest doing at least one of the following:
- If the code looks reasonable, then fix the docs
- If the code may be buggy then file a ticket and update the docs to say what's going on

In this case I'd file a ticket and note that an exception is raised if extra keywords are provided with `line` (but not if positional args are provided; then reference the ticket).

This pybind11 pass is a good opportunity to catch existing problems and fix them or file tickets about them.

python/lsst/afw/table/schema.py
r-owen: I do request a ticket.

python/lsst/afw/math/functionLibrary.cc
```diff
@@ -61,8 +61,8 @@ void declarePolynomialFunctions(py::module &mod, const std::string & suffix) {
                BasePolynomialFunction2<ReturnT>>
                    clsPolynomialFunction2(mod, ("PolynomialFunction2" + suffix).c_str());
     /* PolynomialFunction2 Constructors */
-    clsPolynomialFunction2.def(py::init<unsigned int>());
     clsPolynomialFunction2.def(py::init<std::vector<double> const &>());
+    clsPolynomialFunction2.def(py::init<unsigned int>());
```
r-owen: I'm curious what happened here. It appears that two lines simply got swapped (here and above at line 50 and a definition at about line 81). If so I'd like to understand why the swap helped (so I can better understand C++).

tests/SConscript
```diff
@@ -7,7 +7,7 @@ afwdataDir = env.ProductDir("afwdata")
 
 with open('test.txt', 'r') as f:
     tests = f.readlines()
-pybind11_ported_tests = [t for t in tests if not t.startswith('#')]
+pybind11_ported_tests = [t[6:].strip() for t in tests if not t.startswith('#')]
```
r-owen: Please add a comment to the SConscript file explaining the "6", and please consider using a named variable (in hopes of enhancing readability).

tests/testSpatialCell.py
```diff
@@ -300,7 +300,7 @@ def testInsertCandidate(self):
         """Test that we can use SpatialCellMaskedImageCandidate"""
 
         flux = 10
-        self.cellSet.insertCandidate(testLib.TestImageCandidate(0, 0, flux))
+        self.cellSet.insertCandidate(afwMath.TestImageCandidate(0, 0, flux))
```
r-owen: Could you please explain what happened here? Is testLib gone?

python/lsst/afw/fits/SConscript
fred3m: I'm going to push back on this just because this is a template file that is automatically created in that directory. I do agree with the other places that you are requesting more information but there are SConscript files commented out exactly like this all over afw and are only uncommented when we finally wrap them. The only reason this one showed up is because I started to wrap this directory until I realized that it was unnecessary.

python/lsst/afw/fits/__init__.py
fred3m: Ok

python/lsst/afw/geom/ellipses/__init__.py
fred3m: Ok. But I think that you'll find as you start wrapping that you'll be doing a lot of commenting and uncommenting code that is not yet wrapped and it would be quite burdensome to always add extra comments when doing this.

python/lsst/afw/math/functionLibrary.cc
fred3m: Ah, this is what was causing one of the tests to fail. pybind11 was updated to allow scalar numpy arrays to be cast as integers or floats. Unlike a C++ compiler, which looks for the best overloaded function to match the parameters, pybind11 move linearly through the overloaded wrappers and uses the first function that it can cast the parameters into. So for overloaded functions, in pybind11 order matters.

tests/testSpatialCell.py
fred3m: Yes, testLib is gone. So the test classes were put into the pybind11 wrapper.

python/lsst/afw/table/schema.py
fred3m: This might not be necessary. Pim thinks that pybind11 has added support for operators in enumerated types, so the next rebase might take care of this issue.

PR #112, opened 2016-10-28
afw/DM-7176: proposed new interface for Footprint

TallJimbo:


TallJimbo:
Closing, as this doesn't make for a useful diff.  Will create a new PR after renaming the changed file back to Footprint.h so I can see the changes.

PR #113, opened 2016-10-28
afw/DM-7176: proposed new API for Footprint

TallJimbo:


include/lsst/afw/detection/Footprint.h
```diff
@@ -291,7 +192,7 @@ class Footprint : public lsst::daf::base::Citizen,
      *
      */
     template <typename MaskPixelT>
-    void intersectMask(
+    std::shared_ptr<Footprint> intersectMask(
         image::Mask<MaskPixelT> const & mask,
         MaskPixelT bitmask=~0x0
     );
```
TallJimbo: I would very much like to get rid of this method.  I think I added it (or Martin, a programmer I worked with as a graduate student, did), and it's always had confusing behavior given its name.  And I think it can be replaced by the more intuitive `SpanSet-Mask` operations we discussed a few days ago.  Do you know where it's being used?

include/lsst/afw/detection/Footprint.h
```diff
@@ -305,32 +206,26 @@ class Footprint : public lsst::daf::base::Citizen,
      *                       NOT the same as the footprint's bounding box.
      *  @param[in]  doClip   If true, clip the new footprint to the region bbox before returning it.
      */
-    PTR(Footprint) transform(
+    std::shared_ptr<Footprint> transform(
```
TallJimbo: I wonder if we should return `unique_ptr` here instead of `shared_ptr`.

Argument in favor:
 - `shared_ptr`s can be constructed from `unique_ptr`s, but not vice versa, so `unique_ptr` gives the caller more flexibility.

Argument against:
 - If we return `shared_ptr`, we can use `make_shared` to construct a `shared_ptr` more efficiently than we could if we went through `unique_ptr`, and we'll probably almost always put our `Footprint`s in `shared_ptr`s, since that's how `SourceRecord` holds them.
 - I'm not sure Swig understands `unique_ptr`.

Given that constructing a `Footprint` (let alone transforming one) is much more expensive than the gain we get from using `make_shared` and we're already considering doing this directly in pybind11, I think `unique_ptr` is clearly better.  That holds for any method that you currently have returning a `shared_ptr<Footprint>`.

include/lsst/afw/detection/Footprint.h
```diff
@@ -305,32 +206,26 @@ class Footprint : public lsst::daf::base::Citizen,
...
-     *  @brief Update the Footprint in-place to be the union of itself and all others provided
-     *
-     *  Only spans will be modified; peaks will be left unchanged.  If ignoreSelf is true it
-     *  will only be the union of all its children.
-     *
-     */
-    void include(std::vector<PTR(Footprint)> const & children, bool ignoreSelf=false);
+    std::shared_ptr<Footprint> intersect(std::shared_ptr<Footprint> other) const;
+    std::shared_ptr<Footprint> intersectNot(std::shared_ptr<Footprint> other) const;
+    std::shared_ptr<Footprint> union_(std::shared_ptr<Footprint> other) const;
```
TallJimbo: I believe we discussed adding versions of these that take `(Mask<T> const & mask, T bitmask)`.

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-
- Dist      :  1 0 0 1 2 2 1 0
-
- "argmin" gives the index of the nearest footprint, and "dist" its
- Manhattan (L_1 norm) distance.  The pixel at index 4 is closest to
- footprint 0, and its distance is 2.
- */
-void nearestFootprint(std::vector<PTR(Footprint)> const& foots,
-                      lsst::afw::image::Image<std::uint16_t>::Ptr argmin,
-                      lsst::afw::image::Image<std::uint16_t>::Ptr dist);
```
TallJimbo: Does this function have any current users?  Like `clipToNonzero`, it seems like a fairly specialized algorithm that should perhaps live with the code that uses it rather than `Footprint` itself, but it also seems tricky enough to implement that we shouldn't remove it lightly.

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-PTR(Footprint) growFootprint(Footprint const& foot, int nGrow,
-                             bool left, bool right, bool up, bool down);
-
-/**
- * Return a list of BBox%s, whose union contains exactly the pixels in
- * foot, neither more nor less
- *
- * Useful in generating sets of meas::algorithms::Defects for the ISR
- */
-std::vector<lsst::afw::geom::Box2I> footprintToBBoxList(Footprint const& foot);
```
TallJimbo: We may want to move this to `ip_isr` instead of deleting it, if that's where it's used.

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-
-/**
- * \brief Set all image pixels in a Footprint to a given value
- *
- * \return value
- */
-template<typename ImageT>
-typename ImageT::Pixel setImageFromFootprint(ImageT *image,
-                                             Footprint const& footprint,
-                                             typename ImageT::Pixel const value);
```
TallJimbo: While it's now trivial to do this in C++ to do this with `SpanSet::applyFunctor` and a lambda, we definitely need to preserve the ability to do it from Python.  However, it should probably be moved to `SpanSet`, and it should take the `Image` argument by reference instead of by pointer.  I also wouldn't object to only defining it in the Python interface.

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-
-/**
- * \brief Set all image pixels in a set of Footprint%s to a given value
- *
- * \return value
- */
-template<typename ImageT>
-typename ImageT::Pixel setImageFromFootprintList(ImageT *image,
-                                                 std::vector<PTR(Footprint)> const& footprints,
-                                                 typename ImageT::Pixel  const value);
```
TallJimbo: I suspect this is a fairly important convenience method that we want to preserve somehow, though it might be best reimplemented in pure Python as loop over an arbitrary iterable of `Footprints` (which would let it be used on a e.g. a generator that extracts `Footprint`s from an iterator over `SourceRecord`s).

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-
-/**
- * \brief OR bitmask into all the Mask's pixels that are in the Footprint
- *
- * \return bitmask
- */
-template<typename MaskT>
-MaskT setMaskFromFootprint(lsst::afw::image::Mask<MaskT> *mask,
-                           Footprint const& footprint,
-                           MaskT const bitmask);
```
TallJimbo: Just to confirm: this functionality is being retained, but is now a `SpanSet` method (and the same for `clearMaskFromFootprint`).

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
-/************************************************************************************************************/
-/**
- * \brief OR bitmask into all the Mask's pixels which are in the set of Footprint%s
- *
- * \return bitmask
- */
-template<typename MaskT>
-MaskT setMaskFromFootprintList(lsst::afw::image::Mask<MaskT> *mask,
-                               std::vector<PTR(Footprint)> const& footprints,
-                               MaskT const bitmask);
```
TallJimbo: Like `setImageFromFootprintList`, maybe we should reimplement this in Python so it can take an arbitrary iterable.

include/lsst/afw/detection/Footprint.h
```diff
@@ -340,212 +235,19 @@ class Footprint : public lsst::daf::base::Citizen,
...
- Respects image xy0, so the footprints should be defined in the
- "PARENT" frame.
-
- Pixels that are in the footprint but do not overlap with both
- images are not copied.
- */
-template <typename ImageOrMaskedImageT>
-void copyWithinFootprint(Footprint const& foot,
-                         PTR(ImageOrMaskedImageT) const input,
-                         PTR(ImageOrMaskedImageT) output);
```
TallJimbo: Not used in Python, I assume?  (I can see it's trivial in C++ with `applyFunctor`.)

TallJimbo:
@natelust I'm done with my review, though you'll note that I have a lot of follow-up questions that largely mirror our in-person talk from last week (with perhaps a few more details).

If he has time, it would also be very useful to get @RobertLuptonTheGood's opinion on this; I think this diff is a concise and complete summary of what we plan to change in the Footprint interface.

include/lsst/afw/detection/Footprint.h
```diff
@@ -219,30 +161,15 @@ class Footprint : public lsst::daf::base::Citizen,
      * totally zero, and moves endpoints to non-zero; it does not
      * split spans that have internal zeros.
      */
-    template<typename PixelT>
-    void clipToNonzero(lsst::afw::image::Image<PixelT> const& img);
+     template<typename PixelT>
+     void cliptoNonZero(image::Image<PixelT> const & img);
```
RobertLuptonTheGood: `git blame` points at Dustin Lang, and sure enough it's used twice in meas_deblender

include/lsst/afw/detection/Footprint.h
RobertLuptonTheGood: Also Dustin Lang

include/lsst/afw/detection/Footprint.h
RobertLuptonTheGood: It's used in processing defects, which are represented as sets of rectangles.

include/lsst/afw/detection/Footprint.h
```diff
@@ -251,38 +178,12 @@ class Footprint : public lsst::daf::base::Citizen,
...
-     * oldIds: if non-NULL, set the IDs that were overwritten
-     * region: Footprint's region (default: getRegion())
-     */
-    template<typename PixelT>
-    void insertIntoImage(typename lsst::afw::image::Image<PixelT>& idImage,
-                         std::uint64_t const id,
-                         bool const overwriteId, long const idMask,
-                         typename std::set<std::uint64_t> *oldIds,
-                         geom::Box2I const& region=geom::Box2I()
-    ) const;
```
RobertLuptonTheGood: Please check, but I think it was probably used only in merging footprints (the uint64 is needed as we needed to keep two uint32 numbers from conflicting, although I now forget why I needed uint32 not uint16).

include/lsst/afw/detection/Footprint.h
RobertLuptonTheGood: It's used in `lsst.afw.image.utils.clipImage`

include/lsst/afw/detection/Footprint.h
```diff
@@ -48,55 +42,31 @@
...
- * A Footprint is a set of pixels, usually but not necessarily contiguous.
- * There are constructors to find Footprints above some threshold in an Image
- * (see FootprintSet), or to create Footprints in the shape of various
- * geometrical figures
- */
 class Footprint : public lsst::daf::base::Citizen,
-                  public afw::table::io::PersistableFacade<lsst::afw::detection::Footprint>,
-                  public afw::table::io::Persistable
+                  public table::io::PersistableFacade<lsst::afw::detection::Footprint>,
+                  public table::io::Persistable
```
natelust: fixed

include/lsst/afw/detection/Footprint.h
natelust: It was only used in meas_deblender. I plan on introducing the fix into meas_deblender in DM-8108

include/lsst/afw/detection/Footprint.h
natelust: The only use I could find was in FootprintSets and the unit test for Footprints. As such I think it should be removed and fixed up with a functor where appropriate.

include/lsst/afw/detection/Footprint.h
natelust: There seems to be only one or two places that use this. I am removing it and will address those places in the fix-up ticket.

include/lsst/afw/detection/Footprint.h
natelust: Changed all the pointers to Footprints to be unique

include/lsst/afw/detection/Footprint.h
natelust: Updated to reflect our past discussion

include/lsst/afw/detection/Footprint.h
natelust: There only seems to be one place using this, and it is in the deblender. This will be removed and updated in the make new Footprints work in the stack ticket.

include/lsst/afw/detection/Footprint.h
natelust: This is only used in utils, ip_isr, and obs_lsstSims, and only once in each of those. This will be able to be replaced with a quick python statement in each of those locations, and will be done in the Footprints stack fixup ticket.

include/lsst/afw/detection/Footprint.h
natelust: This will be turned into a python method in the bindings on SpanSet

include/lsst/afw/detection/Footprint.h
natelust: Ok, this will be set in the python bindings

include/lsst/afw/detection/Footprint.h
natelust: again will be taken care of in the clean up. It is only used in one spot in the deblender. It will be easy to do something different.

include/lsst/afw/detection/Footprint.h
natelust: Yes

jdswinbank:
Now obsolete.

PR #114, opened 2016-11-01
DM-8122: SpherePoint tests fail on MacOS

kfindeisen:
This change works around a difference in how C++ standard library implementations treat output of NaN to a stream that requires all nonnegative numbers to have a sign. It should allow unit tests to pass on both Linux and MacOS platforms.


parejkoj:
The new test file passes for me on OSX 10.11, so I think we're good. Thanks for the fix.

PR #115, opened 2016-11-01
afw/DM-8147: Make ExposureCatalog input even more backwards-compatible.

TallJimbo:
Some old ExposureCatalogs saved as part of early CoaddPsfs don't even have the fields for ApCorrMap or ValidPolygon, and we'd like to use those for the PDAC.

### DIFF
--- a/src/table/Exposure.cc
+++ b/src/table/Exposure.cc
@@ -32,7 +32,16 @@ namespace {
 
 int const EXPOSURE_TABLE_CURRENT_VERSION = 2;  // current version of ExposureTable
 std::string const EXPOSURE_TABLE_VERSION_KEY = "EXPTABLE_VER";  // FITS header key for ExposureTable version
-std::string const VISIT_INFO_FIELD_NAME = "visitInfo"; // name of field used for VisitInfo
+
+// Field names used to store the archive IDs of different components (used in
+// multiple places, so we define them here instead of as multiple string
+// literals).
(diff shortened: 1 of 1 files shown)


src/table/Exposure.cc
```diff
@@ -439,18 +466,12 @@ ExposureCatalogT<RecordT> ExposureCatalogT<RecordT>::readFromArchive(
 ) {
     // table metadata is not available, so we don't have a clean way to determine the version;
     // the hack is version = current version if visitInfo key is found, else 1
```
r-owen: This comment needs updating

TallJimbo:
> Does this code scale to new versions of VisitInfo, PSF, etc.? I'm guessing "yes" since it is explicitly working with old and new versions of PSF. Assuming that's right, this looks like a nice improvement.

It doesn't do this automatically, of course - that backwards compatibility for those components has to be added to the persistence code for those components.  But this code doesn't get in the way of that.

src/table/Exposure.cc
```diff
@@ -32,7 +32,12 @@ namespace {
 
 int const EXPOSURE_TABLE_CURRENT_VERSION = 2;  // current version of ExposureTable
 std::string const EXPOSURE_TABLE_VERSION_KEY = "EXPTABLE_VER";  // FITS header key for ExposureTable version
+std::string const WCS_FIELD_NAME = "wcs";
```
TallJimbo: Done.

PR #116, opened 2016-11-02
Load LST into VisitInfo tickets/DM-8052

parejkoj:
Adding LST into VisitInfo, by way of a "fake" ERA.


parejkoj:
I agree that the `*1/*2` split wasn't necessarily an improvement, but I wasn't sure how best to handle the non-visitInfo data (HA/LST) that we're also checking. As you say, it's easy enough to bundle all the `data` values into a tuple if we want.

PR #117, opened 2016-11-08
table.Schema: allow construction from FITS file

PaulPrice:
Sometimes we want the Schema from a catalog without having to read
the entire catalog (e.g., when setting up for concatenating a bunch
of catalogs). This patch adds two methods to help with that.

### DIFF
--- a/include/lsst/afw/table/Schema.h
+++ b/include/lsst/afw/table/Schema.h
@@ -272,6 +272,20 @@ public:
     /// @brief Copy constructor.
     Schema(Schema const & other);
 
+    /// @brief Construct from reading a FITS file.
+    ///
+    /// Reads from the nominated 'hdu' (1=PHU which cannot be a catalog,
+    /// 0 is a special value meaning read from the first HDU with NAXIS != 0).
+    static Schema readFits(std::string const& filename, int hdu=0);
+    static Schema readFits(fits::MemFileManager & manager, int hdu=0);
+    static Schema readFits(fits::Fits & fitsfile);
+
+    /// @brief Construct from reading a FITS header
+    ///
+    /// If 'stripMetadata', then the header will be modified,
+    /// removing the relevant keywords.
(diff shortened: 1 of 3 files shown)


include/lsst/afw/table/Schema.h
```diff
@@ -272,6 +272,17 @@ class Schema {
     /// @brief Copy constructor.
     Schema(Schema const & other);
 
+    /// @brief Construct from reading a FITS file.
+    ///
+    /// Reads from the nominated 'hdu' (1=PHU which cannot be a catalog).
+    static Schema readFits(std::string const& filename, int hdu=2);
```
TallJimbo: `Catalog::readFits` uses a default of `hdu=0`, where 0 is interpreted as "the first HDU with `NAXIS != 0`.

tests/testTableIO.py
```diff
@@ -45,6 +46,25 @@ def testAngleUnitWriting(self):
...
+        row = cat.addNew()
+        row.set(aa, 12345)
+        row.set(bb, 1.2345)
+        with lsst.utils.tests.getTempFilePath(".fits") as temp:
+            cat.writeFits(temp)
+            self.assertEqual(lsst.afw.table.Schema.readFits(temp), schema)
+            # Not testing Schema.fromFitsMetadata because afw.image.readMetadata (which is the only
+            # python-accessible FITS header reader) returns a PropertySet, but we want a PropertyList
+            # and it doesn't up-convert easily.
+
```
TallJimbo: Maybe add an alias to the schema to ensure those are also round-tripped?

include/lsst/afw/table/Schema.h
TallJimbo: Do we want to add the `Fits` and `MemFileManager` overloads too?  I'm not sure they're useful here, but that's the established pattern.

PR #119, opened 2016-11-10
afw/DM-8221: Add test helper for comparing Schemas.

TallJimbo:
This is test support code used to help diagnose DM-8221 that may be useful long-term.

### DIFF
--- a/python/lsst/afw/table/testUtils.py
+++ b/python/lsst/afw/table/testUtils.py
@@ -1,0 +1,95 @@
+#
+# LSST Data Management System
+# Copyright 2016 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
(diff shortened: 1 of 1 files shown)


python/lsst/afw/table/testUtils.py
```diff
@@ -0,0 +1,88 @@
...
+        will be prefixed with "-" in the diff.
+    schema2 : :py:class:`lsst.afw.table.Schema`
+        Second schema to diff.  Items appearing only in this schema
+        will be prefixed with "-" in the diff.
+    flags : `int`
+        A bitwise OR of :py:class:`lsst.afw.table.Schema.ComparisonFlags`
+        indicating which features of schema items to compare.  The returned
+        diff will always show all differences, but no diff will be shown if
+        the only differences are not included in the flags.  Default is
+        `lsst.afw.table.Schema.IDENTICAL`, which checks everything.
```
r-owen: It seems a bit odd to show diffs that are not in the flags, but I'm guessing that made the code simpler.

In any case it's a minor point. This seems like a really useful utility.

python/lsst/afw/table/testUtils.py
```diff
@@ -0,0 +1,88 @@
...
+    diff = "\n".join(difflib.unified_diff(str(schema1).split("\n"), str(schema2).split("\n")))
+    if len(components) == 1:
+        msg = "%s differ:\n%s" % (components[0].capitalize(), diff)
+    elif len(components) == 2:
+        msg = "%s and %s differ:\n%s" % (components[0].capitalize(), components[1], diff)
+    else:
+        msg = "%s, %s, and %s differ:\n%s" % (components[0].capitalize(),
+                                              ", ".join(components[1:-1]),
+                                              components[-1], diff)
+    return msg
```
r-owen: I think this would be a bit cleaner if you made a separate function that created the list of components (with or without capitalization--that could always be done afterwards since it would be a short string). Whether the gain is worth the cost of further coding is a valid question.

python/lsst/afw/table/testUtils.py
TallJimbo: You guessed correctly.  It'd be really hard to actually modify the diff to pay attention to the flags.

python/lsst/afw/table/testUtils.py
TallJimbo: Done.

PR #120, opened 2016-11-14
Tickets/dm 8213

r-owen:


include/lsst/afw/fits.h
```diff
@@ -510,6 +511,19 @@ class Fits {
...
 
+//@{
+/// @brief Read FITS header
+///
+/// If 'strip' is true, common FITS keys that usually have non-metadata intepretations
+/// (e.g. NAXIS, BITPIX) will be ignored.
+///
+/// Includes support for the INHERIT convention: if 'INHERIT = T' is in the header, the
+/// PHU will be read as well, and nominated HDU will override any duplicated values.
+PTR(daf::base::PropertyList) readMetadata(std::string const & fileName, int hdu=0, bool strip=false);
```
r-owen: This seems like a very logical set of new functions.

I'm a bit disappointed that we can't just return a PropertyList instead of a pointer to a property list. I assume the pointer is used to avoid copying the data (though I wonder if C++ really would copy in this situation, or if it could swap).

Please document all three functions independently, so they all show up as documented in the Doxygen. If you want to void repetition then I suppose you could fully document one version and have a one-liner for the others that points to the full docs of the one, but personally I'd just add the repetition.

src/fits.cc
```diff
@@ -1128,6 +1128,47 @@ void Fits::closeFile() {
...
+}
+
+PTR(daf::base::PropertyList) readMetadata(fits::MemFileManager & manager, int hdu, bool strip) {
+    fits::Fits fp(manager, "r", fits::Fits::AUTO_CLOSE | fits::Fits::AUTO_CHECK);
+    fp.setHdu(hdu);
+    return readMetadata(fp, strip);
+}
+
+PTR(daf::base::PropertyList) readMetadata(fits::Fits & fitsfile, bool strip) {
+    PTR(daf::base::PropertyList) metadata(new lsst::daf::base::PropertyList);
```
r-owen: Is there's less repetitive way way to do this (and similar code below) with auto? Would this work?:
```
auto readMetadata = std::make_shared<daf::base::PropertyList>();
```

include/lsst/afw/fits.h
PaulPrice: As I understand it, the `//@{` at the top of the doc and the `//@}` below the last declaration of `readMetadata` group the declarations together, so they all share the same doc.

The pointer is used because that's what we had before and I didn't want to change it without thinking more carefully.

PR #121, opened 2016-11-14
Fixed build error

kfindeisen:
@pschella, trying to build the stack gives the following error:
bq. [2016-11-14T21:46:55.323114Z] python/lsst/afw/math/functionLibrary.cc:30:43: fatal error: lsst/afw/math/functionLibrary.h: No such file or directory


kfindeisen:
Commits moved to `u/kfindeisen/pybind11unix` for testing.

PR #122, opened 2016-11-15
Fixed build errors.

kfindeisen:
All errors are caused by trying to include a file with the wrong capitalization.

@pschella, can you confirm and merge?


pschella:
Confirmed

PR #123, opened 2016-11-23
DM-8374: Add function for Pybind11 wrapping of PersistableFacade

kfindeisen:
This file centralizes wrapping of `lsst::afw::table::io::PersistableFacade`. It is designed to be included in the pybind11 `.cc` files of classes that implement `PersistableFacade` using the curiously recurring template pattern.

The function has been successfully tested with local wrappers for `ApCorrMap` and `ChebyshevBoundedField` written as part of DM-7797.

### DIFF
--- a/include/lsst/afw/table/io/pybind11.h
+++ b/include/lsst/afw/table/io/pybind11.h
@@ -1,0 +1,69 @@
+// -*- lsst-c++ -*-
+/*
+ * LSST Data Management System
+ * See COPYRIGHT file at the top of the source tree.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_AFW_TABLE_IO_PYBIND11_H
+#define LSST_AFW_TABLE_IO_PYBIND11_H
+
+#include <memory>
+#include <string>
+
+#include <pybind11/pybind11.h>
+
+#include "lsst/afw/table/io/Persistable.h"
+
+namespace lsst {
+namespace afw {
+namespace table {
+namespace io {
+
+    /**
+     * @brief Wraps an instantiation of @ref PersistableFacade.
+     *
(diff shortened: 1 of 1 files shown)


include/lsst/afw/table/io/pybind11.h
```diff
@@ -0,0 +1,63 @@
...
+     * @param module The pybind11 module that shall contain `PersistableFacade<T>`
+     * @param suffix A string to disambiguate this class from other
+     *               `PersistableFacades`. The Python name of this class shall be
+     *               `PersistableFacade<suffix>`.
+     */
+    template <typename T>
+    void declarePersistableFacade(pybind11::module & module, std::string const & suffix) {
+        using namespace pybind11::literals;
+        // shared_ptr is used by subclasses of PersistableFacade
+        pybind11::class_<PersistableFacade<T>, std::shared_ptr<PersistableFacade<T>>>
```
pschella: Are we sure that the holder type is always `shared_ptr`? If so then we should probably specify this in the documentation because everything that inherits from the facade needs the same holder type.

pschella: If not all subclasses use `shared_ptr` then it isn't a problem to require that holder type anyway. Just a tiny bit less efficient.

include/lsst/afw/table/io/pybind11.h
```diff
@@ -0,0 +1,63 @@
...
+                clsFacade(module, ("PersistableFacade" + suffix).c_str());
+        clsFacade.def_static("readFits",
+                             (PTR(T) (*)(std::string const &, int)) &PersistableFacade<T>::readFits,
+                             "fileName"_a, "hdu"_a=0);
```
pschella: Is this the only member function defined? Or do we need to add more later?

pschella: Also note that the `PTR` macro is slated for removal.

include/lsst/afw/table/io/pybind11.h
kfindeisen: You said that all classes in an inheritance heirarchy need to use the same holder type, and since `ApCorrMap` and `ChebyshevBoundedField` do, it follows that `Persistable` and `PersistableFacade` need to. Did I misunderstand?

include/lsst/afw/table/io/pybind11.h
kfindeisen: We may need to add more later, but I'm trying to stick to the "only if it shows up in a test case" principle.

Noted, but the current declaration of `PersistableFacade::readFits` uses `PTR` so I thought it best to be consistent.

include/lsst/afw/table/io/pybind11.h
pschella: No you did not. This is fine. But it does mean that everything that inherits from `PersistableFacade` now needs to use a `shared_ptr` holder type. This might be true already, but not nessesarily. Thus just add a sentence to the docstring saying that.

include/lsst/afw/table/io/pybind11.h
pschella: Yes, that is fine.

include/lsst/afw/table/io/pybind11.h
kfindeisen: I think `Persistable` already made that true, but I'll document it.

PR #124, opened 2016-11-28
Tickets/DM-8171 Add keyword arguments to afwDisplay.setDefaultBackend

stargaser:
The extra parameters passed to setDefaultBackend have to be made "sticky" to apply to subsequent definitions of Display objects.

### DIFF
--- a/python/lsst/afw/display/interface.py
+++ b/python/lsst/afw/display/interface.py
@@ -147,11 +147,24 @@ class Display(object):
                     setDefaultBackend("ds9")
                 except RuntimeError:
                     setDefaultBackend("virtualDevice")
+            else: # a default was set, merge optional arguments
+                try:
+                    self._args += args
+                except AttributeError:
+                    self._args = args
+                try:
+                    self._kwargs.update(kwargs)
+                except AttributeError:
+                    self._kwargs = kwargs
 
             backend = Display._defaultBackend
+        else: # a backend was specified, discard default optional arguments
+            self._args = args
+            self._kwargs = kwargs
 
         self.frame = frame
-        self._impl = _makeDisplayImpl(self, backend, *args, **kwargs)
+        
+        self._impl = _makeDisplayImpl(self, backend, *self._args, **self._kwargs)
         self.name = backend
 
         self._xy0 = None                # the data displayed on the frame's XY0
@@ -194,9 +207,23 @@ class Display(object):
     def __del__(self):
         self.close()
 
-    def __getattr__(self, name, *args, **kwargs):
-        """Try to call self._impl.name(*args, *kwargs)"""
-        
+    def __getattr__(self, name):
+        """Return the attribute of self._impl, or ._impl if it is requested
+
+        Parameters:
+        -----------
+            name : string
+                name of the attribute requested
+
+        Returns:
+        --------
+            attribute : object
+                the attribute of self._impl for the requested name
+        """
+
+        if name == '_impl':
+            return object.__getattr__(self, name)
+
         if not (hasattr(self, "_impl") and self._impl):
             raise AttributeError("Device has no _impl attached")
 
@@ -230,17 +257,22 @@ class Display(object):
     # Handle Displays, including the default one (the frame to use when a user specifies None)
     #
     @staticmethod
-    def setDefaultBackend(backend):
+    def setDefaultBackend(backend, *args, **kwargs):
+        Display._args = args
+        Display._kwargs = kwargs
         try:
-            _makeDisplayImpl(None, backend)
+            _makeDisplayImpl(None, backend, *Display._args, **Display._kwargs)
         except Exception as e:
             raise RuntimeError("Unable to set backend to %s: \"%s\"" % (backend, e))
 
         Display._defaultBackend = backend
 
     @staticmethod
-    def getDefaultBackend():
-        return Display._defaultBackend
+    def getDefaultBackend(optional_args=False):
+        if optional_args:
+            return(Display._defaultBackend, Display._args, Display._kwargs)
+        else:
+            return Display._defaultBackend
 
     @staticmethod
     def setDefaultFrame(frame=0):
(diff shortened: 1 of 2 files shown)


python/lsst/afw/display/interface.py
```diff
@@ -230,9 +252,11 @@ def __str__(self):
     # Handle Displays, including the default one (the frame to use when a user specifies None)
     #
     @staticmethod
-    def setDefaultBackend(backend):
+    def setDefaultBackend(backend, *args, **kwargs):
```
jdswinbank: Since you're changing `setDefaultBackend` to not only set the default backend but also to set default arguments for it, doesn't symmetry suggest you also change `getDefaultBackend` to return those arguments as well?

python/lsst/afw/display/interface.py
```diff
@@ -151,7 +151,15 @@ def __init__(self, frame=None, backend=None, *args, **kwargs):
             backend = Display._defaultBackend
 
         self.frame = frame
-        self._impl = _makeDisplayImpl(self, backend, *args, **kwargs)
+        try:
+            self._args += args
+        except AttributeError:
+            self._args = args
```
jdswinbank: Here's an example of something which I think should work, but doesn't:

```
In [1]: import lsst.afw.display as afwDisplay

In [2]: afwDisplay.setDefaultBackend("ds9", host="xxx.yyy.zzz", port=12345)

In [3]: afwDisplay.Display(backend="virtualDevice")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-08c8161175ba> in <module>()
----> 1 afwDisplay.Display(backend="virtualDevice")

/ssd/swinbank/afw/python/lsst/afw/display/interface.py in __init__(self, frame, backend, *args, **kwargs)
    160         except AttributeError:
    161             self._kwargs = kwargs
--> 162         self._impl = _makeDisplayImpl(self, backend, *self._args, **self._kwargs)
    163         self.name = backend
    164

/ssd/swinbank/afw/python/lsst/afw/display/interface.py in _makeDisplayImpl(display, backend, *args, **kwargs)
    101
    102     if display:
--> 103         _impl = _disp.DisplayImpl(display, *args, **kwargs)
    104         if not hasattr(_impl, "frame"):
    105             _impl.frame = display.frame

TypeError: __init__() got an unexpected keyword argument 'host'
```

stargaser:
Revisions made and rebased.

- The Display initializer will now clear the optional arguments when a backend is specified
- getDefaultBackend will optionally return the optional arguments
- Two tests of these functionalities have been added to the afw testDisplay.py suite

python/lsst/afw/display/interface.py
stargaser: Yes, good point! I have added an option to return the extra arguments.

tests/testDisplay.py
```diff
@@ -90,6 +90,36 @@ def testWith(self):
         with afwDisplay.getDisplay(0) as disp:
             self.assertIsNotNone(disp)
 
+    def testDefaultBackend(self):
+        """Test setting and getting of default backend"""
+        if (backend != 'virtualDevice'):
```
jdswinbank: Thank you for adding these tests!

However: when this is run by Jenkins, isn't the `backend` always going to be `"virtualDevice"`? If so, these tests — and particularly `testClearDefaults()`, below — aren't really testing very much.

python/lsst/afw/display/interface.py
```diff
@@ -230,17 +257,22 @@ def __str__(self):
...
+            _makeDisplayImpl(None, backend, *Display._args, **Display._kwargs)
         except Exception as e:
             raise RuntimeError("Unable to set backend to %s: \"%s\"" % (backend, e))
 
         Display._defaultBackend = backend
 
     @staticmethod
-    def getDefaultBackend():
-        return Display._defaultBackend
+    def getDefaultBackend(optional_args=False):
```
jdswinbank: Thanks for adding `optional_args` here.

I'm still worried about this interface, though. In particular, the fact that you can't in general do `afwDisplay.setDefaultBackend(afwDisplay.getDefaultBackend())` seems like a red flag.

I'm wondering if we actually need to adjust how the information about the backend is being encapsulated. Rather than a naked string and some free-floating arguments we need to remember to associate with it, you could bundle it all up into a tiny class:

```python
class DisplayBackend(object):
    @property
    def name(self):
        return (self.__class__.__name__[0].lower() +
                self.__class__.__name__[1:])

class VirtualDevice(DisplayBackend):
    pass

class Firefly(DisplayBackend):
    def __init__(self, host="localhost", port="8080"):
        self.host, self.port = host, port
```

The backend itself then tells you all you need to know about it, and you no longer have to worry about keeping your `args` in sync. This also makes handling arguments in `Display.__init__` easier: you no longer need to worry about the case when `self._args` or `self._kwargs` doesn't exist. In fact, you can just drop the `*args, **kwargs` arguments to `Display.__init__` entirely, and just have the user pass `DisplayBackend` subclasses with all the arguments already provided.

I should add that this isn't a requirement. In fact, I'd need to think harder about whether it's even a good idea: a new class makes life harder for backend implementers, you'd need to rework `_makeDisplayImpl` to handle it, and, logically, I'm not sure whether information about things like hosts and ports belongs with the backend at all (is the host it's running on a property of the backend or a property of the `Display`?).

So I'm not sure how to proceed here, but the current design does give me the cold shivers.

python/lsst/afw/display/interface.py
```diff
@@ -147,11 +147,24 @@ def __init__(self, frame=None, backend=None, *args, **kwargs):
...
+                    self._kwargs = kwargs
 
             backend = Display._defaultBackend
+        else: # a backend was specified, discard default optional arguments
+            self._args = args
+            self._kwargs = kwargs
 
         self.frame = frame
-        self._impl = _makeDisplayImpl(self, backend, *args, **kwargs)
+
```
stargaser: I didn't notice the whitespace -- I can take it out. 

I'm using Sublime Text 3 -- I wonder if there's a way to configure it to point out issues like this? Or would it be covered by `pep8`?

python/lsst/afw/display/interface.py
stargaser: These are excellent points about the deficiencies in the interface. I agree that the implemented solution is brittle. 

Logically, the host and port belong to the `Display`. A user can specify host and port for each Firefly Display along with the frame number, and optionally, a channel name. 

After some discussion with @xiuqin , we are thinking it is better to not implement this change to `setDefaultBackend` and to just tell people to use the method that already works, using the `Display` constructor.  This pull request just provides an alternative path to defining displays, that would probably be better left to a future redesign of the afwDisplay interface.

tests/testDisplay.py
stargaser: Yes, the new tests will not be run by Jenkins.  A proper test using `testDisplay.py` requires specifying a backend (`ds9` or `firefly` currently) and at least one user click to test everything. Since the virtual backend takes no optional arguments, it is not possible to test the defaultBackend setting of these with it.

jdswinbank:
This was Won't Fix-ed.

PR #126, opened 2016-11-29
Tickets/dm 8264

r-owen:


python/lsst/afw/table/match.cc
```diff
@@ -22,16 +22,78 @@
 
 #include <pybind11/pybind11.h>
 //#include <pybind11/operators.h>
-//#include <pybind11/stl.h>
+#include <pybind11/stl.h>
+
+#include "lsst/pex/config/pybind11.h"  // for LSST_DECLARE_CONTROL_FIELD
```
fred3m: Thank you for this comment. That will make it much more useful to find out where this is declared in the future.

python/lsst/afw/table/source.cc
```diff
@@ -59,8 +60,101 @@ PYBIND11_PLUGIN(_source) {
...
+    //clsSourceRecord.def("getShape", &SourceTable::getShape);
+    //clsSourceRecord.def("getShapeErr", &SourceTable::getShapeErr);
+    //clsSourceRecord.def("getShapeFlag", &SourceTable::getShapeFlag);
+
+    //clsSourceRecord.def("getX", &SourceTable::getX);
+    //clsSourceRecord.def("getY", &SourceTable::getY);
+    //clsSourceRecord.def("getIxx", &SourceTable::getIxx);
+    //clsSourceRecord.def("getIyy", &SourceTable::getIyy);
+    //clsSourceRecord.def("getIxy", &SourceTable::getIxy);
+
```
fred3m: There's a lot of code here commented out, is there a reason for that?

tests/test.txt
```diff
@@ -80,16 +80,16 @@ tests/testSpline.py
 #tests/testStatistics.py
```
fred3m: Should tests/testSourceMatch.py be uncommented? I know that there is one test that doesn't run properly yet, so could that test be wrapped with a ```@unittest.skip("Skip test until pickling is fixed")``` decorator so that we ensure that nothing else breaks the other tests in it that pass?

tests/testAmpInfoTable.py
```diff
@@ -115,7 +115,7 @@ def testBasics(self):
             self.assertEquals(saturation, record.getSaturation())
             self.assertEquals(readNoise, record.getReadNoise())
             self.assertEquals(readoutCorner, record.getReadoutCorner())
-            self.assertEquals(linearityCoeffs, record.getLinearityCoeffs())
+            self.assertEquals(linearityCoeffs, tuple(record.getLinearityCoeffs()))
```
fred3m: I would recommend switching this to ```self.assertEquals(list(linearityCoeffs), record.getLinearityCoeffs())``` so that we are casting the type of the comparison tuple to the expected type of record.getLinearityCoeffs().

tests/testMatchFits.py
```diff
@@ -44,20 +44,20 @@ def setUp(self):
         self.cat1 = afwTable.SimpleCatalog(self.schema)
         self.cat2 = afwTable.SimpleCatalog(self.schema)
         for i in range(self.size):
-            record1 = self.cat1.table.makeRecord()
-            record2 = self.cat2.table.makeRecord()
+            record1 = self.cat1.makeRecord()
+            record2 = self.cat2.makeRecord()
```
fred3m: In general I'm not sure it's a good idea to effectively change the API in a test. Is there a reason this method was moved from the table to the catalog?

include/lsst/afw/table/pybind11Catalog.h
```diff
@@ -0,0 +1,194 @@
...
+    cls.def("writeFits",
+                    (void (Catalog::*)(std::string const &, std::string const &, int) const)
+                        &Catalog::writeFits,
+                    "filename"_a, "mode"_a="w", "flags"_a = 0);
+    cls.def("writeFits",
+                    (void (Catalog::*)(fits::MemFileManager &,
+                                                 std::string const &,
+                                                 int) const) &Catalog::writeFits,
+                    "manager"_a, "mode"_a="w", "flags"_a=0);
+    //cls.def("writeFits", (void (Catalog::*)() const) &Catalog::writeFits);
```
r-owen: I have been intentionally leaving other methods wrapped but commented out so they can trivially be enabled if/when the need for them appears. This had been done in some of the existing pybind11 code and I found it *very* useful -- it made my job much faster. I hope we can continue to do this. I agree we should have a final pass at some point to remove all the remaining lines or add tests for them and uncomment them.

include/lsst/afw/table/pybind11SortedCatalog.h
```diff
@@ -0,0 +1,88 @@
...
+    /* Overridden Methods */
+    cls.def("subset",
+            (Catalog (Catalog::*)(ndarray::Array<bool const,1> const &) const) &Catalog::subset);
+    cls.def("subset",
+            (Catalog (Catalog::*)(std::ptrdiff_t, std::ptrdiff_t, std::ptrdiff_t) const) &Catalog::subset);
+};
+
+}}} // lsst::afw::table
+
+#endif // !AFW_TABLE_PYBIND11_CATALOG_H_INCLUDED
```
r-owen: I'm not very comfortable combining these functions in one .h file, especially since we may find we need additional wrappers, so I'll make a new directory.

python/lsst/afw/table/aliasMap.py
```diff
@@ -2,12 +2,20 @@
...
+def _values(self):
+    """Return an iterator over AliasMap values"""
+    for key, value in self.items():
+        yield value
+
+AliasMap.__iter__ = _keys
+
+AliasMap.keys = _keys
+
+AliasMap.values = _values
```
r-owen: I'm pretty sure something wasn't working, but I don't remember what it was. Unfortunately I don't remember -- if it was in a commit message that info was lost when I squashed everything.

tests/testAmpInfoTable.py
r-owen: Good suggestion.

tests/testMatchFits.py
r-owen: I believe this was ancient code from before we could call `makeRecord` directly on catalogs, and I prefer that unit tests demonstrate the preferred way to do things. However @TallJimbo may know better so I'll ask him to weigh in.

TallJimbo: Omost forwarded methods, I'd agree with @r-owen, but this particular one is a reminder that maybe we shouldn't be forwarding *all* of Table's methods to Catalog (though I'm not suggesting we try to fix that now).  `catalog.makeRecord` is a bit confusing, because it  creates a record without adding it to the catalog, so I think this was probably better as it was.

python/lsst/afw/geom/affineTransform.cc
```diff
@@ -81,8 +82,8 @@ PYBIND11_PLUGIN(_affineTransform) {
     clsAffineTransform.def("setParameterVector", &AffineTransform::setParameterVector);
     clsAffineTransform.def_static("makeScaling", (AffineTransform (*)(double)) &AffineTransform::makeScaling);
     clsAffineTransform.def_static("makeScaling", (AffineTransform (*)(double, double)) &AffineTransform::makeScaling);
-    clsAffineTransform.def_static("makeRotation", (AffineTransform (*)(Angle)) &AffineTransform::makeRotation);
-    clsAffineTransform.def_static("makeRotation", (AffineTransform (*)(Extent2D)) &AffineTransform::makeRotation);
+    clsAffineTransform.def_static("makeRotation", &AffineTransform::makeRotation, "angle"_a);
+    clsAffineTransform.def_static("makeTranslation", &AffineTransform::makeTranslation, "translation"_a);
```
r-owen: I'm dismayed that the original code built. This suggests a bug in pybind11.

tests/test.txt
r-owen: Unfortunately many tests call `checkPickle`, which is what segfaults. I prefer to leave the test commented out for now, but it's easy to tweak it so it runs by adding `return` as the first statement of `checkPickle`. If I do that the test runs, so I'm confident it can be enabled when pickling is fixed.

r-owen:
Thanks to @TallJimbo for feedback about `catalog.makeRecord` vs. `catalog.table.makeRecord`. As per his recommendation I changed `testMatchFits.py` to use the latter (as it originally did).

python/lsst/afw/geom/affineTransform.cc
kfindeisen: Hardly. The expression `(AffineTransform (*)(Extent2D)) &AffineTransform::makeRotation` is not evaluated by Pybind11, but by the compiler. My guess is that the rules for function pointer casts in C/C++ are quite loose.

PR #127, opened 2016-11-30
Tickets/dm 7797

pschella:


python/lsst/afw/geom/ellipses/separable.cc
```diff
@@ -24,22 +24,80 @@
...
+    using Class = Separable<Ellipticity_, Radius_>;
+
+    py::class_<Class, std::shared_ptr<Class>, BaseCore> cls(mod, ("Separable" + suffix).c_str());
 
-    /* Module level */
+//    py::enum_<typename Class::ParameterEnum>(mod, "ParameterEnum")
+//        .value("E0", Class::ParameterEnum::E0)
+//        .value("E1", Class::ParameterEnum::E1)
+//        .value("RADIUS", Class::ParameterEnum::RADIUS)
+//        .export_values();
```
fred3m: Is this not needed?

python/lsst/afw/geom/extent.cc
```diff
@@ -86,8 +86,8 @@ py::class_<Extent<T,N>> declareExtent(py::module &mod, const std::string & suffi
     cls.def("__sub__", [](Extent<T,N> &e, Extent<double,N> &o) { return e - o; });
     cls.def("__eq__", [](Extent<T,N> &e, Extent<T,N> &o) { return e == o; });
     cls.def("__eq__", [](Extent<T,N> &e, py::none) { return false; });
-    cls.def("__neq__", [](Extent<T,N> &e, Extent<T,N> &o) { return e != o; });
-    cls.def("__neq__", [](Extent<T,N> &e, py::none) { return true; });
+    cls.def("__ne__", [](Extent<T,N> &e, Extent<T,N> &o) { return e != o; });
+    cls.def("__ne__", [](Extent<T,N> &e, py::none) { return true; });
```
fred3m: Good catch, ```__neq__``` and ```__ne__``` are not equal.

python/lsst/afw/geom/polygon/polygon.cc
```diff
@@ -22,24 +22,65 @@
...
+//    clsPolygon.def("symDifference_", (std::vector<PTR(Polygon)> (Polygon::*)(Polygon::Box const &) const) &Polygon::symDifference_);
+//    clsPolygon.def("simplify", &Polygon::simplify);
+//    clsPolygon.def("convexHull", &Polygon::convexHull);
+//    clsPolygon.def("transform" (PTR(Polygon) (Polygon::*)(CONST_PTR(XYTransform) const &) const) &Polygon::transform);
+//    clsPolygon.def("transform" (PTR(Polygon) (Polygon::*)(AffineTransform const &) const) &Polygon::transform);
+//    clsPolygon.def("subSample", (PTR(Polygon) (Polygon::*)(size_t) const) &Polygon::subSample);
+//    clsPolygon.def("subSample", (PTR(Polygon) (Polygon::*)(double) const) &Polygon::subSample);
+//    clsPolygon.def("createImage", (PTR::afw::image::image<float>) (Polygon::*)(Polygon::Box2I const &) const) &Polygon::createImage);
+//    clsPolygon.def("createImage", (PTR::afw::image::image<float>) (Polygon::*)(Extent2I const &) const) &Polygon::createImage);
+//    clsPolygon.def("isPersistable", &Polygon::isPersistable);
```
fred3m: Is Polygon initialized somewhere but not used? You commented out everything but the constructors and it doesn't inherit from anything!

python/lsst/afw/image/SConscript
```diff
@@ -1,4 +1,19 @@
...
+				  'apCorrMap',
+				  'calib',
+				  'wcs',
+				  'tanWcs',
+				  'exposure',
+				  'exposureInfo',
+				  'mask',
+				  'maskedImage',
+                  'imageUtils',
+                  'imageSlice',
```
fred3m: Please fix the indentation

python/lsst/afw/image/apCorrMap.cc
```diff
@@ -32,14 +40,62 @@ PYBIND11_PLUGIN(_apCorrMap) {
...
+    clsApCorrMap.def("keys", [](ApCorrMap const & self) {
+        // Can't create a vector of const element type, not sure why
+        using KeyType = std::remove_const<ApCorrMap::Iterator::value_type::first_type>::type;
+        auto r = std::vector<KeyType>();
+        r.reserve(self.size());
+        for (ApCorrMap::Iterator i = self.begin(); i != self.end(); ++i) {
+            r.push_back(i->first);
+        }
+        return r;
+    });
```
fred3m: Since this is an iterator, does it need to have a ```keep_alive``` macro added?

python/lsst/afw/image/apCorrMap.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import absolute_import, division, print_function
+
+from ._apCorrMap import ApCorrMap
+
+ApCorrMap.values = lambda self: [self[name] for name in self.keys()]
+ApCorrMap.items = lambda self: [(name, self[name]) for name in self.keys()]
+ApCorrMap.__iter__ = lambda self: iter(self.keys())
```
fred3m: I agree that this is the best way to do this, but I think the style guide generally forbids using lambdas. Although one could argue that in this case the functions are very clear and it is a much simpler implementation.

python/lsst/afw/image/exposure.cc
```diff
@@ -24,13 +24,83 @@
 //#include <pybind11/operators.h>
 //#include <pybind11/stl.h>
 
+#include "lsst/afw/image/Wcs.h"
+#include "lsst/afw/image/Exposure.h"
+#include "lsst/afw/detection/Psf.h"
+
 namespace py = pybind11;
 
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
```
fred3m: I think this can be removed

python/lsst/afw/image/exposure.py
```diff
@@ -0,0 +1,14 @@
...
+
+def __reduce__(self):
+    from lsst.afw.fits import reduceToFits
+    return reduceToFits(self)
+
+ExposureI.__reduce__ = __reduce__
+ExposureF.__reduce__ = __reduce__
+ExposureD.__reduce__ = __reduce__
+ExposureU.__reduce__ = __reduce__
+#ExposureL.__reduce__ = __reduce__
```
fred3m: Is there a reason that Longs are not implemented?

python/lsst/afw/image/exposure.cc
```diff
@@ -24,13 +24,83 @@
...
+    cls.def_static("readFits", (Exposure<PixelT> (*)(lsst::afw::fits::MemFileManager &)) Exposure<PixelT>::readFits);
+}
+
 PYBIND11_PLUGIN(_exposure) {
     py::module mod("_exposure", "Python wrapper for afw _exposure library");
 
+    declareExposure<int>(mod, "I");
+    declareExposure<float>(mod, "F");
+    declareExposure<double>(mod, "D");
+    declareExposure<std::uint16_t>(mod, "U");
```
fred3m: You never declare ExposureL, but it is defined in SWIG. Make sure that pybind11 isn't casting longs to ints.

python/lsst/afw/image/image.cc
```diff
@@ -21,25 +21,158 @@
...
+#include "lsst/afw/fits.h"
+
 namespace py = pybind11;
+using namespace pybind11::literals;
 
 using namespace lsst::afw::image;
 
-PYBIND11_PLUGIN(_image) {
-    py::module mod("_image", "Python wrapper for afw _image library");
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
```
fred3m: I think you can remove

python/lsst/afw/image/apCorrMap.cc
kfindeisen: `keys` itself is not an iterator, it creates a copy of the map's keyset as a list. This means iterators are unaffected by the deallocation of the parent object, at the cost of silently failing any attempt to modify the keys.

This is by no means a good long-term solution, but it's a straight copy of what the swig interface did, so it works no worse than the old code.

python/lsst/afw/image/apCorrMap.py
kfindeisen: Indeed, I missed that. So what's the best idiom for this kind of code (answer should go in the tutorial)?

python/lsst/afw/image/exposure.py
kfindeisen: Not my code, but could this be related to [RFC-227](https://jira.lsstcorp.org/browse/RFC-227)?

python/lsst/afw/image/SConscript
kfindeisen: Replaced tabs with spaces, per style guidelines.

timj: Ordinarily spaces and tabs in a file will break python 3 but in this case you'd never know because scons always runs with python 2.

python/lsst/afw/image/maskedImage.cc
```diff
@@ -22,24 +22,188 @@
...
+#include "lsst/afw/image/MaskedImage.h"
 
 namespace py = pybind11;
 
 using namespace lsst::afw::image;
+using namespace pybind11::literals;
 
-PYBIND11_PLUGIN(_maskedImage) {
-    py::module mod("_maskedImage", "Python wrapper for afw _maskedImage library");
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
```
fred3m: I think this can be removed

python/lsst/afw/image/utils.cc
```diff
@@ -24,22 +24,21 @@
...
 //#include <pybind11/stl.h>
 
+#include "lsst/afw/image/Utils.h"
+
 namespace py = pybind11;
 
 using namespace lsst::afw::image;
+using namespace pybind11::literals;
+
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
```
fred3m: Please remove

python/lsst/afw/image/wcs.cc
```diff
@@ -21,25 +21,89 @@
...
+
+#include "lsst/afw/table/io/pybind11.h"
+
 namespace py = pybind11;
 
 using namespace lsst::afw::image;
 
+using lsst::afw::table::io::PersistableFacade;
+
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
```
fred3m: Please remove

python/lsst/afw/math/chebyshevBoundedField.cc
```diff
@@ -28,6 +28,10 @@
 #include "ndarray/pybind11.h"
 #include "ndarray/converter.h"
 
+#include "lsst/pex/config/pybind11.h"
```
fred3m: It would be nice to have a comment here to let future developers know this was included for LSST_DECLARE_CONTROL_FIELD, so they know where the macro originates

python/lsst/afw/math/convolveImage.cc
```diff
@@ -24,22 +24,45 @@
...
+    mod.def("scaledPlus", (void (*)(ImageType1 &, double, ImageType2 const &, double, ImageType2 const &)) scaledPlus);
+}
 
-    /* Module level */
+template <typename PixelType1, typename PixelType2>
+void declareConvolve(py::module & mod) {
+    using lsst::afw::image::Image;
+    using lsst::afw::image::MaskedImage;
+    using lsst::afw::image::MaskPixel;
+    using lsst::afw::image::VariancePixel;
```
fred3m: Perhaps just ```using lsst::afw::image;``` can be included, otherwise it will be confusing as to which  header file to find objects in the various namespaces.

tests/testApCorrMap.py
```diff
@@ -70,9 +70,7 @@ def compare(self, a, b):
             value2 = b.get(name)
             self.assertIsNotNone(value2)
             self.assertEqual(value.getBBox(), value2.getBBox())
-            self.assertClose(lsst.afw.math.ChebyshevBoundedField.cast(value).getCoefficients(),
-                             lsst.afw.math.ChebyshevBoundedField.cast(value2).getCoefficients(),
-                             rtol=0.0)
+            self.assertClose(value.getCoefficients(), value2.getCoefficients(), rtol=0.0)
```
fred3m: It's nice to see all of these tests simplified without having to cast!

tests/testRandom1.py
```diff
@@ -91,8 +91,8 @@ def testPolicy(self):
         pol = pexPolicy.Policy()
         seed = getSeed()
         pol.set("rngSeed", str(seed))
-        pol.set("rngAlgorithm", afwMath.Random.getAlgorithmNames()[afwMath.Random.RANLXD2])
-        r1 = afwMath.Random(afwMath.Random.RANLXD2, seed)
+        pol.set("rngAlgorithm", afwMath.Random.getAlgorithmNames()[int(afwMath.Random.Algorithm.RANLXD2)])
+        r1 = afwMath.Random(afwMath.Random.Algorithm.RANLXD2, seed)
```
fred3m: Is there a reason the API changed here?

tests/testRandom1.py
```diff
@@ -139,7 +139,7 @@ def testRandomGaussianImage(self):
     def testRandomChisqImage(self):
         nu = 10
         afwMath.randomChisqImage(self.image, self.rand, nu)
-        stats = afwMath.makeStatistics(self.image, afwMath.MEAN | afwMath.VARIANCE)
+        stats = afwMath.makeStatistics(self.image, int(afwMath.MEAN) | int(afwMath.VARIANCE))
```
fred3m: Is there a reason we have to cast this now?

tests/testRandom1.py
```diff
@@ -149,7 +149,7 @@ def testRandomChisqImage(self):
     def testRandomPoissonImage(self):
         mu = 10
         afwMath.randomPoissonImage(self.image, self.rand, mu)
-        stats = afwMath.makeStatistics(self.image, afwMath.MEAN | afwMath.VARIANCE)
+        stats = afwMath.makeStatistics(self.image, int(afwMath.MEAN) | int(afwMath.VARIANCE))
```
fred3m: Here too.

python/lsst/afw/image/apCorrMap.cc
fred3m: You're right, I should have noticed that.

python/lsst/afw/image/apCorrMap.py
fred3m: Usually we have been using something like 
```
def _values(self):
    return [self[name] for name in self.keys()]

ApCorrMap.values = _values
```
etc.

python/lsst/afw/image/exposure.py
fred3m: I don't think the two are related. RFC-227 is talking about how afw::table allows users to alias the C++ types (I, L, F, etc) using python types (int, long, float, etc). But in python 3 all ints are longs, so the C++ types can no longer be aliased using int and long, but instead np.int32 and np.int64. I don't think that this is related to that.

python/lsst/afw/image/apCorrMap.py
pschella: Don't forget the `del _values` either. Yes it is really ugly syntax...
At some point we may want to investigate writing some magic for it instead. Unfortunately `continue class` was not accepted upstream :( Anyway, I suspect that most cases can be pushed up to the C++ wrapper layer. But lets not do that now.

python/lsst/afw/image/exposure.py
pschella: Don't remember, but there was a reason. I'll try reenabling and see what happens.

python/lsst/afw/image/exposureInfo.cc
```diff
@@ -32,6 +35,7 @@ PYBIND11_PLUGIN(_exposureInfo) {
     py::module mod("_exposureInfo", "Python wrapper for afw _exposureInfo library");
 
     /* Module level */
+    py::class_<ExposureInfo, PTR(ExposureInfo)> clsExposureInfo(mod, "ExposureInfo");
```
pschella: Yes, but copy paste was easier. I intend to run `sed` on the whole codebase at some point to remove `PTR` and `CONST_PTR` as part of DM-6056.

python/lsst/afw/image/image.cc
```diff
@@ -21,25 +21,158 @@
...
+void addImageSliceOperators(py::class_<Image<PixelT>> & cls) {
+    cls.def("__add__", [](Image<PixelT> const & self, ImageSlice<PixelT> const & other) { return self + other; }, py::is_operator());
+    cls.def("__sub__", [](Image<PixelT> const & self, ImageSlice<PixelT> const & other) { return self - other; }, py::is_operator());
+    cls.def("__mul__", [](Image<PixelT> const & self, ImageSlice<PixelT> const & other) { return self * other; }, py::is_operator());
+    cls.def("__truediv__", [](Image<PixelT> const & self, ImageSlice<PixelT> const & other) { return self / other; }, py::is_operator());
+    cls.def("__iadd__", [](Image<PixelT> & self, ImageSlice<PixelT> const & other) { self += other; return self; }, py::is_operator());
+    cls.def("__isub__", [](Image<PixelT> & self, ImageSlice<PixelT> const & other) { self -= other; return self; }, py::is_operator());
+    cls.def("__imul__", [](Image<PixelT> & self, ImageSlice<PixelT> const & other) { self *= other; return self; }, py::is_operator());
+    cls.def("__itruediv__", [](Image<PixelT> & self, ImageSlice<PixelT> const & other) { self /= other; return self; }, py::is_operator());
+}
```
pschella: Hmm, there better be a nice way to do this because otherwise it is going to be a mess.

python/lsst/afw/math/chebyshevBoundedField.cc
pschella: Indeed.

python/lsst/afw/detection/psf.cc
```diff
@@ -24,13 +24,18 @@
 //#include <pybind11/operators.h>
 //#include <pybind11/stl.h>
 
+#include "lsst/afw/detection/Psf.h"
+
 namespace py = pybind11;
 
+PYBIND11_DECLARE_HOLDER_TYPE(MyType, std::shared_ptr<MyType>);
+
```
pschella: done

python/lsst/afw/geom/ellipses/ellipse.cc
```diff
@@ -37,13 +42,26 @@ using namespace ellipses;
...
 
-    /* Member types and enums */
+//    /* Member types and enums */
+//    py::class_<Ellipse::GridTransform> clsEllipseGridTransform(clsEllipse, "GridTransform");
+//
+//    clsEllipseGridTransform.def(py::init<Ellipse const &>());
+//    clsEllipseGridTransform.def("getMatrix", &Ellipse::GridTransform::getMatrix);
+//    clsEllipseGridTransform.def("d", &Ellipse::GridTransform::d);
+//    clsEllipseGridTransform.def("getDeterminant", &Ellipse::GridTransform::getDeterminant);
+//    clsEllipseGridTransform.def("invert", &Ellipse::GridTransform::invert);
```
pschella: done

python/lsst/afw/geom/ellipses/radii.cc
```diff
@@ -24,22 +24,40 @@
...
+    
+    clsLogDeterminantRadius.def(py::init<double>(), "value"_a=0.0);
+    clsLogDeterminantRadius.def("normalize", &LogDeterminantRadius::normalize);
+    clsLogDeterminantRadius.def_static("getName", LogDeterminantRadius::getName);
+
+    py::class_<LogTraceRadius> clsLogTraceRadius(mod, "LogTraceRadius");
+    
+    clsLogTraceRadius.def(py::init<double>(), "value"_a=0.0);
+    clsLogTraceRadius.def("normalize", &LogTraceRadius::normalize);
+    clsLogTraceRadius.def_static("getName", LogTraceRadius::getName);
```
pschella: I don't understand.

python/lsst/afw/geom/extent.cc
pschella: :)

python/lsst/afw/geom/polygon/polygon.cc
pschella: Yes, it is just passed around in the tests that have currently been wrapped. Will uncomment / add later as needed.

python/lsst/afw/geom/ellipses/radii.cc
fred3m: When I was working on afw::table I wrapped objects together like this instead of using the 
```
/* Module level */
 -
 -
 -    /* Constructors */
 -
 -    /* Operators */
 -
 -    /* Members */
```
format, and you asked me to change it since that wasn't the convention we were using at the time.

python/lsst/afw/geom/ellipses/separable.cc
fred3m: Sorry, I meant to remove these comments

python/lsst/afw/geom/polygon/polygon.cc
fred3m: That's what I figured, it was just really surprising.

python/lsst/afw/image/exposure.cc
pschella: done

python/lsst/afw/image/exposure.cc
pschella: Added

python/lsst/afw/image/exposure.py
pschella: Apparently no problem.

python/lsst/afw/image/image.cc
pschella: Done

python/lsst/afw/image/mask.cc
```diff
@@ -22,24 +22,119 @@
...
+#include "lsst/afw/fits.h"
 
 namespace py = pybind11;
 
 using namespace lsst::afw::image;
+using namespace pybind11::literals;
 
-PYBIND11_PLUGIN(_mask) {
-    py::module mod("_mask", "Python wrapper for afw _mask library");
+PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>);
```
pschella: Done

python/lsst/afw/image/maskedImage.cc
pschella: Done

python/lsst/afw/image/utils.cc
pschella: Done

python/lsst/afw/image/wcs.cc
pschella: Done

python/lsst/afw/math/convolveImage.cc
pschella: No, because I don't want to potentially override things from the current math namespace.

tests/testApCorrMap.py
pschella: Yes, casting in Python is ugly and very unpythonic.

tests/testRandom1.py
pschella: Yes to the cast, but no to the scope (which I now changed back). In Swig enums are really just ints whereas in pybind11 they are actual enums which means casts are sometimes needed. But on the plus side, two different enum types can now no longer accidentally be exchanged in function calls. Pybind11 is safer.

fred3m: Ok. We should probably add this to the list of things to make sure downstream users know about the new pipeline.

tests/testRandom1.py
pschella: Yes, and no. Yes it was needed at the time. But now pybind11 supports the `py::is_arithmetic()` tag to `py::enum_` which adds binary operators and removes the need for the cast in these cases. All enums that are to be compared as such should get this tag. I changed it here now.

tests/testRandom1.py
pschella: And here.

tests/testRandom1.py
kfindeisen: That sounds useful. Exactly what operations are added by `py::is_arithmetic()`?

PR #130, opened 2016-12-06
Wrap for of afw table and image

r-owen:
There are two commits: the first one has everything of importance. The second one has some temporary additional code in CoaddInputs that may help diagnose the issue with wrapping that in pybind11.


include/lsst/afw/image/CoaddInputs.h
```diff
@@ -62,6 +62,15 @@ class CoaddInputs : public table::io::PersistableFacade<CoaddInputs>, public tab
...
     CoaddInputs(table::ExposureCatalog const & visits_, table::ExposureCatalog const & ccds_);
 
+    // This code is temporary diagnostic code; an attempt to work around
+    // a pointer error seen in python when trying to read "visits" (or, presumably, "ccds").
+    // These added methods allow one to wrap access using def_property instead of def_readwrite.
+    // Once the problem is resolved please remove this code.
+    table::ExposureCatalog & _getVisits() { return visits; };
+    table::ExposureCatalog & _getCcds() { return ccds; };
+    void _setVisits(table::ExposureCatalog & visits) { this->visits = visits; };
+    void _setCcds(table::ExposureCatalog & ccds) { this->ccds = ccds; };
```
pschella: I assume this is the code that I should diagnose in DM-8597?

include/lsst/afw/table/Exposure.h
```diff
@@ -381,7 +381,7 @@ class ExposureCatalogT : public SortedCatalogT<RecordT> {
      *
      *  The returned array's records are shallow copies, and hence will not in general be contiguous.
      */
-    ExposureCatalogT<RecordT> subset(ndarray::Array<bool const,1> const & mask) const {
+    ExposureCatalogT subset(ndarray::Array<bool const,1> const & mask) const {
```
pschella: Why change this?

python/lsst/afw/coord/observatory.cc
```diff
@@ -21,30 +21,41 @@
...
 PYBIND11_PLUGIN(_observatory) {
     py::module mod("_observatory", "Python wrapper for afw _observatory library");
 
-    py::class_<lsst::afw::coord::Observatory> clsObservatory(mod, "Observatory");
+    py::class_<Observatory> cls(mod, "Observatory");
 
     /* Constructors */
-    clsObservatory.def(py::init<lsst::afw::geom::Angle const, lsst::afw::geom::Angle const,  double const>());
-    clsObservatory.def(py::init<std::string const, std::string const, double const>());
+    cls.def(py::init<lsst::afw::geom::Angle const, lsst::afw::geom::Angle const,  double const>());
```
pschella: double space

python/lsst/afw/coord/observatory.cc
```diff
@@ -21,30 +21,41 @@
...
+    cls.def(py::init<std::string const, std::string const, double const>());
 
     /* Operators */
-    clsObservatory.def("__str__", &lsst::afw::coord::Observatory::toString);
-    clsObservatory.def("__repr__", &lsst::afw::coord::Observatory::toString);
+    cls.def(py::self == py::self);
+    cls.def(py::self != py::self);
```
pschella: Prefer lambdas for operators (with self as first argument).

python/lsst/afw/geom/box.cc
```diff
@@ -49,6 +49,7 @@ PYBIND11_PLUGIN(_box) {
         py::arg("minimum"), py::arg("dimensions"), py::arg("invert")=true);
     clsBox2I.def(py::init<Box2D const &, Box2I::EdgeHandlingEnum>(),
         py::arg("other"), py::arg("edgeHandling")=Box2I::EXPAND);
+    clsBox2I.def(py::init<Box2I const &>(), py::arg("other"));
```
pschella: Use literals for arguments (yes I know it is old code :))

python/lsst/afw/geom/polygon/polygon.cc
```diff
@@ -20,67 +20,96 @@
...
 
-using namespace lsst::afw::geom::polygon;
+namespace lsst { namespace afw { namespace geom { namespace polygon {
 
 PYBIND11_PLUGIN(_polygon) {
     py::module mod("_polygon", "Python wrapper for afw _polygon library");
 
-    py::class_<Polygon, std::shared_ptr<Polygon>> clsPolygon(mod, "Polygon");
+    // TO DO: Commented-out code is waiting until needed and is untested.
+    // Add tests for it and enable it or remove it before the final pybind11 merge.
```
pschella: Rename `TODO:` easier to find. But note that I haven't put in such a nice descriptive note in many other places so We will probably end up grepping for `//` anyway, or look at the final diff before merge to master.

python/lsst/afw/image/coaddInputs.cc
```diff
@@ -19,27 +19,66 @@
...
+    // Also note that using py::return_value_policy::reference (which is not as safe,
+    // but emulates how we wrapped using SWIG shows the same error.
+    bool const useReadWrite = true;
+    if (useReadWrite) {
+        cls.def_readwrite("visits", &CoaddInputs::visits, py::return_value_policy::reference_internal);
+        cls.def_readwrite("ccds", &CoaddInputs::ccds, py::return_value_policy::reference_internal);
+    } else {
+        cls.def_property("visits", &CoaddInputs::_getVisits, &CoaddInputs::_setVisits, py::return_value_policy::reference_internal);
+        cls.def_property("ccds", &CoaddInputs::_getCcds, &CoaddInputs::_setCcds, py::return_value_policy::reference_internal);
+    }
```
pschella: Ok, will remove when fixed.

python/lsst/afw/image/makeVisitInfo.py
```diff
@@ -73,25 +55,12 @@ def makeVisitInfo(
...
-        era,
-        boresightRaDec.toIcrs(),
-        boresightAzAlt,
-        boresightAirmass,
-        boresightRotAngle,
-        rotType,
-        observatory,
-        weather,
-    )
+    warnings.warn("makeVisitInfo is deprecated; use VisitInfo directly, instead", DeprecationWarning)
```
pschella: Nice

pschella: But wait. Are we deprecating all factories? Or just this one?

python/lsst/afw/image/visitInfo.cc
```diff
@@ -0,0 +1,129 @@
+/*
```
pschella: This file didn't exist before?

pschella: Does the corresponding header exist?

python/lsst/afw/image/wcs.cc
```diff
@@ -55,7 +59,15 @@ PYBIND11_PLUGIN(_wcs) {
     mod.def("makeWcs", (PTR(Wcs) (*)(lsst::afw::coord::Coord const &, lsst::afw::geom::Point2D const &,
             double, double, double, double)) makeWcs);
 
-    py::class_<Wcs, std::shared_ptr<Wcs>, lsst::afw::table::io::Persistable, PersistableFacade<Wcs>> cls(mod, "Wcs");
+    py::class_<Wcs,
+               std::shared_ptr<Wcs>,
+               lsst::daf::base::Citizen,
+               lsst::afw::table::io::Persistable,
+               PersistableFacade<Wcs>> cls(mod, "Wcs");
```
pschella: Not the only class in file, so should be `clsWcs`.

python/lsst/afw/image/wcs.cc
```diff
@@ -55,7 +59,15 @@ PYBIND11_PLUGIN(_wcs) {
...
+
+    py::class_<XYTransformFromWcsPair,
+               std::shared_ptr<XYTransformFromWcsPair>,
+               lsst::afw::geom::XYTransform> clsXYTranFromWcs(mod, "XYTransformFromWcsPair");
```
pschella: Don't abbreviate class name. Makes it harder to grep.

python/lsst/afw/table/exposure.cc
```diff
@@ -25,23 +25,152 @@
...
+typedef py::class_<CatalogT<ExposureRecord>,
+                   std::shared_ptr<CatalogT<ExposureRecord>>> PyBaseExposureCatalog;
+
+typedef py::class_<SortedCatalogT<ExposureRecord>,
+                   std::shared_ptr<SortedCatalogT<ExposureRecord>>,
+                   CatalogT<ExposureRecord>> PySortedBaseExposureCatalog;
+
+typedef py::class_<ExposureCatalogT<ExposureRecord>,
+                   std::shared_ptr<ExposureCatalogT<ExposureRecord>>,
+                   SortedCatalogT<ExposureRecord>> PyExposureCatalog;
```
pschella: I typically prefer `using` but it doesn't really matter. I do suspect that at some point we will want to use `using` everywhere, and so far I think I have always used it in pybind11.

python/lsst/afw/table/exposure.cc
```diff
@@ -25,23 +25,152 @@
...
+
+/**
+Declare constructors and member and static functions for a pybind11 ExposureCatalog
+
+@tparam RecordT  Record type, e.g. BaseRecord or SimpleRecord.
+
+@param[in] cls  Catalog pybind11 class.
+*/
+void declareExposureCatalog(PyExposureCatalog & cls) {
```
pschella: Should be in anonymous namespace.

pschella: But really, why does it exist? It is not templated so can go directly into the module section.

pschella: Ok, never mind...

python/lsst/afw/table/slots.cc
```diff
@@ -24,26 +24,49 @@
...
 //#include <pybind11/stl.h>
 
+#include "lsst/afw/table/slots.h"
+
 namespace py = pybind11;
 
-namespace lsst {
-namespace afw {
-namespace table {
+using namespace lsst::afw::table;
```
pschella: Change back :)

python/lsst/afw/table/slots.cc
```diff
@@ -24,26 +24,49 @@
...
+using namespace pybind11::literals;
+
+/**
+Declare standard methods for subclasses of SlotDefinition (but not SlotDefinition itself).
+*/
+template<typename SlotDefSubclass>
+void declareSlotDefinitions(py::class_<SlotDefSubclass, SlotDefinition> & pyClass) {
```
pschella: In anonymous namespace.

pschella: Argument should be named `cls`. Same convention for class naming applies in declare functions.

tests/testExposureTable.py
```diff
@@ -89,23 +89,21 @@ def createVisitInfo():
...
         self.assertIsNotNone(psf1)
         self.assertIsNotNone(psf2)
         self.assertEqual(psf1.getValue(), psf2.getValue())
 
     def setUp(self):
         np.random.seed(1)
         schema = lsst.afw.table.ExposureTable.makeMinimalSchema()
-        self.ka = schema.addField("a", type=float, doc="doc for a")
-        self.kb = schema.addField("b", type=int, doc="doc for b")
+        self.ka = schema.addField("a", type=np.float64, doc="doc for a")
```
pschella: This is different from `float`. Why change it?

tests/testExposureTable.py
```diff
@@ -89,23 +89,21 @@ def createVisitInfo():
...
+        self.kb = schema.addField("b", type=np.int64, doc="doc for b")
```
pschella: Same here, but I assume the reason is the `int` vs `long` discussion?

tests/testTableIO.py
```diff
@@ -52,8 +54,8 @@ def testSchemaReading(self):
         Per DM-8211.
         """
         schema = lsst.afw.table.Schema()
-        aa = schema.addField("a", type=int, doc="a")
-        bb = schema.addField("b", type=float, doc="b")
+        aa = schema.addField("a", type=np.int64, doc="a")
+        bb = schema.addField("b", type=np.float64, doc="b")
```
pschella: ?

tests/testVisitInfo.py
```diff
@@ -33,10 +33,10 @@
...
 
 RotTypeEnumNameDict = {
-    afwImage.RotType_UNKNOWN: "UNKNOWN",
-    afwImage.RotType_SKY: "SKY",
-    afwImage.RotType_HORIZON: "HORIZON",
-    afwImage.RotType_MOUNT: "MOUNT",
+    afwImage.RotType.UNKNOWN: "UNKNOWN",
+    afwImage.RotType.SKY: "SKY",
+    afwImage.RotType.HORIZON: "HORIZON",
+    afwImage.RotType.MOUNT: "MOUNT",
```
pschella: I wonder what we should do with this. One the one hand I prefer using `.` and this should definitely be the future. However, I wonder how much code expects `_` to also exist. Perhaps it is better to enforce this sooner rather than later, but it is an API change (one that we have made in a number of places already for sure though).

include/lsst/afw/table/Exposure.h
r-owen: Good question. I did it to improve uniformity with most of the rest of our code and internal self-consistency. I found it confusing to have the same return type spelled two different ways within the same file. I chose the shorter version because @TallJimbo prefers it and because more of our code uses it.

include/lsst/afw/image/CoaddInputs.h
r-owen: Almost. I figured it might help you diagnose DM-8497, but I want to remove it once DM-8497 is fixed. The current pybind11 wrapper (coaddInputs.cc) has an if/else that will use the usual `def_readwrite` to wrap `exposures` and `ccds` directly (ignoring these new methods), but that can be toggled to use `def_property` and use these new methods. Neither works, so it wasn't much help, but I left it in case you find it of any use.

Note that this extra code was added as a special separate commit. My plan is to merge this ticket without that extra commit, but leave the commit on the ticket branch.

python/lsst/afw/coord/observatory.cc
r-owen: Yes. Good catch.

python/lsst/afw/geom/polygon/polygon.cc
```diff
@@ -20,67 +20,96 @@
...
+
+    py::class_<Polygon,
+               std::shared_ptr<Polygon>,
+               table::io::PersistableFacade<Polygon>,
+               table::io::Persistable> clsPolygon(mod, "Polygon");
+
+    auto clsSinglePolygonException = lsst::pex::exceptions::python::declareException<
+        SinglePolygonException, lsst::pex::exceptions::RuntimeError>(
+            mod, "SinglePolygonException", "RuntimeError");
+    clsSinglePolygonException.def(py::init<std::string const &>());
```
r-owen: Great. We could then almost always ignore the return value.

python/lsst/afw/geom/polygon/polygon.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import absolute_import, division, print_function
+
+from ._polygon import Polygon
+
+__all__ = []
```
r-owen: I don't know about SWIG. It's clearly appropriate for our code though I see I left off my usual comment, which I will add.

python/lsst/afw/image/makeVisitInfo.py
r-owen: This one was always documented as a temporary hack until we switched to pybind11. I can't speak to any others.

python/lsst/afw/image/visitInfo.cc
r-owen: Yes to both questions. What happened is that my metadata ticket was merged after you started the pybind11 work, so some new header files were not picked up.

python/lsst/afw/image/wcs.cc
r-owen: I'm going to push back a bit here. Why can't use the style shown in the pybind11 manual and declare the class and follow it with a series of `.def` instead of repeating the name everywhere. This name is so long that it makes the code hard to read if it is repeated for every def.

python/lsst/afw/table/exposure.cc
r-owen: I'm afraid I don't understand this comment. How would you change this line?

tests/testApCorrMap.py
```diff
@@ -70,7 +70,7 @@ def compare(self, a, b):
             value2 = b.get(name)
             self.assertIsNotNone(value2)
             self.assertEqual(value.getBBox(), value2.getBBox())
-            self.assertClose(value.getCoefficients(), value2.getCoefficients(), rtol=0.0)
+            self.assertFloatsAlmostEqual(value.getCoefficients(), value2.getCoefficients(), rtol=0.0)
```
r-owen: The old code causes a deprecation warning

tests/testExposureTable.py
r-owen: Unfortunately the change is required: `float` breaks with "KeyError: <class 'float'>"

tests/testVisitInfo.py
r-owen: I'm firmly in favor of the change. I don't like having two ways to do something and the SWIG way was ugly. I don't think it will be hard to propagate the change. But I understand it adds work and may need to be discussed more widely.

tests/testVisitInfo.py
```diff
@@ -195,10 +195,10 @@ def testSetVisitInfoMetadata(self):
 
     def testSetVisitInfoMetadataMissingValues(self):
         """If a value is unknown then it should not be written to the metadata"""
-        visitInfo = afwImage.makeVisitInfo()  # only rot type is known
+        visitInfo = afwImage.VisitInfo()  # only rot type is known
```
r-owen: If we don't make this change then the test will give deprecation warnings. Your call but I would like your opinion given the warnings.

python/lsst/afw/image/wcs.cc
r-owen: OK, I admit it's way too late to change to following the `pybind11` manual.

But I'm really not willing to use such a crazy long name, either, so I'll add two functions, one per class, and use `cls` internally in each. It may improve readability a bit in any case.

python/lsst/afw/table/exposure.cc
r-owen: @kfindeisen thank you. I'll change the code.

python/lsst/afw/image/calib.cc
```diff
@@ -19,60 +19,100 @@
...
 #include <pybind11/pybind11.h>
 #include <pybind11/operators.h>
 #include <pybind11/stl.h>
 
+#include "numpy/arrayobject.h"
+#include "ndarray/pybind11.h"
+#include "ndarray/converter.h"
+#include "ndarray_fwd.h"
+
+#include "lsst/base.h"  // for CONST_PTR
```
r-owen: We could decide to not use them in our pybind11 wrappers. I'm happy to write out the std::shared_ptr equivalents if you prefer.

tests/testApCorrMap.py
pschella: Yes, I would change it back for now.

tests/testExposureTable.py
pschella: So why change it?

tests/testVisitInfo.py
pschella: I would keep it too. But it might cause some problems later on. Lets keep it for now. We can always add back the old syntax as aliasses if needed.

tests/testVisitInfo.py
pschella: We get the same warnings with SWIG now. And if someone decides to fix them in master we get merge conflicts. Better alternative would be to fix in master instead of here and then pull it in with the next rebase.

include/lsst/afw/image/CoaddInputs.h
pschella: Sounds like a good plan.

include/lsst/afw/table/Exposure.h
pschella: I'd rather not merge these kinds of changes into the pybind11 branch when not really needed. Perhaps this falls into the category of one-line fixes that don't need ticketing and review? If so, then rather push it to master and we'll pick it up with the next rebase.

python/lsst/afw/image/calib.cc
pschella: Perhaps just do that for this file (so we can remove the `base.h` header include). In the rest of the code use whatever is most convenient for copy and paste reasons and we'll run a regexp search / replace on the whole thing before (or after) merge to master.

python/lsst/afw/image/makeVisitInfo.py
pschella: Ok

python/lsst/afw/image/visitInfo.cc
pschella: Ok

tests/testExposureTable.py
r-owen: Because the unit test fails with a `KeyError` using `float`.

tests/testVisitInfo.py
r-owen: We do not get the same warnings with SWIG now. With SWIG we have to use makeVisitInfo, so it doesn't print a deprecation warning. The change in pybind11 is the replace `makeVisitInfo` with `VisitInfo` (because pybind11 allows a wrapping with keyword arguments that have sensible defaults) and add a deprecation warning to `makeVisitInfo`. That is why I'm so keen to keep this new code. It is directly relevant to the work i did on this ticket.

python/lsst/afw/image/calib.cc
r-owen: I will go through the pybind11 .cc files that I touched on this ticket and eliminate `PTR`, `CONST_PTR` and `#include "lsst/base.h"`.

PR #132, opened 2016-12-09
Tickets/dm 7801

pschella:


include/lsst/afw/cameraGeom/Detector.h
```diff
@@ -150,6 +150,28 @@ class Detector {
     lsst::afw::table::AmpInfoRecord const & operator[](std::string const &name) const;
 
     /**
+     * Get the amplifier specified by index as a shared_ptr
+     *
+     * @warning Intended only for internal use by pybind11
```
r-owen: I added this comment: "This exists because Operator[] returns a data type that is harder for pybind11 to use."

include/lsst/afw/geom/pybind11/transformMap.h
```diff
@@ -0,0 +1,84 @@
...
+    const std::string className = prefix + "TransformMap";
+    py::class_<Map, std::shared_ptr<Map>> cls(mod, className.c_str());
+
+    /* Constructors */
+    cls.def(py::init<>());
+    cls.def(py::init<CoordSysT const &, typename Map::Transforms const &>(),
+            "nativeCoordSys"_a, "transforms"_a);
+
+    /* Operators */
+    cls.def("__getitem__", &Map::operator[], "coordSys"_a);
```
r-owen: In my opinion we concluded that consistency within a file is most important, so I will stubbornly stick to just wrapping the function directly. `py::is_operator()` is important for operators because it handles type errors gracefully by returning `NotImplemented`.

python/lsst/afw/cameraGeom/cameraFactory.py
```diff
@@ -39,7 +39,7 @@ def makeDetector(detectorConfig, ampInfoCatalog, focalPlaneToPupil):
     return Detector(
         detectorConfig.name,
         detectorConfig.id,
-        detectorConfig.detectorType,
+        DetectorType(detectorConfig.detectorType),
```
r-owen: pybind11 is pickier about enums than SWIG. This change is needed because `DetectorConfig.detectorType` is an int, but the value wanted is a `DetectorType` enum.

python/lsst/afw/cameraGeom/detector.py
```diff
@@ -0,0 +1,39 @@
...
+from ._detector import Detector, DetectorType
+
+__all__ = ["Detector", "DetectorType", "SCIENCE", "FOCUS", "GUIDER", "WAVEFRONT"]
+
+# export DetectorType enums as module globals for SWIG compatibility;
+# this should be removed when possible
+SCIENCE = DetectorType.SCIENCE
+FOCUS = DetectorType.FOCUS
+GUIDER = DetectorType.GUIDER
+WAVEFRONT = DetectorType.WAVEFRONT
```
r-owen: For backwards compatibility with SWIG. I'll add a note to that effect.

python/lsst/afw/cameraGeom/fitsUtils.py
```diff
@@ -180,7 +182,7 @@ def _makeDefaultAmpMap(self):
                    ('GAIN', 'setGain', 1.),
                    ('RDNOISE', 'setReadNoise', 0.),
                    ('SATURATE', 'setSaturation', 2<<15),
-                   ('RDCRNR', 'setReadoutCorner', afwTable.LL),
+                   ('RDCRNR', 'setReadoutCorner', int(afwTable.ReadoutCorner.LL), afwTable.ReadoutCorner),
```
r-owen: @pschella did improve pybind11 support for enums but I don't remember the details. This change was definitely required to avoid type errors.

python/lsst/afw/geom/point.cc
```diff
@@ -48,7 +48,6 @@ py::class_<PointBase<T,N>> declarePointBase(py::module &mod, const std::string &
     cls.def("le", [](PointBase<T,N> &p, T value) { return p.le(value); });
     cls.def("gt", [](PointBase<T,N> &p, T value) { return p.gt(value); });
     cls.def("ge", [](PointBase<T,N> &p, T value) { return p.ge(value); });
-    cls.def("__str__", &PointBase<T,N>::toString);
```
r-owen: SWIG had a very different wrapper for `__str__` and `__repr__` and my changes restore the old ones. Not only were the old ones nice, but at least one was required for a unit test to pass.

In my opinion our pybind11 wrappers are somewhat lacking in this area; I have seen several instances where we used to have nice readable representations that are now gone. I'll try to remember to document those and file a ticket. That said, the most worrisome changes are when the default representation is used, which is essentially useless. This one is more subtle since it was simply a change from one reasonable representation to another.

kfindeisen: That might be a good thing to add to our procedure/tutorial -- unless a test case explicitly tests `__str__` and `__repr__`, we're not likely to notice if the `.i` files had code for this...

python/lsst/afw/image/filter.cc
```diff
@@ -20,26 +20,86 @@
...
+namespace image {
 
 PYBIND11_PLUGIN(_filter) {
     py::module mod("_filter", "Python wrapper for afw _filter library");
 
     /* Module level */
+    py::class_<FilterProperty, std::shared_ptr<FilterProperty>> clsFilterProperty(mod, "FilterProperty");
+    py::class_<Filter> clsFilter(mod, "Filter");
+
+    mod.def("stripFilterKeywords", &detail::stripFilterKeywords, "metadata"_a);
```
r-owen: You are right. However, that is how our SWIG did it, so it was required for backwards compatibility.

python/lsst/afw/image/filter.cc
```diff
@@ -20,26 +20,86 @@
...
+    /* FilterProperty Operators */
+    clsFilterProperty.def("__eq__",
+                  [](FilterProperty const & self, FilterProperty const & other) { return self == other; },
+                  py::is_operator());
+    clsFilterProperty.def("__ne__",
+                  [](FilterProperty const & self, FilterProperty const & other) { return self != other; },
+                  py::is_operator());
 
-    /* Operators */
+    // /* FilterProperty Members */
```
r-owen: I can't find this.

python/lsst/afw/image/exposure.cc
```diff
@@ -20,86 +20,138 @@
...
+            "src"_a, "deep"_a);
+}
+
+
+template <typename PixelT>  // only the image type varies; mask and variance are fixed
+py::class_<Exposure<PixelT, MaskPixel, VariancePixel>,
+           std::shared_ptr<Exposure<PixelT, MaskPixel, VariancePixel>>>
+                declareExposure(py::module & mod, const std::string & suffix) {
+    using ExposureT = Exposure<PixelT, MaskPixel, VariancePixel>;
+    using MaskedImageT = typename ExposureT::MaskedImageT;
```
kfindeisen: Actually, on re-reading Effective Modern C++, whether it's a typedef or not is irrelevant. You do need `typename` after all.

python/lsst/afw/image/filter.cc
kfindeisen: https://github.com/lsst/afw/blob/tickets/DM-7801/python/lsst/afw/image/filter.cc#L68

python/lsst/afw/image/exposure.cc
```diff
@@ -20,86 +20,138 @@
...
+*/
+template <typename FromPixelT, typename ToPixelT>
+void declareCastConstructor(py::class_<Exposure<ToPixelT, MaskPixel, VariancePixel>,
+                                       std::shared_ptr<Exposure<ToPixelT, MaskPixel, VariancePixel>>> & cls) {
+    cls.def(py::init<Exposure<FromPixelT, MaskPixel, VariancePixel> const &, bool const>(),
+            "src"_a, "deep"_a);
+}
+
+
+template <typename PixelT>  // only the image type varies; mask and variance are fixed
```
kfindeisen: I guess I don't understand how the template is instantiated. When you call `declareExposure<int>(mod, "I");`, `PixelT` is obviously `int`, but `Exposure<int, MaskPixel, VariancePixel>` seems like it ought to give an undeclared identifier error.

tests/testColor.py
```diff
@@ -148,7 +148,7 @@ def testCalibFromCalibs(self):
         """Test creating a Calib from an array of Calibs"""
         mag0, mag0Sigma = 1.0, 0.01
 
-        calibs = afwImage.vectorCalib()
+        calibs = []
```
r-owen: This is a standard part of SWIG to pybind11 conversion. SWIG usually allowed either use of explicitly named vector templates or standard python lists, but pybind11 only supports the latter.

tests/testDetector.py
```diff
@@ -74,7 +74,7 @@ def testBasics(self):
         self.assertEquals(nativeCoordSys,
                           cameraGeom.CameraSys(cameraGeom.PIXELS.getSysName(), detectorName))
 
-    def testConstructorErrors(self):
+    def xtestConstructorErrors(self):
```
r-owen: It disables the test. I didn't mean to commit that. Thanks for catching it.

include/lsst/afw/cameraGeom/Detector.h
```diff
@@ -150,6 +150,28 @@ class Detector {
     lsst::afw::table::AmpInfoRecord const & operator[](std::string const &name) const;
 
     /**
+     * Get the amplifier specified by index as a shared_ptr
```
r-owen: I'm very glad you pointed this out. That triggered a discussion on Slack dm-docs about enabling autobrief in doxygen, which prompted https://jira.lsstcorp.org/browse/DM-8647. Having merged that, I'm going to omit `@brief` in the new code.

python/lsst/afw/image/exposure.cc
r-owen: `Exposure<int, MaskPixel, VariancePixel>` is a normal integer exposure: in Python this is `ExposureI`. I believe we have 4 types that we make available in Python: `int`, `std::uint16_t`, `float` and `double`. If that doesn't clarify then perhaps we should talk off line, as I probably don't understand your question.

python/lsst/afw/geom/point.cc
r-owen: Agreed. Note that the SWIG wrappers aren't always easy to find, either. In this case the `__str__` and `__repr__`for `PointX` and `ExtentX` were in `coordinateBase.i`.

PR #135, opened 2016-12-12
Tickets/dm 8608

natelust:


include/lsst/afw/geom/SpanSet.h
```diff
@@ -617,6 +620,23 @@ class SpanSet {
...
     std::vector<std::shared_ptr<geom::SpanSet>> split() const;
 
+protected:
+    /* Returns the name used by the persistence layer to identify the SpanSet class
+     */
+    virtual std::string getPersistenceName() const;
+
+    /* Return a string corresponding to the python module that SpanSets lives in
+     */
+    virtual std::string getPythonModule() const;
```
pschella: I assume this is standard. But I find this explicit mixing of C++ and Python API extremely ugly.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -617,6 +620,23 @@ class SpanSet {
      */
     std::vector<std::shared_ptr<geom::SpanSet>> split() const;
 
+protected:
```
pschella: Why protected?

include/lsst/afw/geom/SpanSet.h
```diff
@@ -617,6 +620,23 @@ class SpanSet {
...
+    /* Returns the name used by the persistence layer to identify the SpanSet class
+     */
+    virtual std::string getPersistenceName() const;
```
pschella: Why virtual (applies to all below)?

src/geom/SpanSet.cc
```diff
@@ -731,4 +734,82 @@ std::shared_ptr<geom::SpanSet> geom::SpanSet::transformedBy(geom::XYTransform co
...
+// that does the work is in the base class ctor)
+SpanSetFactory registration(getSpanSetPersistenceName());
+
+} // end anonymous
+
+
+
+std::string geom::SpanSet::getPersistenceName() const { return getSpanSetPersistenceName(); }
+
+std::string geom::SpanSet::getPythonModule() const { return "lsst.afw.geom"; }
```
pschella: Why not inline (e.g. in class)?

src/geom/SpanSet.cc
```diff
@@ -731,4 +734,82 @@ std::shared_ptr<geom::SpanSet> geom::SpanSet::transformedBy(geom::XYTransform co
...
+        // No Moving
+        SpanSetPersistenceHelper (SpanSetPersistenceHelper &&) = delete;
+        SpanSetPersistenceHelper & operator=(SpanSetPersistenceHelper &&) = delete;
+
+    private:
+        SpanSetPersistenceHelper() :
+            spanSetSchema(),
+            spanY(spanSetSchema.addField<int>("y", "row position of span", "pixel")),
+            spanX0(spanSetSchema.addField<int>("x0", "first column of span (inclusive)", "pixel")),
+            spanX1(spanSetSchema.addField<int>("x1", "first column of span (inclusive)", "pixel")) {}
```
pschella: Better hope nobody changes the order of the data members.

jdswinbank: I'm guessing `x0` and `x1` are not both the first column of the span.

include/lsst/afw/geom/SpanSet.h
TallJimbo: There's really no choice.  We have to store a string that tells us the module to `import` in order to unpersist the object, or we wouldn't be able to persist objects defined in plug-in packages.  The only way to avoid that would be to move all persistence down to Python.  That might be a good idea, but I don't think this issue is a big player in that cost/benefit analysis.

include/lsst/afw/geom/SpanSet.h
natelust: Possibly as we don't expect anything to inherit from SpanSets, this is not needed?

natelust: after discussion it has been decided to use the override identifier instead

src/geom/SpanSet.cc
natelust: moved to header as an inline.

src/geom/SpanSet.cc
natelust: Changed wording

include/lsst/afw/geom/SpanSet.h
natelust: As we are not expecting anything to inherit from SpanSet, this functionality was moved to private.

include/lsst/afw/geom/SpanSet.h
TallJimbo: Probably good practice, but we should take this as a sign that we should add `override` to our coding conventions.

PR #137, opened 2016-12-14
Tickets/dm 8619

r-owen:


python/lsst/afw/math/warpExposure.cc
```diff
@@ -20,109 +20,199 @@
...
 PYBIND11_PLUGIN(_warpExposure) {
     py::module mod("_warpExposure", "Python wrapper for afw _warpExposure library");
 
-    py::class_<WarpingControl, std::shared_ptr<WarpingControl>> clsWarpingControl(mod, "WarpingControl");
+    /* Module level */
+    auto clsLanczosWarpingKernel = declareWarpingKernel<LanczosWarpingKernel>(mod, "LanczosWarpingKernel");
+    declareSimpleWarpingKernel<BilinearWarpingKernel>(mod, "BilinearWarpingKernel");
+    declareSimpleWarpingKernel<NearestWarpingKernel>(mod, "NearestWarpingKernel");
+
+    py::class_<WarpingControl> clsWarpingControl(mod, "WarpingControl");
```
fred3m: From the looks of the deleted line 
``py::class_<WarpingControl, std::shared_ptr<WarpingControl>> clsWarpingControl(mod, "WarpingControl");``,
 it looks like this needs at std::shared_ptr

python/lsst/afw/image/pixel/pixel.cc
```diff
@@ -0,0 +1,83 @@
...
+PYBIND11_PLUGIN(_pixel) {
+    py::module mod("_pixel", "Python wrapper for afw _pixel library");
+
+    /* Module level */
+    declareSinglePixel<float>(mod, "SinglePixelF");
+    declareSinglePixel<double>(mod, "SinglePixelD");
+    declareSinglePixel<int>(mod, "SinglePixelI");
+    declareSinglePixel<std::uint16_t>(mod, "SinglePixelU");
+    // TODO: fix or remove; presently this fail with complaint about being ambiguous
+    //declareSinglePixel<std::uint64_t>(mod, "SinglePixelL");
```
r-owen: pybind11 certainly declares `std::uint64_t` versions of `MaskedImage` and `Exposure` and uses the suffix "L" for them.

The problem with `SinglePixel<std::uint64_t>` is ambiguity between the `double` and `float` constructor (where the type refers to the first argument: the image value). I am very puzzled why `SinglePixel` has those two constructors instead of one that takes `ImageT` for the image value. I asked @RobertLuptonTheGood that question on Slack in the dm room, but have not heard back. I am also puzzled why only the `std::uint64_t` version of `SinglePixel` produces that compiler error.

PR #140, opened 2016-12-21
Tickets/dm 8674

pschella:


include/lsst/afw/table/pybind11/catalog.h
```diff
@@ -118,45 +126,29 @@ void declareCatalog(py::class_<CatalogT<RecordT>, std::shared_ptr<CatalogT<Recor
...
-    cls.def("append", [](Catalog & self, PTR(RecordT) const & rec) {
+    cls.def("_append", [](Catalog & self, std::shared_ptr<RecordT> const & rec) {
         self.push_back(rec);
     });
-    cls.def("reserve", &Catalog::reserve);
-    cls.def("subset", (Catalog (Catalog::*)(ndarray::Array<bool const,1> const &) const) &Catalog::subset);
-    cls.def("subset",
-            (Catalog (Catalog::*)(std::ptrdiff_t, std::ptrdiff_t, std::ptrdiff_t) const) &Catalog::subset);
-    cls.def("__delitem__", [](Catalog & self, std::ptrdiff_t i) {
+    cls.def("_delitem_", [](Catalog & self, std::ptrdiff_t i) {
```
pschella: Was this intentional? If so, why the trailing underscore?

include/lsst/afw/table/pybind11/catalog.h
```diff
@@ -168,8 +160,34 @@ void declareCatalog(py::class_<CatalogT<RecordT>, std::shared_ptr<CatalogT<Recor
             throw LSST_EXCEPT(lsst::pex::exceptions::InvalidParameterError, os.str());
         }
         self.erase(self.begin() + i);
-
     });
+ 
+    cls.def("set", &Catalog::set);
+    cls.def("_getitem_", [](Catalog & self, int i) {
```
pschella: Why no double underscores? And if intentional, why a trailing one?

include/lsst/afw/table/pybind11/catalog.h
```diff
@@ -168,8 +160,34 @@ void declareCatalog(py::class_<CatalogT<RecordT>, std::shared_ptr<CatalogT<Recor
...
+        return self.get(i);
+    });
+    cls.def("isContiguous", &Catalog::isContiguous);
+    cls.def("writeFits",
+            (void (Catalog::*)(std::string const &, std::string const &, int) const) &Catalog::writeFits,
+            "filename"_a, "mode"_a="w", "flags"_a = 0);
+    cls.def("writeFits",
+            (void (Catalog::*)(fits::MemFileManager &, std::string const &, int) const) &Catalog::writeFits,
+            "manager"_a, "mode"_a="w", "flags"_a=0);
+    //cls.def("writeFits", (void (Catalog::*)() const) &Catalog::writeFits);
```
pschella: Remove?

python/lsst/afw/detection/peak.cc
```diff
@@ -35,40 +37,75 @@ namespace lsst {
...
+    cls.def("getPeakValue", &PeakRecord::getPeakValue);
+    cls.def("setPeakValue", &PeakRecord::setPeakValue);
+}
 
-    /* Member types and enums */
+/**
+Declare constructors and member and static functions for a pybind11 PeakTable
+*/
+void declarePeakTable(PyPeakTable & cls) {
+    table::pybind11::addCastFrom<table::BaseTable, PeakTable>(cls);
```
pschella: Perhaps the naming is confusing. A plain `addCast` would be more readable. Or you could make it really nice and have the `To` type deduced from the `cls` (i.e. `py::class_<To, ...>`) argument.

python/lsst/afw/detection/peak.cc
```diff
@@ -35,40 +37,75 @@ namespace lsst {
...
+    PyPeakTable clsPeakTable(mod, "PeakTable");
+    PyPeakCatalog clsPeakCatalog(mod, "PeakCatalog", py::dynamic_attr());
 
     /* Members */
+    declarePeakRecord(clsPeakRecord);
+    declarePeakTable(clsPeakTable);
+    table::pybind11::declareCatalog(clsPeakCatalog);
+
+    clsPeakCatalog.attr("Record") = clsPeakRecord;
+    clsPeakCatalog.attr("Table") = clsPeakTable;
```
pschella: This looks good to me.

python/lsst/afw/table/aggregates.cc
```diff
@@ -156,17 +160,24 @@ PYBIND11_PLUGIN(_aggregates) {
     clsEllipseKey.def(py::init<SubSchema const &>());
 
     /* Operators */
+    // Unlike the other keys, CoordKey uses a free function to test equality
```
pschella: Any particular reason for this? Perhaps we could just change it.

python/lsst/afw/table/ampInfo.cc
```diff
@@ -34,16 +36,112 @@ using namespace pybind11::literals;
...
+using PyAmpInfoRecord = py::class_<AmpInfoRecord, std::shared_ptr<AmpInfoRecord>, BaseRecord>;
+using PyAmpInfoTable = py::class_<AmpInfoTable, std::shared_ptr<AmpInfoTable>, BaseTable>;
+using PyAmpInfoCatalog =  py::class_<CatalogT<AmpInfoRecord>, std::shared_ptr<AmpInfoCatalog>>;
+
+void declareAmpInfoRecord(PyAmpInfoRecord & cls) {
+    table::pybind11::addCastFrom<BaseRecord, AmpInfoRecord>(cls);
+
+    cls.def("getName", &AmpInfoRecord::getName);
+    cls.def("setName", &AmpInfoRecord::setName, "name"_a,
+                         "Set name of amplifier location in camera");
```
pschella: Docstrings? I thought we didn't do that yet? I don't mind, but I just want to avoid setting a precedent.

python/lsst/afw/table/ampInfo.cc
```diff
@@ -34,16 +36,112 @@ using namespace pybind11::literals;
...
+    cls.def("getBBox", &AmpInfoRecord::getBBox);
+    cls.def("setBBox", &AmpInfoRecord::setBBox, "bbox"_a);
+    cls.def("getGain", &AmpInfoRecord::getGain);
+    cls.def("setGain", &AmpInfoRecord::setGain, "gain"_a, "Set amplifier gain in e-/ADU");
```
pschella: Not sure this matches astropy unit naming convention (which we have adopted).

python/lsst/afw/table/base.cc
```diff
@@ -0,0 +1,213 @@
...
+namespace table {
+
+namespace {
+
+/**
+Unlike most pybind11 wrapper classes, which have one .cc file per header file,
+this module wraps both BaseRecord.h and BaseTable.h (as well as CatalogT<BaseRecord> from Catalog.h).
+
+This allows us to define BaseCatalog.Table = clsBaseTable, which is need to support `cast` in Python,
+and makes wrapping Base catalogs more similar to all other types of catalog.
```
pschella: Ok

pschella: Thanks for commenting.

tests/testSourceTable.py
```diff
@@ -182,6 +182,7 @@ def testCanonical2(self):
         self.table.defineShape("c")
         self.checkCanonical()
 
+    @unittest.skip("TODO support pickling with pybind11")
```
pschella: Almost done ;)

include/lsst/afw/table/pybind11/catalog.h
pschella: Fine

python/lsst/afw/table/catalog.py
```diff
@@ -88,8 +120,11 @@ def __getattribute__(self, name):
         try:
             return object.__getattribute__(self, name)
         except AttributeError:
+            # self._columns is created the when self.columns is accessed -
+            # looking for it in self.columns below would trigger infinite
+            # recursion.
```
pschella: Ok, so this is a little sad. But I think we can live with it.

tests/testSourceTable.py
```diff
@@ -275,7 +275,6 @@ def testCast(self):
         baseCat = self.catalog.cast(lsst.afw.table.BaseCatalog)
         baseCat.cast(lsst.afw.table.SourceCatalog)
 
-    @unittest.skip("TO DO fix this")
```
pschella: Good catch

python/lsst/afw/table/ampInfo.cc
r-owen: I'm afraid I don't understand what you are suggesting. Are you just commenting on the string used to describe the units? If so, what do you suggest? We can't change the actual units, of course.

python/lsst/afw/table/base.cc
```diff
@@ -0,0 +1,213 @@
...
+        */
+        return self[key];
+    });
+}
+
+template <typename U>
+void declareBaseRecordArrayOverloads(PyBaseRecord clsBaseRecord, std::string const & suffix) {
+    typedef lsst::afw::table::Array<U> T;
+    clsBaseRecord.def(("_get_"+suffix).c_str(),
+                      (typename Field<T>::Value (BaseRecord::*)(Key<T> const &) const) &BaseRecord::get);
```
r-owen: I'll give it a try and see what breaks.

python/lsst/afw/table/aggregates.cc
r-owen: @TallJimbo may know why CoordKey defines `operator==` differently than all other keys in `aggregates.h` (defining it as a free function instead of an operator method). I agree it would be nice to change this if it can be changed. But do we want to do that now, or wait until after the pybind11 merge? If the latter I can open a ticket (though it is likely to get lost and forgotten).

python/lsst/afw/detection/peak.cc
r-owen: Thanks. Personally I hate the fact that it is so easy to forget, and if forgotten, then casting will do the wrong thing. But I could not find a safe way to make declareCatalog handle it: I could pass the python record and table classes as argument, but I could not think of any way of detecting if they were passed in the wrong order!

include/lsst/afw/table/pybind11/catalog.h
r-owen: I'd rather wait until the final merge, to see if we've needed it.

python/lsst/afw/table/ampInfo.cc
```diff
@@ -34,16 +36,112 @@ using namespace pybind11::literals;
...
+    cls.def("getReadNoise", &AmpInfoRecord::getReadNoise);
+    cls.def("setReadNoise", &AmpInfoRecord::setReadNoise, "readNoise"_a, "Set read noise in e-");
```
r-owen: I'll change the unit to `electron`

python/lsst/afw/table/base.py
```diff
@@ -0,0 +1,219 @@
...
+            # Try using a functor
+            return key.get(self)
+
+        except Exception:
+            raise TypeError("Argument to BaseRecord.set must be a string or Key.")
+        if prefix != "Key":
+            raise TypeError("Argument to BaseRecord.set must be a string or Key.")
+        method = getattr(self, '_get_'+suffix)
+        return method(funcKey)
+    cls.get = get
```
TallJimbo: I don't see any of the `type(funcKey)` stuff or the lookup of typed methods from the suffixes in Swig (https://github.com/lsst/afw/blob/master/python/lsst/afw/table/Base.i#L486), so I suspect we don't need any of that.

python/lsst/afw/table/base.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def addBaseRecordMethods(cls):
+    """Add pure Python methods to BaseRecord
+    """
+    addRecordMethods(cls)
+
+    def _set(self, key, value):
+        """
+        Given a string or `lsst.afw.table.Key`, find the `BaseRecord` for the given key and set its value.
```
TallJimbo: Docstring here (and below) is incorrect: we're not "finding" a `BaseRecord`; this is a a method on `BaseRecord`, and we're just setting the value of a field with it.

python/lsst/afw/table/base.py
```diff
@@ -0,0 +1,219 @@
...
+        for name, schemaItem in list(d.items()):  # must use list because we might be adding/deleting elements
+            key = schemaItem.key
+            if split and key.HAS_NAMED_SUBFIELDS:
+                for subname, subkey in zip(key.subfields, key.subkeys):
+                    d["%s.%s" % (name, subname)] = self.get(subkey)
+                del d[name]
+            else:
+                d[name] = self.get(schemaItem.key)
+        return d
+    cls.extract = extract
```
TallJimbo: I don't suppose there's any way we can keep all of these `extract` methods in one file?  I realize it's a little weird to have methods for different classes separate from those classes, but these are all quite tightly coupled and have more in common with each other than they do with the rest of the classes they belong to.

r-owen: They all seem enough different to me that I would not know how to refactor the code. Perhaps you would like to take a pass at it?

tests/testSourceTable.py
```diff
@@ -278,7 +278,7 @@ def testCast(self):
         baseCat = self.catalog.cast(lsst.afw.table.BaseCatalog)
         baseCat.cast(lsst.afw.table.SourceCatalog)
 
-    @unittest.skip("TO DO wrap Footprint")
+    @unittest.skip("TO DO fix this")
```
r-owen: Thanks for catching this.

python/lsst/afw/table/catalog.py
r-owen: I think the best solution is to cache in C++. It would dramatically clean up the wrapper code with little added complexity to the C++.

python/lsst/afw/table/base.py
r-owen: I filed a ticket https://jira.lsstcorp.org/browse/DM-8716 and originally thought @fred3m was going to take it, but he's too busy so it is unassigned.

python/lsst/afw/table/base.py
TallJimbo: Ah, looks like the changes that split up the original happened before this PR, so there's no point in me complaining about it now.  Sorry about the noise.

r-owen: It was an excellent suggestion. But it would help to have somebody more familiar with the code attempt reconciliation. I think the various implementations have grown apart.

PR #141, opened 2016-12-21
Tickets/dm 8415

pschella:


python/lsst/afw/coord/coord.py
```diff
@@ -51,21 +51,23 @@ def __len__(self):
...
-    return (TYPE, (self.getLongitude(), self.getLatitude()))
+def _reduceGalacticCoord(self):
+    return (GalacticCoord, (self.getLongitude(), self.getLatitude()))
+GalacticCoord.__reduce__ = _reduceGalacticCoord
 
-IcrsCoord.__reduce__ = __reduce2__
-GalacticCoord.__reduce__ = __reduce2__
+def _reduceIcrsCoord(self):
+    return (IcrsCoord, (self.getLongitude(), self.getLatitude()))
+IcrsCoord.__reduce__ = _reduceIcrsCoord
```
r-owen: Looks good. I confess I would have been tempted to use a functor, but I doubt it would save any lines of code.

python/lsst/afw/fits.cc
```diff
@@ -1,55 +0,0 @@
...
-    clsMemFileManager.def(py::init<>());
-
-    /* Operators */
-
-    /* Members */
-    clsMemFileManager.def("getData", &MemFileManager::getData);
-    clsMemFileManager.def("getLength", &MemFileManager::getLength);
-
-    return mod.ptr();
-}
```
r-owen: Why did you remove this file? My suspicion is that you moved it down into the fits subdirectory, but if so, I think you forgot to commit it because I am not seeing it there.

python/lsst/afw/geom/ellipses/axes.py
```diff
@@ -0,0 +1,19 @@
...
+__all__ = []
+
+def __repr__(self):
+    return "Axes(a=%r, b=%r, theta=%r)" % (self.getA(), self.getB(), self.getTheta())
+
+def __reduce__(self):
+    return (Axes, (self.getA(), self.getB(), self.getTheta()))
+
+def __str__(self):
+    return "(a=%s, b=%s, theta=%s)" % (self.getA(), self.getB(), self.getTheta())
```
r-owen: picky but...please keep repr and str together (e.g. put reduce before or after both of them). That keeps related code together and makes it easier to see how str and repr differ.

Also, I suggest assigning the function right after declaring it. That keeps related code together.

python/lsst/afw/geom/ellipses/ellipsesLib.py
```diff
@@ -4,10 +4,12 @@
...
 from ._radii import *
 from ._ellipticityBase import *
 from ._distortion import *
 from ._conformalShear import *
 from ._reducedShear import *
 from ._separable import *
-from ._quadrupole import *
\ No newline at end of file
+from ._quadrupole import *
+from .quadrupole import *
```
r-owen: (please add a trailing linefeed)

python/lsst/afw/image/wcs.py
```diff
@@ -15,6 +15,7 @@ def __reduce__(self):
     import pickle
     exposure = ExposureU(1, 1)
     exposure.setWcs(self)
+    print(type(exposure))
```
r-owen: Debugging statement accidentally left in?

python/lsst/afw/math/background.py
```diff
@@ -1,223 +1,12 @@
...
-import lsst.daf.base as dafBase
-from lsst.log import Log
-import lsst.afw.geom as afwGeom
-import lsst.afw.image as afwImage
-from lsst.afw.fits import FitsError, MemFileManager, reduceToFits
-from . import mathLib as afwMath
 
-class BackgroundList(object):
-    """A list-like class to contain a list of (afwMath.Background, interpStyle, undersampleStyle) tuples
+from ._background import *
```
r-owen: Please don't use `from foo import *`. In this case you need only import `Background`.

Also, I hate to say this, but that licensing boilerplate probably ought to be restored. But suit yourself.

tests/testSourceMatch.py
```diff
@@ -302,20 +302,31 @@ def checkPickle(self, matches, checkSlots=True):
...
+#            self.assertEqual(m1.first.getId(), m2.first.getId())
+#            self.assertEqual(m1.first.getRa(), m2.first.getRa())
+#            self.assertEqual(m1.first.getDec(), m2.first.getDec())
+#            self.assertEqual(m1.second.getId(), m2.second.getId())
+#            self.assertEqual(m1.second.getRa(), m2.second.getRa())
+#            self.assertEqual(m1.second.getDec(), m2.second.getDec())
+#            self.assertEqual(m1.distance, m2.distance)
+#            if checkSlots:
+#                self.assertEqualFloat(m1.first.getPsfFlux(), m2.first.getPsfFlux())
+#                self.assertEqualFloat(m1.second.getPsfFlux(), m2.second.getPsfFlux())
```
r-owen: I agree with the idea of not testing pickling, but suggest skipping the test rather than disabling it by commenting it out. The explanation is still helpful.

python/lsst/afw/fits.cc
pschella: There should only be one `fits.cc`.

python/lsst/afw/geom/ellipses/axes.py
pschella: Ok

python/lsst/afw/geom/ellipses/ellipsesLib.py
pschella: done

python/lsst/afw/image/wcs.py
pschella: yes, removed.

python/lsst/afw/math/background.py
pschella: Ok

tests/testSourceMatch.py
pschella: This isn't a test. It is a simple member that is called by multiple tests. So I cannot skip it.

PR #143, opened 2016-12-25
Tickets/dm 8434

r-owen:
Please see the comments on https://jira.lsstcorp.org/browse/DM-8434


python/lsst/afw/detection/peak.cc
```diff
@@ -111,11 +111,9 @@ PYBIND11_PLUGIN(_peak) {
     table::pybind11::declareCatalog(clsPeakCatalog);
 
     clsPeakRecord.attr("Table") = clsPeakTable;
-    clsPeakRecord.attr("ColumnView") = clsPeakColumnView;
```
pschella: How can you be sure it isn't used anywhere? Might be used by user code.

python/lsst/afw/table/schema.cc
```diff
@@ -177,6 +177,11 @@ PYBIND11_PLUGIN(_schema) {
                                            std::string const &,
                                            std::string const &) const) &Schema::join,
                   "a"_a, "b"_a, "c"_a, "d"_a);
+    clsSchema.def("__repr__", [](Schema const & self) {
+        std::ostringstream os;
+        os << self;
+        return os.str();
+    });
```
pschella: We will want to have some policy on `__str__` vs `__repr__` in particular I think the latter should always include the type because pybind11 prints `__repr__` when overload resolution fails to find a candidate.

python/lsst/afw/detection/peak.cc
pschella: I see in the other commit that it does exist for `BaseRecord` why the discrepancy?

python/lsst/afw/cameraGeom/cameraPoint.cc
```diff
@@ -50,8 +50,10 @@ PYBIND11_PLUGIN(_cameraPoint) {
     cls.def("__ne__",
             [](CameraPoint const & self, CameraPoint const & other) { return self != other; },
             py::is_operator());
-    cls.def("__str__", [](CameraPoint &e){std::ostringstream os; os << e; return os.str();});
-    cls.def("__repr__", [](CameraPoint &e){std::ostringstream os; os << e; return os.str();});
+    cls.def("__str__",
+            [](CameraPoint const & self) { std::ostringstream os; os << self; return os.str(); });
+    cls.def("__repr__",
+            [](CameraPoint & self) { std::ostringstream os; os << self; return os.str(); });
```
pschella: See other comment about `__repr__`.

python/lsst/afw/geom/SConscript
```diff
@@ -8,6 +8,7 @@ scripts.BasicSConscript.pybind11(['affineTransform',
 				  'extent',
 				  'functor',
 				  'point',
+                  'spherePoint',
```
pschella: Tabs?

python/lsst/afw/geom/spherePoint.py
```diff
@@ -0,0 +1,64 @@
...
+#  *
+#  * This program is distributed in the hope that it will be useful,
+#  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#  * GNU General Public License for more details.
+#  *
+#  * You should have received a copy of the LSST License Statement and
+#  * the GNU General Public License along with this program. If not,
+#  * see <http://www.lsstcorp.org/LegalNotices/>.
+#  */
```
pschella: Python and C++ comments?

python/lsst/afw/geom/spherePoint.py
```diff
@@ -0,0 +1,64 @@
...
+        if index < -len(self) or index > 1:
+            raise IndexError("Invalid index: %d" % index)
+        try:
+            if index >= 0:
+                return self._getitem(index)
+            else:
+                return self._getitem(len(self) + index)
+        except lsst.pex.exceptions.OutOfRangeError as e:
+            raise_from(IndexError(), e)
+    cls.__getitem__ = __getitem__
```
pschella: Didn't I see a C++ utility function for this pop by somewhere?

tests/testSpherePoint.py
```diff
@@ -123,16 +123,21 @@ def testInit1ArgErrors(self):
...
             SpherePoint()
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(TypeError):
             SpherePoint("Rotund", "Bovine")
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(TypeError):
             SpherePoint(42)
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(TypeError):
             SpherePoint("ICRS", 34.0, -56.0)
```
pschella: Hmm, this brings to mind we should be careful when we use `TypeError` and where a check for `NotImplemented` is more appropriate. Keep this change here for now, but we will need to reconsider.

tests/testSpherePoint.py
```diff
@@ -773,9 +778,9 @@ def testReprValue(self):
             self.assertIn("degrees", pointRepr)
             self.assertEqual(2, len(pointRepr.split(",")))
 
-            copy = eval(pointRepr)
-            self.assertAnglesNearlyEqual(point.getLongitude(), copy.getLongitude())
-            self.assertAnglesNearlyEqual(point.getLatitude(), copy.getLatitude())
+            spcopy = eval(pointRepr)
+            self.assertAnglesNearlyEqual(point.getLongitude(), spcopy.getLongitude())
+            self.assertAnglesNearlyEqual(point.getLatitude(), spcopy.getLatitude())
```
pschella: Name clash?

python/lsst/afw/math/spatialCell.cc
```diff
@@ -89,105 +222,30 @@ void declareTestClasses(py::module &mod){
...
-
-    /* Operators */
-
-    /* Members */
+    wrapSpatialCellCandidate(mod);
+    wrapSpatialCellCandidateIterator(mod);
+    wrapSpatialCell(mod);
+    wrapSpatialCellSet(mod);
+    wrapCandidateVisitor(mod);
+    wrapSpatialCellImageCandidate(mod);
```
pschella: Why change from the normal procedure of wrapping and introduce separate functions?

python/lsst/afw/detection/peak.cc
r-owen: I added it (but not for `Catalog`) in one commit and removed it in the last commit because it's just clutter. I can leave it in if desired. I don't think it is useful in Python and adds clutter.

python/lsst/afw/table/schema.cc
r-owen: The current policy is "replicate SWIG", which this does. For most classes I agree that repr is more useful if it includes the class. For Schema it doesn't matter so much because the printout is very verbose and obviously a schema.

python/lsst/afw/cameraGeom/cameraPoint.cc
r-owen: Again, I just copied SWIG. And some unit tests need that. I thought we were going to postpone changes to the API until later?

tests/testFunction.py
```diff
@@ -628,42 +628,30 @@ def testDFuncDParameters(self):
 
             self.assertAlmostEqual(f(x, y), sum([params[i]*dFdC[i] for i in range(len(params))]))
 
-    @unittest.skip("temporary skip while wrapping")
-    def testCast(self):
+    def testAutoCast(self):
```
pschella: In the past we have simply removed the test and assume that this is covered by pybind11's own unit tests.

python/lsst/afw/cameraGeom/cameraPoint.cc
pschella: Yes, that is fine. Just mentioned this here because I saw @TallJimbo question about introducing a common utility function.

python/lsst/afw/math/spatialCell.cc
```diff
@@ -20,26 +20,129 @@
...
+        cls.def("setStatus", &SpatialCellCandidate::setStatus);
+        cls.def("getId", &SpatialCellCandidate::getId);
+    }
+
+    // Wrap SpatialCellCandidateIterator
+    void wrapSpatialCellCandidateIterator(py::module & mod) {
+        py::class_<SpatialCellCandidateIterator> cls(mod, "SpatialCellCandidateIterator");
+        cls.def("__incr__", &SpatialCellCandidateIterator::operator++, py::is_operator());
+        cls.def("__deref__",
+                [](SpatialCellCandidateIterator &it) -> std::shared_ptr<SpatialCellCandidate> { return *it; },
```
pschella: I usually prefer to have the return type deduced, but either way is fine.

python/lsst/afw/math/spatialCell.cc
```diff
@@ -20,26 +20,129 @@
...
+            .value("UNKNOWN",SpatialCellCandidate::Status::UNKNOWN)
+            .export_values();
+
+        cls.def("getXCenter", &SpatialCellCandidate::getXCenter);
+        cls.def("getYCenter", &SpatialCellCandidate::getYCenter);
+        cls.def("setStatus", &SpatialCellCandidate::setStatus);
+        cls.def("getId", &SpatialCellCandidate::getId);
+    }
+
+    // Wrap SpatialCellCandidateIterator
```
pschella: We can't use pybind11's `make_iterator` here I assume?

python/lsst/afw/table/schema.cc
pschella: Ok

python/lsst/afw/geom/spherePoint.py
r-owen: Yes. I wrote one for STL collections but it takes the collection as an argument instead of the size. I should rewrite it to take the size and use it here.

tests/testSpherePoint.py
r-owen: I'm also not very happy about this, but didn't see an easy way to restore the old behavior. Fortunately production code usually tests for `Exception`.

tests/testSpherePoint.py
r-owen: Yes. `copy` hides a global function name so I picked one of the other names that was already being used for a copy of a point (in a different test in this file).

python/lsst/afw/math/spatialCell.cc
r-owen: Because there were too darn many classes. I felt it made the code very hard to read, partly due to clutter and partly due to the long class names being repeated many times, making for long lines.

python/lsst/afw/math/spatialCell.cc
r-owen: I agree, but this code was moved from elsewhere. I guessed that automatic deduction would give the wrong type in this case, but I did not test it to see.

python/lsst/afw/math/spatialCell.cc
r-owen: No. The class is not STL-like. It doesn't even have `begin` and `end` methods. Thanks to @TallJimbo for pointing this out on Slack.

python/lsst/afw/geom/spherePoint.py
r-owen: I changed `lsst::utils::cppIndex`` to accept size instead of an STL collection as its first parameter, and to raise `lsst::pex::exceptions::OutOfRangeError` instead of `...::InvalidParameterError` because that is more standard for index errors.

Unfortunately this does mean that the Python raises an LSST exception instead of the standard `IndexError`, but I think it is justified by using a utility function.

PR #144, opened 2016-12-27
Tickets/dm 8751

pschella:


python/lsst/afw/table/schema.py
```diff
@@ -27,329 +31,331 @@
...
-    # Check for astropy compatible unit string
-    astropy.units.Unit(units, parse_strict=parse_strict)
-    if type is None:
-        try:
-            prefix, suffix = __builtins__['type'](field).__name__.split("_")
-        except Exception:
-            raise TypeError("First argument to Schema.addField must be a Field if 'type' is not given.")
-        if prefix != "Field":
-            raise TypeError("First argument to Schema.addField must be a Field if 'type' is not given.")
+def addSchemaMethods(cls):
```
pschella: Is this really needed? It is only called once. I'm not opposed to having it, but if we are changing the wrapping like this in many places (which we seem to be doing more and more) I agree with @TallJimbo that we should standardise the approach.

PR #145, opened 2017-01-09
tickets/DM-8693: Add pybind11 wrappers for spansets

fred3m:


python/lsst/afw/geom/geomLib.py
```diff
@@ -15,4 +15,4 @@
 from ._affineTransform import *
 from .affineTransform import *
 from ._linearTransform import *
-from .linearTransform import *
```
fred3m: I don't think that this line should be deleted.

python/lsst/afw/geom/spanSet.cc
```diff
@@ -0,0 +1,225 @@
...
+                    return maskToSpanSet(mask);
+                });
+        mod.def("maskToSpanSet",
+                []
+                (lsst::afw::image::Mask<pixel> mask, pixel const & bitmask)
+                {
+                    auto functor = [&bitmask](pixel const & pixval){ return (pixval & bitmask) == bitmask; };
+                    return maskToSpanSet(mask, functor);
+                });
+    }
```
natelust: Yes, because the second parameter must be a functor, and we have decided that for the time being we are not going to expose functionality that makes use of functors to python. This would involve doing a call back system and is not currently worth the effort.

python/lsst/afw/geom/spanSet.cc
```diff
@@ -0,0 +1,225 @@
...
+    clsSpanSet.def("erode", (std::shared_ptr<SpanSet> (SpanSet::*)(SpanSet const &) const) &SpanSet::erode);
+    clsSpanSet.def("intersect",
+                   (std::shared_ptr<SpanSet> (SpanSet::*)(SpanSet const &) const) &SpanSet::intersect);
+    clsSpanSet.def("intersectNot",
+                    (std::shared_ptr<SpanSet> (SpanSet::*)(SpanSet const &) const) &SpanSet::intersectNot);
+    clsSpanSet.def("union", (std::shared_ptr<SpanSet> (SpanSet::*)(SpanSet const &) const) &SpanSet::union_);
+    clsSpanSet.def_static("spanSetFromShape",
+                          (std::shared_ptr<SpanSet> (*)(int, Stencil)) &SpanSet::spanSetFromShape);
+    clsSpanSet.def("split", &SpanSet::split);
+    //clsSpanSet.def("isPersistable", &SpanSet::isPersistable);
```
natelust: removed

python/lsst/afw/geom/geomLib.py
natelust: You are correct, fixed

python/lsst/afw/geom/spanSet.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    if(_import_array() < 0) {
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return nullptr;
+    }
+
+    py::enum_<Stencil>(mod, "Stencil")
+        .value("CIRCLE", Stencil::CIRCLE)
+        .value("BOX", Stencil::BOX)
+        .value("MANHATTAN", Stencil::MANHATTAN);
```
natelust: This works becuase the enum is supposed to be explicitly accessed as Stencil.(shape), and not be available at a higher namespace scope.

python/lsst/afw/geom/spanSet.cc
RobertLuptonTheGood: Does this remove the current functionality to get a set of Footprints from a mask with at least one of a given set of bits set?  If so, I think you need to wrap this in a C++ function and bind it to python.

tests/testSpanSets.py
```diff
@@ -0,0 +1,298 @@
...
+        spanSetMask.clearMask(mask, 2)
+        mskArray = mask.getArray()
+        for i in range(mskArray.shape[0]):
+            for j in range(mskArray.shape[1]):
+                self.assertEqual(mskArray[i, j], 1)
+
+    def makeOverlapSpanSets(self):
+        firstSpanSet = afwGeom.SpanSet.spanSetFromShape(2, afwGeom.Stencil.BOX).shiftedBy(2, 4)
+        secondSpanSet = afwGeom.SpanSet.spanSetFromShape(2, afwGeom.Stencil.BOX).shiftedBy(2, 2)
+        return firstSpanSet, secondSpanSet
```
fred3m: For this (and the next) methods I would offset the SpanSets in the x direction, just to verify that intersection, union, etc are working properly.

python/lsst/afw/geom/spanSet.cc
natelust: No, it does not, the default functor has the functionality to build a SpanSet based on any bit being set, and there is also an additional overload to the function that allows specifying a specific bit pattern to look for as well. There is a ticket to supply a convince function that will not only turn a mask to a SpanSet, but split the SpanSet up into a list of isolated SpanSets, producing exactly the behavior you spoke of.

natelust:
To make a note for the future, Fred and I spoke directly about how the c++ unit test covers the behavior of SpanSets much more extensively. This python test is a minimal unit test designed to verify the python bindings were working.

PR #146, opened 2017-01-11
DM-6391 afw display needs to be able to display a catalog

parejkoj:


tests/testDisplay.py
```diff
@@ -183,7 +212,7 @@ def setup_module(module):
 
     parser = argparse.ArgumentParser(description="Run the image display test suite")
 
-    parser.add_argument('backend', type=str, nargs="?", default="virtualDevice",
+    parser.add_argument('backend', type=str, nargs="?", default="ds9",
```
gpdf: I seem to remember a past conversation about keeping `virtualDevice` as the default for tests, but I can't find it right now.

RobertLuptonTheGood: I think it has to be the virtual device;  it's the only one that we know is available.

PR #147, opened 2017-01-12
DM-8419: Wrap meas_base with pybind11

kfindeisen:
Wraps afw components that are needed in `meas_base`. Tickets should be opened to add afw unit tests for all these components.


python/lsst/afw/geom/linearTransform.cc
```diff
@@ -57,6 +57,20 @@ PYBIND11_PLUGIN(_linearTransform) {
...
 
     /* Operators */
+    clsLinearTransform.def("__mul__", [](LinearTransform const & self,
+                                         LinearTransform const & other) {
+            return self*other;
+        }, py::is_operator());
+    // __getitem__ implemented in linearTransform.py
+    clsLinearTransform.def("_getitem_nochecking", [](LinearTransform const & self,
+                                                     int i) {
+            return self[i];
```
r-owen: I suggest you implement this using `cppIndex` in `pybind11.h` in the `utils` package. It saves a lot of trouble because it handles negative indices in the usual Python way and does full range checking with a nice verbose error message. That's how we do almost all of our Python indexing. It raises `lsst::pex::exceptions::OutOfRangeError` which I realize is a change (so a minor tweak to a test or two may be needed), but as I say...it's what most of our code does. This is all you need (edited as you like), where `2` is the number of elements:

    { auto cind = utils::cppIndex(2, i); return self[i]; }

And similarly for the next overload of this method but you have to pass the appropriate dimension to each call to `cppIndex`...

python/lsst/afw/math/boundedField.cc
```diff
@@ -69,11 +69,14 @@ PYBIND11_PLUGIN(_boundedField) {
...
-    clsBoundedField.def("evaluate", (ndarray::Array<double,1,1>
-        (BoundedField::*)(ndarray::Array<double const,1> const &,
-                          ndarray::Array<double const,1> const &) const) &BoundedField::evaluate);
+    clsBoundedField.def("evaluate", (double (BoundedField::*)
+        (double, double) const) &BoundedField::evaluate);
+    clsBoundedField.def("evaluate", (ndarray::Array<double,1,1> (BoundedField::*)
+        (ndarray::Array<double const,1> const &,
+         ndarray::Array<double const,1> const &) const) &BoundedField::evaluate);
+    clsBoundedField.def("evaluate", (double (BoundedField::*)
+        (lsst::afw::geom::Point2D const &) const) &BoundedField::evaluate);
```
r-owen: If you are going to reformat existing code please be sure to use `clang-format` (as per a recent RFC that specifies "google" format and a very few config overrides). That's our new standard. In this case I suspect the old code was already formatted that way. Technically you ought to run that on all C++ code you write (I was tempted to mention that earlier but worried it would be too picky. Oh well...).

python/lsst/afw/table/source.cc
```diff
@@ -112,6 +112,8 @@ void declareSourceRecord(PySourceRecord & cls) {
     cls.def("updateCoord",
             (void (SourceRecord::*)(image::Wcs const &, PointKey<double> const &)) &SourceRecord::updateCoord,
             "wcs"_a, "key"_a);
+
+    cls.def_property_readonly("table", &SourceRecord::getTable);
```
r-owen: Please also do this in `peak.cc`

Please consider moving all of these to just under the wrapper for "getTable" so it's clear they are two different ways of getting the same data.

Consider also doing it in `baseColumnView.cc'. Admittedly that is on a column view instead of a record. My argument in favor it makes all objects that have "getTable" the same.

python/lsst/afw/geom/ellipses/quadrupole.cc
```diff
@@ -70,6 +73,11 @@ PYBIND11_PLUGIN(_quadrupole) {
     clsQuadrupole.def("transformInPlace", [](Quadrupole & self, lsst::afw::geom::LinearTransform const & t) {
        self.transform(t).inPlace();
     });
+    // TODO: pybind11 based on swig wrapper for now. Hopefully can be removed once pybind11 gets smarter
+    // handling of implicit conversions
```
pschella: Perhaps file a Jira ticket for implementing this upstream? Precisely describing the current and desired behaviours.

python/lsst/afw/geom/linearTransform.cc
pschella: @r-owen do we have something similar for 2d indexing? As is often used.

python/lsst/afw/geom/linearTransform.py
```diff
@@ -16,6 +16,18 @@ def __reduce__(self):
...
+def __getitem__(self, k):
+    try:
+        i,j = k
+        if i < 0 or i > 1: raise IndexError
+        if j < 0 or j > 1: raise IndexError
+        return self._getitem_nochecking(i, j)
+    except TypeError:
+        if k < 0 or k > 3: raise IndexError
+        return self._getitem_nochecking(k)
+
```
pschella: Like here.

python/lsst/afw/math/boundedField.cc
pschella: It is not our new standard yet I'm afraid. @ktlim hasn't signed off on it yet.

kfindeisen: This was a mistake and shouldn't have been committed. Re: using `clang_format` in general, I agree with @pschella that it's probably better to wait until the new rules are finalized.

python/lsst/afw/geom/linearTransform.py
kfindeisen: Sorry, I don't understand this comment.

r-owen: I'm guessing that @pschella was suggesting this as a use case for a 2-d indexing utility function. But @pschella should weigh in.

python/lsst/afw/geom/linearTransform.cc
r-owen: Not yet. The main advantage over using the 1d function on each axis separately is a better error message. It might be worth adding for that reason. If you and @kfindeisen want it now for use on this ticket I can whip it together (or feel free).

r-owen: Done as DM-9014

python/lsst/afw/table/source.cc
kfindeisen: Should I just add this to `BaseRecord` and all its subclasses?

r-owen: Yes, but I thought you already had.

r-owen: I think `baseColumnView.cc` already has this property. Sorry. But yes, all record types should have `getTable` and `table` as a property, preferably defined adjacent to each other (as in `baseColumnView.cc` so that it's trivial to see that both exist.

PR #148, opened 2017-01-13
Tickets/dm 8453

pschella:


python/lsst/afw/detection/footprintArray.cc
```diff
@@ -21,17 +21,53 @@
...
+    if (_import_array() < 0) {
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return nullptr;
+    }
+
     /* Module level */
+    declareFlattenArray<std::uint16_t>(mod);
+    declareFlattenArray<int>(mod);
+    declareFlattenArray<float>(mod);
+    declareFlattenArray<double>(mod);
```
r-owen: @kfindeisen has already wrapped this on his ticket. As I said in his review: could one of you please also wrap `expandArray`? It is required by at least two packages downstream. I believe it is wrapped similarly (perhaps identically) to `flattenArray` and if so, I suggest renaming this to `declareAll` and declaring both here.

It appears that many of the remaining changes are also duplicated in DM-8419...

PR #150, opened 2017-01-18
Wrap math::stringToInterpStyle

r-owen:
Wrap stringToInterpStyle

Also eliminate use of PTR in the same file and reformat
that bit of code using clang-format.

### DIFF
--- a/python/lsst/afw/math/interpolate.cc
+++ b/python/lsst/afw/math/interpolate.cc
@@ -71,25 +71,16 @@ PYBIND11_PLUGIN(_interpolate) {
                        (ndarray::Array<double, 1> (Interpolate::*) (ndarray::Array<double const, 1> const&)
                            const) &Interpolate::interpolate);
 
-    mod.def("makeInterpolate", 
-                       (PTR(Interpolate) (*)(std::vector<double> const &,
-                                             std::vector<double> const &,
-                                             Interpolate::Style const)) makeInterpolate,
-                       "x"_a, "y"_a, "style"_a=Interpolate::AKIMA_SPLINE);
-    mod.def("makeInterpolate", 
-                       (PTR(Interpolate) (*)(ndarray::Array<double const, 1> const &,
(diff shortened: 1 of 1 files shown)


python/lsst/afw/math/interpolate.cc
```diff
@@ -71,25 +71,16 @@ PYBIND11_PLUGIN(_interpolate) {
...
-                                             Interpolate::Style const)) makeInterpolate,
-                       "x"_a, "y"_a, "style"_a=Interpolate::AKIMA_SPLINE);
-    mod.def("makeInterpolate", 
-                       (PTR(Interpolate) (*)(ndarray::Array<double const, 1> const &,
-                                             ndarray::Array<double const, 1> const &y,
-                                             Interpolate::Style const)) makeInterpolate,
-                       "x"_a, "y"_a, "style"_a=Interpolate::AKIMA_SPLINE);
-    /* Module level */
+    mod.def("makeInterpolate",
+            (std::shared_ptr<Interpolate>(*)(std::vector<double> const &, std::vector<double> const &,
```
pschella: Please add a space (after return value) in cast.

python/lsst/afw/math/interpolate.cc
```diff
@@ -71,25 +71,16 @@ PYBIND11_PLUGIN(_interpolate) {
...
+                                             Interpolate::Style const))makeInterpolate,
+            "x"_a, "y"_a, "style"_a = Interpolate::AKIMA_SPLINE);
+    mod.def("makeInterpolate", (std::shared_ptr<Interpolate>(*)(ndarray::Array<double const, 1> const &,
```
pschella: Same here.

python/lsst/afw/math/interpolate.cc
r-owen: I think you mean adding a space before "makeInterpolate"? I agree it looks funny without, but unfortunately clang-format did it this way, so I'm inclined to live with it. (@pschella agreed in private conversation)

PR #151, opened 2017-01-18
Wrap LinearTransform::operator()

r-owen:
Wrap `LinearTransform::operator()`

Also update the file as follows:
- Remove commented-out `#include <pybind11/operators.h>`
- Use `namespace lsst {...` in the usual way.
- Wrap `LinearTransform::getParameterVector` (which was commented out)

### DIFF
--- a/python/lsst/afw/geom/linearTransform.cc
+++ b/python/lsst/afw/geom/linearTransform.cc
@@ -34,6 +34,7 @@
 #include "lsst/afw/geom/LinearTransform.h"
 
 namespace py = pybind11;
+using namespace pybind11::literals;
 
 namespace lsst {
 namespace afw {
@@ -59,7 +60,7 @@ PYBIND11_PLUGIN(_linearTransform) {
 
     /* Constructors */
     clsLinearTransform.def(py::init<>());
-    clsLinearTransform.def(py::init<typename LinearTransform::Matrix const &>());
+    clsLinearTransform.def(py::init<typename LinearTransform::Matrix const &>(), "matrix"_a);
 
     /* Operators */
     clsLinearTransform.def("__call__",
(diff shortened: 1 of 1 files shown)


python/lsst/afw/geom/linearTransform.cc
```diff
@@ -59,37 +62,50 @@ PYBIND11_PLUGIN(_linearTransform) {
...
+    clsLinearTransform.def(
+        "__call__", (Extent2D(LinearTransform::*)(Extent2D const &) const) & LinearTransform::operator(),
+        py::is_operator());
+    clsLinearTransform.def(
+        "__getitem__", [](LinearTransform const &self, int i) { return self[lsst::utils::cppIndex(4, i)]; },
+        py::is_operator());
+    clsLinearTransform.def("__getitem__",
+                           [](LinearTransform const &self, std::pair<int, int> i) {
+                               auto row = lsst::utils::cppIndex(2, i.first);
+                               auto col = lsst::utils::cppIndex(2, i.second);
```
pschella: I don't really like the magic numbers for the size of the collection. Not sure if it can be avoided either. But perhaps a simple `constexpr std::ptrdiff_t size = 2;` will clarify the intent enough.

pschella: Same goes for other invocations of `cppIndex`.

pschella: Although in the single return value one-liner it is debatable if it would make things clearer. Perhaps just leave it.

pschella: Would be nicest if size is a template parameter that can somehow be deduced from the container type. But that is hard to do in general.

r-owen: Originally I had it templated on an STL container type and used `size()` but then it would not work with `afw:geom::Point`, `ndarray` arrays, `Eigen` matrices... so that was a bust. Given the lack of a standard way to get size or dimensions I think taking the size as an int-like argument is best. I take your point about magic numbers, but I feel they are fairly harmless in this context.

Note that this is pre-existing code. I just ran it through clang-format because some of it was too long.

PR #153, opened 2017-01-19
Tickets/DM-8973

natelust:


python/lsst/afw/detection/SConscript
```diff
@@ -1,6 +1,7 @@
 # -*- python -*-
 from lsst.sconsUtils import scripts
 scripts.BasicSConscript.pybind11(['psf',
+				  'bootprint',
```
pschella: space

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <https://www.lsstcorp.org/LegalNotices/>.
+ */
+
+
+#include <cstdint>
```
pschella: newline

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
```
pschella: newline

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+#include "numpy/arrayobject.h"
+#include "ndarray/pybind11.h"
+#include "ndarray/converter.h"
```
pschella: newline

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+    clsBootprint.def(py::init<std::shared_ptr<geom::SpanSet>,
+                              afw::table::Schema const &,
+                              geom::Box2I const &>(),
+                     "inputSpans"_a, "peakSchema"_a, "region"_a = geom::Box2I());
+
+    /* Bootprint Methods */
+    clsBootprint.def("getSpans", &Bootprint::getSpans);
+    clsBootprint.def("setSpans", &Bootprint::setSpans);
+    clsBootprint.def("getPeaks", (PeakCatalog & (Bootprint::*)()) &Bootprint::getPeaks,
+                     py::return_value_policy::reference);
```
pschella: Are you sure? What happens if the `Bootprint` object is destroyed before the return value is? Perhaps you want `reference_internal`?

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+                                       (Bootprint const & self)->geom::ellipses::Quadrupole {
+                                           return self.getShape();
+                                       });
+    clsBootprint.def_property_readonly("bbox",
+                                       []
+                                       (Bootprint const & self)->geom::Box2I {
+                                           return self.getBBox();
+                                       });
+
+    /* Python magic functions */
```
pschella: These are usually just in the `Operators` section.

src/detection/Bootprint.cc
```diff
@@ -129,6 +129,32 @@ void Bootprint::removeOrphanPeaks() {
     }
 }
 
+bool Bootprint::operator==(Bootprint const & other) const {
+    /* If the peakCatalogs are not the same length the Bootprints can't be equal */
+    if (getPeaks().size() != other.getPeaks().size()) {
```
pschella: I assume `::size` is cheap?

src/detection/Bootprint.cc
```diff
@@ -129,6 +129,32 @@ void Bootprint::removeOrphanPeaks() {
...
+        return false;
+    }
+    /* Check that for every peak in the PeakCatalog there is a corresponding peak
+     * in the other, and if not return false
+     */
+    for (auto const & selfPeak : getPeaks()) {
+        int counter = 0;
```
pschella: `bool match = false`

src/detection/Bootprint.cc
```diff
@@ -129,6 +129,32 @@ void Bootprint::removeOrphanPeaks() {
...
+        for (auto const & otherPeak : other.getPeaks()) {
+            if (selfPeak.getI() == otherPeak.getI() &&
+                selfPeak.getF() == otherPeak.getF() &&
+                selfPeak.getPeakValue() == otherPeak.getPeakValue()) {
```
pschella: `match = true;`
`break;`

src/detection/Bootprint.cc
```diff
@@ -129,6 +129,32 @@ void Bootprint::removeOrphanPeaks() {
...
+                counter += 1;
+            }
+        }
+        if (counter == 0) {
```
pschella: `if (!match) {`

src/detection/Bootprint.cc
pschella: delete counter and break out early, you only care about one match right?

src/detection/Bootprint.cc
```diff
@@ -158,7 +184,7 @@ class LegacyBootprintPersistenceHelper {
         spanY(spanSchema.addField<int>("y", "The row of the span", "pixel")),
         spanX0(spanSchema.addField<int>("x0", "First column of span (inclusive)", "pixel")),
         spanX1(spanSchema.addField<int>("x1", "Second column of span (inclusive)", "pixel"))
-        {}
+        {spanSchema.getCitizen().markPersistent();}
```
pschella: Isn't this making it liver forever?

src/detection/Bootprint.cc
```diff
@@ -168,6 +194,7 @@ std::pair<afw::table::Schema&, table::Key<int>&> spanSetPersistenceHelper() {
     if (initialize) {
         idKey = spanSetIdSchema.addField<int>("id", "id of the SpanSet catalog");
         initialize = false;
+        spanSetIdSchema.getCitizen().markPersistent();
```
pschella: same here

src/geom/SpanSet.cc
```diff
@@ -761,7 +761,9 @@ namespace {
             spanSetSchema(),
             spanY(spanSetSchema.addField<int>("y", "The row of the span", "pixel")),
             spanX0(spanSetSchema.addField<int>("x0", "First column of span (inclusive)", "pixel")),
-            spanX1(spanSetSchema.addField<int>("x1", "Second column of span (inclusive)", "pixel")) {}
+            spanX1(spanSetSchema.addField<int>("x1", "Second column of span (inclusive)", "pixel")) {
+                spanSetSchema.getCitizen().markPersistent();
```
pschella: Again living forever?

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def testIsHeavy(self):
+        self.assertFalse(self.footprint.isHeavy)
+
+    def testGetSetSpans(self):
+        '''
+        Test getting and setting the SpanSet member of the Bootprint with both
+        the getters and setters and the python property accessor
+        '''
+        self.assertTrue(self.footprint.getSpans() == self.spans)
```
pschella: `assertEquals`?

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+        self.footprint.spans = self.spans
+        self.assertTrue(self.footprint.spans == self.spans)
+
+    def testPeakFunctionality(self):
+        newSchema = afwDet.PeakTable.makeMinimalSchema()
+        newField = afwTable.Field_I("extra", "doc", "na")
+        newSchema.addField(newField)
+        self.footprint.setPeakSchema(newSchema)
+        names = self.footprint.getPeaks().getSchema().getNames()
+        self.assertTrue("extra" in names)
```
pschella: `assertIn`?

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+        self.spanRad = 4
+        self.regionRad = 10
+        self.spans = afwGeom.SpanSet.spanSetFromShape(self.spanRad,
+                                                      afwGeom.Stencil.BOX)
+        minPoint = afwGeom.Point2I(-self.regionRad, -self.regionRad)
+        maxPoint = afwGeom.Point2I(self.regionRad, self.regionRad)
+        self.region = afwGeom.Box2I(minPoint, maxPoint)
+        self.schema = afwDet.PeakTable.makeMinimalSchema()
+        # Run the the constructors test to ensure the Bootprints are setUp
+        self.testConstructors()
```
pschella: Isn't this now run twice?

tests/testBootprint.py
pschella: here and everywhere below.

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+        '''
+        Test getting and setting the SpanSet member of the Bootprint with both
+        the getters and setters and the python property accessor
+        '''
+        self.assertTrue(self.footprint.getSpans() == self.spans)
+        self.assertTrue(self.footprint.spans == self.spans)
+        tempSpanSet = afwGeom.SpanSet.spanSetFromShape(2, afwGeom.Stencil.BOX)
+        self.footprint.setSpans(tempSpanSet)
+        self.assertTrue(self.footprint.spans == tempSpanSet)
+        # reset back to origional with property
```
pschella: original :)

pschella: Can tests run in parallel? If so this will be a problem.

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+        sortKey = self.footprint.peaks.getSchema()['peakValue'].asKey()
+        self.footprint.sortPeaks(sortKey)
+        for i, peak in enumerate(self.footprint.peaks):
+            self.assertEqual(peak['i_x'], i)
+
+        # Test that peaks outside the Bootprint are removed
+        self.footprint.removeOrphanPeaks()
+        self.assertEqual(len(self.footprint.peaks), 2)
+        for peak in self.footprint.peaks:
+            self.assertNotEqual(peak['i_x'], 1)
```
pschella: reset the peaks?

pschella: perhaps add a general `self.resetFootprint` function and always call that at the beginning? Or always construct a new one?

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def testGeometry(self):
+        # Move the base footprint by 2 in x and 2 in y
+        offset = 2
+        self.footprint.shift(offset, offset)
+        # verify that this shifts the center from 0,0 as the default
+        # constructed footprint has
+        center = self.footprint.getCentroid()
+        self.assertEqual(center.getX(), offset)
+        self.assertEqual(center.getY(), offset)
```
pschella: Use different offset for `x` and `y`. And perhaps also a negative one?

python/lsst/afw/detection/bootprint.cc
```diff
@@ -0,0 +1,123 @@
...
+                                     (Bootprint const & self,
+                                      geom::Point2I const & point)->bool {
+                                          return self.contains(point);
+                                      });
+    clsBootprint.def("__eq__", []
+                               (Bootprint const & self,
+                                Bootprint const & other)->bool {
+                                    return self == other;
+                                }, py::is_operator());
+}
```
natelust: Note to self, need to return mod here

tests/testBootprint.py
```diff
@@ -0,0 +1,246 @@
...
+        bBox = self.footprint.getBBox()
+        self.assertEqual(bBox.getMinX(), -self.spanRad)
+        self.assertEqual(bBox.getMinY(), -self.spanRad)
+        self.assertEqual(bBox.getMaxX(), self.spanRad)
+        self.assertEqual(bBox.getMaxY(), self.spanRad)
+
+        # Test the point membership in a Bootprint
+        memberPoint = afwGeom.Point2I(1, 1)
+        self.assertTrue(self.footprint.contains(memberPoint))
+        self.assertTrue(memberPoint in self.footprint)
```
pschella: Also test non membership.

tests/testBootprint.py
natelust: It is run twice for the testConstructors test. All tests only run themselves, setup and tearDown. I could separate it out into a not test function, or have the code sucked into the setUp, but then I would not have a test that clearly tests the constructors for the benefit of future readers.

tests/testBootprint.py
natelust: no, Each test is in its own environment.

tests/testBootprint.py
natelust: I talked with Fred about this, and there seemed to be no single place to look at all the assert statements, so in lie of wasting more time, I just coded something that worked.

tests/testBootprint.py
natelust: a new one is constructed at the beginning of each test function by the setUp function

tests/testBootprint.py
natelust: negative is a good idea, but why the need for different numbers?

python/lsst/afw/detection/SConscript
natelust: fixed, was actually a mixed tab issue vim inserted :(

python/lsst/afw/detection/bootprint.cc
natelust: added

python/lsst/afw/detection/bootprint.cc
natelust: added

python/lsst/afw/detection/bootprint.cc
natelust: added

python/lsst/afw/detection/bootprint.cc
natelust: changed

python/lsst/afw/detection/bootprint.cc
natelust: I didn't have any other operators so I renamed this section to operators

python/lsst/afw/detection/bootprint.cc
natelust: I returned mod.ptr()

tests/testBootprint.py
natelust: I changed it anyway as there was no reason not to, and it introduced more variation in behavior to test

tests/testBootprint.py
natelust: added, and changed to assetIn / assertNotIn

tests/testBootprint.py
pschella: ok

tests/testBootprint.py
pschella: ok

src/detection/Bootprint.cc
natelust: Yes, it is just calling size on a vector ultimately

src/detection/Bootprint.cc
natelust: yes this is a much smarter way to do this

src/detection/Bootprint.cc
natelust: This was discussed in person, but leaving a not here for future readers. This is a standard part of the persistence layer. This basically just keeps Citizen from mistakenly thinking memory is leaking as a variable is static and does not get cleared at the time the memory examiner expects it to.

PR #161, opened 2017-01-27
DM-8461: Wrap pipe_tasks with pybind11

kfindeisen:
Various issues that occur in packages upstream from `afw`:
- The changes to polygonLib.py are to let `Polygon` be imported without needing other imports first.
- Making `Exposure` a `Persistable` is needed to allow calls to methods that expect a `Persistable`.
- The changes to `kernel.cc` take advantage of the fact that a method with a default argument is equivalent to two methods with no defaults. Using `py::init` is a bit risky, since strictly speaking it's only guaranteed to work if a constructor with a matching signature exists. Using a lambda to make a [custom constructor](http://pybind11.readthedocs.io/en/master/advanced/classes.html#custom-constructors) may be more forward-compatible.

### DIFF
--- a/python/lsst/afw/geom/polygon/polygonLib.py
+++ b/python/lsst/afw/geom/polygon/polygonLib.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
 
+from lsst.afw.table.io import Persistable
 from ._polygon import *
 from .polygon import *
--- a/python/lsst/afw/math/kernel.cc
+++ b/python/lsst/afw/math/kernel.cc
@@ -91,10 +91,13 @@ PYBIND11_PLUGIN(_kernel) {
 
     py::class_<AnalyticKernel, std::shared_ptr<AnalyticKernel>, Kernel> clsAnalyticKernel(mod, "AnalyticKernel");
     clsAnalyticKernel.def(py::init<>());
+    // Workaround for NullSpatialFunction and py::arg not playing well with Citizen
+    clsAnalyticKernel.def(py::init<int, int, AnalyticKernel::KernelFunction const &>(),
+                          "width"_a, "height"_a, "kernelFunction"_a);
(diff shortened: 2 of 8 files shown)


python/lsst/afw/math/kernel.cc
```diff
@@ -91,10 +91,13 @@ PYBIND11_PLUGIN(_kernel) {
...
     clsAnalyticKernel.def(py::init<>());
+    // Workaround for NullSpatialFunction and py::arg not playing well with Citizen
+    clsAnalyticKernel.def(py::init<int, int, AnalyticKernel::KernelFunction const &>(),
+                          "width"_a, "height"_a, "kernelFunction"_a);
     clsAnalyticKernel.def(py::init<int, int, AnalyticKernel::KernelFunction const &, Kernel::SpatialFunction const &>(),
-                          "widht"_a, "height"_a, "kernelFunction"_a, "specialFuction"_a=Kernel::NullSpatialFunction());
+                          "width"_a, "height"_a, "kernelFunction"_a, "spatialFunction"_a);
     clsAnalyticKernel.def(py::init<int, int, AnalyticKernel::KernelFunction const &, std::vector<Kernel::SpatialFunctionPtr> const &>(),
-                          "widht"_a, "height"_a, "kernelFunction"_a, "specialFuctionList"_a);
+                          "width"_a, "height"_a, "kernelFunction"_a, "spatialFunctionList"_a);
```
r-owen: Clever workaround!

python/lsst/afw/geom/polygon/polygonLib.py
```diff
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
 
+from lsst.afw.table.io import Persistable
+from ._polygon import PersistableFacadePolygon
```
r-owen: Why add this line? The next line pulls it in along with everything else in `_polygon`. I suggest deleting this (but keeping the line above, of course).

PR #163, opened 2017-01-30
tickets/DM-9105

natelust:


include/lsst/afw/geom/SpanSet.h
```diff
@@ -344,99 +344,123 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
...
                      starting from the end
      */
-    template <typename Pixel, int outC, int inC>
+    template <typename PixelIn, typename PixelOut, int inA, int outC, int inC>
     void flatten(
-            ndarray::Array<Pixel, 1, outC> const & output,
-            ndarray::Array<Pixel, 2, inC> const & input,
+            ndarray::Array<PixelOut, inA-1, outC> const & output,
+            ndarray::Array<PixelIn, inA, inC> const & input,
             Point2I const & xy0 =  Point2I()) const {
```
TallJimbo: Spurious extra space after `=`.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -344,99 +344,123 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
...
-                           typename details::ImageNdGetter<Pixel, 2, inC>::Reference in)
+                           typename details::FlatNdGetter<PixelOut, inA-1, outC>::Reference out,
+                           typename details::ImageNdGetter<PixelIn, inA, inC>::Reference in)
                           {out = in;};
         // Populate array output with values from input at positions given by SpanSet
         applyFunctor(ndAssigner, ndarray::ndFlat(output), ndarray::ndImage(input, xy0));
     }
 
-    /** @brief Create 2d array at pionts given by SpanSet
+    /** @brief Expand an array by One spatial dimension at pionts given by SpanSet
```
TallJimbo: Unnecessary capitalization of `One`.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -344,99 +344,123 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
...
+     * where the x,y coordinates of the SpanSet have been flattened to one dimension. First two
+     * dimensions of the input array must be the h,w which correspond to the SpanSet coordinates. Any
+     * number of remaining dimensions is permissable.
      *
      * @param input - The ndarray from which the values will be taken
      * @param xy0 - A point object with is used as the origin point for the SpanSet coordinate system
      *
      * @tparam Pixel - The datatype for the ndarray
+     * @tparam inN - The number of dimensions in the array object
      * @tparam inC - Number of guaranteed row-major contiguous dimensions, starting from the end
```
TallJimbo: I'm a little bothered by the fact that we're documenting these function template parameters, as they should never need to be specified directly by the caller, and hence they feel like an implementation detail (especially because you could imagine defining `inN - 1` as the parameter rather than `inN` in this case).  I wonder if @pschella or @kfindeisen has an opinion on whether we should document these.

include/lsst/afw/geom/SpanSet.h
```diff
@@ -344,99 +344,123 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
...
     }
 
-    /** @brief Populate 2d array at points given by SpanSet
+     /** @brief Expand an array by One spatial dimension at pionts given by SpanSet
      *
-     * Take values from a 1d array and insert them in a new 2D array at points defined by the SpanSet,
-     * offset by the xy0 parameter
+     * Take values from a lower dimensional array and insert them in an output array with one additional
+     * dimension at points defined by the SpanSet, offset by the lower left hand corner of the
+     * bounding box of the SpanSet. The first two dimensions of the output array will correspond with the
```
TallJimbo: Should be "correspond to", not "correspond with".

include/lsst/afw/geom/SpanSet.h
```diff
@@ -344,99 +344,123 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
...
+    /** @brief Reduce the pixel dimensionality from 2 to 1 of an array at points given by SpanSet
      *
-     * Take values from an array at points defined by SpanSet and use them to populate a new 1D ndarray
+     * Take values from an array at points defined by SpanSet and use them to populate a new array
+     * where the x,y coordinates of the SpanSet have been flattened to one dimension. First two
+     * dimensions of the input array must be the h,w which correspond to the SpanSet coordinates. Any
+     * number of remaining dimensions is permissable.
      *
      * @param input - The ndarray from which the values will be taken
      * @param xy0 - A point object with is used as the origin point for the SpanSet coordinate system
```
TallJimbo: I don't believe we should be using dashes to separate terms in Doxygen `@param` calls - I think that they'll actually be rendered as dashes in the HTML, which looks weird. If you can confirm that's not the case, it's okay to leave them.

include/lsst/afw/geom/SpanSet.h
TallJimbo: Also repeated below.

PR #164, opened 2017-02-01
tickets/DM-9233

natelust:


include/lsst/afw/detection/Bootprint.h
```diff
@@ -74,6 +74,10 @@ class Bootprint : public lsst::daf::base::Citizen,
     explicit Bootprint(std::shared_ptr<geom::SpanSet> inputSpans,
                        afw::table::Schema const & peakSchema,
                        geom::Box2I const & region=geom::Box2I());
+    explicit Bootprint(): lsst::daf::base::Citizen(typeid(this)),
+                          _spans(std::make_shared<geom::SpanSet>()),
+                          _peaks(PeakTable::makeMinimalSchema()),
+                          _region(geom::Box2I()) {}
```
TallJimbo: Could the other constructor that just uses `PeakTable::makeMinimalSchema()` initialize it the same way, or does it have to use `setPeakSchema`?  Probably best to make them consistent, and if this approach works, it's the better one.

natelust: I remember doing in the body when I came across some error, but since this works I can not figure out what could have stopped the other from working. I will make them consistent.

PR #169, opened 2017-02-11
Tickets/dm 9156

pschella:


python/lsst/afw/detection/footprint.cc
```diff
@@ -86,8 +86,13 @@ PYBIND11_PLUGIN(_footprint) {
 
     clsFootprint.def("assign", &Footprint::operator=);
 
-    // No good way to wrap the bitmask, not sure if it is needed
-    clsFootprint.def("intersectMask", [](Footprint & self, image::Mask<image::MaskPixel> const & mask) { return self.intersectMask(mask); });
+    // Default argument for bitmask doesn't work here.
```
r-owen: A bit more information as to why the default argument does not work would be appreciated.

python/lsst/afw/detection/footprint.cc
```diff
@@ -86,8 +86,13 @@ PYBIND11_PLUGIN(_footprint) {
...
+    clsFootprint.def("intersectMask", [](Footprint &self, image::Mask<image::MaskPixel> const &mask) {
+        return self.intersectMask(mask);
+    });
+    clsFootprint.def("intersectMask",
+                     [](Footprint &self, image::Mask<image::MaskPixel> const &mask,
+                        image::MaskPixel bitmask) { return self.intersectMask(mask, bitmask); });
```
r-owen: Does this finish wrapping all of Footprint? I hope so.

python/lsst/afw/image/mask.cc
```diff
@@ -78,6 +78,7 @@ void declareMask(py::module & mod, std::string const & suffix) {
     cls.def("__ixor__", [](Mask<MaskPixelT> & self, int other) { return self ^= other; }, py::is_operator());
 
     /* Members */
+    cls.def("swap", (void (Mask<MaskPixelT>::*)(Mask<MaskPixelT>&)) &Mask<MaskPixelT>::swap);
```
r-owen: I have been omitting `swap` because I thought it was a C++ detail. Do we typically want it in Python, or is this a special case? (Such information might be useful in our wrapping guide or at least as a posting to the pybind11 Slack room).

python/lsst/afw/math/statistics.cc
```diff
@@ -86,7 +86,16 @@ PYBIND11_PLUGIN(_statistics) {
...
+    py::enum_<StatisticsControl::WeightsBoolean>(clsStatisticsControl, "WeightsBoolean")
+        .value("WEIGHTS_FALSE", StatisticsControl::WeightsBoolean::WEIGHTS_FALSE)
+        .value("WEIGHTS_TRUE", StatisticsControl::WeightsBoolean::WEIGHTS_TRUE)
+        .value("WEIGHTS_NONE", StatisticsControl::WeightsBoolean::WEIGHTS_NONE)
+        .export_values();
+
+    clsStatisticsControl.def(py::init<double, int, lsst::afw::image::MaskPixel, bool,
+                                      typename StatisticsControl::WeightsBoolean>(),
+                             "numSigmaClip"_a = 3.0, "numIter"_a = 3, "andMask"_a = 0x0, "isNanSafe"_a = true,
+                             "useWeights"_a = StatisticsControl::WEIGHTS_NONE);
```
r-owen: As you have seen I already added this code. I also modernized the file, which means you should probably just undo your changes. However, I did not wrap the `useWeights` argument because the WeightsBoolean enum was private. Since you have made it public I encourage you to retain this feature (e.g. by adding it to my code).

include/lsst/afw/math/Statistics.h
```diff
@@ -90,8 +90,8 @@ Property stringToStatisticsProperty(std::string const property);
  *
  */
 class StatisticsControl {
-    typedef enum { WEIGHTS_FALSE=0, WEIGHTS_TRUE=1, WEIGHTS_NONE } WeightsBoolean; // initial state is NONE
 public:
+    enum WeightsBoolean { WEIGHTS_FALSE=0, WEIGHTS_TRUE=1, WEIGHTS_NONE }; // initial state is NONE
```
r-owen: You were bolder than I, by making this public and thus expanding the API slightly. I approve. You might also have a look at the code that implements the StatisticsControl constructor; I believe you can replace some ugly trinary logic with a simple `_useWeights(useWeights)`.

python/lsst/afw/detection/footprint.cc
pschella: Changed to:

    // Default argument for bitmask doesn't work here, presumably because the Python `int`
    // generated doesn't cast to the image::MaskPixel type. But I am not 100% sure.
    // Therefore, this workaround until it can be investigated in more detail (DM-9401).

python/lsst/afw/detection/footprint.cc
pschella: I don't think so. There are some more functions commented out in the file.

python/lsst/afw/image/mask.cc
pschella: In this case it was needed. I don't know if this always applies.

python/lsst/afw/math/statistics.cc
pschella: Will do.

python/lsst/afw/detection/footprint.cc
r-owen: Please consider enabling the rest (or at least doing a global search through our Python to see if any are needed). It is very inefficient to have to keep going back to afw to wrap additional functions, as it takes so long to rebuild everything and risks duplicate work if two of us are both adding the wrappers.

r-owen: Please consider wrapping the remaining methods (and reporting the lack of unit tests)

python/lsst/afw/math/statistics.cc
r-owen: I'm puzzled what happened here. Did you adopt my modernization of the file (the diff suggests not) or just use your version of the file.

PR #170, opened 2017-02-14
Fix obsolete uses of cast and vector

r-owen:
These changes fall into 3 categories:
- Fix examples (or in the case of one test, fix what I can and file a ticket to fix the rest)
- Remove obsolete code that was commented out
- Remove a skipped test that will pybind11 compatible.

### DIFF
--- a/examples/estimateBackground.py
+++ b/examples/estimateBackground.py
@@ -38,7 +38,7 @@ except NameError:
 
 def getImage():
     imagePath = os.path.join(lsst.utils.getPackageDir("afwdata"),
-                             "DC3a-Sim", "sci", "v5-e0", "v5-e0-c011-a00.sci_img.fits")
+                             "DC3a-Sim", "sci", "v5-e0", "v5-e0-c011-a00.sci.fits")
     return afwImage.MaskedImageF(imagePath)
 
 def simpleBackground(image):
--- a/examples/estimateBackground.py
+++ b/examples/estimateBackground.py
@@ -49,7 +49,7 @@ def simpleBackground(image):
 
     bkgd = afwMath.makeBackground(image, bctrl)
 
-    statsImage = afwMath.cast_BackgroundMI(bkgd).getStatsImage()
+    statsImage = bkgd.getStatsImage()
 
     image  -= bkgd.getImageF(afwMath.Interpolate.NATURAL_SPLINE)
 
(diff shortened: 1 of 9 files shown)


examples/estimateBackground.py
```diff
@@ -38,7 +38,7 @@
 
 def getImage():
     imagePath = os.path.join(lsst.utils.getPackageDir("afwdata"),
-                             "DC3a-Sim", "sci", "v5-e0", "v5-e0-c011-a00.sci_img.fits")
+                             "DC3a-Sim", "sci", "v5-e0", "v5-e0-c011-a00.sci.fits")
```
kfindeisen: Why is this change needed? It doesn't seem to be related either to casting or to Swig container wrappers...

examples/spatialCellExample.py
```diff
@@ -62,13 +62,13 @@ def readImage(filename=None):
         except Exception:
             raise RuntimeError("You must provide a filename or setup afwdata to run these examples")
 
-        filename = os.path.join(afwDataDir, "CFHT", "D4", "cal-53535-i-797722_1")
+        filename = os.path.join(afwDataDir, "CFHT", "D4", "cal-53535-i-797722_1.fits")
```
kfindeisen: Again, how is this pybind11-related?

examples/spatialCellExample.py
```diff
@@ -62,13 +62,13 @@ def readImage(filename=None):
...
 
         bbox = afwGeom.Box2I(afwGeom.Point2I(270, 2530), afwGeom.Extent2I(512, 512))
     else:
         bbox = None
 
-    mi = afwImage.MaskedImageF(filename, 0, None, bbox, afwImage.LOCAL)
+    mi = afwImage.MaskedImageF(filename, bbox=bbox, origin=afwImage.LOCAL)
```
kfindeisen: Nice use of keywords. :+1:

examples/estimateBackground.py
r-owen: This change is needed because the file that was being loaded disappeared years ago (we used to persist MaskedImage and Exposure as three separate FITS files with suffixes `_img.fits, `_var.fits` and `_mask.fits`.

I will make this change (and the other one like it) as a separate commit. That should clear things up.

PR #171, opened 2017-02-16
DM-8715: Fix or remove pickling support for Match lists with pybind11

pschella:


PaulPrice:
We need to be able to pickle a matchlist. How is that functionality being tested?

TallJimbo:
@PaulPrice Can you point me at where we need to do that?  I didn't think we needed to send them over the wire, and I'm a fair bit concerned about the *conceptual* difficulty of pickling a normalized match table as a denormalized match table, which (IIRC) was what was happening before.

PaulPrice:
On the HSC side, we've been gathering match lists for each CCD to do a global astrometric fit (solvetansip).  However, that's changed on the LSST side, and I'm not sure we actually did it by pickling a `SourceMatchVector`.  So maybe it doesn't matter.  But *please* won't you let me know in the future before dumping my code?  Especially when it's a test for functionality that we may be using.

TallJimbo:
Yeah, sorry about the poor communication on this.  I'd intended to check with you before we actually removed it, but I'd also told @pschella that we would probably be able to remove it, and I think I failed to tell him that you were the person who could best verify that.

jdswinbank:
Note that it was actually me who told Pim to get rid of it, so you can send all your complaints in my direction. Sorry!

That said, I'm actually not sure the process here is a terribly bad one. I encourage Pim and others to move fast and make (or be complicit in) the occasional mistake: git makes it easy to recover when necessary.

PR #172, opened 2017-02-16
DM-9392: Wrap meas_extensions_convolved with pybind11 instead of swig

pschella:


include/lsst/afw/table/pybind11/functorKey.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "lsst/afw/table/FunctorKey.h"
+
+namespace py = pybind11;
+using namespace py::literals;
+
+namespace lsst {
+namespace afw {
+namespace table {
+namespace pybind11 {
```
r-owen: Our coding standard calls for using a subdirectory named `python` not `pybind11`, and using the same rule for the sub-namespace. I believe this is to avoid conflict with the `pybind11` namespace.

PR #173, opened 2017-02-16
DM-9433 ds9.py error code not working as intended

r-owen:
When display_ds9 is not set up the code tries to print the exception. However, in Python 3 the exception no longer exists by the time it is needed, so explicitly bind the exception.

The code is a bit clumsy due to limitations of Python 2, but I include instructions for eventual simplification.

I tried using functools.partial but felt the results were less clear (thought slightly shorter) than the code I adopted.

### DIFF
--- a/python/lsst/afw/display/ds9.py
+++ b/python/lsst/afw/display/ds9.py
@@ -42,12 +42,20 @@ except NameError:
         # Let's define a version of getDisplay() which will throw an exception.
         e.args = ["%s (is display_ds9 setup?)" % e]
 
-        def getDisplay(*args, **kwargs):
-            raise e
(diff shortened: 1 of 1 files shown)


timj:
I'm surprised that such a big change was needed to fix this. I thought we'd be able to get away with
```python
    except Exception as exc:
        # Copy the exception to retain it in Python 3
        e = exc
        # No usable version of display_ds9.
        # Let's define a version of getDisplay() which will throw an exception.
        e.args = ["%s (is display_ds9 setup?)" % e]
```
i.e. copy the exception to a new local variable so that it won't disappear as soon as the block exits (granted the diff would be bigger if you retained `e` and wrote `exc = e` instead).

PR #175, opened 2017-02-18
Use extracted name, not field name

brianv0:


TallJimbo:
Already fixed on master in a separate commit.

PR #177, opened 2017-02-20
DM-9504: lsst_py3 CI failure due to meas_extensions_ngmix

jdswinbank:
In DM-9504 we saw that storing the original exception thrown when failing to
import ds9 in a closure for the duration of execution causes test failures and
wastes memory. Instead, we generate a new exception which contains only a
minimal error message.

### DIFF
--- a/python/lsst/afw/display/ds9.py
+++ b/python/lsst/afw/display/ds9.py
@@ -40,14 +40,18 @@ except NameError:
     except Exception as e:
         # No usable version of display_ds9.
         # Let's define a version of getDisplay() which will throw an exception.
-        e.args = ["%s (is display_ds9 setup?)" % e]
 
-        # TODO: once we abandon Python 2 use the following simpler code:
-        # def getDisplay(*args, exception=e, **kwargs):
-        #     raise exception
+        # Prior to DM-9433, we used a closure to implicitly capture the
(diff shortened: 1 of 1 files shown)


python/lsst/afw/display/ds9.py
```diff
@@ -40,19 +40,14 @@
...
         # No usable version of display_ds9.
         # Let's define a version of getDisplay() which will throw an exception.
-        e.args = ["%s (is display_ds9 setup?)" % e]
 
-        # TODO: once we abandon Python 2 use the following simpler code:
-        # def getDisplay(*args, exception=e, **kwargs):
-        #     raise exception
-        class _RaiseException(object):
-            def __init__(self, exception):
-                self.exception = exception
```
timj: @rowen, do you mean you would prefer to change this line to something like:
```python
self.exception = type(e)("%s (is display_d9 setup?)" % e)
```
(and remove the `e.args` bit from above)?

r-owen: Yes, or better yet `RuntimeError("%s (is display_d9 setup?)" % e)` so the exception type is more predictable.

Another obscure issue for Python 2 is that `str(exception)` can raise an exception, but it may not be worth worrying about that.

PR #180, opened 2017-02-24
DM-8465: wrap meas_modelfit with pybind11

TallJimbo:
This adds wrappers for two afw methods that are apparently first used in meas_modelfit (lack of test coverage has been noted).

### DIFF
--- a/python/lsst/afw/geom/ellipses/ellipse.cc
+++ b/python/lsst/afw/geom/ellipses/ellipse.cc
@@ -33,6 +33,7 @@
 #include "lsst/afw/geom/Point.h"
 #include "lsst/afw/geom/ellipses/GridTransform.h"
 #include "lsst/afw/geom/ellipses/Ellipse.h"
+#include "lsst/afw/geom/ellipses/Transformer.h"
 
 namespace py = pybind11;
 using namespace pybind11::literals;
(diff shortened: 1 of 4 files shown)


python/lsst/afw/geom/ellipses/ellipse.cc
```diff
@@ -85,7 +86,13 @@ PYBIND11_PLUGIN(_ellipse) {
     clsEllipse.def("shift", &Ellipse::shift);
     clsEllipse.def("getParameterVector", &Ellipse::getParameterVector);
     clsEllipse.def("setParameterVector", &Ellipse::setParameterVector);
-//    clsEllipse.def("transform", (Transformer const (Ellipse::*)(AffineTransform const &) const) &Ellipse::transform);
+    clsEllipse.def(
+        "transform",
+        [](Ellipse const & self, AffineTransform const & t) -> Ellipse {
```
pschella: Why does this not return a `Transformer` anymore?

TallJimbo: `Transformer`, `Convolution`, and `GridTransform` are expression objects that don't translate to Python well (they require on assignment operators being able to invoke type conversion logic).  They shouldn't be seen in Python at all.

PR #182, opened 2017-02-28
DM-8439: Add wrapper on astshim to take point lists

r-owen:


include/lsst/afw/geom/SpherePoint.h
```diff
@@ -64,6 +64,13 @@ class SpherePoint
 {
 public:
     /**
+     * Construct a SpherePoint with longitude, latitude = 0, 0
+     *
+     * @exceptsafe Provides strong exception guarantee.
+     */
+    SpherePoint();
```
kfindeisen: I'd still prefer not to let `SpherePoint` be default-constructible if it can be avoided; the stated default, while better than any of the alternatives, is not as obvious as it could be because the default point is `(0, 0)` in lon-lat representation but `(1, 0, 0)` in vector representation.

Do you use default constructibility anywhere? I tried looking for explicit variable declarations but didn't find any.

If you do keep the default constructor, consider declaring it `noexcept`.

tests/test_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        point = endpoint.pointFromData(pointData)
+        self.assertEqual(type(point), pointType)
+        for axis in range(nAxes):
+            if isAngle:
+                self.assertAlmostEqual(point[axis].asRadians(), pointData[axis])
+            else:
+                self.assertAlmostEqual(point[axis], pointData[axis])
+
+        pointDataRoundTrip = endpoint.dataFromPoint(point)
+        self.assertTrue(np.allclose(pointData, pointDataRoundTrip))
```
parejkoj: Another case for `self.assertFloatsAlmostEqual`

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+    Note that only Generic has more than one possible number of axes
+    """
+    return {
+        "Generic": (1, 2, 3, 4, 5),
+        "Point3": (3,),
+        "Point2": (2,),
+        "SpherePoint": (2,),
+    }[name]
+
+def getBadNAxes(name):
```
parejkoj: These two methods seem a bit excessive, since only Generic has more than one element, and it has no possible "bad" values. It would be simpler to me to just define a dict for each and use that, unless you plan to add other endpoint types.

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+    return np.array(forwardCoeffs, dtype=float)
+
+def makePolyMap(nIn, nOut):
+    """Make an astShim.PolyMap suitable for testing
+
+    The forward transform is as follows:
+    fj(x) = C0j x0 + C1j x1 + C2j x2... where Cij = 0.001 i (j+1)
+
+    The reverse transform is the same equation with i and j reversed
+    thus it is NOT the inverse of the forward direction,
```
parejkoj: Good to make that clear.

ups/afw.table
```diff
@@ -15,6 +15,7 @@ setupRequired(astropy)
 setupRequired(python_future)
 setupRequired(log)
 setupRequired(pybind11)
+setupRequired(astshim)
```
parejkoj: And they never thought there'd be a day when afw depended on AST!

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+        fromAxes  number of axes in fromFrame
+        toAxes  number of axes in toFrame
+        """
+        transformClassName = "TransformFrom{}To{}".format(fromName, toName)
+        transformClass = getattr(afwGeom, transformClassName)
+        for nIn in getNaxes(fromName):
+            # check invalid numbers of output for the given toName
+            for badNout in getBadNAxes(toName):
+                badPolyMap = makePolyMap(nIn, badNout)
+                with self.assertRaises(Exception):
```
parejkoj: Which Exception?

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+                fromEndpoint = transform.getFromEndpoint()
+                toEndpoint = transform.getToEndpoint()
+                frameSet = transform.getFrameSet()
+
+                # forward transformation of one point
+                rawInPoint = makeRawPointData(nIn)
+                inPoint = fromEndpoint.pointFromData(rawInPoint)
+                outPoint = transform.tranForward(inPoint)
+                rawOutPoint = toEndpoint.dataFromPoint(outPoint)
+                self.assertTrue(np.allclose(rawOutPoint, polyMap.tranForward(rawInPoint)))
```
parejkoj: `self.assertFloatsAlmostEqual` and `msg=` here, and the other asserts below.

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+
+                # inverse transformation of one point;
+                # remember that the inverse will not give the original values!
+                inversePoint = transform.tranInverse(outPoint)
+                rawInversePoint = fromEndpoint.dataFromPoint(inversePoint)
+                self.assertTrue(np.allclose(rawInversePoint, polyMap.tranInverse(rawOutPoint)))
+                self.assertTrue(np.allclose(rawInversePoint, frameSet.tranInverse(rawOutPoint)))
+
+                # forward transformation of an array of points
+                nPoints = 7  # arbitrary
```
parejkoj: Thanks for the comment.

tests/test_transform.py
```diff
@@ -0,0 +1,187 @@
...
+                inverseArray = transform.tranInverse(outArray)
+                rawInverseArray = fromEndpoint.dataFromArray(inverseArray)
+                self.assertTrue(np.allclose(rawInverseArray, polyMap.tranInverse(rawOutArray)))
+                self.assertTrue(np.allclose(rawInverseArray, frameSet.tranInverse(rawOutArray)))
+
+        # check invalid numbers of inputs with valid and invalid #s of inputs
+        for badNin in getBadNAxes(fromName):
+            for nIn in getNaxes(toName) + getBadNAxes(toName):
+                badPolyMap = makePolyMap(badNin, nOut)
+                with self.assertRaises(Exception):
```
parejkoj: Which Exception?

include/lsst/afw/geom/Endpoint.h
```diff
@@ -0,0 +1,303 @@
...
+Virtual base class for endpoints, which are helper classes for Transform
+
+Endpoints transform points and lists of points from LSST-specific data types,
+such as Point2D and SpherePoint, to a form accepted by ast::Mapping.tran.
+Each type of endpoint is used for a particular LSST data type, for example:
+- Point2Endpoint is used for Point2D data
+- Point3Endpoint is used for Point3D data
+- SpherePointEndpoint for SpherePoint data
+- GenericEndpoint is used when no other form will do; its LSST data type
+  is identical to the type used for ast::Mapping.tran.
```
r-owen: I prefer my wording. I consider it an explanation of the design and a suggestion of a place to look for more information.

include/lsst/afw/geom/Endpoint.h
```diff
@@ -0,0 +1,303 @@
...
+- makeFrame creates a new frame with the correct type and settings
+
+@tparam PointT  LSST data type for one point
+@tparam ArrayT  LSST data type for an array of points
+*/
+
+template <typename PointT, typename ArrayT>
+class BaseEndpoint {
+public:
+    using Point = PointT;
```
r-owen: This makes `PointT` available to users as `BaseEndpoint::Point`. I suspect the template type name and the class attribute can have the same name, but I worry that `using Point = Point` looks too strange. @kfindeisen may have an opinion about this.

include/lsst/afw/geom/Endpoint.h
```diff
@@ -0,0 +1,303 @@
...
+    Create a Frame that can be used with this end point in a Transform
+    */
+    virtual std::shared_ptr<ast::Frame> makeFrame() const;
+
+    /**
+    Adjust and check the frame as needed.
+
+    Do not obother to check the number of axes because that is done elsewhere.
+
+    The base implementation does nothing.
```
r-owen: This is a default behavior that derived classes inherit and can choose to override. As such it might do something useful, but in this case it does nothing.

include/lsst/afw/geom/Endpoint.h
```diff
@@ -0,0 +1,303 @@
...
+
+    int _getNAxes(ndarray::Array<double, 1, 1> const & data) const {
+        return data.getSize<0>();
+    }
+
+    int _getNAxes(std::vector<double> const & data) const {
+        return data.size();
+    }
+
+    int _getNPoints(ndarray::Array<double, 2, 2> const & data) const {
```
r-owen: No because `<double, 1, 1>` is used to represent a single point. (Also these private methods are simply used as convenience methods for other code; the fact that the code compiles and passes unit tests is a fairly strong indication that it is not missing such methods).

include/lsst/afw/geom/Endpoint.h
```diff
@@ -0,0 +1,303 @@
...
+
+@tparam N  number of axes; the only allowed values are 2 or 3
+*/
+template <int N>
+class PointEndpoint : public BaseVectorEndpoint<Point<double, N>> {
+public:
+    /// Construct a PointEndpoint; template parameter N specifies the number of axes
+    explicit PointEndpoint() : BaseVectorEndpoint<Point<double, N>>(N) {}
+
+    /// This constructor is used by Transform; nAxes must be 2
```
r-owen: The note is wrong (a holdover from when I was only supporting Point2D); thanks for catching this.

include/lsst/afw/geom/Endpoint.h
parejkoj: Ah, ok. That makes sense as a technical reason why to do it this way.

include/lsst/afw/geom/Transform.h
```diff
@@ -0,0 +1,139 @@
...
+
+namespace lsst {
+namespace afw {
+namespace geom {
+
+/**
+Transform LSST spatial data, such as Point2D and SpherePoint, using an AST transform.
+
+This class contains two Endpoints, to specify the "from" and "to" LSST data type,
+and an astshim::FrameSet or astshim::Mapping to specify the transformation.
```
r-owen: A `FrameSet` may contain multiple transformations, but Transform only exposes the mapping that connects the `BASE` frame to the `CURRENT` frame. I added the following line to the comment in hopes of making this clearer:
```
In the case of a FrameSet the transformation is from the `BASE` frame to the `CURRENT` frame.
```

include/lsst/afw/geom/Endpoint.h
kfindeisen: I don't actually know whether `using Point = Point` would compile (though it probably would). @parejkoj, the [LSST style guide](https://developer.lsst.io/coding/cpp_style_guide.html#names-representing-template-parameters-may-be-a-single-uppercase-letter-or-a-mixed-case-phrase-first-letter-capitalized-in-each-word-indicating-the-desired-type) recommends the `PointT` name (and disambiguating that we *don't* mean `afw::geom::Point` is certainly relevant here).

python/lsst/afw/geom/endpoint.cc
```diff
@@ -0,0 +1,165 @@
...
+    // return a deep copy so Python cannot modify the internal state
+    cls.def("makeFrame", [](Class const & self) {
+        auto frame = self.makeFrame();
+        return frame->copy();
+    });
+}
+}  // <anonymous>
+
+PYBIND11_PLUGIN(_endpoint) {
+    // TODO uncomment these imports during the cleanup pass
```
r-owen: It is waiting for the wrapper library name to not have the leading underscore, which is part of cleaning up the afw.geom pybind11 wrapper. I'll clarify the comment.

python/lsst/afw/geom/transform.cc
```diff
@@ -0,0 +1,105 @@
...
+
+    py::module mod("_transform");
+
+    // Need to import numpy for ndarray and eigen conversions
+    if (_import_array() < 0) {
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return nullptr;
+    }
+
+    declareTransform<GenericEndpoint, GenericEndpoint>(mod, "Generic", "Generic");
```
r-owen: I don't know of any other than m4 which I'd rather not use. I did write a script for the purpose, but it was trivial to write and likely to go out of date, so I am not keen to add it to the repository.

src/geom/Endpoint.cc
```diff
@@ -0,0 +1,148 @@
...
+    this->_assertNAxes(_getNAxes(point));
+    return point;
+}
+
+ndarray::Array<double, 2, 2> GenericEndpoint::dataFromArray(Array const& arr) const {
+    this->_assertNAxes(_getNAxes(arr));
+    return ndarray::copy(arr);
+}
+
+std::vector<double> GenericEndpoint::pointFromData(std::vector<double> const& data) const {
```
r-owen: std::vector is the generic form used to represent one point of data: one element per axis. I'm not entirely happy with it but the other alternatives I considered are:
- Use a 1-d ndarray. That would require a change to astshim. It may be worth considering because vector is copied and ndarray is not. But it's only a few numbers, and vectors are easier to work with both in C++ and Python -- in Python they can be a list or any collection of numbers.
- Use a 2-d ndarray with only one row. This would work, but it seems clumsy and unnatural since the number of rows must always be 1.

src/geom/Endpoint.cc
```diff
@@ -0,0 +1,148 @@
...
+    if (skyFramePtr->getLonAxis() != 1) {
+        // axes are swapped to Lat, Lon; swap them back to the usual Lon, Lat
+        std::vector<int> perm = {2, 1};
+        skyFramePtr->permAxes(perm);
+    }
+}
+
+
+// explicit instantiations
+// TODO understand why all but the last two are needed (to avoid a complaint
```
r-owen: I'll remove the note.

tests/test_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        badFrame = astshim.SkyFrame()
+        with self.assertRaises(Exception):
+            endpoint.normalizeFrame(badFrame)
+
+    def testPoint3Endpoint(self):
+        endpoint = afwGeom.Point3Endpoint()
+        self.checkEndpointBasics(endpoint=endpoint, pointType=afwGeom.Point3D, nAxes=3)
+
+        frame1 = astshim.Frame(2)
+        endpoint.normalizeFrame(frame1)  # should not raise
```
r-owen: OK, though `SpecificException` in this case seems out of place. It should pass and anything else is an error.

parejkoj: There are usually only a small number exceptions you might expect from a call, which are Failures. Other (unexpected) exceptions are Errors, which are treated slightly differently in the test reports.

tests/test_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+
+class EndpointTestCase(unittest.TestCase):
+
+    def testSpherePointEndpoint(self):
+        endpoint = afwGeom.SpherePointEndpoint()
+        self.checkEndpointBasics(endpoint=endpoint, pointType=afwGeom.SpherePoint, nAxes=2)
+
+        for doPermute in (False, True):
+            frame = astshim.SkyFrame()
+            if doPermute:
```
r-owen: I'm going to push back here. I think the code is more readable as written. I worry that by moving it to a subroutine it will be harder to see what's going on and far more verbose for so few lines of code. I could probably use variables to and always call permute (a no-op in one case and permutation in the other), but I fear the results will be harder to read.

tests/test_endpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        with self.assertRaises(Exception):
+            endpoint.normalizeFrame(badFrame)
+
+    def testGnericEndpoint(self):
+        for nAxes in (1, 2, 3, 4, 5):
+            endpoint = afwGeom.GenericEndpoint(nAxes)
+            self.checkEndpointBasics(endpoint=endpoint, pointType=list, nAxes=nAxes)
+
+            newFrame = endpoint.makeFrame()
+            self.assertEqual(type(newFrame), astshim.Frame)
```
r-owen: Great idea! I didn't know about setting `longMessage before`. This could really clean up some of my old tests in afw.

python/lsst/afw/geom/transform.cc
parejkoj: I was thinking "programmatically in C++" (can't the template system basically do anything you want, given sufficient cleverness?). I agree, such a script probably shouldn't live in this repo.

src/geom/Endpoint.cc
parejkoj: God it, thanks for the clarification.

parejkoj:
I think you still need to convert all the `with self.assertRaises(Exception)` into catching only the specific exception that should be raised in those cases. Otherwise, strange errors that shouldn't occur would be masked by the test.

parejkoj:
And I guess the remaining `assertTrue(np.allclose())` are due to the assertFloatsAlmostEqual bug re: lists that you filed earlier? In that case, I'd suggest using `np.testing.assert_almost_equal()`, with a note about why that includes the ticket number (and mention it in the ticket you filed).

python/lsst/afw/geom/transform.cc
```diff
@@ -0,0 +1,120 @@
...
+
+    cls.def("getFromEndpoint", &Class::getFromEndpoint);
+    cls.def("getFrameSet", &Class::getFrameSet);
+    cls.def("getToEndpoint", &Class::getToEndpoint);
+
+    cls.def("tranForward", (ToArray (Class::*)(FromArray const &) const) &Class::tranForward, "array"_a);
+    cls.def("tranForward", (ToPoint (Class::*)(FromPoint const &) const) &Class::tranForward, "point"_a);
+    cls.def("tranInverse", (FromArray (Class::*)(ToArray const &) const) &Class::tranInverse, "array"_a);
+    cls.def("tranInverse", (FromPoint (Class::*)(ToPoint const &) const) &Class::tranInverse, "point"_a);
+    /// repr format is lsst.afw.geom.<ClassName>[<nIn>-><nOut>]
```
kfindeisen: Doxygen comments should not be used inside function bodies. Did you mean `//`? [Same for `__str__`]

python/lsst/afw/geom/endpoint.cc
```diff
@@ -0,0 +1,182 @@
...
+    using Class = typename PyClass::type;  // C++ class associated with pybind11 wrapper class
+    cls.def("__repr__", [](Class const& self) {
+        std::ostringstream os;
+        os << "lsst.afw.geom." << self;
+        return os.str();
+    });
+    cls.def("__str__", [](Class const& self) {
+        std::ostringstream os;
+        os << self;
+        return os.str();
```
r-owen: I'd be surprised and disappointed if we felt we had to dictate the variable name in our standards.

python/lsst/afw/geom/endpoint.cc
```diff
@@ -0,0 +1,182 @@
...
+
+    // abstract base class so do not wrap the constructor
+
+    cls.def("getNAxes", &Class::getNAxes);
+    cls.def("normalizeFrame", &Class::normalizeFrame);
+    // return a deep copy so Python cannot modify the internal state
+    cls.def("makeFrame", [](Class const & self) {
+        auto frame = self.makeFrame();
+        return frame->copy();
+    });
```
r-owen: I didn't wrap `getNPoints` and etc. because they were not implemented in this base class, but I believe it's OK to wrap them anyway so I'll try that.

I don't anticipate us making Python-only subclasses, and the protected methods aren't essential to doing so.

python/lsst/afw/geom/endpoint.cc
```diff
@@ -0,0 +1,182 @@
...
+void declareBaseVectorEndpoint(py::module& mod, std::string const& suffix) {
+    using Class = BaseVectorEndpoint<Point>;
+    using Array = typename Class::Array;
+
+    declareBaseEndpoint<Point, Array>(mod, suffix);
+
+    py::class_<Class, std::shared_ptr<Class>, BaseEndpoint<Point, Array>> cls(
+            mod, ("_BaseVectorEndpoint" + suffix).c_str());
+
+    cls.def(py::init<int>(), "nAxes"_a);
```
r-owen: I'll remove it. Also, I'll make the constructor protected, since it's only intended to be called by subclasses.

python/lsst/afw/geom/endpoint.cc
kfindeisen: Not that kind of standard. Pim expressed an interest in reducing the code duplication in our current `__str__` and `__repr__` implementations, presumably by creating a single centralized function to replace these lambdas.

src/geom/Endpoint.cc
```diff
@@ -0,0 +1,148 @@
...
+    return Point(data.data());
+}
+
+template <typename Point>
+std::vector<Point> BaseVectorEndpoint<Point>::arrayFromData(ndarray::Array<double, 2, 2> const& data) const {
+    this->_assertNAxes(this->_getNAxes(data));
+    int const nPoints = this->_getNPoints(data);
+    Array arr;
+    arr.reserve(nPoints);
+    for (auto dataRowIter = data.begin(), dataRowEnd = data.end(); dataRowIter != dataRowEnd; ++dataRowIter) {
```
r-owen: I think that will work. It's worth a try.

python/lsst/afw/geom/endpoint.cc
r-owen: Got it. I think that will work out to not wrapping `__repr__` at all (which I am violating to provide a slightly nicer class name) and making `__str__` = operator<< (which it is, so it can trivially modified to call the standard function when we have one).

PR #184, opened 2017-03-01
DM-7913: Implement RFC-240: make Angle's named methods const...

kfindeisen:
In addition to the changes requested in [DM-7913](https://jira.lsstcorp.org/browse/DM-7913), this change includes some code cleanup of `Angle`... which I might have gotten a bit carried away with:
- `clang-format` pass
- splitting declaration/definition of complex methods in `Angle`
- added unary `-` operator (surprising omission in previous API, came up when writing tests for `separation`)
- added support for C++11 features
- reformatted and greatly expanded documentation


include/lsst/afw/geom/Angle.h
```diff
@@ -3,241 +3,294 @@
...
 
-ANGLE_OPUP_TYPE(*=, double)
-ANGLE_OPUP_TYPE(*=, int)
-ANGLE_OPUP_TYPE(+=, double)
-ANGLE_OPUP_TYPE(+=, int)
-ANGLE_OPUP_TYPE(-=, double)
-ANGLE_OPUP_TYPE(-=, int)
+    //@{
+    /// Multiply this angle by the given factor.
+    ANGLE_OPUP_TYPE(*=, double)
```
r-owen: Any chance of getting rid of these? It is another unwelcome instance of treating `radians` as special. I suppose that should be another RFC.

include/lsst/afw/geom/Angle.h
```diff
@@ -3,241 +3,294 @@
...
 
 #undef ANGLE_COMP
 
 private:
     double _val;
 };
 
-Angle const NullAngle = Angle(-1000000., degrees);
-
+Angle constexpr NullAngle = Angle(-1000000., degrees);
```
r-owen: Please get rid of this. A global search turned up no usage of this in the stack and we use `nan` for an unknown angle.

include/lsst/afw/geom/Angle.h
```diff
@@ -246,73 +299,182 @@ Angle const NullAngle = Angle(-1000000., degrees);
  * N.b. We need both int and double versions to avoid ambiguous overloading due to implicit conversion of
  * Angle to double
```
r-owen: `ANGLE_OP` does not appear to do anything with `int`; I suspect this comment refers to `ANGLE_OP_TYPE`; if so please move it.

include/lsst/afw/geom/Angle.h
```diff
@@ -246,73 +299,182 @@ Angle const NullAngle = Angle(-1000000., degrees);
...
-const Angle operator *(T lhs,              ///< the value to convert
-                       AngleUnit const rhs ///< the conversion coefficient
-                      ) {
-    static_assert(std::numeric_limits<T>::is_specialized,
-                            "Only numeric types may be converted to Angles using degrees/radians!");
-    return Angle(lhs*rhs._val);
+template <typename T>
+inline constexpr Angle operator*(T lhs, AngleUnit rhs) noexcept {
+    static_assert(std::is_arithmetic<T>::value,
+                  "Only numeric types may be converted to Angles using degrees/radians!");
```
r-owen: This error message could be improved. Perhaps something like "Only numeric types may multiplied by an AngleUnit such as degrees or radians to create an Angle".

include/lsst/afw/geom/Angle.h
```diff
@@ -246,73 +299,182 @@ Angle const NullAngle = Angle(-1000000., degrees);
...
+    return Angle(lhs * rhs._val);
+}
+
+/************************************************************************************************************/
+// Inline method definitions, placed last in order to benefit from Angle's full API
+
+inline double Angle::toUnitSphereDistanceSquared() const noexcept {
+    return 2. * (1. - std::cos(asRadians()));
+    // == 4.0 * pow(std::sin(0.5 * asRadians()), 2.0)
```
r-owen: Is one preferred over the other in a particular regime? If so, is there any point to a conditional?

In any case it might be worth a brief comment as to why one was chosen and the other not

include/lsst/afw/geom/Angle.h
```diff
@@ -246,73 +299,182 @@ Angle const NullAngle = Angle(-1000000., degrees);
...
+    double const refAngRad = refAng.asRadians();
+    double wrapped = (*this - refAng).wrapCtr().asRadians() + refAngRad;
+
+    // roundoff can cause slightly out-of-range values; fix those
+    if (wrapped - refAngRad >= PI) {
+        wrapped -= TWOPI;
+    }
+    // maximum relative roundoff error for subtraction is 2 epsilon
+    if (wrapped - refAngRad < -PI) {
+        wrapped -= wrapped * 2.0 * std::numeric_limits<double>::epsilon();
```
r-owen: Shouldn't this be +=?

python/lsst/afw/geom/angle/angle.cc
```diff
@@ -90,20 +90,22 @@ PYBIND11_PLUGIN(angle) {
...
     clsAngle.def("__iadd__", [](Angle& self, Angle const& other) { return self += other; });
     clsAngle.def("__isub__", [](Angle& self, Angle const& other) { return self -= other; });
     clsAngle.def("__truediv__", [](Angle const& self, double other) { return self / other; },
                  py::is_operator());
+    // Without an explicit wrapper, Python lets Angle / Angle -> Angle
+    clsAngle.def("__truediv__", [](Angle const& self, Angle const& other) {
+        throw py::type_error("unsupported operand type(s) for /: 'Angle' and 'Angle'");
+    });
 
     clsAngle.def("__float__", &Angle::operator double);
```
r-owen: Can we get rid of this? I suppose it may require an RFC (though if it passed Jenkins...)

include/lsst/afw/geom/Angle.h
kfindeisen: I don't see anything wrong with the `*=` operator; 2.5 times an angle is a well-defined geometric quantity. I'd agree with you on, e.g., `Angle += double`, but that's [DM-9630](https://jira.lsstcorp.org/browse/DM-9630).

include/lsst/afw/geom/Angle.h
kfindeisen: There might be, but this is preexisting code, and the initial commit (2b49698aac5268bff502c3c48061f1e547db9a29) doesn't provide any clues.

kfindeisen: Hmm... both methods are used only in `afw/table/Match.cc`. Perhaps they could at least be moved there, since they aren't really operations on angles in the general sense.

include/lsst/afw/geom/Angle.h
kfindeisen: Actually, this code doesn't make sense at all -- it's rescaling `wrapped` by a factor of (1 - 2&epsilon;), but depending on the value of `refAngRad` the absolute value of `wrapped` may be well outside the ~&pi; range. Shouldn't the correction be scaled to `|this - refAng|`?

kfindeisen: I think this is equivalent to `wrapped *= (1.0 - 2ε)`, regardless of whether `wrapped` is positive or negative. So neither += nor -= really make sense (e.g., consider if `wrapped = 0` and `refAngRad = π + ε`).

kfindeisen: Try as I might, I can't come up with a test case that exposes any problems with the code as-is (although _removing_ this block does cause existing tests to fail). I'm not comfortable trying to fix this if I can't see the effects, so at most I can file a ticket to investigate later... :disappointed:

PR #187, opened 2017-03-04
DM-8716: cleanup (and in many cases, rewrite) afw::table pybind11 wrappers

TallJimbo:


include/lsst/afw/table/python/catalog.h
```diff
@@ -21,102 +19,117 @@
...
+@tparam Record  Record type, e.g. BaseRecord or SimpleRecord.
+
+@param[in] mod    Module object class will be added to.
+@param[in] name   Name prefix of the record type, e.g. "Base" or "Simple".
+@param[in] isBase Whether this instantiation is only being used as a base class (used to set the class name).
 */
-template <typename RecordT>
-void declareCatalog(pybind11::class_<CatalogT<RecordT>, std::shared_ptr<CatalogT<RecordT>>> & cls) {
+template <typename Record>
+PyCatalog<Record> declareCatalog(pybind11::module & mod, std::string const & name, bool isBase=false) {
```
pschella: New idiom?

include/lsst/afw/table/python/catalog.h
```diff
@@ -21,102 +19,117 @@
...
+    using Table = typename Record::Table;
+
+    std::string fullName;
+    if (isBase) {
+        fullName = "_" + name + "CatalogBase";
+    } else {
+        fullName = name + "Catalog";
+    }
+
+    PyCatalog<Record> cls(mod, fullName.c_str(), py::dynamic_attr());
```
pschella: A comment, why `py::dynamic_attr` is needed here, would be helpful.

include/lsst/afw/table/python/sortedCatalog.h
```diff
@@ -85,6 +88,39 @@ void declareSortedCatalog(
             (Catalog (Catalog::*)(ndarray::Array<bool const,1> const &) const) &Catalog::subset);
     cls.def("subset",
             (Catalog (Catalog::*)(std::ptrdiff_t, std::ptrdiff_t, std::ptrdiff_t) const) &Catalog::subset);
+
+    cls.def(
```
pschella: Comment would be appreciated.

python/lsst/afw/table/aggregates/aggregatesContinued.py
```diff
@@ -0,0 +1,53 @@
...
+from lsst.utils import TemplateMeta
+from . import aggregates
+
+
+class PointKey(with_metaclass(TemplateMeta, object)):
+    pass
+
+
+PointKey.register(np.int32, aggregates.Point2IKey)
+PointKey.register(np.float64, aggregates.Point2DKey)
```
pschella: Doesn't this exclude the future possibility of having a `Point3IKey` (by dropping the `2` in the abc)?

python/lsst/afw/table/arrays/arrays.cc
```diff
@@ -0,0 +1,122 @@
...
+                py::size_t start=0, stop=0, step=0, length=0;
+                bool valid = slice.compute(self.getSize(), &start, &stop, &step, &length);
+                if (!valid) throw py::error_already_set();
+                if (step != 1) {
+                    PyErr_SetString(PyExc_IndexError, "Step for ArrayKey must be 1.");
+                    throw py::error_already_set();
+                }
+                return py::cast(self.slice(start, stop));
+            } else {
+                return py::cast(self[py::cast<int>(index)]);
```
pschella: `size_t`?

include/lsst/afw/table/Schema.h
```diff
@@ -171,7 +184,7 @@ class Schema {
      *  @brief Add a new field to the Schema, and return the associated Key.
      *
      *  This is simply a convenience wrapper, equivalent to:
-     *  @code
+     *  @codeinspect
```
r-owen: What is the `codeinspect` command? It is not listed in https://www.stack.nl/~dimitri/doxygen/manual/commands.html and a google search didn't turn up anything.

include/lsst/afw/table/Schema.h
```diff
@@ -341,10 +354,52 @@ class SubSchema {
...
+    /**
+     *  @brief Find a nested SchemaItem by name and run a functor on it.
+     *
+     *  Names corresponding to named subfields are not accepted.
+     *  The given functor must have an overloaded function call
+     *  operator that accepts any SchemaItem type (the same as
+     *  a functor provided to apply or Schema::forEach).
+     */
+    template <typename F>
+    void find(std::string const & name, F && func) const {
```
r-owen: I'm curious why you didn't use std::function here.

include/lsst/afw/table/python/catalog.h
r-owen: @pschella how so?

python/lsst/afw/table/base/base.cc
```diff
@@ -0,0 +1,195 @@
...
+    cls.def("preallocate", &BaseTable::preallocate);
+
+    return cls;
+}
+
+PYBIND11_PLUGIN(base) {
+    py::module mod("base");
+
+    py::module::import("lsst.afw.table.schema");
+    py::module::import("lsst.afw.table.baseColumnView");
```
pschella: Typically above module.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+
+#include "numpy/arrayobject.h"
+#include "ndarray/pybind11.h"
+
+#include "lsst/afw/table/Key.h"
+#include "lsst/afw/table/BaseColumnView.h"
+
+namespace ndarray { namespace detail {
+
+template <> struct NumpyTraits<lsst::afw::geom::Angle> : public NumpyTraits<double> {};
```
pschella: Comment please.

include/lsst/afw/table/BaseTable.h
```diff
@@ -40,12 +40,7 @@ namespace table {
  *  Each table class should be associated with a particular record class (1-to-1).  Each table instance may
  *  be associated with many record instances.
  */
-class BaseTable
-#ifndef SWIG // swig complains about these not being %shared_ptrs, but it doesn't need to know about them
-: public std::enable_shared_from_this<BaseTable>,
-  public daf::base::Citizen
-#endif
-{
+class BaseTable : public std::enable_shared_from_this<BaseTable>, public daf::base::Citizen {
```
kfindeisen: @r-owen `enable_shared_from_this` is to prevent objects from being managed by two independent `shared_ptr` groups, particularly as a result of naively calling `shared_ptr(this)`. See Effective Modern C++ (can't name a specific Item right now) for details.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+template <typename PyClass>
+static void declareBaseColumnViewFlagOverloads(PyClass & cls) {
+    cls.def("_basicget", [](BaseColumnView & self, Key<Flag> const & key)
+                          ->ndarray::Array<bool const,1,1> const {
+        return ndarray::copy(self[key]);
+    });
+};
+
+static void declareBaseColumnView(py::module & mod) {
+    PyBaseColumnView cls(mod, "_BaseColumnViewBase"); // "BaseColumnView" = "ColumnViewT<BaseRecord>"
```
pschella: `_BaseColumnViewBase`?
And comments should go above.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+
+namespace lsst {
+namespace afw {
+namespace table {
+namespace {
+
+using PyBaseColumnView = py::class_<BaseColumnView, std::shared_ptr<BaseColumnView>>;
+
+using PyBitsColumn = py::class_<BitsColumn, std::shared_ptr<BitsColumn>>;
+
```
pschella: anonymous namespace

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+template <typename T, typename PyClass>
+static void declareBaseColumnViewOverloads(PyClass & cls) {
```
pschella: no static

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+    cls.def("_basicget", [](BaseColumnView & self, Key<T> const & key)
+                          ->typename ndarray::Array<T,1> const {
+        return self[key];
+    });
+};
+
+template <typename U, typename PyClass>
+static void declareBaseColumnViewArrayOverloads(PyClass & cls) {
```
pschella: no static

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+    cls.def("_basicget", [](BaseColumnView & self, Key<lsst::afw::table::Array<U>> const & key)
+                          ->typename ndarray::Array<U,2,1> const {
+        return self[key];
+    });
+};
+
+template <typename PyClass>
+static void declareBaseColumnViewFlagOverloads(PyClass & cls) {
```
pschella: no static

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+    cls.def("_basicget", [](BaseColumnView & self, Key<Flag> const & key)
+                          ->ndarray::Array<bool const,1,1> const {
+        return ndarray::copy(self[key]);
+    });
+};
+
+static void declareBaseColumnView(py::module & mod) {
```
pschella: no static

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -0,0 +1,132 @@
...
+    // std::string columns are not supported, because numpy string arrays
+    // do not have the same memory model as ours.
+    declareBaseColumnViewOverloads<lsst::afw::geom::Angle>(cls);
+    declareBaseColumnViewArrayOverloads<std::uint16_t>(cls);
+    declareBaseColumnViewArrayOverloads<int>(cls);
+    declareBaseColumnViewArrayOverloads<float>(cls);
+    declareBaseColumnViewArrayOverloads<double>(cls);
+}
+
+static void declareBitsColumn(py::module & mod) {
```
pschella: no static

include/lsst/afw/table/FieldBase.h
```diff
@@ -58,18 +58,16 @@ struct FieldBase {
...
+    // Only the first of these constructors is valid for this specializations, but
+    // it's convenient to be able to instantiate both, since the other is used
+    // by other specializations.
     FieldBase() {}
     FieldBase(int) {
         throw LSST_EXCEPT(
             lsst::pex::exceptions::LogicError,
-            "Constructor disabled (it only appears to exist as a workaround for a SWIG bug)."
+            "Constructor disabled (this Field type is not sized)."
         );
```
kfindeisen: Can you use `FieldBase(int) = delete` here? I'm not clear on how `delete` works when applied to general templates but not template specializations.

python/lsst/afw/table/baseColumnView/baseColumnViewContinued.py
```diff
@@ -0,0 +1,157 @@
...
+from past.builtins import basestring
+
+import numpy as np
+
+from lsst.utils import continueClass
+from ..schema import KeyFlag
+from .baseColumnView import _BaseColumnViewBase
+
+@continueClass
+class _BaseColumnViewBase:
```
pschella: still private?

include/lsst/afw/table/Flag.h
```diff
@@ -37,18 +37,16 @@ struct FieldBase<Flag> {
...
+    // it's convenient to be able to instantiate both, since the other is used
+    // by other specializations.
     FieldBase() {}
     FieldBase(int) {
         throw LSST_EXCEPT(
             lsst::pex::exceptions::LogicError,
-            "Constructor disabled (it only appears to exist as a workaround for a SWIG bug)."
+            "Constructor disabled (this Field type is not sized)."
         );
     }
```
kfindeisen: Definitely don't see the benefit of keeping this constructor for `FieldBase<Flag>`.

python/lsst/afw/table/exposure/exposure.cc
```diff
@@ -125,45 +109,26 @@ void declareExposureTable(PyExposureTable & cls) {
...
-@param[in] cls  Catalog pybind11 class.
-*/
-void declareExposureCatalog(PyExposureCatalog & cls) {
+PyExposureCatalog declareExposureCatalog(py::module & mod) {
     using Catalog = ExposureCatalogT<ExposureRecord>;
+    table::python::declareSortedCatalog<ExposureRecord>(mod, "Exposure", true);
 
-    // TODO: Commented-out code is waiting until needed and is untested.
-    // Add tests for it and enable it or remove it before the final pybind11 merge.
+    PyExposureCatalog cls(mod, "ExposureCatalog", py::dynamic_attr());
```
pschella: Please add comment why this is needed.

python/lsst/afw/table/exposure/exposure.cc
```diff
@@ -176,29 +141,22 @@ void declareExposureCatalog(PyExposureCatalog & cls) {
...
-    declareExposureTable(clsExposureTable);
-    table::python::declareColumnView(clsExposureColumnView);
-    table::python::declareCatalog(clsBaseExposureCatalog);
-    table::python::declareSortedCatalog(clsSortedBaseExposureCatalog);
-    declareExposureCatalog(clsExposureCatalog);
+PYBIND11_PLUGIN(exposure) {
+    py::module mod("exposure");
+    py::module::import("lsst.afw.table.simple");
+    py::module::import("lsst.afw.geom");
+    // afw.image and afw.detection cannot be imported due to circular dependencies
```
pschella: Are they imported in Python or not needed at all? In the latter case, please just remove the comment.

python/lsst/afw/table/io/persistable.cc
```diff
@@ -20,37 +20,35 @@
...
-    py::class_<Persistable, std::shared_ptr<Persistable>> clsPersistable(mod, "Persistable");
+namespace lsst { namespace afw { namespace table { namespace io {
 
-    /* Member types and enums */
+using PyPersistable = py::class_<Persistable, std::shared_ptr<Persistable>>;
 
-    /* Constructors */
+PYBIND11_PLUGIN(persistable) {
+    py::module mod("persistable");
+    py::module::import("lsst.afw.fits");
```
pschella: Import first?

include/lsst/afw/table/Schema.h
```diff
@@ -46,7 +46,7 @@ class Schema {
 public:
 
     // This variable is defined in SchemaImpl, but is replicated here as
-    // a static so that it is available to SWIG.
+    // a static so that it is available to Python.
     static int const VERSION = detail::SchemaImpl::VERSION;
```
kfindeisen: It should clearly be `static` (i.e., be a class-level constant), but perhaps the comment is clearer if "as a static" is removed from the text.

python/lsst/afw/table/match/match.cc
```diff
@@ -35,93 +34,80 @@ using namespace pybind11::literals;
...
 }
 
-PYBIND11_PLUGIN(_match) {
-    py::module mod("_match", "Python wrapper for afw _match library");
+PYBIND11_PLUGIN(match) {
+    py::module mod("match");
+    py::module::import("lsst.afw.geom");
+    py::module::import("lsst.afw.coord");
+    py::module::import("lsst.afw.table.simple");
+    py::module::import("lsst.afw.table.source");
```
pschella: import before module?

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+using namespace pybind11::literals;
+
+namespace lsst {
+namespace afw {
+namespace table {
+
+namespace {
+
+using PySchema = py::class_<Schema>;
+
```
pschella: remove newline

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+template <typename T>
+using PyKey = py::class_<Key<T>, KeyBase<T>, FieldBase<T>>;
+
+template <typename T>
+using PySchemaItem = py::class_<SchemaItem<T>>;
+
+
+// Specializations for FieldBase
+
```
pschella: namespace?

include/lsst/afw/table/Schema.h
```diff
@@ -417,7 +417,7 @@ class SubSchema {
     /**
      *  @brief Implicit conversion to the appropriate Key type.
      *
-     *  Implicit conversion operators cannot be translated to Python.  Instead, the SWIG
+     *  Implicit conversion operators cannot be translated to Python.  Instead, the Python
      *  wrappers provide an equivalent asKey() method.
      */
     template <typename T>
```
kfindeisen: Since implicit conversions *can* be replicated with pybind11, will this be removed in the future?

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+                if (!valid) throw py::error_already_set();
+                if (step != 1) {
+                    throw LSST_EXCEPT(
+                        pex::exceptions::InvalidParameterError,
+                        "Step for array Key indexing must be 1."
+                    );
+                }
+                return py::cast(self.slice(start, stop));
+            } else {
+                return py::cast(self[py::cast<int>(index)]);
```
pschella: `size_t`?

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+        [](Field<T> const & self, Schema & schema, bool doReplace) -> Key<T> {
+            return schema.addField(self, doReplace);
+        }
+    );
+    clsField.def("getName", &Field<T>::getName);
+    clsField.def("getDoc", &Field<T>::getDoc);
+    clsField.def("getUnits", &Field<T>::getUnits);
+    clsField.def("copyRenamed", &Field<T>::copyRenamed);
+    clsField.def(
+        "__repr__",
```
pschella: Should probably be `__str__` after rfc.

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+            std::ostringstream os;
+            os << self;
+            return os.str();
+        }
+    );
+    // The Key methods below actually wrap templated methods on Schema and
+    // SchemaMapper.  Rather than doing many-type overload resolution by
+    // wrapping those methods directly, we use the visitor pattern by having
+    // the wrappers for those methods delegate back to these non-templated
+    // methods on the templated Key classes.
```
pschella: nice :)

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+        "__getitem__",
+        [](py::object const & self, int index) -> py::object {
+            if (index == 0) {
+                return self.attr("key");
+            } else if (index == 1) {
+                return self.attr("field");
+            }
+            // Have to raise IndexError not some LSST exception to get the
+            // right behavior when unpacking.
+            PyErr_SetString(PyExc_IndexError, "Index to SchemaItem must be 0 or 1.");
```
pschella: This does not look very pythonic to me. Neither does it look like good C++. I think we should either take a `KEY / FIELD` enum, or a `"key" / "field"` string argument.

pschella: Or is this for iteration? Then I still don't like this interface.

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+    clsSchemaItem.def(
+        "__repr__",
+        [](py::object const & self) -> py::str {
+            return py::str("SchemaItem(key={0.key}, field={0.field})").format(self);
+        }
+    );
+}
+
+// Helper class for Schema::find(name, func) that converts the result to Python.
+// In C++14, this should be converted to a universal lambda.
```
pschella: Yes, please C++14 👍

python/lsst/afw/table/schema/schemaContinued.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    for k, v in types.items():
+        dtype = _dtypes.get(k, None)
+        if dtype is not None:
+            abc.register(dtype, v)
+            abc.alias(k, v)
+        else:
+            abc.register(k, v)
+
+
```
pschella: Please add comment here explaining where `_Key`, `_Field` and `_SchemaItem` come from and what they contain.

python/lsst/afw/table/schemaMapper/schemaMapper.cc
```diff
@@ -0,0 +1,82 @@
...
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+namespace lsst {
+namespace afw {
+namespace table {
+namespace {
+
+using PySchemaMapper = py::class_<SchemaMapper, std::shared_ptr<SchemaMapper>>;
+
```
pschella: namespace

include/lsst/afw/table/python/catalog.h
TallJimbo: Perhaps (in the sense that I've used this on all the `Catalog` declare functions), but the C++ pattern it's wrapping (multiple levels of CRTP base classes) is not a common one, and in fact one I somewhat regret and would not like to see used again.

python/lsst/afw/table/source/source.cc
```diff
@@ -261,4 +245,4 @@ PYBIND11_PLUGIN(_source) {
     return mod.ptr();
 }
 
-}}}  // namespace lsst::afw::table
+}}}}  // namespace lsst::afw::table::<anonymous>
```
pschella: Why is the module in the anonymous namespace? Only the declare functions should be.

include/lsst/afw/table/python/catalog.h
```diff
@@ -144,17 +157,22 @@ void declareCatalog(pybind11::class_<CatalogT<RecordT>, std::shared_ptr<CatalogT
...
- 
+
     cls.def("set", &Catalog::set);
     cls.def("_getitem_", [](Catalog & self, int i) {
-        return self.get(utils::python::cppIndex(self.size(), i));
+        try { // try/catch is a workaround for DM-9715
+            return self.get(utils::python::cppIndex(self.size(), i));
+        } catch (pex::exceptions::OutOfRangeError & err) {
+            PyErr_SetString(PyExc_IndexError, err.what());
+            throw py::error_already_set();
```
kfindeisen: This looks awfully Swiggish. Is this really better than throwing `IndexError` from a python wrapper?

include/lsst/afw/table/Schema.h
```diff
@@ -341,6 +341,19 @@ class SubSchema {
     typedef detail::SchemaImpl Impl;
 public:
 
+    //@{
+    /// Join strings using the field delimiter appropriate for this Schema
```
pschella: Why not have a nice variadic?

include/lsst/afw/table/Schema.h
pschella: Not needed, and is potentially more costly I assume.

pschella: More explicit though.

python/lsst/afw/table/schema/schema.cc
```diff
@@ -0,0 +1,522 @@
...
+        "__contains__",
+        [](py::object const & self, py::object const & key) {
+            try {
+                self.attr("find")(key);
+            } catch (py::error_already_set & err) {
+                return false;
+            }
+            return true;
+        }
+    );
```
r-owen: Can the `__contains__` method match the `contains` method? I think that would better match user expectations (especially if we also deprecate `contains`)

include/lsst/afw/table/Schema.h
pschella: And this is what STL does ;)

include/lsst/afw/table/python/functorKey.h
```diff
@@ -1,54 +0,0 @@
...
-template <typename T>
-void declareFunctorKeys(pybind11::module & mod, std::string const & suffix) {
-    pybind11::class_<OutputFunctorKey<T>, std::shared_ptr<OutputFunctorKey<T>>>
-        clsOutputFunctorKey(mod, ("OutputFunctorKey"+suffix).c_str());
-    pybind11::class_<InputFunctorKey<T>, std::shared_ptr<InputFunctorKey<T>>>
-        clsInputFunctorKey(mod, ("InputFunctorKey"+suffix).c_str());
-    pybind11::class_<FunctorKey<T>, std::shared_ptr<FunctorKey<T>>, OutputFunctorKey<T>, InputFunctorKey<T>>
-        clsFunctorKey(mod, ("FunctorKey"+suffix).c_str());
-    
-};
```
kfindeisen: Where does this code go? I don't see any wrapping of `FunctorKey` in other files.

python/lsst/afw/table/aggregates/aggregatesContinued.py
TallJimbo: I'll add the dimension as a separate key.

include/lsst/afw/table/python/sortedCatalog.h
```diff
@@ -1,73 +1,76 @@
...
-    from lsst.afw.table.catalog import addCatalogMethods
-    from lsst.afw.table.sortedCatalog import addSortedCatalogMethods
-    from *yourmodule* import *YourCatalogClass*
-    addCatalogMethods(*YourCatalogClass*)
-    addSortedCatalogMethods(*YourCatalogClass*)
+In addition to calling this method (which also instantiates and wraps the CatalogT base class),
+you must call addCatalogMethods on the class object in Python.
 
-@tparam RecordT  Record type, e.g. BaseRecord or SimpleRecord.
+@tparam Record  Record type, e.g. BaseRecord or SimpleRecord.
```
kfindeisen: Note this goes against [LSST recommendations](https://developer.lsst.io/coding/cpp_style_guide.html#names-representing-template-parameters-may-be-a-single-uppercase-letter-or-a-mixed-case-phrase-first-letter-capitalized-in-each-word-indicating-the-desired-type) (though I personally prefer the new form).

python/lsst/afw/table/arrays/arrays.cc
TallJimbo: Yup.  Also made it use `cppIndex`.

python/lsst/afw/table/schema/schemaContinued.py
```diff
@@ -0,0 +1,204 @@
...
+from builtins import str as futurestr
+from future.utils import with_metaclass
+
+import numpy as np
+import fnmatch
+import re
+import collections
+import astropy.units
+
+from ...geom import Angle
```
r-owen: I suggest `from lsst.afw.geom import Angle` as being clearer

include/lsst/afw/table/BaseTable.h
TallJimbo: And yes, this is why `BaseTable::make` exists and you can't call the ctors directly - to use `enable_shared_from_this` we need to guarantee the instances are always in `shared_ptrs`.

include/lsst/afw/table/Schema.h
TallJimbo: I guess we don't expose it to Python (maybe we don't need it, but I'll add it just in case; clearly we needed it at one point).  Agree the comment is clearer without "as a static".

python/lsst/afw/table/schema/schemaContinued.py
r-owen: In my opinion "where they come from" is already sufficiently clear, since they are imported from schema. I won't comment on what they contain.

include/lsst/afw/table/Schema.h
TallJimbo: Also, it could be constexpr, but I don't think there's much value for it to be, and I'd rather not change the pure C++ more than I have to.

include/lsst/afw/table/Schema.h
TallJimbo: Looks like I thought I was typing into another window and committed it.  I'll revert this entirely.

python/lsst/afw/table/schema/schemaContinued.py
```diff
@@ -0,0 +1,204 @@
...
+if futurestr is not str:
+    Key.alias(futurestr, _Key["String"])
+    Field.alias(futurestr, _Field["String"])
+    SchemaItem.alias(futurestr, _SchemaItem["String"])
+
+
+@continueClass
+class Schema:
+
+    def getOrderedNames(self):
```
r-owen: Do we need "Ordered" in the method name? Definition order seems a reasonable default. (If this is pre-existing then no point changing it now)

include/lsst/afw/table/Schema.h
TallJimbo: This function needs to have a templated `operator()`, which `std::function` doesn't support.

python/lsst/afw/table/base/base.cc
TallJimbo: @r-owen was saying he liked it better this way, and I'm inclined to agree.  I don't think it's a big enough deal to put a lot of effort into standardizing, though.

include/lsst/afw/table/Schema.h
```diff
@@ -100,6 +100,19 @@ class Schema {
...
     /**
+     *  @brief Find a SchemaItem by name and run a functor on it.
+     *
+     *  Names corresponding to named subfields are not accepted.
+     *  The given functor must have an overloaded function call
+     *  operator that accepts any SchemaItem type (the same as
+     *  a functor provided to forEach).
+     */
+    template <typename F>
+    void find(std::string const & name, F && func) const {
```
pschella: I don't like the name of this function. `find` does not imply "run a functor on it".

`applyTo` / `findAndApply` / `transformItem`? I don't know, but not `find`.

python/lsst/afw/table/slots.cc
```diff
@@ -31,48 +30,41 @@ using namespace pybind11::literals;
...
+    cls.def("getMeasKey", &Class::getMeasKey);
+    cls.def("getErrKey", &Class::getErrKey);
+    cls.def("getFlagKey", &Class::getFlagKey);
+    cls.def("setKeys", &Class::setKeys, "alias"_a, "schema"_a);
+}
 
-    /* Members */
-    clsSlotDefinition.def("getName", &SlotDefinition::getName);
-    clsSlotDefinition.def("getAlias", &SlotDefinition::getAlias);
+PYBIND11_PLUGIN(slots) {
```
r-owen: Why is this in anonymous namespace? A code comment would be helpful (here and anywhere else it's done -- I noted one other instance above)

include/lsst/afw/table/detail/SchemaImpl.h
```diff
@@ -102,6 +102,20 @@ class SchemaImpl : public daf::base::Citizen {
     /// Find an item by key (used to implement Schema::find).
     SchemaItem<Flag> find(Key<Flag> const & key) const;
 
+    /// Find an item by name and run the given functor on it.
+    template <typename F>
+    void find(std::string const & name, F && func) const {
```
pschella: Again, I don't like the name.

include/lsst/afw/detection/Peak.h
```diff
@@ -194,7 +200,7 @@ class PeakTable : public afw::table::BaseTable {
     friend class afw::table::io::FitsWriter;
 
      // Return a writer object that knows how to save in FITS format.  See also FitsWriter.
-    virtual PTR(afw::table::io::FitsWriter) makeFitsWriter(fits::Fits * fitsfile, int flags) const;
+    std::shared_ptr<afw::table::io::FitsWriter> makeFitsWriter(fits::Fits * fitsfile, int flags) const override;
```
pschella: `override` implies `virtual`. Policy is to not specify both.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
TallJimbo: `BaseColumnView` is more logically mapped to `ColumnViewT<BaseRecord>` in Python, so I'm using the same naming convention I use for invisible base classes in the `declare` methods.  Comment expanded.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
TallJimbo: I don't follow - all of this *is* in an anonymous namespace.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
TallJimbo: Wait, we changed our mind on this again?

python/lsst/afw/table/baseColumnView/baseColumnViewContinued.py
TallJimbo: Yup.  I'll add the same comment here.

src/table/Exposure.cc
```diff
@@ -417,7 +384,7 @@ PTR(ExposureTable) ExposureTable::make(Schema const & schema) {
             "Schema for Exposure must contain at least the keys defined by makeMinimalSchema()."
         );
     }
-    return std::make_shared<ExposureTableImpl>(schema);
+    return std::shared_ptr<ExposureTable>(new ExposureTable(schema));
```
r-owen: Why this change?

src/table/Exposure.cc
```diff
@@ -448,6 +415,14 @@ ExposureTable::makeFitsWriter(fits::Fits * fitsfile, PTR(io::OutputArchive) arch
     return std::make_shared<ExposureFitsWriter>(fitsfile, archive, flags);
 }
 
+std::shared_ptr<BaseTable> ExposureTable::_clone() const {
+    return std::shared_ptr<ExposureTable>(new ExposureTable(*this));
```
r-owen: Why not `std::make_shared`?

python/lsst/afw/table/exposure/exposure.cc
TallJimbo: They are not imported in Python, and I believe they cannot be unless it's in `afw/__init__.py`, which I'm not sure we want to modify.  This is a defect of our poor in-afw package organization.

python/lsst/afw/table/schema/schema.cc
TallJimbo: This is just a comment heading for the code that follows.

python/lsst/afw/table/schema/schema.cc
TallJimbo: Afraid not.  C++ interface here is `int`.

include/lsst/afw/table/Schema.h
TallJimbo: The kind we need here - a conversion invoked by assignment - cannot be replicated in Python, regardless of the wrapper tool used.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
kfindeisen: Why not?

python/lsst/afw/table/match/match.cc
```diff
@@ -35,93 +34,80 @@ using namespace pybind11::literals;
...
     // Free Functions
     mod.def("unpackMatches", &unpackMatches<Catalog1, Catalog2>, "matches"_a, "cat1"_a, "cat2"_a);
-
     mod.def("matchRaDec",
             (MatchList (*)(Catalog1 const &, Catalog2 const &, geom::Angle, MatchControl const &))
              matchRaDec<Catalog1, Catalog2>, "cat1"_a, "cat2"_a, "radius"_a, "mc"_a=MatchControl());
-    // The following is deprecated; consider changing the code instead of wrapping it:
-    // mod.def("matchRaDec",
-    //         (MatchList (*)(Catalog1 const &, Catalog2 const &, geom::Angle, bool))
-    //          matchRaDec<Catalog1, Catalog2>, "cat1"_a, "cat2"_a, "radius"_a, "closest"_a);
```
kfindeisen: Consider leaving this comment in until we can properly  deprecate the method -- judging from how often the old forms were used in the stack, they were quire popular.

python/lsst/afw/table/match/match.cc
```diff
@@ -35,93 +34,80 @@ using namespace pybind11::literals;
...
 template <typename Catalog>
 void declareMatch1(py::module &mod) {
     typedef std::vector<Match<typename Catalog::Record, typename Catalog::Record>> MatchList;
     mod.def("matchRaDec",
             (MatchList (*)(Catalog const &, geom::Angle, MatchControl const &))
              matchRaDec<Catalog>, "cat"_a, "radius"_a, "mc"_a=MatchControl());
-    // The following is deprecated; consider changing the code instead of wrapping it:
-    // mod.def("matchRaDec",
-    //         (MatchList (*)(Catalog const &, geom::Angle, bool))
-    //          &matchRaDec<Catalog1>, "cat"_a, "radius"_a, "symmetric"_a);
```
kfindeisen: Same

python/lsst/afw/table/match/match.cc
```diff
@@ -35,93 +34,80 @@ using namespace pybind11::literals;
...
-    mod.def("matchXy",
-            (SourceMatchVector (*)(SourceCatalog const &, double, MatchControl const &))
-             matchXy, "cat"_a, "radius"_a, "mc"_a=MatchControl());
-    // The following are deprecated; consider changing the code instead of wrapping them:
-    // mod.def("matchXy",
-    //         (SourceMatchVector (*)(SourceCatalog const &, SourceCatalog const &, double, bool))
-    //          matchXy, "cat1"_a, "cat2"_a, "radius"_a, "closest"_a);
-    // mod.def("matchXy",
-    //         (SourceMatchVector (*)(SourceCatalog const &, double, bool))
-    //          &matchXy, "cat"_a, "radius"_a, "symmetric"_a);
```
kfindeisen: Same

include/lsst/afw/detection/Peak.h
```diff
@@ -173,6 +175,10 @@ class PeakTable : public afw::table::BaseTable {
 
     PeakTable(PeakTable const & other);
 
+    std::shared_ptr<afw::table::BaseTable> _clone() const override;
+
+    std::shared_ptr<afw::table::BaseRecord> _makeRecord() override;
```
TallJimbo: Hmm.  I don't disagree in general, but that creates a "what to call this" problem for this particular pattern, in which a nonvirtual public method delegates to a protected virtual of (effectively) the same name.  In any case, that's out of scope here - all I'm changing here is the addition of `override` and spelling out `shared_ptr` instead of using the deprecated `PTR` macros.

src/table/Exposure.cc
pschella: Why not `std::make_shared<ExposureTable>(schema);`?

include/lsst/afw/detection/Peak.h
TallJimbo: @pschella's new guidelines for using `override` say to not use `virtual` as well.  I don't personally have a strong opinion.

src/table/Exposure.cc
pschella: Same here.

src/table/Exposure.cc
```diff
@@ -448,6 +415,14 @@ ExposureTable::makeFitsWriter(fits::Fits * fitsfile, PTR(io::OutputArchive) arch
...
+}
+
+std::shared_ptr<BaseRecord> ExposureTable::_makeRecord() {
+    return std::shared_ptr<ExposureRecord>(new ExposureRecord(getSelf<ExposureTable>()));
```
pschella: and here

src/detection/Peak.cc
```diff
@@ -185,9 +141,7 @@ PTR(PeakTable) PeakTable::make(
         }
     }
     // No match: we create a new table and put it in the cache
-    PTR(PeakTable) newTable = std::make_shared<PeakTableImpl>(
-        schema, afw::table::IdFactory::makeSimple()
-    );
+    std::shared_ptr<PeakTable> newTable(new PeakTable(schema, afw::table::IdFactory::makeSimple()));
```
TallJimbo: The problem is how to declare it as a friend so it can access a private/protected constructor.  I don't think C++11/14 has an answer to that - I think the exact signature you need to friend is not mandated by the standard.  I hope C++17 has a solution, but I haven't been paying attention besides seeing that important people consider it a problem.

src/table/Source.cc
```diff
@@ -504,6 +460,16 @@ PTR(io::FitsWriter) SourceTable::makeFitsWriter(fits::Fits * fitsfile, int flags
 }
 
 
+std::shared_ptr<BaseTable> SourceTable::_clone() const {
+    return std::shared_ptr<SourceTable>(new SourceTable(*this));
+}
+
+std::shared_ptr<BaseRecord> SourceTable::_makeRecord() {
+    std::shared_ptr<SourceRecord> record(new SourceRecord(getSelf<SourceTable>()));
```
pschella: I assume this mismatch is deliberate (factory).

include/lsst/afw/table/Schema.h
kfindeisen: Then you might want to clarify in the comment that the issue is supporting assignments rather than e.g. passing to a method.

include/lsst/afw/table/FieldBase.h
TallJimbo: I don't think you can use `delete` on methods the compiler doesn't synthesize.

src/detection/Peak.cc
kfindeisen: Sorry, can you clarify *what* is being friended? I don't see how the use of `make_shared` puts any restrictions on something like that.

include/lsst/afw/table/Flag.h
TallJimbo: It's essentially expediency.  I tried removing it, things I didn't want to mess with now broke, and so I put it back int.

python/lsst/afw/table/schema/schema.cc
TallJimbo: It's for element unpacking: `key, field = item`.

I agree it's pretty dangerous (like any unpacking), since you could do `field, key = item` and get something incorrect with no error (yet).  But this was in the old interface and I think we don't want to change it now.

include/lsst/afw/table/FieldBase.h
kfindeisen: Yes, you can -- it's a useful trick for preventing misuse of implicit conversions, for example. See the Effective Modern C++ Item on `delete` for specific examples.

src/detection/Peak.cc
TallJimbo: `make_shared` needs to call a class constructor, forwarding the arguments you pass to it.  It that constructor is private or protected, `make_shared` can't call it.

I avoided that in the past here by making a hidden subclass (in an anoynmous namespace) with a public constructor that delegated to the protected one on its base class. But that hidden subclass confused pybind11's automatic downcasting abiliity, because when it used `typeid` on an instance, it doesn't return something in pybind11's registry of known types to try to cast to.

python/lsst/afw/table/schema/schema.cc
kfindeisen: Note that [DM-9140](https://jira.lsstcorp.org/browse/DM-9140) restores there being both `__str__` and `__repr__` for this class.

python/lsst/afw/table/schemaMapper/schemaMapper.cc
TallJimbo: I don't understand.

python/lsst/afw/table/source/source.cc
TallJimbo: I didn't think we had a convention for this, but I believe we did confirm that it doesn't matter in terms of code correctness (the `PYBIND11_PLUGIN` macro's `extern "C"` negates any namespaces it is defined in).

include/lsst/afw/table/Schema.h
TallJimbo: Just because the other overloads predate our use of C++11, and I didn't want to change too much.

include/lsst/afw/table/Schema.h
TallJimbo: I'll go with `findAndApply`.

src/table/Exposure.cc
TallJimbo: The constructor is protected, and I do not know how to make `make_shared` a friend (see also conversation with @kfindeisen elsewhere on this PR).

src/table/Source.cc
TallJimbo: Correct.

python/lsst/afw/table/base/baseContinued.py
```diff
@@ -0,0 +1,292 @@
...
+            )
+        return cls(columns, meta=meta, copy=False)
+
+    def __getattr__(self, name):
+        # Catalog forwards unknown method calls to its table and column view
+        # for convenience.  (Feature requested by RHL; complaints about magic
+        # should be directed to him.)
+        if name == "_columns":
+            self._columns = None
+            return None
```
TallJimbo: If `__getattr__` is being called at all, it means `_columns` did not already exist as a regular attribute, so we initialize it here.  Will comment to that effect.

python/lsst/afw/table/exposure/exposure.cc
```diff
@@ -54,41 +53,29 @@ namespace {
...
+        "contains",
+        (bool (ExposureRecord::*)(coord::Coord const &, bool) const) &ExposureRecord::contains,
+        "coord"_a, "includeValidPolygon"_a=false
+    );
+    cls.def(
+        "contains",
+        (bool (ExposureRecord::*)(geom::Point2D const &, image::Wcs const &, bool) const)
+            &ExposureRecord::contains,
+        "point"_a, "wcs"_a, "includeValidPolygon"_a=false
+    );
```
TallJimbo: Can't do that here, because `contains` takes multiple arguments.

python/lsst/afw/table/idFactory.cc
```diff
@@ -31,23 +30,21 @@ using namespace py::literals;
...
 
-PYBIND11_PLUGIN(_idFactory) {
-    py::module mod("_idFactory", "Python wrapper for afw _idFactory library");
-
-    py::class_<IdFactory, std::shared_ptr<IdFactory>> clsIdFactory(mod, "IdFactory");
+using PyIdFactory = py::class_<IdFactory, std::shared_ptr<IdFactory>>;
 
-    /* Operators */
-    clsIdFactory.def("__call__", &IdFactory::operator());
+PYBIND11_PLUGIN(idFactory) {
```
TallJimbo: I didn't think we had a convention, and I'm pretty sure it doesn't matter in terms of code correctness.

python/lsst/afw/table/schema/schema.cc
TallJimbo: Unfortunately, that'd be an API change I don't want to make here.

python/lsst/afw/table/schema/schemaContinued.py
TallJimbo: Hmm.  I agree it is a bit easier to read, but I don't think relative imports stop being a good idea within a package just because the number of levels increases.

python/lsst/afw/table/schema/schemaContinued.py
TallJimbo: Yup, this is pre-existing.

src/table/Exposure.cc
TallJimbo: The `ExposureTable` constructor is protected, and hence `make_shared` can't access it (and I don't know how to friend it).

include/lsst/afw/table/python/catalog.h
TallJimbo: IMO, this is definitely better, but I may be a weird bird.  In any case, I expect it to be removed in the future.

include/lsst/afw/detection/Peak.h
r-owen: I didn't know about that. I've been using both for clarity. But I guess I can live with it.

include/lsst/afw/table/python/functorKey.h
TallJimbo: It's just been removed entirely - the interface it defines is meaningless in Python.  From the python in `aggregates.cc` and `arrays.cc`:
```
\
// We don't expose FunctorKey base classes to Python, since they're just used to
// define a CRTP interface in C++ and in Python that's just duck-typing.
```

include/lsst/afw/table/Exposure.h
```diff
@@ -126,11 +126,14 @@ class ExposureRecord : public BaseRecord {
...
 protected:
 
-    ExposureRecord(PTR(ExposureTable) const & table);
+    explicit ExposureRecord(PTR(ExposureTable) const & table);
 
     virtual void _assign(BaseRecord const & other);
 
 private:
+
+    friend class ExposureTable;
```
r-owen: Where should we put them? I've tended to put them right after `class`

include/lsst/afw/table/python/sortedCatalog.h
TallJimbo: Grr.  I didn't think the recommendations had a preference here, and I (obviously) prefer this too.

python/lsst/afw/table/source/source.cc
r-owen: Given it works either way I'd be happier always putting it in the anonymous namespace. The advantage is it all wrapper code goes in the same namespace, making it easier to read.

include/lsst/afw/table/FieldBase.h
TallJimbo: Ah, neat!  I don't want to try that now, but definitely good to know.

include/lsst/afw/table/python/functorKey.h
kfindeisen: Might want to amend that comment -- I read it and thought you meant "base classes of `FunctorKey`", not "base classes, aka `FunctorKey`".

python/lsst/afw/table/base/baseContinued.py
r-owen: But I thought we should initialize it to the correct camera view and return that?

python/lsst/afw/table/schema/schemaContinued.py
r-owen: I'll push back gently. I prefer to keep relative imports within one subpackage. But it's just one opinion.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
r-owen: I think he means you should add static (and I don't agree so I ignored it)

include/lsst/afw/table/Exposure.h
TallJimbo: I don't think we have a policy.  I've put the ones here in `private` because that's where most of them were already.  I don't have a strong opinion.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
TallJimbo: But I *did* add static already!

PR #188, opened 2017-03-06
DM-9182: pybind11 cleanup for afw.image and afw.geom

TallJimbo:
I'm not completely done making changes here yet (want to bring over some SpanSets changes and ellipses and polygons), but I believe what I have done is ready to be looked at.

Will be lots of squashing later, of course.

### DIFF
--- a/python/lsst/afw/image/exposureInfo.cc
+++ b/python/lsst/afw/image/exposureInfo.cc
@@ -77,6 +77,7 @@ PYBIND11_PLUGIN(exposureInfo) {
         "coaddInputs"_a = std::shared_ptr<CoaddInputs>(),
         "apCorrMap"_a = std::shared_ptr<ApCorrMap>(),
         "visitInfo"_a = std::shared_ptr<VisitInfo const>());
+    cls.def(py::init<>());
     cls.def(py::init<ExposureInfo>(), "other"_a);
     cls.def(py::init<ExposureInfo, bool>(), "other"_a, "copyMetadata"_a);
 


python/lsst/afw/geom/angle/angle.cc
```diff
@@ -0,0 +1,137 @@
...
+
+    /* AngleUnit */
+
+    PyAngleUnit clsAngleUnit(mod, "AngleUnit");
+
+    clsAngleUnit.def("__eq__", [](AngleUnit const& self, AngleUnit const& other) { return self == other; },
+                     py::is_operator());
+    clsAngleUnit.def("__ne__", [](AngleUnit const& self, AngleUnit const& other) { return !(self == other); },
+                     py::is_operator());
+    clsAngleUnit.def("_mul", [](AngleUnit const& self, double other) { return other * self; },
```
r-owen: A brief comment explaining these would be helpful. I see they are used by a python continuation to implement `__mul__` and `__rmul__`, but why is that needed (why can't the code be written here)?

python/lsst/afw/geom/box.cc
```diff
@@ -147,8 +191,37 @@ PYBIND11_PLUGIN(_box) {
...
+        }
+    );
+    clsBox2D.def(
+        "__str__", [](Box2D const & self) {
+            return py::str("(minimum={}, maximum={})").format(
+                py::str(py::cast(self.getMin())),
+                py::str(py::cast(self.getMax()))
+            );
+        }
+    );
```
kfindeisen: Why are these methods so different from the deleted `.py` versions?

python/lsst/afw/geom/angle/angleContinued.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@continueClass
+class Angle:
+
+    def __abs__(self):
+        return abs(self.asRadians())*radians
+
+    def __reduce__(self):
+        return (Angle, (self.asRadians(),))
```
r-owen: Can't these two methods be in the C++?

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+
+template <typename Derived, typename T, int N>
+void declareCoordinateBase(py::module & mod, std::string const & suffix) {
+    static std::string const name = "CoordinateBase" + suffix;
+    PyCoordinateBase<Derived,T,N> cls(mod, name.c_str());
+
+    /* Operators */
+    cls.def("__getitem__", [](CoordinateBase<Derived,T,N> &c, int i) -> T {
+        if (i < 0 || i >= N) {
+            throw py::index_error{};
```
kfindeisen: Should we be using this instead of the `py::error_already_set` convention you used at https://github.com/lsst/afw/pull/187/files#diff-e11406535f01337997f15de456b1ae3bR176?

python/lsst/afw/geom/box.cc
```diff
@@ -53,35 +62,33 @@ PYBIND11_PLUGIN(_box) {
...
+    clsBox2I.def("getEnd", &Box2I::getEnd);
+    clsBox2I.def("getEndX", &Box2I::getEndX);
+    clsBox2I.def("getEndY", &Box2I::getEndY);
+    clsBox2I.def("getDimensions", &Box2I::getDimensions);
+    clsBox2I.def("getWidth", &Box2I::getWidth);
+    clsBox2I.def("getHeight", &Box2I::getHeight);
+    clsBox2I.def("getArea", &Box2I::getArea);
+    clsBox2I.def("isEmpty", &Box2I::isEmpty);
+    clsBox2I.def("contains", (bool (Box2I::*)(Point2I const &) const) &Box2I::contains);
+    clsBox2I.def("contains", (bool (Box2I::*)(Box2I const &) const) &Box2I::contains);
```
r-owen: Please add `__contains__`

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+
+template <int N>
+void declareCoordinateExpr(py::module & mod, std::string const & suffix) {
+    static std::string const name = "CoordinateExpr" + suffix;
+    declareCoordinateBase<CoordinateExpr<N>,bool,N>(mod, name);
+
+    PyCoordinateExpr<N> clsCoordinateExpr(mod, name.c_str());
+
+    /* Constructors */
+    clsCoordinateExpr.def(py::init<bool>(), py::arg("val")=false);
```
kfindeisen: `py::arg` is discouraged

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+
+template <typename T, int N>
+void declareExtentBase(py::module & mod, std::string const & suffix) {
+    declareCoordinateBase<Extent<T,N>,T,N>(mod, "Extent" + suffix);
+    static std::string const name = "ExtentBase" + suffix;
+
+    PyExtentBase<T,N> cls(mod, name.c_str());
+
+    /* Operators */
+    cls.def("eq", [](ExtentBase<T, N> const &self, Extent<T, N> other) { return self.eq(other); });
```
r-owen: A brief code comment explaining why these are not spelled `__eq__` etc. would be appreciated. Readers will wonder.

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+    clsI.def("__ifloordiv__", [](Extent<int, N> &self, int other) -> Extent<int, N> & {
+        self = floor(self / static_cast<double>(other));
+        return self;
+    });
+
+    clsI.def("__idiv__", [](Extent<int, N> &self, int other) -> Extent<int, N> & {
+        self = floor(self / static_cast<double>(other));
+        return self;
+    });
+    clsD.def("__idiv__", [](Extent<double, N> &self, double other) -> Extent<double, N> & {
```
r-owen: Please get rid of these non-true-div variants

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+
+
+template <typename T, int N>
+void declarePointBase(py::module &mod, std::string const & suffix) {
+    declareCoordinateBase<Point<T,N>,T,N>(mod, "Point" + suffix);
+    static std::string const name = "PointBase" + suffix;
+    PyPointBase<T,N> cls(mod, name.c_str());
+
+    /* Operators */
+    cls.def("eq", [](PointBase<T, N> const & self, Point<T, N> const & other) { return self.eq(other); });
```
r-owen: Again...a brief code comment explaining the decision to not spell these `__eq__`, etc. would be appreciated.

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -0,0 +1,509 @@
...
+    });
+    clsD.def("__itruediv__", [](Extent<double, N> &self, double other) -> Extent<double, N> & {
+        self /= other;
+        return self;
+    });
+
+    clsI.def("__iadd__", [](Extent<int, N> &self, Extent<int, N> const &other) -> Extent<int, N> & {
+        self += other;
+        return self;
+    });
```
kfindeisen: This is giving me a feeling of deja vu, but what's our convention on using lambdas vs. casted method pointers?

python/lsst/afw/geom/span.cc
```diff
@@ -1,96 +1,98 @@
...
+    SpanIterator(const Span &s) : _it{s.begin()}, _end{s.end()} {};
+    Point2I next() {
+        if (_it == _end) {
+            throw py::stop_iteration();
+        }
+        return *_it++;
+    };
+private:
+    Span::Iterator _it;
+    Span::Iterator _end;
```
r-owen: I had not seen this C++ version of Python iterators before. Fascinating. I hope we can make this a utility function or class for wider use (not in this ticket).

python/lsst/afw/geom/span.cc
```diff
@@ -1,96 +1,98 @@
...
-
-    py::class_<SpanIterator> clsSpanIterator(mod, "SpanIterator");
-
-    /* Operators */
-    clsSpanIterator.def("__iter__", [](SpanIterator &it) -> SpanIterator& { return it; });
-    clsSpanIterator.def("__next__", &SpanIterator::next);
+static void declareSpanIterator(py::module & mod) {
+    py::class_<SpanIterator> cls(mod, "SpanIterator");
+    cls.def("__iter__", [](SpanIterator &it) -> SpanIterator& { return it; });
+    cls.def("__next__", &SpanIterator::next);
```
r-owen: Is it practical to add `"__len__"`?

python/lsst/afw/image/apCorrMap/apCorrMap.cc
```diff
@@ -0,0 +1,89 @@
...
+    table::io::python::declarePersistableFacade<ApCorrMap>(mod, "ApCorrMap");
+
+    PyApCorrMap cls(mod, "ApCorrMap");
+
+    /* Constructors */
+    cls.def(py::init<>());
+
+    /* Operators */
+    cls.def("__imul__", &ApCorrMap::operator*=);
+    cls.def("__idiv__", &ApCorrMap::operator/=);
```
r-owen: Please delete and consider a global search and delete for wrappers of `"__div__"` and `"__idiv__"`

python/lsst/afw/image/calib.cc
```diff
@@ -75,9 +78,9 @@ PYBIND11_PLUGIN(_calib) {
     /* Operators */
     cls.def("__eq__", &Calib::operator==, py::is_operator());
     cls.def("__ne__", &Calib::operator!=, py::is_operator());
-    cls.def("__imul__", &Calib::operator*=, py::is_operator());
-    cls.def("__itruediv__", &Calib::operator/=, py::is_operator());
-    cls.def("__idiv__", &Calib::operator/=, py::is_operator());
+    cls.def("__imul__", &Calib::operator*=);
+    cls.def("__itruediv__", &Calib::operator/=);
+    cls.def("__idiv__", &Calib::operator/=);
```
r-owen: Delete

python/lsst/afw/geom/separableXYTransform.cc
```diff
@@ -20,29 +20,42 @@
...
 
-PYBIND11_PLUGIN(_separableXYTransform) {
-    py::module mod("_separableXYTransform", "Python wrapper for afw _separableXYTransform library");
+using PySeparableXYTransform =
+    py::class_<SeparableXYTransform, std::shared_ptr<SeparableXYTransform>, XYTransform>;
 
-    py::class_<SeparableXYTransform, std::shared_ptr<SeparableXYTransform>, XYTransform> clsSeparableXYTransform(mod, "SeparableXYTransform");
+PYBIND11_PLUGIN(separableXYTransform) {
+    py::module mod("separableXYTransform");
+ py::module::import("lsst.afw.geom.functor");
```
kfindeisen: Messy indentation.

python/lsst/afw/image/image/image.cc
```diff
@@ -188,28 +185,34 @@ py::class_<Image<PixelT>,
...
+    cls.def("get0", [](Image<PixelT> &img, int x, int y) {
+        return img.get0(x, y, lsst::afw::image::CheckIndices(true));
+    });
 
     return cls;
 }
 
 template <typename PixelT>
-void declareDecoratedImage(py::module & mod, std::string const & suffix) {
-    using Class = DecoratedImage<PixelT>;
```
r-owen: Why remove this?

python/lsst/afw/image/imageSlice.cc
```diff
@@ -1,39 +1,37 @@
...
+#include "pybind11/pybind11.h"
 
 #include "lsst/afw/image/ImageSlice.h"
 
 namespace py = pybind11;
 
 using namespace py::literals;
 
-using namespace lsst::afw::image;
+namespace lsst { namespace afw { namespace image { namespace {
```
r-owen: Perhaps too picky but...do we even allow putting all of these on the same line? clang-format will certainly change it.

python/lsst/afw/image/maskedImage/maskedImage.cc
```diff
@@ -0,0 +1,227 @@
...
+            (MI& (MI::*)(math::Function2<double> const &)) &MI::operator-=);
+    cls.def("scaledMinus", &MI::scaledMinus);
+    cls.def("__imul__", (MI& (MI::*)(ImagePixelT const)) &MI::operator*=);
+    cls.def("__imul__", (MI& (MI::*)(MI const &)) &MI::operator*=);
+    cls.def("__imul__",
+            (MI& (MI::*)(Image<ImagePixelT> const &)) &MI::operator*=);
+    cls.def("scaledMultiplies", &MI::scaledMultiplies);
+    cls.def("__idiv__", (MI& (MI::*)(ImagePixelT const)) &MI::operator/=);
+    cls.def("__idiv__", (MI& (MI::*)(MI const &)) &MI::operator/=);
+    cls.def("__idiv__", (MI& (MI::*)(Image<ImagePixelT> const &)) &MI::operator/=);
```
r-owen: delete

python/lsst/afw/image/pixel/pixel.cc
```diff
@@ -38,37 +37,35 @@ namespace image {
...
 
     declareSinglePixel<float>(mod, "SinglePixelF");
     declareSinglePixel<double>(mod, "SinglePixelD");
     declareSinglePixel<int>(mod, "SinglePixelI");
     declareSinglePixel<std::uint16_t>(mod, "SinglePixelU");
-    // TODO: fix or remove; presently this fail with complaint:
-    // call to constructor of 'SinglePixel<unsigned long long, unsigned short, float>' is ambiguous
-    // and it points out the "int" and "double" constructors
-    //declareSinglePixel<std::uint64_t>(mod, "SinglePixelL");
+    declareSinglePixel<std::uint64_t>(mod, "SinglePixelL");
```
r-owen: I'm glad this works now. Any idea why it was broken and now works?

python/lsst/afw/image/apCorrMap/apCorrMap.cc
```diff
@@ -0,0 +1,89 @@
...
+    cls.def("set", &ApCorrMap::set);
+    cls.def(
+        "items",
+        [](ApCorrMap const & self) {
+            return py::make_iterator(self.begin(), self.end());
+        },
+        py::keep_alive<0, 1>()
+    );
+    cls.def("keys", [](ApCorrMap const & self) {
+    });
```
kfindeisen: Why do we need an empty lambda here?

python/lsst/afw/geom/angle/angle.cc
kfindeisen: I think this would be a good question to defer to [DM-9630](https://jira.lsstcorp.org/browse/DM-9630) -- `Angle` has enough inconsistencies in its operators even before this ticket.

include/lsst/afw/image/Exposure.h
```diff
@@ -150,9 +150,9 @@ class Exposure : public lsst::daf::base::Persistable,
 
     // Get Members
     /// Return the MaskedImage
-    MaskedImageT getMaskedImage() { return _maskedImage; }
+    MaskedImageT & getMaskedImage() { return _maskedImage; }
     /// Return the MaskedImage
-    MaskedImageT getMaskedImage() const { return _maskedImage; }
+    MaskedImageT const & getMaskedImage() const { return _maskedImage; }
```
TallJimbo: Three shared pointers isn't all that small IMO, at least for an attribute we get all the time.  I believe we may it a by-value return originally to avoid dangling-pointer problems with Swig, so it seemed like the time to revert that.

include/lsst/afw/image/Exposure.h
```diff
@@ -150,9 +150,9 @@ class Exposure : public lsst::daf::base::Persistable,
...
 
     CONST_PTR(Wcs) getWcs() const { return _info->getWcs(); }
     PTR(Wcs) getWcs() { return _info->getWcs(); }
```
TallJimbo: If I'm not already touching the line, I'd like to keep changes minimal.

python/lsst/afw/geom/angle/angle.cc
TallJimbo: I had hoped to clean this up at one point too but ran out of ambition as the cleanup work dragged on.  I'm glad there's already another ticket for it.

python/lsst/afw/geom/span.cc
TallJimbo: There already is an upstream utility function for this: `pybind11::make_iterator`.  It does not work here, for reasons we do not understand (but may be related to the fact that `SpanPixelIterator` is not a legal `RandomAccessIterator` because it yields proxies).  This is [DM-7658](https://jira.lsstcorp.org/browse/DM-7658)

python/lsst/afw/image/apCorrMap/apCorrMap.cc
TallJimbo: Done here, and in other places in geom and image.  I'd rather not do it in the rest of afw, yet.

include/lsst/afw/image/Exposure.h
TallJimbo: The fact that it didn't return an internal handle before was a workaround for Swig wrapping this into a Python object that didn't hold a reference to the Exposure to keep it alive, yielding dangling pointers.

python/lsst/afw/geom/angle/angleContinued.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@continueClass
+class AngleUnit:
+
+    def __mul__(self, other):
+        if isinstance(other, (Angle, AngleUnit)):
+            raise NotImplementedError
+        return AngleUnit._mul(self, other)
```
TallJimbo: I'm really not sure; this is code I moved but did not write.  Since it seems you have ambitions to clean up Angle's maddeningly inconsistent set of operators, I'm hoping you'll be able to fix this later.

python/lsst/afw/geom/box.cc
```diff
@@ -31,19 +31,28 @@ using namespace py::literals;
...
 namespace geom {
+namespace {
+
+using PyBox2I = py::class_<Box2I, std::shared_ptr<Box2I>>;
+using PyBox2D = py::class_<Box2D, std::shared_ptr<Box2D>>;
+
+PYBIND11_PLUGIN(box) {
+    py::module mod("box");
+
+    py::object modCoordinates = py::module::import("lsst.afw.geom.coordinates");
```
TallJimbo: It could be.  But I'm not using any methods besides the ones on `object`, so I figured this was marginally better in the sense that it'd be harder to accidentally add new content to the `coordinates` module this way.

python/lsst/afw/geom/angle/angleContinued.py
kfindeisen: You're right, it was added by @pschella as part of a `meas_algorithms` fix: https://github.com/lsst/afw/commit/a004bb0a7c599b30566a6a088981c5daf629c266.

@pschella, do you remember why you added the `isinstance` checks?

python/lsst/afw/geom/box.cc
TallJimbo: When moving the code I realized it could very confusing to no know whether one was looking at the dimensions of the box or its maximum point, so I added the kwarg-like syntax (which is now happily `eval`-able in the case of `__repr__`).  The rest of the changes are just the difference between the Python and C++ interfaces for string formatting.

python/lsst/afw/geom/coordinates/coordinates.cc
TallJimbo: Probably.  Looks like I knew about this at one point and then forgot about it.  I'll go back and switch them later - if @r-owen merges DM-9715 first, I can just remove above of those explicit `IndexError` throws.

python/lsst/afw/geom/coordinates/coordinates.cc
TallJimbo: I think we landed on "either is okay", and I just haven't changed much in this file.

python/lsst/afw/image/apCorrMap/apCorrMap.cc
TallJimbo: Yikes.  That's a mistake.  Removed.  I think I must have been trying to move the implementation from C++ to Python and didn't realize I'd left this artifact when I gave up because it was shadowed by the pure Python.

PR #190, opened 2017-03-10
DM-9111 Add rotator information to HSC VisitInfo

jmeyers314:


include/lsst/afw/image/VisitInfo.h
```diff
@@ -160,9 +162,10 @@ class VisitInfo : public table::io::PersistableFacade<VisitInfo>, public table::
     /**
     * Get rotation angle at boresight at middle of exposure
     *
-    * Rotation angle is angle of coordinate system specified by boresightRotType with respect to detector.
-    * For example if boresightRotType is SKY then at a rotation angle of 0 North is along detector Y axis
-    * (never flipped) and East is along the detector X axis (flipped if necessary)
+    * Rotation angle is position angle of focal plane +Y with respect to the coordinate system
+    * specified by rotType.  For example, if rotType is SKY, then at a rotation angle of 0,
+    * North is along focal plane +Y and East is along +X or -X (flipped to match handedness if
+    * necessary).  At rotation angle of 90, +Y will be aligned with E.
```
r-owen: The above is nice and clean, but wrong for `MOUNT`. Please try again. I suppose you could back off to something like "The meaning of the angle is defined by `rotType`, which see.", but if you can find a compact way to say more without misleading for `MOUNT` that would be nice.

jmeyers314: Sure.  How about:

```
    /**
    * Get rotation angle at boresight at middle of exposure
    *
    * The meaning of rotation angle depends on rotType.  If rotType is SKY (HORIZON), then rotation
    * angle is position angle of focal plane +Y with respect to North (+Alt).  If rotType is MOUNT,
    * then rotation angle indicates the position sent to the rotator which depends on the specific
    * rotator.  Finally, if rotType is UNKNOWN, then the meaning of rotation angle is undefined.
    *
    * As an example, if rotType is SKY, rotation angle 0 means North is along focal plane +Y and
    * East is along +X or -X (flipped to match handedness if necessary).  At rotation angle of 90,
    * +Y is along East.
    */
```

r-owen: I worry about the duplication of information with rotType. That can lead to the information going out of date in one place or the other. My first inclination would be to rely on a Doxygen link to the rotation type information and put very little information here. A compromise that might suffice is to provide the most common example (other than UNKNOWN, which is all too common, alas, but self-explanatory) but not complete information, for instance:
```
* The meaning...depends on @ref RotType "rotType". For example, if `rotType` is SKY,
* the angle is the position angle of the focal plane +Y with respect to North.
```

jmeyers314: Good point.  I've switched the doc to your suggestion.

PR #192, opened 2017-03-17
DM-9595: Allow Transform to return its inverse

kfindeisen:
This PR implements `Transform::getInverse()`, adds a Python-only equality test for `Endpoint`, and updates the style of the Transform code to match the output of `clang-format` (mostly changing whitespace around `&`).

### DIFF
--- a/python/lsst/afw/geom/endpoint.cc
+++ b/python/lsst/afw/geom/endpoint.cc
@@ -83,7 +83,7 @@ template <typename PyClass>
 void addMakeFrame(PyClass& cls) {
     using Class = typename PyClass::type;  // C++ class associated with pybind11 wrapper class
     // return a deep copy so Python cannot modify the internal state
-    cls.def("makeFrame", [](Class const & self) {
+    cls.def("makeFrame", [](Class const& self) {
         auto frame = self.makeFrame();
         return frame->copy();
     });
--- a/python/lsst/afw/geom/endpoint.cc
+++ b/python/lsst/afw/geom/endpoint.cc
@@ -22,6 +22,7 @@
 #include <ostream>
 #include <memory>
 #include <string>
+#include <typeinfo>
 
 #include "pybind11/pybind11.h"
 #include "pybind11/stl.h"
@@ -88,6 +89,29 @@ void addMakeFrame(PyClass& cls) {
     });
 }
 
+// Comparison of different Endpoints useful in Python but counterproductive
+//     in C++: point2Endpoint == spherePoint should not compile instead of
+//     returning `false`. Therefore, implemented only on the Python side.
+// Two Endpoints are defined to be equal if and only if they have the same
+//     implementation type and the same number of dimensions
+template <typename SelfClass, typename OtherClass, typename PyClass>
+void addEquals(PyClass& cls) {
+    auto pyEquals = [](SelfClass const& self, OtherClass const& other) {
+        return self.getNAxes() == other.getNAxes() && typeid(self) == typeid(other);
+    };
+    cls.def("__eq__", pyEquals);
+    cls.def("__ne__",
+            [pyEquals](SelfClass const& self, OtherClass const& other) { return !pyEquals(self, other); });
+}
+
+template <typename SelfClass, typename PyClass>
+void addAllEquals(PyClass& cls) {
+    addEquals<SelfClass, GenericEndpoint>(cls);
+    addEquals<SelfClass, PointEndpoint<2>>(cls);
+    addEquals<SelfClass, PointEndpoint<3>>(cls);
+    addEquals<SelfClass, SpherePointEndpoint>(cls);
+}
+
 /*
 * Declare BaseVectorEndpoint<Point, Array>;
 * this is meant to be called by other `declare...` functions;
@@ -103,6 +127,7 @@ void declareBaseEndpoint(py::module& mod, std::string const& suffix) {
     addDataConverters(cls);
     addMakeFrame(cls);
     cls.def("normalizeFrame", &Class::normalizeFrame);
+    addAllEquals<Class>(cls);
 }
 
 // Declare BaseVectorEndpoint and all subclasses (the corresponding BaseEndpoint)
--- a/src/geom/Transform.cc
+++ b/src/geom/Transform.cc
@@ -55,8 +55,8 @@ Transform<FromEndpoint, ToEndpoint>::Transform(ast::FrameSet const &frameSet, bo
     // and normalize the frame set as a frame (i.e. normalize the frame "in situ").
     // The obvious alternative of normalizing a shallow copy of the frame does not work;
     // the frame is altered but not the associated mapping!
-    auto frameSetCopy = simplify ? std::dynamic_pointer_cast<ast::FrameSet>(frameSet.simplify())
-                                 : frameSet.copy();
+    auto frameSetCopy =
+            simplify ? std::dynamic_pointer_cast<ast::FrameSet>(frameSet.simplify()) : frameSet.copy();
 
     // Normalize the current frame by normalizing the frameset as a frame
     _toEndpoint.normalizeFrame(frameSetCopy);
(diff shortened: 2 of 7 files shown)


tests/test_transform.py
```diff
@@ -180,10 +180,94 @@ def checkTransform(self, fromName, toName):
...
+                         invFrameSet.tranForward(rawOutPoint), msg=msg)
+
+        # transformations of arrays of points
+        nPoints = 7  # arbitrary
+        rawInArray = makeRawArrayData(nPoints, nIn)
+        inArray = fromEndpoint.arrayFromData(rawInArray)
+        outArray = transform.tranForward(inArray)
+        rawOutArray = toEndpoint.dataFromArray(outArray)
+        # Assertions must work with both lists and numpy arrays
+        # TODO replace with assertFloatsAlmostEqual once DM-9707 is fixed
```
jdswinbank: My suggestion would be to avoid embedding TODOs like this in the code — we have literally hundreds of them scattered across the codebase, and nobody has time to regularly audit old code to see what actually needs to be done. Perhaps you could file a ticket on JIRA to do this work, and mark it as triggered by DM-9707? (Obviously there are _thousands_ of tickets in JIRA which need keeping track of, but at least that means we only need to track a single source.)

kfindeisen: That seems reasonable for DM-9707. Note that the similar comments for DM-9826 represent work to  be done _as part of_ DM-9826, they're just there to justify omissions to the reviewer.

tests/test_transform.py
```diff
@@ -180,10 +180,94 @@ def checkTransform(self, fromName, toName):
...
+                # Transform with both forward and inverse transforms
+                self.checkGetInverseWithMap(
+                    transformClass(makePolyMap(nIn, nOut)),
+                    nIn,
+                    "{}, Map={}, nIn={}, nOut={}".format(
+                        baseMsg, "Invertible", nIn, nOut))
+                # TODO: add more test cases once DM-9826 resolved
+
+    def checkGetInverseWithMap(self, transform, nIn, msg):
+        """Test Transform<fromName>To<toName>.getInverse with a specific mapping.
```
r-owen: What do you mean by "with a specific mapping"? A transform can be constructed with a mapping or a frame set, but the results is that it contains a frame set in either case.

To my mind the name suggests that the user will provide a mapping as an argument to this method (I am not suggesting you do that; the current test seems excellent).

Can you call this `checkGetInverse` and skip the `WithMap`?

python/lsst/afw/geom/endpoint.cc
```diff
@@ -82,12 +83,34 @@ template <typename PyClass>
...
     using Class = typename PyClass::type;  // C++ class associated with pybind11 wrapper class
     // return a deep copy so Python cannot modify the internal state
-    cls.def("makeFrame", [](Class const & self) {
+    cls.def("makeFrame", [](Class const& self) {
         auto frame = self.makeFrame();
         return frame->copy();
     });
 }
 
+// Endpoint comparison not needed(?) in C++
```
kfindeisen: Agreed, I've already decided that it definitely should not be present in C++ (it's redundant with the type system).

tests/test_transform.py
kfindeisen: I need to give it a distinct name from the method immediately above... `checkGetInverseWithFrameSet`?

tests/test_transform.py
kfindeisen: @r-owen does that also apply to your note about replacing `np.allclose`?

src/geom/Transform.cc
```diff
@@ -105,16 +105,22 @@ typename FromEndpoint::Array Transform<FromEndpoint, ToEndpoint>::tranInverse(
 }
 
 template <typename FromEndpoint, typename ToEndpoint>
+Transform<ToEndpoint, FromEndpoint> Transform<FromEndpoint, ToEndpoint>::getInverse() const {
+    std::shared_ptr<ast::Mapping> inverse = _frameSet->getInverse();
```
kfindeisen: Hmm... is there a way to unit test for the information you say is currently missing?

tests/test_transform.py
```diff
@@ -180,10 +180,94 @@ def checkTransform(self, fromName, toName):
...
+        baseMsg = "transformClass={}".format(transformClass.__name__)
+        for nIn in self.goodNaxes[fromName]:
+            for nOut in self.goodNaxes[toName]:
+                # Transform with both forward and inverse transforms
+                self.checkGetInverseWithMap(
+                    transformClass(makePolyMap(nIn, nOut)),
+                    nIn,
+                    "{}, Map={}, nIn={}, nOut={}".format(
+                        baseMsg, "Invertible", nIn, nOut))
+                # TODO: add more test cases once DM-9826 resolved
```
kfindeisen: I was trying to be better about organizing work, avoiding scope creep, etc., but fine. I will ask for a re-review of those changes, though.

tests/test_transform.py
r-owen: `np.allclose` should be replaced with `np.testing.assert_allclose`. Do you want to do that while working on the ticket or hand it back to me to do. I didn't know about the `np.testing` asserts when I wrote the code.

src/geom/Transform.cc
r-owen: Sure...create a FrameSet and add some frames. Then look for them after constructing the Transform and retrieving the FrameSet from it.

tests/test_transform.py
kfindeisen: I'll do it, and remove all mentions of DM-9707.

tests/test_transform.py
r-owen: If I am reading the code correctly:
- `checkGetInverse` creates all possible transforms for a given pair of endpoint names, creating each frame set or mapping as it sees fit.
- `checkGetInverseWithMap` checks one transform.

I'm asking for a less confusing name scheme. To be fair I'll propose one, but do as you like:
- Rename `checkGetInverseWithMap` to `checkGetInverse` and ditch old version of the latter.
- Make `checkTransform` call `checkGetInverseWithMap`

Or if that's too much work for `checkTransform` you could make a new method `testGetInverse` and have it build all valid transforms that you want to test, then call `checkTransform` from there.

I wonder if I should break `checkTransform` up a bit, e.g. I could imagine a function by that name which takes a valid transform as an argument and checks basics. Then `testTransform` would have to construct the transforms. I'm not sure where I would move the code that tries to construct invalid transforms -- perhaps a new `testX` method. Comments?

src/geom/Transform.cc
```diff
@@ -105,16 +106,28 @@ typename FromEndpoint::Array Transform<FromEndpoint, ToEndpoint>::tranInverse(
 }
 
 template <typename FromEndpoint, typename ToEndpoint>
+Transform<ToEndpoint, FromEndpoint> Transform<FromEndpoint, ToEndpoint>::getInverse() const {
+    auto inverse = std::dynamic_pointer_cast<ast::FrameSet>(_frameSet->getInverse());
+    if (!inverse) {
+        // std::bad_cast more appropriate, but doesn't aid in debugging
```
r-owen: I like the code but find the note confusing. This situation should never occur, so logic_error does seem the best choice.

python/lsst/afw/geom/endpoint.cc
```diff
@@ -82,12 +83,34 @@ template <typename PyClass>
...
     using Class = typename PyClass::type;  // C++ class associated with pybind11 wrapper class
     // return a deep copy so Python cannot modify the internal state
-    cls.def("makeFrame", [](Class const & self) {
+    cls.def("makeFrame", [](Class const& self) {
         auto frame = self.makeFrame();
         return frame->copy();
     });
 }
 
+// Endpoint comparison useful in Python but not C++
```
r-owen: I'm not convinced by this assertion. If it would not be a lot more work I recommend making this available to C++.

tests/test_endpoint.py
```diff
@@ -158,6 +158,13 @@ def checkEndpointBasics(self, endpoint, pointType, nAxes):
         pointDataRoundTrip = endpoint.dataFromPoint(point)
         assert_allclose(pointData, pointDataRoundTrip, err_msg=baseMsg)
 
+    def testEndpointEquals(self):
+        """Test Endpoint == Endpoint
+        """
+        for i1, point1 in enumerate(self.makeEndpoints()):
+            for i2, point2 in enumerate(self.makeEndpoints()):
+                self.assertEqual(i1 == i2, point1 == point2)
```
r-owen: Cute! But the message in case of an error will not be clear. I suggest formatting a useful error message to be passed as the `msg` argument to `self.assertEqual`. (This might be easier if you handled the `i1 == i2` case separately from the other case, but I'll leave that to you).

Also, please test `!=` as well as `==`.

tests/test_transform.py
```diff
@@ -400,11 +513,107 @@ def checkTransformFromFrameSet(self, fromName, toName):
 
                     self.checkTransformation(permTransform, mapping=polyMap, msg=msg)
 
+    def checkAllGetInverse(self, fromName, toName):
```
r-owen: A nicely written method, but what does the "All" in the name tell me? Why not "checkGetInverse"?

PR #195, opened 2017-03-18
implement PhotoCalib tickets/DM-9192

parejkoj:


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,369 @@
...
+     *
+     * @param[in]  counts       The source flux in ADU.
+     * @param[in]  countsSigma  The counts error (sigma).
+     * @param[in]  point        The point that flux is measured at (must be within the domain of the
+     *                          BoundedField of this PhotoCalib).
+     * @param[in]  throwOnNegativeFlux  Raise an exception when passed negative counts.
+     *
+     * @return     The flux in maggies and error (sigma).
+     */
+    std::pair<double, double> countsToMaggies(double counts, double countsSigma,
```
kfindeisen: Is there a "value with error" class? Pair is really not the best API for this.

I won't ask if you really need *eight* overloads...

parejkoj: Pair is basically a python tuple containing two things, right? Other suggestions are welcome.

I'd be happy to flatten the overloads, if someone has a suggestion of how. About half of these are already used somewhere in the stack, the others (e.g. `SourceRecord` or `SourceCatalog`) are ones we know we want to use.

kfindeisen: Pair is a bit generic, and it can be hard to remember what the first and second elements are after it's been passed around some. But the appropriateness/inappropriateness of tuples as return types is a controversial subject, so I won't belabor the point.

I'll have to think about the overloads; part of the problem is that apart from `Point` vs. average I don't see much of a pattern to them.

parejkoj: Maybe when I've got some more of the unittest written the use cases will be clearer?

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+
+#ifndef LSST_AFW_IMAGE_PHOTOCALIB_H
+#define LSST_AFW_IMAGE_PHOTOCALIB_H
+
+/**
+ * The photometric calibration of an exposure.
+ *
+ * A PhotoCalib is a BoundedField (a function with a specified domain) that converts between calibrated
+ * counts-on-chip (ADU) to flux and magnitude. It is defined in terms of "maggies", which are a linear
+ * unit defined in SDSS: http://www.sdss.org/dr12/algorithms/magnitudes/#nmgy
```
jdswinbank: maggies or nanomaggies?

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+        static PersistenceHelper const instance;
+        return instance;
+    }
+
+    PersistenceHelper(int tableVersion=PHOTOCALIB_TABLE_CURRENT_VERSION) :
+    schema(), countsMag0(), countsMag0Sigma(), field() {
+        countsMag0 = schema.addField<double>("countsMag0", "counts of a zero-magnitude object", "count");
+        countsMag0Sigma = schema.addField<double>("countsMag0Sigma", "1-sigma error on countsmag0", "count");
+        isConstant = schema.addField<table::Flag>("isConstant", "Is this spatially-constant?");
+        field = schema.addField<int>("field", "archive ID of the BoundedField object");
```
jdswinbank: Again per JIRA, you need to mark the schema as persistent.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+ *     sqrt((countsSigma/counts)^2 + (zeroPointSigma/zeroPoint)^2)*flux = fluxSigma (in maggies)
+ * @f]
+ * This implies that the conversions from counts and counts error to magnitude and magnitude error
+ * are as follows:
+ * @f[
+ *     -2.5 * log_10(counts / zeroPoint(x,y)) = magnitude
+ * @f]
+ * and
+ * @f[
+ *     2.5*log(10) * sqrt((countsSigma/counts)^2 + (zeroPointSigma/zeroPoint)^2) = magSigma
```
jdswinbank: (If it is not confused, then obviously this needs to be changed in the code below too.)

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+} // anonymous namespace
+
+// ------------------- Conversions to Maggies -------------------
+
+double PhotoCalib::countsToMaggies(double counts,
+                       const afw::geom::Point<double, 2> &point,
+                       bool throwOnNegativeFlux) const {
+    throwOnNegativeCounts(counts, throwOnNegativeFlux);
+    if (_isConstant)
+        return toMaggies(counts, _countsMag0);
```
parejkoj: (I did forget to run clang-format before submitting for review: I'll fix these manually, and also run clang-format when I'm done with your review comments.)

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+     * @param[in]  countsMag0       The constant counts/magnitude zero point (counts at magnitude 0).
+     * @param[in]  countsMag0Sigma  The error on the zero point.
+     * @param[in]  bbox             The bounding box on which this PhotoCalib is valid.
+     */
+    explicit PhotoCalib(double countsMag0, double countsMag0Sigma=0,
+                        afw::geom::Box2I const & bbox=afw::geom::Box2I()):
+        _countsMag0(countsMag0), _countsMag0Sigma(countsMag0Sigma), _isConstant(true) {
+            ndarray::Array<double, 2, 2> coeffs = ndarray::allocate(ndarray::makeVector(1, 1));
+            coeffs[0][0] = countsMag0;
+            _zeroPoint = std::make_shared<afw::math::ChebyshevBoundedField>(
```
parejkoj: It's about a factor of 2, in my testing. The time appears to be mostly independent of the complexity of the BoundedField, so it's likely due to the lookup of the centroids.

There's more to it than that, though: a constant PhotoCalib can be created without a bounding box, in which case an "empty" bbox gets created. Empty bboxes have some strange properties, which one would have to work around separately. By treating known-constant PhotoCalibs separately, we can avoid that issue, while keeping the empty bbox.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+
+    //@{
+    /**
+     * Convert counts in ADU to maggies.
+     *
+     * If passed point, use the exact calculation at that point, otherwise, use the mean scaling factor.
+     *
+     * @param[in]  counts  The source counts in ADU.
+     * @param[in]  point   The point that counts is measured at (must be within the domain of the
+     *                     BoundedField of this PhotoCalib).
```
parejkoj: Oh, good point. I don't actually do any bounds checking in any of this code, (perhaps naively) assuming the BoundedField will take care of it automatically. However, there's not any obvious bounds checking in `ChebyshevBoundedField` that I can see.

@TallJimbo : should `BoundedField`s do bounds checking (e.g. on evaluation), or does that have to happen in the code that uses them?

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+     * @param[in]  throwOnNegativeCounts  Raise an exception when passed negative counts.
```
parejkoj: To your second point: that's exactly the sort of issue that lead me to file RFC-322. I clearly missed a few of the `Flux`->`Counts` conversions when I was cleaning up the code. Fixed now, thanks!

As to the first comment: I don't disagree. Ideally, we could make it a context manager at the python level (`with PhotoCalib.throwOnNegativeFlux: doStuff()`), but I don't know if that's possible with this model. Maybe with a `PhotoCalibControl`? It looks like we don't actually use the current `Calib.setThrowOnNegativeFlux` much at all, so maybe we could just get rid of it?

include/lsst/afw/image/PhotoCalib.h
TallJimbo: From the `BoundedField` base class:
```
    /**
     *  Evaluate the field at the given point.
     *
     *  This is the only abstract method to be implemented by subclasses.
     *
     *  Subclasses should not provide bounds checking on the given position; this is the responsibility
     *  of the user, who can almost always do it more efficiently.
     */
    virtual double evaluate(geom::Point2D const & position) const = 0;
```

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+                             bool throwOnNegativeCounts=false) const;
+    //@}
+
+    /**
+     * Convert sourceRecord[countsField_counts] (ADU) at location
+     *             (sourceRecord.get('x'), sourceRecord.get('y')) (pixels) to maggies and maggie error.
+     *
+     * @param[in]  sourceRecord  The source record to get counts and position from.
+     * @param[in]  countsField   The counts field: Keys of the form "*_counts" and "*_countsSigma" must exist.
+     *                           For example: countsField = "PsfFlux" -> "PsfFlux_counts", "PsfFlux_countsSigma"
```
parejkoj: This definitely depends on the results of RFC-322. If that's rejected, there's no obvious place to put such values. Also, as noted in DM-10155, we might actually want to merge the maggies and magnitudes versions into a single method that sets both values in the catalog.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+    void countsToMaggies(const afw::table::SourceCatalog &sourceCatalog,
+                         const std::string &countsField,
+                         ndarray::Array<double,2,2> result,
+                         bool throwOnNegativeCounts=false) const;
+    //@}
+
+    /**
+     * Convert sourceCatalog[countsField_counts] (ADU) at locations
+     *             (sourceCatalog.get('x'), sourceCatalog.get('x')) (pixels) to maggies
+     *             and write the results back to sourceCatalog[outField_mag].
```
parejkoj: If we're iterating over the sourceCatalog already, might as well set the values as we compute them.

include/lsst/afw/math/BoundedField.h
```diff
@@ -175,8 +175,16 @@ class BoundedField : public table::io::PersistableFacade<BoundedField>, public t
...
 
+    /// BoundedFields (of the same sublcass) are equal if their bounding boxes and parameters are equal.
+    virtual bool operator==(BoundedField const& rhs) const = 0;
+    /// @copydoc operator==
+    bool operator!=(BoundedField const& rhs) const { return !(*this == rhs); };
+
     virtual ~BoundedField() {}
 
+    friend std::ostream & operator<<(std::ostream & os, BoundedField const & bf)
+    { return os << bf.getPersistenceName() << " on " << bf.getBBox(); }
```
parejkoj: Unfortunately, there's no other way that I could come up with to determine the "type" of the `BoundedField` that this PhotoCalib contains. Other suggestions are welcome.

TallJimbo: I think it would make sense to give `BoundedField` an `operator<<` with a polymorphic implementation that this could delegate to.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+    return (_countsMag0 == rhs._countsMag0 && _countsMag0Sigma == rhs._countsMag0Sigma &&
+        (*_zeroPoint) == *(rhs._zeroPoint));
+}
+
+double PhotoCalib::computeCountsMag0(std::shared_ptr<afw::math::BoundedField> zeroPoint) const
+{
+    return zeroPoint->mean();
+}
+
+std::ostream & operator<<(std::ostream & os, PhotoCalib const & photoCalib) {
```
parejkoj: I've reworked all of those to match the coding standard.

include/lsst/afw/image/PhotoCalib.h
parejkoj: I've defined PhotoCalib in terms of Maggies. I've clarified the docstring here to note that the SDSS link gives the definition in terms of nanomaggies.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+        if (_isConstant)
+            countsMag0 = _countsMag0;
+        else
+            countsMag0 = _zeroPoint->evaluate(rec.getCentroid());
+        maggies = toMaggies(counts, countsMag0);
+        (*iter)[0] = maggies;
+        (*iter)[1] = toMaggiesSigma(counts, countsSigma, countsMag0, _countsMag0Sigma, maggies);
+        iter++;
+    }
+    if (negatives > 0  && throwOnNegativeFlux) {
```
parejkoj: Per our discussion, I've completely removed the `throwOnNegative` functionality, thus mooting this question.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+        if (_isConstant)
+            countsMag0 = _countsMag0;
+        else
+            countsMag0 = _zeroPoint->evaluate(rec.getCentroid());
+        maggies = toMaggies(counts, countsMag0);
+        (*iter)[0] = toMagnitude(counts, countsMag0);
+        (*iter)[1] = toMagnitudeSigma(counts, countsSigma, countsMag0, _countsMag0Sigma, maggies);
+        iter++;
+    }
+    if (negatives > 0  && throwOnNegativeFlux) {
```
parejkoj: throw functionality removed, so moot.

include/lsst/afw/image/PhotoCalib.h
parejkoj: Thanks for checking this! I did indeed mis-translate that from my mathematica notebook. Good catch! Fixed in 3 places (test, docstring, code).

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+#include "lsst/afw/geom/Point.h"
+#include "lsst/afw/geom/Box.h"
+#include "lsst/afw/table/Source.h"
+#include "lsst/afw/table/io/Persistable.h"
+
+namespace lsst { namespace afw { namespace image {
+
+/// A value and its error.
+struct Quantity {
+    Quantity(double value, double sigma): value(value), sigma(sigma) {}
```
parejkoj: Whether or not we are currently consistent about it, ensuring that our internal code is consistent with our outward catalogs is a "good thing"(tm). I've replaced all "sigma" with "err" throughout PhotoCalib.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+#include "lsst/afw/math/ChebyshevBoundedField.h"
+#include "lsst/afw/geom/Point.h"
+#include "lsst/afw/geom/Box.h"
+#include "lsst/afw/table/Source.h"
+#include "lsst/afw/table/io/Persistable.h"
+
+namespace lsst { namespace afw { namespace image {
+
+/// A value and its error.
+struct Quantity {
```
parejkoj: This was @r-owen and @kfindeisen 's idea, and they both though it was something we should probably have more generically. I'm not sure that's in scope for this ticket, but I filed DM-10260 so it doesn't get forgotten.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -0,0 +1,426 @@
...
+     */
+    PhotoCalib(double countsMag0, double countsMag0Sigma,
+               std::shared_ptr<afw::math::BoundedField> zeroPoint,
+               bool isConstant):
+        _zeroPoint(zeroPoint), _countsMag0(countsMag0), _countsMag0Sigma(countsMag0Sigma),
+        _isConstant(isConstant) {}
+
+    //@{
+    /**
+     * Convert counts in ADU to maggies.
```
parejkoj: Here are the existing variations. I'm happy to entertain suggestions as to which could be removed, but they all seem to have obvious utility to me:

1. counts, position independent
2. counts, position dependent
3. counts+error, position independent
4. counts+error, position dependent
5. source record (counts+error, position dependent)
6. source catalog, return new array (counts+error, position dependent)
7. source catalog, pass in pre-allocated array (counts+error, position dependent)
8. (not yet implemented) source catalog, put answer into catalog (counts+error, position dependent)

I could unify 1+2 and 3+4 if we had an "empty" Point object, but the default Point constructor is `(0,0)`, which is a valid position (when we want the "mean" PhotoCalib for an "empty" point).

5. is probably the least obviously useful of those in a pipeline, but it saves a user doing analysis on a handful of things from having to extract the values from the record themselves.

7. could go away if we're willing to always return a new array, but I get the impression that common C++ behavior is to allow the user to pre-allocate.

include/lsst/afw/image/PhotoCalib.h
parejkoj: As per my above comment, I've removed all `throwOnNegative` stuff.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+
+    PersistenceHelper(int tableVersion=PHOTOCALIB_TABLE_CURRENT_VERSION) :
+    schema(), countsMag0(), countsMag0Sigma(), field() {
+        countsMag0 = schema.addField<double>("countsMag0", "counts of a zero-magnitude object", "count");
+        countsMag0Sigma = schema.addField<double>("countsMag0Sigma", "1-sigma error on countsmag0", "count");
+        isConstant = schema.addField<table::Flag>("isConstant", "Is this spatially-constant?");
+        field = schema.addField<int>("field", "archive ID of the BoundedField object");
+    }
+
+    PersistenceHelper(table::Schema const & s) :
```
parejkoj: I mostly stole all of this code from @r-owen 's VisitInfo implementation. I don't know what best practices for persistence are, at all. Do we have any guide on the topic?

include/lsst/afw/image/PhotoCalib.h
parejkoj: Ah, thanks for that note! So, now the question is, should we throw in that case (and not throw for constant `PhotoCalib`s?), or should we just do the calculation whatever the case?

I had originally written that comment assuming the `BoundedField` would manage the testing of the bbox, but since that isn't true, I'm not sure whether we actually should when evaluating at the position(s). Only `computeScalingTo` would actually test the bbox then...

src/image/PhotoCalib.cc
jdswinbank: The most useful document I found was https://github.com/lsst/afw/blob/master/doc/tablePersistence.dox, which (for reasons I've not had chance to investigate) doesn't actually seem to be being published by Doxygen. It's by no means a complete reference though.

kfindeisen: @jdswinbank it's published at https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/afw_table_persistence.html. Maybe a cross-reference needs to be added from the persistence classes?

jdswinbank: DM-10265.

include/lsst/afw/image/PhotoCalib.h
parejkoj: And as I think about this a bit more, the thing we'd probably most want to protect against (using one `Exposure`'s `PhotoCalib` to convert the fluxes of points from a different `Exposure`) won't be protected by checking a bbox, since both `PhotoCalib`s will have the "same" bbox: `(0,4000)`.

include/lsst/afw/image/PhotoCalib.h
kfindeisen: I already deleted my notes on the subject because I thought you weren't interested, but going from memory:

The danger of having lots of convenience methods is that it not only makes more work for future class maintainers, it makes the API harder to learn by the user, who needs to carefully compare each method to figure out which one they need (case in point, it's impossible to have a constructive discussion about this without a list of the overloads in hand). A good rule of thumb is to only add a convenience overload if you expect *lots* of people to find it easier than doing without it, and to leave it out if only a handful will get a strong advantage.

I think the output parameter used in number 7 is more a holdover from C than something that's still justified in C++ (I assume the original motivation was that arrays need a lot more manual management than container objects do). BTW, what's the behavior if `result` doesn't have the right dimensions for the output? The documentation doesn't say.

I assume versions 6 and 7 are equivalent to just iterating over the catalog and calling 5 on each record? Then, per the rule of thumb above, they probably don't add enough extra functionality to be worth the human overhead. (Yes, this is the exact opposite of what you said above re: 5 being redundant with 4; consider it an alternative point of view.)

A similar argument would apply to version 8 -- would it be more useful to have an update-inplace version of 5, and let the user decide how many records they want to update?

parejkoj: 6 and 7 are there because having vectorized methods is extremely useful when working with catalogs, and because

```python
result = np.array(len(catalog))
for i,source in enumerate(catalog):
    result[i] = photoCalib.countsToMagnitude(source)
```
is going to be *much* slower than
```python
result = photoCalib.countsToMagnitude(catalog)
```
That said, I'm totally on board with removing 7, for exactly the reasons you suggest. I think @TallJimbo suggested it during the initial design discussion.

8\. isn't really "update in place", it's "put values in probably empty slots, in place". The immediate use case I see is in jointcal, where I'd run that version after fitting the new `PhotoCalib` to update the catalog with flux and magnitude for analysis by the verification system.

parejkoj: Separately, re:8, I'm now leaning towards `countsToMaggiesAndMagnitudes` that puts both of them in the catalog, since that's the use case I'm thinking of above for jointcal.

kfindeisen: Curses, foiled by Python. I'll concede we probably do need the catalog versions, then.

parejkoj: The reason I like keeping 5. is that I personally get annoyed every time I have to do something like `photoCalib.countsToMagnitude(source['PsfFlux_instFlux'], source['PsfFlux_instFluxErr'], source.getCentroid())`: why can't I just hand it my source? But I don't have a good sense for how often one would use the single-record version. I suspect it would be much more common when someone's working at a console/jupyter-notebook, which is harder to track.

I guess I'm not sure what the "helper function" for `countsToMaggiesAndMagnitudes` functionality should be: it's doing operations directly on the catalog. And it's the sort of thing I'd think we want after anything (ProcessCcd, jointcal) produces a new PhotoCalib for an exposure; I've just got it in the front of my mind for jointcal.

RobertLuptonTheGood: I think that the ability to treat rows in a catalog as objects is extremely useful, and we should keep the syntactic sugar that makes it easy.  Treating columns as numpy arrays is useful too, of course.

parejkoj:
I've renamed things to use "err" and also renamed `counts` to `instFlux`.

include/lsst/afw/image/PhotoCalib.h
jdswinbank: I don't have a strong opinion on this, but I think it'd be fine to just do the calculation (and note in the documentation that that's what's happening).

include/lsst/afw/image/PhotoCalib.h
parejkoj: After some extensive discussion offline, I've removed 5., but kept all the others because there's no good way to collapse them, and they all have utility.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+    // No moving
+    PersistenceHelper (PersistenceHelper&&) = delete;
+    PersistenceHelper& operator=(PersistenceHelper&&) = delete;
+
+    static PersistenceHelper const & get() {
+        static PersistenceHelper const instance;
+        return instance;
+    }
+
+    PersistenceHelper(int tableVersion=PHOTOCALIB_TABLE_CURRENT_VERSION) :
```
parejkoj: Table version has been removed after discussion with @TallJimbo who said that we typically only do that when it becomes necessary. That may not be optimal, but then I can also wait and see if we implement a more persistence versioning generic system in the future.

src/image/PhotoCalib.cc
parejkoj: I switched to using the `get()`, removed the schema-based constructor, and made the empty constructor private.

include/lsst/afw/math/BoundedField.h
parejkoj: Ok, I've delegated to a private `toString()` for the math part of the `BoundedField`, and jazzed it up a bit.

src/image/PhotoCalib.cc
```diff
@@ -0,0 +1,409 @@
...
+    countsToMaggiesArray(sourceCatalog, countsField, result, throwOnNegativeFlux);
+}
+
+void PhotoCalib::countsToMaggies(afw::table::SourceCatalog &sourceCatalog,
+                     const std::string &countsField,
+                     const std::string &outField,
+                     bool throwOnNegativeFlux) const
+{
+    throw LSST_EXCEPT(pex::exceptions::LogicError,
+                      "Will not be implemented until RFC-322 is implemented. See DM-10155");
```
parejkoj: RFC-322 is adopted, and I've reworded everything to match the `instFlux` name.

include/lsst/afw/image/PhotoCalib.h
parejkoj: Per discussion and after further thought, I've removed the comments about bounds checking from the docstrings of the methods that take a `Point`, and clarified the meaning of an "empty" `BBox` in the relevant constructor. If it turns out that we do want to do bounds checking on the `BoundedField`, we can add it in the future.

src/math/ChebyshevBoundedField.cc
```diff
@@ -427,15 +427,12 @@ PTR(BoundedField) ChebyshevBoundedField::operator*(double const scale) const {
...
-        auto& rhsCasted = dynamic_cast<ChebyshevBoundedField const&>(rhs);
-        bool coefficientsEqual = _coefficients.getShape() == rhsCasted._coefficients.getShape() &&
-                                  all(equal(_coefficients, rhsCasted._coefficients));
-        return (getBBox() == rhsCasted.getBBox()) && coefficientsEqual;
-    }
-    catch (std::bad_cast&) {
-        return false;
-    }
+    auto rhsCasted = dynamic_cast<ChebyshevBoundedField const *>(&rhs);
+    if (!rhsCasted) return false;
```
kfindeisen: Is the "efficiency" just sidestepping the exception handler, or is there something more subtle going on?

TallJimbo: It's just avoiding (possible) overhead from catch/throw.  I'm not sure if the compiler should be able to optimize any of that away in this case, but I'm pretty sure some major ones wouldn't, at least as of a couple of years ago.

include/lsst/afw/image/PhotoCalib.h
TallJimbo: FWIW, I found myself annoyed this week when I wanted to use (5) but found it wasn't present, and I do think we may want to add it back in at some point.  One explicit use case is that it would let you calibrate an array of fluxes stored in some other table-like object, like an Astropy table (which, being column-based instead of row-based is architecturally better suited than `SourceCatalog` for many analysis tasks).  But it's more that I think the overall calculus on interface clutter changes when the C++/Python border is involved: because it's so critical to perform any loops in C++ for performance, additional signatures are frequently much more than just a matter of convenience.  The result is that I think it's more appropriate to have many signatures in hybrid C++/Python projects, and the trick to mitigating the bad aspects of that is to design them so that the C++ overloads seem like a multitude of optional keyword arguments in Python.

All that said, in this case I could have just used (6), if it had been implemented to use the current naming conventions for flux fields instead of the envisioned future ones.  So for now, I'm just going to change that (on DM-10728), and we can consider adding (5) back in the future if/when a more concrete use case comes up.

PR #197, opened 2017-03-21
DM-9882 Add integrate interface to BoundedField/ChebyshevBoundedField

parejkoj:


tests/testChebyshevBoundedField.py
```diff
@@ -113,6 +113,123 @@ def testEvaluate(self):
...
+        coeffs[0, 2] = 3.
+        self._testIntegrateBox(bbox, coeffs, 4.*coeffs[0, 0] - 4./3.*coeffs[2, 0] - 4./3.*coeffs[0, 2])
+
+    def testIntegrateBox(self):
+        """Test integrating over an "interesting" box.
+
+        The values of these integrals were checked in Mathematica. The code
+        block below can be pasted into Mathematica to re-do those calculations.
+
+        .. code:: mathematica
```
jonathansick: This may not work as I don't see `mathematica` listed on Pygment's supported languages. You could just do a general pre-formatted block:

```
.. code:: mathematica
```

to

```
::
```

On the other hand, we're not planning rendering unit test docstrings in Sphinx, so this isn't a significant point.

include/lsst/afw/math/BoundedField.h
```diff
@@ -37,6 +37,9 @@ namespace lsst { namespace afw { namespace math {
 /**
  *  @brief An abstract base class for 2-d functions defined on an integer bounding boxes
  *
+ *  Integer bounding boxes (afw.geom.Box2I) are inclusive of the end pixels, thus a BoundedField
+ *  is defined on the inclusive box [x0, x1] x [y0, y1].
```
TallJimbo: I think you need something stronger than this if you really want to get at what's not intuitive about these boxes.  Maybe something like this "because integer positions correspond to the centers of pixels, and these boxes include the entirety of those pixels, the actual range covered by the `BoundedField` is `[x0 - 0.5, x1 + 0.5] x [y0 - 0.5, y1 + 0.5]`."

include/lsst/afw/math/BoundedField.h
```diff
@@ -85,6 +88,20 @@ class BoundedField : public table::io::PersistableFacade<BoundedField>, public t
     ) const;
 
     /**
+     * @brief      Compute the integral of this function over its bounding-box.
+     *
+     * @return     The value of the integral.
+     */
```
TallJimbo: I don't think we normally do all these extra spaces between the Doxygen tag and the text they correspond to.  I also don't think the `@brief` tags are actually necessary; Doxygen will automatically treat the first line as the brief summary.

src/math/BoundedField.cc
```diff
@@ -38,6 +38,16 @@ ndarray::Array<double,1,1> BoundedField::evaluate(
     return out;
 }
 
+double BoundedField::integrate() const
+{
```
TallJimbo: Open parens should go on the same line as the function declaration.  Same comment in several places below.

tests/testChebyshevBoundedField.py
```diff
@@ -113,6 +113,123 @@ def testEvaluate(self):
...
+        coeffs = np.array([[5.]])
+        self._testIntegrateBox(bbox, coeffs, 4.*coeffs[0, 0])
+
+        # 1st order polynomial: odd orders drop out of integral
+        coeffs = np.array([[5., 2.], [3., 4.]])
+        self._testIntegrateBox(bbox, coeffs, 4.*coeffs[0, 0])
+
+        # 2nd order polynomial in x, 0th in y
+        coeffs = np.array([[5., 0., 7.]])
+        self._testIntegrateBox(bbox, coeffs, 4.*coeffs[0, 0] - 4./3.*coeffs[0, 2])
```
TallJimbo: Please add some parens around the `4./3.` here for clarity.  Same below.

src/math/ChebyshevBoundedField.cc
```diff
@@ -313,6 +313,30 @@ double ChebyshevBoundedField::evaluate(geom::Point2D const & position) const {
...
+double integrateTn(int n) {
+    if (n % 2 == 1)
+        return 0;
+    else
+        return 2./(1. - double(n*n));
+}
+
+double ChebyshevBoundedField::integrate() const {
+    double result = 0;
+    double determinant = getBBox().getArea() / 4.;
```
TallJimbo: Coding standards for C++ say we should use `4.0`, not `4.` (I think we're free to do either in Python, formally, but I think it'd be best to use the same formatting there, too).

include/lsst/afw/math/BoundedField.h
parejkoj: I need to reconfigure my DoxyDoxygen settings probably. Fixed.

PR #198, opened 2017-03-22
DM-9895 FrameSet frames not preserved by Transform(frameSet) constructor

r-owen:


tests/test_transform.py
```diff
@@ -37,6 +38,122 @@ def makeRawArrayData(nPoints, nAxes, delta=0.123):
...
+def permFrameSetIter(frameSet):
+    """Iterator over 0 or more permuted versions of frameSet
+
+    Each returned value is tuple:
+    - permFrameSet: a copy of frameSet with the base and/or current frame permuted
+    - isBasePermuted: a boolean
+    - isCurrPermuted: a boolean
+
+    Only SkyFrames are permuted.
+    Only permuted versions are returned (the original frameSet is never returned)
```
kfindeisen: These two statements would make more sense immediately after the line about "0 or more permuted versions". Otherwise, the reader is left wondering when it's zero or non-zero.

tests/test_transform.py
```diff
@@ -146,15 +262,15 @@ def checkTransform(self, fromName, toName):
                 outPoint = transform.tranForward(inPoint)
                 rawOutPoint = toEndpoint.dataFromPoint(outPoint)
                 # TODO replace all use of np.allclose with assertFloatsAlmostEqual once DM-9707 is fixed
```
kfindeisen: Forgot to remove the comment.

tests/test_transform.py
```diff
@@ -180,10 +296,126 @@ def checkTransform(self, fromName, toName):
...
+                fromEndpoint = transform.getFromEndpoint()
+                toEndpoint = transform.getToEndpoint()
+                polyMap = makePolyMap(nIn, nOut)
+
+                # forward transformation of one point
+                rawInPoint = makeRawPointData(nIn)
+                inPoint = fromEndpoint.pointFromData(rawInPoint)
+                outPoint = transform.tranForward(inPoint)
+                rawOutPoint = toEndpoint.dataFromPoint(outPoint)
+                # TODO replace all use of np.allclose with assertFloatsAlmostEqual once DM-9707 is fixed
```
kfindeisen: Copy-and-pasted comment.

Can you factor out the common code here and in `checkTransformFromMapping` into a method that takes a `Transform`?

tests/test_transform.py
```diff
@@ -180,10 +296,126 @@ def checkTransform(self, fromName, toName):
...
+                # inverse transformation of an array of points;
+                # remember that the inverse will not give the original values!
+                inverseArray = transform.tranInverse(outArray)
+                rawInverseArray = fromEndpoint.dataFromArray(inverseArray)
+                self.assertFloatsAlmostEqual(rawInverseArray, polyMap.tranInverse(rawOutArray), msg=msg)
+                self.assertFloatsAlmostEqual(rawInverseArray, frameSet.tranInverse(rawOutArray), msg=msg)
+
+                # If frameSet has a SkyFrame as base or current, try permuting that (in place);
+                # the Transform constructor should undo the permutation.
+                # Warning: you must permute the axes of the SkYFrame once it is in the frameSet,
```
kfindeisen: Capitalization error in `SkYFrame`

tests/test_transform.py
```diff
@@ -180,10 +296,126 @@ def checkTransform(self, fromName, toName):
...
+                # rather than before adding it to the FrameSet, else the behavior is wrong
+                for permFrameSet, isBasePermuted, isCurrPermuted in permFrameSetIter(frameSet):
+                    if isBasePermuted:
+                        self.assertEqual(permFrameSet.getFrame(astshim.FrameSet.BASE).getLonAxis(), 2)
```
kfindeisen: Do you need an `else` case for these setup assertions?

tests/test_transform.py
```diff
@@ -180,10 +296,126 @@ def checkTransform(self, fromName, toName):
...
+                    if isCurrPermuted:
+                        self.assertEqual(permFrameSet.getFrame(astshim.FrameSet.CURRENT).getLonAxis(), 2)
+                    transformBasePerm = transformClass(permFrameSet)
+                    # the base frame set in the transform should NOT be permuted
```
kfindeisen: I don't understand this comment. It looks below like you're requiring that neither the base nor the current frame be permuted.

tests/test_transform.py
```diff
@@ -180,10 +296,126 @@ def checkTransform(self, fromName, toName):
...
+                    if toName == "SpherePoint":   # it is a SkyFrame and so has getLonAxis
+                        self.assertEqual(permFrameSetCopy.getFrame(astshim.FrameSet.CURRENT).getLonAxis(), 1)
+
+                    # reconstruct transform, just in case
+                    frameSet = makeFrameSet(baseFrame, currFrame)
+                    transform = transformClass(frameSet)
+                    fromEndpoint = transform.getFromEndpoint()
+
+                    # the transform should work identically to `transform`
+                    nPoints = 7  # arbitrary
```
kfindeisen: Again, can this duplicate code be factored out?

src/geom/Endpoint.cc
```diff
@@ -142,16 +159,19 @@ std::shared_ptr<ast::Frame> SpherePointEndpoint::makeFrame() const {
 }
 
 void SpherePointEndpoint::normalizeFrame(std::shared_ptr<ast::Frame> framePtr) const {
-    auto skyFramePtr = std::dynamic_pointer_cast<ast::SkyFrame>(framePtr);
+    // call getFrame in case framePtr points to a FrameSet
+    auto skyFramePtr = std::dynamic_pointer_cast<ast::SkyFrame>(getFrame(framePtr));
     if (!skyFramePtr) {
         std::ostringstream os;
-        os << "frame is a \"" << framePtr->getClass() << "\", not a SkyFrame";
+        os << "frame is a " << getFrame(framePtr)->getClass() << ", not a SkyFrame";
```
r-owen: Thanks for catching this. I'll fix `PointEndpoint<N>::normalizeFrame` to show the frame class (not `FrameSet`)

tests/test_transform.py
```diff
@@ -37,6 +38,122 @@ def makeRawArrayData(nPoints, nAxes, delta=0.123):
...
+    """Return the appropriate frame for the given name and nAzes
+
+    @param[in] name  one of "Generic", "Point2", "Point3" or "SpherePoint"
+    @param[in] nAxes  number of axes; ignored if the name is not "Generic"
+    """
+    return makeEndpoint(name, nAxes).makeFrame()
+
+
+def makeBadFrames(name, nAxes):
+    """Return a list of 0 or more frames that are not a valid match for name and nAxes
```
r-owen: Thank you for realizing `nAxes` was irrelevant. The point of this subroutine is to construct a collection of 0 or more astshim frames that are not compatible with the named endpoint. Since all frames are compatible with `GenericEndpoint` and the number of axes is known for all other endpoints, nAxes is indeed irrelevant.

tests/test_transform.py
```diff
@@ -37,6 +38,122 @@ def makeRawArrayData(nPoints, nAxes, delta=0.123):
...
+    nIn = baseFrame.getNaxes()
+    nOut = currFrame.getNaxes()
+    polyMap = makePolyMap(nIn, nOut)
+
+    baseFrame.setIdent("baseFrame")
+    currFrame.setIdent("currFrame")
+
+    frameSet = astshim.FrameSet(baseFrame)
+    frame2 = astshim.Frame(nIn, "Ident=frame2")
+    frameSet.addFrame(frameSet.getCurrent(), astshim.UnitMap(nIn), frame2)
```
r-owen: I use `getCurrent()` when it is convenient (e.g. a frameset is readily available). Otherwise I use `CURRENT`. I find myself wishing that `BASE` and `CURRENT` were module level constants. They certainly could be.

tests/test_transform.py
r-owen: I think the code is right. If `isBasePermuted` then the code permutes the base frame (which is a bit messy because of the need to temporarily set it to be the current frame). If `isCurrPermuted` then the code permutes the current frame (which is trivial). There is no interaction between these flags, though at least one of them will be true.

Am I missing something? Can you say more about what you had in mind?

tests/test_transform.py
r-owen: What I am testing is that `SpherePoint.normalizeFrame` correctly unpermutes the axes into the standard order. I changed the comment as follows in hopes of clarifying this:
```
# Verify that the base and current frames in the frame set in the *transform*
# are *not* permuted (i.e. that SpherePointEndpoint.normalizeFrame correctly
# set the axes to the standard order long, lat)
```

tests/test_transform.py
r-owen: Yes. Done as part of the earlier refactoring. I'm really glad you suggested that. It was a huge help.

tests/test_transform.py
kfindeisen: I assume the purpose of this block is to double-check the output of `permFrameSetIter` before using it to actually test `FrameSet` or `Transform`.

I was thinking that if you want to check `getLonAxis() == 2` when `isBasePermuted`, then you would also want to check that `getLonAxis() == 1` when `!isBasePermuted`.

r-owen: The important part of the test is that after adding a `SkyFrame` to a `Transform` that its axes are *not* permuted, and to be sure this is true I must verify that the axes actually were permuted.

So I don't think the case you are describing is very important. It is also difficult to test because one can only call `getLonAxis` on a `SkyFrame`, so an extra conditional would be required.

r-owen: It's a bit of a nuisance, since if `isXPermuted` is false then the axis may not be a `SkyFrame`, but I see your point. I'll do it.

kfindeisen: Actually, I completely missed the part about `getLonAxis()` not being defined for all frames... that does make my suggestion rather pointless.

PR #199, opened 2017-03-28
Make it possible to read PropertyLists as persistables

RobertLuptonTheGood:


src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+{
+    LOGL_DEBUG(_log, "PropertyListFormatter write start");
+    auto ip = dynamic_cast<lsst::daf::base::PropertyList const*>(persistable);
+    if (ip == 0) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, "Persisting non-PropertyList");
+    }
+    if (typeid(*storage) == typeid(lsst::daf::persistence::FitsStorage)) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError,
+                      "Writing a PropertyList to FitsStorage is not supported");
+    }
```
TallJimbo: I was confused by this until I realized that *writing* (as opposed to *FitsStorage*) is what's not supported; might be clearer if you just say that `PropertyList` persistence is read-only.

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+}
+
+PropertyListFormatter::~PropertyListFormatter(void) {
+}
+
+void PropertyListFormatter::write(
+        lsst::daf::base::Persistable const* persistable,
+        lsst::daf::persistence::Storage::Ptr storage,
+        lsst::daf::base::PropertySet::Ptr
+                                                )
```
TallJimbo: More weird indentation and `::Ptr` usage here (and in most other signatures in this file).

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+        fitsfile.setHdu(hdu);
+        
+        auto metadata = new lsst::daf::base::PropertyList;
+        fitsfile.readMetadata(*metadata, strip);
+        // if INHERIT=T, we want to also include header entries from the primary HDU
+        if (fitsfile.getHdu() != 0 && metadata->exists("INHERIT")) {
+            LOGLS_WARN(_log, "Ignoring \"INHERIT\" keyword in " << fileName);
+        }
+        return metadata;
+    }
```
TallJimbo: I think it'd be better to have this function return by `std::unique_ptr<daf::base::PropertyList>` and then use `release()` at the end of `PropertyListFormatter::read`.

tests/testPropertyListPersistence.py
```diff
@@ -0,0 +1,71 @@
...
+        logicalLocation = dafPers.LogicalLocation(os.path.join("tests", "data", "HSC-0908120-056-small.fits"))
+
+        # Create a FitsStorage and put it in a StorageList.
+        storage = self.persistence.getRetrieveStorage("FitsStorage", logicalLocation)
+        storageList = dafPers.StorageList([storage])
+
+        # Let's do the retrieval!
+        propertyList = self.persistence.unsafeRetrieve("PropertyList", storageList, None)
+
+        assert propertyList.get("AR_HDU") == 5
```
TallJimbo: Should be `self.assertEqual(propertyList.get("AR_HDU"), 5)`.

include/lsst/afw/formatters/PropertyListFormatter.h
```diff
@@ -0,0 +1,91 @@
...
+        lsst::daf::base::PropertySet::Ptr additionalData
+    );
+    virtual void update(
+        lsst::daf::base::Persistable* persistable,
+        lsst::daf::persistence::Storage::Ptr storage,
+        lsst::daf::base::PropertySet::Ptr additionalData
+    );
+
+    static lsst::daf::persistence::Formatter::Ptr createInstance(
+        lsst::pex::policy::Policy::Ptr policy
```
RobertLuptonTheGood: I changed all these to PTR() and had a template match error (that was actually caused by changing `PropertySet` to `PropertyList`).  I'll go through and change them all to `std::shared_ptr<T>`.

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+ * @brief Implementation of PropertyListFormatter class
+ *
+ * @ingroup afw
+ */
+
+#ifndef __GNUC__
+#  define __attribute__(x) /*NOTHING*/
+#endif
+static char const* SVNid __attribute__((unused)) =
+    "$Id$";
```
RobertLuptonTheGood: Yes; I didn't remove it as it was an unnecessary change to the` DecoratedImageFormatter`, but that hardly matters in a copy.

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+namespace lsst {
+namespace afw {
+namespace formatters {
+
+lsst::daf::persistence::FormatterRegistration PropertyListFormatter::registration(
+    "PropertyList", typeid(lsst::daf::base::PropertyList), createInstance);
+
+PropertyListFormatter::PropertyListFormatter(
+        lsst::pex::policy::Policy::Ptr
+                                                             )
```
RobertLuptonTheGood: 8 characters is copied from the `DecoratedImageFormatter`; what would you prefer?  The ) is weird and I've fixed it.

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,157 @@
...
+        lsst::daf::base::Persistable const* persistable,
+        lsst::daf::persistence::Storage::Ptr storage,
+        lsst::daf::base::PropertySet::Ptr
+                                                )
+{
+    LOGL_DEBUG(_log, "PropertyListFormatter write start");
+    auto ip = dynamic_cast<lsst::daf::base::PropertyList const*>(persistable);
+    if (ip == 0) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, "Persisting non-PropertyList");
+    }
```
RobertLuptonTheGood: Cargo culted from `DecoratedImageFormatter`.  I'll leave it unless you insist.

src/formatters/PropertyListFormatter.cc
TallJimbo: 4 chars is standard.

PR #200, opened 2017-03-29
DM-9598: Allow Transform to return a matrix of derivatives

kfindeisen:
This commit implements `Transform::getJacobian`. The unit tests in `test_transform.py` were modified slightly to make the test transform non-linear, so that the Jacobian is not constant. A supplementary test was added as `test_transform.cc` to make absolutely sure that the matrix dimensions are what they should be, since conversion to ndarray loses some of that information.

### DIFF
--- a/src/geom/Transform.cc
+++ b/src/geom/Transform.cc
@@ -20,6 +20,7 @@
  * see <http://www.lsstcorp.org/LegalNotices/>.
  */
 
+#include <exception>
 #include <memory>
 #include <ostream>
 #include <sstream>
@@ -30,6 +31,7 @@
 #include "lsst/afw/geom/Point.h"
 #include "lsst/afw/geom/SpherePoint.h"
 #include "lsst/afw/geom/Transform.h"
+#include "lsst/pex/exceptions/Exception.h"
 
 namespace lsst {
 namespace afw {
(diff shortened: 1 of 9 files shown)


tests/test_transform.py
```diff
@@ -183,16 +183,45 @@ def makeCoeffs(nIn, nOut):
...
+    coeffs = np.empty((nOut, nIn))
+    for iOut in range(nOut):
+        coeffOffset = baseCoeff * iOut
+        for iIn in range(nIn):
+            coeffs[iOut, iIn] = baseCoeff * (iIn + 1) + coeffOffset
+            coeffs[iOut, iIn] *= inArray[iIn]
+    assert coeffs.ndim == 2
+    assert coeffs.shape == (nOut, nIn)
+    # Avoid spurious errors when comparing to a simplified array
+    return coeffs.squeeze()
```
r-owen: I think it will be less confusing having `makeJacobian` return the expected 2-dimensional array. The fact that the pybind11 wrapper may return something else for the Jacobian is a misfeature (one I hope we can eventually figure out a workaround for) that we need not emulate here. Instead consider putting the `squeeze` where the two matrices are compared.

tests/test_transform.py
```diff
@@ -609,12 +636,45 @@ def checkInverseFrameSet(self, clsTransform, frameIn, frameOut):
         self.assertEqual(reverseFrames.getFrame(4).getIdent(), "currFrame",
                          msg=baseMsg)
 
+    def checkGetJacobian(self, fromName, toName):
+        """Test Transform<fromName>To<toName>.getInverse
```
r-owen: Copy/paste error?

src/geom/Transform.cc
```diff
@@ -118,6 +120,25 @@ Transform<ToEndpoint, FromEndpoint> Transform<FromEndpoint, ToEndpoint>::getInve
...
+
+        Eigen::MatrixXd jacobian(nOut, nIn);
+        for (size_t i = 0; i < nOut; ++i) {
+            for (size_t j = 0; j < nIn; ++j) {
+                jacobian(i, j) = _frameSet->rate(point, i + 1, j + 1);
+            }
+        }
+        return jacobian;
+    } catch (std::bad_alloc const &e) {
+        std::throw_with_nested(pex::exceptions::MemoryError("Could not allocate Jacobian."));
```
kfindeisen: This is mainly for API consistency. We have a dedicated `MemoryError` exception, so presumably we should emit it in preference to `std::bad_alloc`. Although perhaps now that pybind11 gives us better exception wrapping than Swig did, we should push for deprecation of `MemoryError` as part of DM-9435.

As for constructing the exception, it looks like `LSST_EXCEPT` evaluates to an exception object rather than a throw statement, so I can do that.

tests/test_transform.py
TallJimbo: numpy generally (or at least occasionally) considers that a feature; I'm not sure I do, but I didn't want to try to argue with numpy's conventions.

r-owen: @TallJimbo I filed https://jira.lsstcorp.org/browse/DM-10011 in hopes you might change your mind about fighting this. It is a terrible nuisance, at the minimum.

PR #202, opened 2017-03-29
Tickets/dm 9990

RobertLuptonTheGood:


python/lsst/afw/cameraGeom/cameraFactory.py
```diff
@@ -48,6 +48,28 @@ def makeDetector(detectorConfig, ampInfoCatalog, focalPlaneToPupil):
         transforms,
     )
 
+def copyDetector(detector, ampInfoCatalog=None):
```
r-owen: What is the point of the function is if `ampInfoCatalog` is omitted? This feels like API clutter. If the only use case is making a new detector for an untrimmed raw image then I suggest writing a function that does that (though most of the trimming work probably belongs in or near `AmpInfoRecord`, as I said in my other review). Then this code could moved there (either as a separate private function or incorporated into the body).

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,152 @@
...
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError,
+                      "FitsStorage for PropertyList read-only (writing is not supported)");
+    }
+    throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError,
+                      "Unrecognized Storage for PropertyList");
+}
+
+namespace {
+    // I'd use a unique_ptr, except that PropertyList won't allow me to merge non-shared_ptrs
+    std::unique_ptr<daf::base::PropertyList>
```
r-owen: You *are* returning a unique pointer, so I think the comment above should be deleted.

Is it necessary to return a smart pointer at all? I realize you want to avoid a copy, but am wondering if move semantics would suffice.

src/formatters/PropertyListFormatter.cc
```diff
@@ -0,0 +1,152 @@
...
+        auto metadata = std::unique_ptr<daf::base::PropertyList>(new lsst::daf::base::PropertyList);
+        //
+        // We need a shared_ptr to be able to call PropertyList.combine()
+        //
+        auto inheritedMetadata = std::shared_ptr<daf::base::PropertyList>(new daf::base::PropertyList);
+
+        fits::Fits fitsfile(fileName, "r", fits::Fits::AUTO_CLOSE | fits::Fits::AUTO_CHECK);
+        fitsfile.setHdu(hdu);
+        fitsfile.readMetadata(*inheritedMetadata, strip);
+        metadata->combine(inheritedMetadata);
```
r-owen: Why `inheritedMetadata` and why are you combining anything? I thought perhaps one of the two metadata variables would hold HDU 0 and the other the specified HDU but that does not seem to be the case.

include/lsst/afw/formatters/PropertyListFormatter.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "lsst/daf/base.h"
+#include "lsst/daf/persistence.h"
+#include "lsst/pex/policy/Policy.h"
+
+namespace lsst {
+namespace afw {
+namespace formatters {
+
+class PropertyListFormatter : public daf::persistence::Formatter {
```
r-owen: Please document the class and methods.

tests/testDetector.py
```diff
@@ -217,6 +217,36 @@ def testGetCenter(self):
...
+    def testCopyDetector(self):
+        """Test copyDetector() method
+        """
+        #
+        # Make a copy without any modifications
+        #
+        detector  = DetectorWrapper().detector
+        ndetector = cameraGeom.copyDetector(detector)
+
+        assert detector.getName() == ndetector.getName()
```
r-owen: Please use `self.assertEquals` and such, so that test failures show up in the usual fashion. Likewise for testPropertyListPersistence.py below.

Also please consider renaming tests from `testFoo.py` to `test_foo.py` when you work on them. Thats the new standard (to aid automatic discovery by pytest).

RobertLuptonTheGood: Sorry, I had thought that pytest handled assert specially;  I'll go back to the unittest conventions.

src/formatters/PropertyListFormatter.cc
RobertLuptonTheGood: This is not part of this ticket (I rebased afw DM-9986 to master before realising that it wouldn't build against w.2017.13) and it'll all get cleaned up when w.2017.14 becomes available.

To answer the question, I knew about that typo but given the number of tickets hanging off this foolish rebase I'd rather fix it when I'm fully merged.  The reason I'm returning a unique pointer is that Jim asked me to; there is a potential exception in the calling code that could have resulted in a leak.

python/lsst/afw/cameraGeom/cameraFactory.py
RobertLuptonTheGood: This isn't the only time that I've wanted to modify a Detector (e.g. to update gains while investigating TS-8 data), so I think that the ability to clone a Detector is valuable.

jdswinbank:
Has been merged.

PR #203, opened 2017-03-29
DM-9952: 0-based HDUs

RobertLuptonTheGood:
Switch to 0-indexed HDUs; RFC-304

N.b. there are also changes in obs_base


timj:
`obs_base` PR is lsst/obs_base#29

PR #208, opened 2017-04-01
Tickets/dm 9952  switch to 0-indexed HDUs

RobertLuptonTheGood:
The weird base is because I need to merge DM-9990 to master and afw then won't build against w.2017.13


python/lsst/afw/math/backgroundList.py
```diff
@@ -152,6 +152,7 @@ def readFits(fileName, hdu=0, flags=0):
 
         self = BackgroundList()
 
+        hdu -= 1                        # we want to start at 0 (post RFC-304), but are about to increment
         while True:
```
parejkoj: Two things here: 

1. Please make the comment 2 spaces from the code.
2. Ugh, really? This `while`-loop looks all kinds of wonky to me. I guess redoing it is out of scope, though.

src/image/ExposureInfo.cc
```diff
@@ -160,8 +160,12 @@ ExposureInfo::_startWriteFits(afw::geom::Point2I const & xy0) const {
...
 
     // In the future, we might not have exactly three image HDUs, but we always do right now,
-    // so 1=primary, 2=image, 3=mask, 4=variance, 5+=archive
-    data.metadata->set("AR_HDU", 5, "HDU containing the archive used to store ancillary objects");
+    // so 0=primary, 1=image, 2=mask, 3=variance, 4+=archive
+    //
+    // Historically the AR_HDU keyword was 1-indexed (see RFC-304), and to maintain file compatibility
+    // this is still the case so we're setting AR_HDU to 5 == 4 + 1
+    //
+    data.metadata->set("AR_HDU", 5, "One more than HDU containing the archive used to store ancillary objects");
```
parejkoj: Thank you for the clarifying comment above, but I'm not so keen on the wording of the metadata comment. Maybe "HDU (1-indexed) containing ..."? Though I'm not sure about that either.

This also seems like it's going to bite us in the future (having a place where we 1-index). Do we need to maintain this backwards compatibility, or can we require reprocessing after this merge?

src/image/MaskedImage.cc
```diff
@@ -655,11 +655,11 @@ void image::MaskedImage<ImagePixelT, MaskPixelT, VariancePixelT>::writeFits(
             "MaskedImage::writeFits can only write to an empty file"
         );
     }
-    if (fitsfile.getHdu() <= 1) {
+    if (fitsfile.getHdu() < 1) {
```
parejkoj: Good catch!

src/table/Source.cc
```diff
@@ -92,11 +92,14 @@ void SourceFitsWriter::_writeTable(CONST_PTR(BaseTable) const & t, std::size_t n
...
-        // HDU 3 is the index for the afw::table::io archive that holds more complex objects
+        // HDU 0 is empty (primary HDU can't be a table)
+        // HDU 1 is the SourceCatalog's records
+        // HDU 2 is the index for the afw::table::io archive that holds more complex objects
+        //
+        // Historically the AR_HDU keyword was 1-indexed (see RFC-304), and to maintain file compatibility
+        // this is still the case so we're setting AR_HDU to 3 == 2 + 1
         metadata->set(
-            "AR_HDU", 3, "HDU containing the archive index for non-record data (e.g. Footprints)"
+            "AR_HDU", 3, "One more than HDU containing the archive index for non-record data (e.g. Footprints)"
```
parejkoj: Same as I had for AR_HDU above: I'm not sure about the wording, and I worry about the backwards compatibility here. We probably get only one chance to fix this, and that chance is now.

tests/testPropertyListPersistence.py
```diff
@@ -58,10 +58,7 @@ def testFitsPersistence(self):
         # Let's do the retrieval!
         propertyList = self.persistence.unsafeRetrieve("PropertyList", storageList, None)
 
-        if False:                       # this test requires that DM-9952 be merged
```
parejkoj: 👍

tests/testTicket2352.py
```diff
@@ -51,15 +54,18 @@ def testExtName(self):
         self.checkExtName("THREE", 3, 4)
 
     def checkExtNum(self, hdu, extNum):
+        if hdu is None:
+            hdu = -(1 << 31)            # == INT_MIN
```
parejkoj: Can we get the `INT_MIN` -> `None` translation at the pybind11 level, so we don't have to do this?

src/image/ExposureInfo.cc
RobertLuptonTheGood: I discussed this with @TallJimbo while writing the code.  The code is 0-indexed, the only thing that isn't is the `AR_HDU` keyword in the files.  I could complicate the code to handle backwards compatibility, but it seemed simpler to leave the files alone.  That way old and new versions of afw can both read the old and new files.  I'll give him another chance to weigh in, but I still feel that this is the correct solution.

src/table/Source.cc
RobertLuptonTheGood: I prefer your comment (and it's shorter).  Fixing both.

tests/testTicket2352.py
RobertLuptonTheGood: This is the only place I had to do it, and chatting with {J,P}im it would be a pain -- the default value appears in many many places in the pybind11ings.  With the advent of named arguments I don't think this is going to be a problem -- if you disagree, please file a ticket.

src/image/ExposureInfo.cc
TallJimbo: Yeah, I think it's okay for the convention used in files to be different from the convention used in code (just as it is with our 0- or 1-based indexing of pixel positions).  And it's highly desirable to keep the format the same whenever we can.  Backwards compatible I/O code can build up quickly and get in the way of refactoring.

parejkoj: Alright. I'm still not entirely keen on it, but I guess I'm convinced.

jdswinbank:
Has been merged (in rebased form).

PR #209, opened 2017-04-01
Tickets/dm 9917 add a callback to showCamera

RobertLuptonTheGood:
Note the base;  that's DM-9990 pre the rebase to master needed to merge it;  I need this code to build with respect to w.2017.13


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -336,64 +343,116 @@ def _prepareImage(self, ccd, im, binSize, allowRotate=True):
...
+    ccdImage = im.Factory(bbox)
+
+    ampImages = []
+    for a in ccd:
+        if isTrimmed:
             data = im[a.getRawDataBBox()]
+        else:
+            data = im
+
+        if subtractBias:
```
mfisherlevine: Is this order of operations correct? (Note that not I'm asserting that it isn't, I'm just not sure). Also, should an option to correct for linearity be inserted here as well, or do you feel that's going too far? If so, then that factors in to the answer to the first question, as the definitions need to be consistent.

RobertLuptonTheGood: It doesn't make sense to be Trimmed and subtract the bias, so I'll add a check.

I think that linearity would be totally inappropriate.  We're not running the ISR here (if we were, we'd register a callback that did just that).

mfisherlevine: OK, I agree on the linearity, but actually not about the trimming vs. bias subtraction. I can imagine use-cases for wanting all combinations of trimmed and untrimmed, and subtracted and un-subtracted.

RobertLuptonTheGood: The trimming says that the image has already been trimmed for display.  

This is a callback from showCamera, and if you want other functionality you can write your own;  I don't think that the callback can modify the size of the image without breaking things, but if you want to do that you should file a new ticket.

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -336,64 +343,116 @@ def _prepareImage(self, ccd, im, binSize, allowRotate=True):
...
-            ampImages.append(data)
+        ampImages.append(data)
 
-        ccdImage = imageFactory(bbox)
-        for ampImage, amp in zip(ampImages, ccd):
-            if self.isTrimmed:
-                assembleAmplifierImage(ccdImage, ampImage, amp)
-            else:
-                assembleAmplifierRawImage(ccdImage, ampImage, amp)
+    ccdImage = type(ccdImage)(bbox)
```
mfisherlevine: This line seems to duplicate L428 (and indeed overwrite `ccdImage`). Should one be deleted?

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -336,64 +343,116 @@ def _prepareImage(self, ccd, im, binSize, allowRotate=True):
...
+        if isTrimmed:
+            assembleAmplifierImage(ccdImage, ampImage, amp)
+        else:
+            assembleAmplifierRawImage(ccdImage, ampImage, amp)
 
-        return ccdImage
+    return ccdImage
 
-def overlayCcdBoxes(ccd, untrimmedCcdBbox, nQuarter, isTrimmed, ccdOrigin, display, binSize):
+def overlayCcdBoxes(ccd, untrimmedCcdBbox=None, nQuarter=0,
```
mfisherlevine: Should add a description of what `nQuarter` is/does, as I don't feel that the parameter name is sufficiently self-explanatory.

RobertLuptonTheGood: Added a phrase of clarification

jdswinbank:
Has been merged to master (but not to the user branch).

PR #211, opened 2017-04-05
DM-10096 Add unit test asserts for SpherePoint, SpherePointList and PointList

r-owen:


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -767,13 +782,13 @@ def showCamera(camera, imageSource=FakeImageDataSource(), imageFactory=afwImage.
 
     if showWcs:
         if originAtCenter:
-            #Can't divide SWIGGED extent type things when division is imported
-            #from future.  This is DM-83
+            # Can't divide SWIGGED extent type things when division is imported
+            # from future.  This is DM-83
```
timj: I wonder about this comment now that we don't use SWIG.

tests/testHeavyFootprint.py
```diff
@@ -1,6 +1,6 @@
 #
 # LSST Data Management System
-# Copyright 2008, 2009, 2010 LSST Corporation.
+# Copyright 2008-2017 LSST Corporation.
```
timj: Changes like this aren't technically correct but we are fudging pending RFC-45.

python/lsst/afw/cameraGeom/utils.py
r-owen: Agreed. DM-83 was fixed long ago. I'll have a look at the code.

tests/testHeavyFootprint.py
r-owen: Exactly

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -559,14 +568,14 @@ def showAmp(amp, imageSource=FakeImageDataSource(isTrimmed=False), display=None,
     if overlay:
         with display.Buffering():
             if amp.getHasRawInfo() and ampImSize == amp.getRawBBox().getDimensions():
-                bboxes = [(amp.getRawBBox(), 0.49, afwDisplay.GREEN),]
+                bboxes = [(amp.getRawBBox(), 0.49, afwDisplay.GREEN), ]
```
natelust: Im not going to insist that you change cases like these, but I think it would be preferable to remove the trailing unneeded comma rather than insert and extra unneeded space after it, just to then close the list. I have noticed this kind of change in a few places.

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -25,8 +25,7 @@
 """
 Support for cameraGeom
 """
-from __future__ import division
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function
```
natelust: sense you already have commits on this ticket for fixing formatting issues, this change should be moved to one of those commits.

python/lsst/afw/geom/utils.py
```diff
@@ -25,19 +24,23 @@
 
 In the case of the assert functions, importing them makes them available in lsst.utils.tests.TestCase
 """
+__all__ = ["assertAnglesAlmostEqual", "assertPairsAlmostEqual",
```
natelust: __all__ should come after the imports

python/lsst/afw/image/basicUtils.py
```diff
@@ -22,11 +22,12 @@
 from __future__ import absolute_import, division, print_function
 
 __all__ = ["makeImageFromArray", "makeMaskFromArray", "makeMaskedImageFromArrays",
```
natelust: if you are modifying the all line, it should come after the imports

python/lsst/afw/geom/utils.py
timj: PEP-8 says the opposite. https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

I think there was a recent discussion on this topic with the style guide.

python/lsst/afw/image/basicUtils.py
```diff
@@ -94,8 +95,8 @@ def _compareWcsOverBBox(wcs0, wcs1, bbox, maxDiffSky=0.01*afwGeom.arcseconds,
     xList = numpy.linspace(bboxd.getMinX(), bboxd.getMaxX(), nx)
     yList = numpy.linspace(bboxd.getMinY(), bboxd.getMaxY(), ny)
     # we don't care about measured error unless it is too large, so initialize to max allowed
-    measDiffSky = (maxDiffSky, "?") # (sky diff, pix pos)
-    measDiffPix = (maxDiffPix, "?") # (pix diff, sky pos)
+    measDiffSky = (maxDiffSky, "?")  # (sky diff, pix pos)
+    measDiffPix = (maxDiffPix, "?")  # (pix diff, sky pos)
```
natelust: sense you have formatting changes these should probably go on one of those commits. I wont be too strict on this, if there is only one change here or there you don't have to recraft all your commits, but please keep this in mind.

python/lsst/afw/image/testUtils.py
```diff
@@ -23,8 +23,11 @@
 
 # the asserts are automatically imported so unit tests can find them without special imports;
 # the other functions are hidden unless explicitly asked for
-__all__ = ["assertImagesNearlyEqual", "assertMasksEqual", "assertMaskedImagesNearlyEqual",
-           "assertImagesEqual", "assertMaskedImagesEqual"]
+__all__ = ["assertImagesAlmostEqual", "assertImagesEqual", "assertMasksEqual",
```
natelust: when modifying the all line, you might as well move it to the proper place after imports

tests/testMethods.py
```diff
@@ -40,106 +40,116 @@ class TestTestUtils(lsst.utils.tests.TestCase):
...
-    def testAssertAnglesNearlyEqual(self):
-        """Test assertAnglesNearlyEqual"""
+    def testAssertAnglesAlmostEqual(self):
+        """Test assertAnglesAlmostEqual"""
         for angDeg in (0, 45, -75):
             ang0 = angDeg*afwGeom.degrees
+            self.assertAnglesAlmostEqual(
+                ang0,
+                ang0 + 0.01*afwGeom.arcseconds,
+                maxDiff=0.010001*afwGeom.arcseconds,
```
natelust: consider moving magic numbers to their own variables so they can be reused.

natelust: This comment applies to any of the tests below as well

tests/testAstropyTableViews.py
```diff
@@ -183,6 +185,7 @@ class TestMemory(lsst.utils.tests.MemoryTestCase):
 def setup_module(module):
     lsst.utils.tests.init()
 
+
```
natelust: consider putting formatting changes onto the formatting commits

tests/testFunctorKeys.py
```diff
@@ -427,6 +430,7 @@ class MemoryTester(lsst.utils.tests.MemoryTestCase):
 def setup_module(module):
     lsst.utils.tests.init()
 
+
```
natelust: consider moving formatting change, not necessary though

tests/testMethods.py
```diff
@@ -156,6 +156,58 @@ def testAssertCoordsNearlyEqual(self):
             coord2 = afwCoord.IcrsCoord(raDecDeg[0] + 360*afwGeom.degrees, raDecDeg[1])
             self.assertCoordsNearlyEqual(coord0, coord2, maxDiff=1e-7*afwGeom.arcseconds)
 
+    def testAssertSpherePointsAlmostEqual(self):
```
natelust: overall pleas consider not having magic numbers in function calls, separate them out to variables so they can be reused

python/lsst/afw/coord/utils.py
```diff
@@ -24,10 +24,11 @@
 
 In the case of the assert functions, importing them makes them available in lsst.utils.tests.TestCase
 """
+__all__ = ["assertCoordsNearlyEqual"]
```
natelust: __all__ should come after imports

python/lsst/afw/geom/utils.py
natelust: Yeah, Russell and I discussed this, I was going on the dev guide which had not been updated at the point of me reviewing this. He has now updated the dev guide

PR #212, opened 2017-04-05
DM-10105: Avoid loss of precision in Wcs persistence.

TallJimbo:
The old implementation round-tripped CRVAL through radians because it used Angle, which resulted in a loss of precision and inexactly persisted WCSs.

### DIFF
--- a/src/image/Wcs.cc
+++ b/src/image/Wcs.cc
@@ -1082,7 +1082,8 @@ void Wcs::write(OutputArchiveHandle & handle) const {
     WcsPersistenceHelper const & keys = WcsPersistenceHelper::get();
     afw::table::BaseCatalog catalog = handle.makeCatalog(keys.schema);
     PTR(afw::table::BaseRecord) record = catalog.addNew();
-    record->set(keys.crval, getSkyOrigin()->getPosition(afw::geom::degrees));
+    record->set(keys.crval.getX(), _wcsInfo[0].crval[0]);
+    record->set(keys.crval.getY(), _wcsInfo[0].crval[1]);
     record->set(keys.crpix, getPixelOrigin());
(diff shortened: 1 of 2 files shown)


tests/testWcsFitsTable.py
```diff
@@ -192,6 +192,32 @@ def testWcsWhenNonPersistable(self):
...
+        metadata.add('CTYPE2', 'DEC--TAN')
+        metadata.add('CUNIT1', 'deg')
+        metadata.add('CUNIT2', 'deg')
+        metadata.add('NAXIS', 2)
+        metadata.add('RADESYS', 'ICRS')
+        wcs1 = lsst.afw.image.makeWcs(metadata)
+        with lsst.utils.tests.getTempFilePath(".fits") as fileName:
+            wcs1.writeFits(fileName)
+            wcs2 = lsst.afw.image.Wcs.readFits(fileName)
+            self.assertEqual(wcs1, wcs2)
```
parejkoj: 👍 Exact equality here is probably good.

src/image/Wcs.cc
```diff
@@ -1082,7 +1082,8 @@ void Wcs::write(OutputArchiveHandle & handle) const {
     WcsPersistenceHelper const & keys = WcsPersistenceHelper::get();
     afw::table::BaseCatalog catalog = handle.makeCatalog(keys.schema);
     PTR(afw::table::BaseRecord) record = catalog.addNew();
-    record->set(keys.crval, getSkyOrigin()->getPosition(afw::geom::degrees));
+    record->set(keys.crval.getX(), _wcsInfo[0].crval[0]);
```
parejkoj: It's sad that we can't use our Sky methods, but I guess it makes sense to do it this way in the internals.

PR #214, opened 2017-04-07
DM-9599: Support concatenation of Transforms

kfindeisen:
This PR implements `Transform::of` in terms of `ast::prepend`. It deviates slightly from the DM-9599 spec in that the C++ code checks for compatible endpoints at compile time; this turned out to be less work than the original run-time-only design because it reduces the number of overloads that need to be wrapped in pybind11. The need to explicitly consider one overload per endpoint implementation class already complicates things enough.

### DIFF
--- a/include/lsst/afw/geom/Transform.h
+++ b/include/lsst/afw/geom/Transform.h
@@ -60,7 +60,7 @@ since the base and current frames in the FrameSet can be checked against by the 
 because data must be copied when converting from LSST data types to the type used by astshim,
 so it didn't seem worth the bother.
 */
-template <typename FromEndpoint, typename ToEndpoint>
+template <class FromEndpoint, class ToEndpoint>
 class Transform {
 public:
     using FromArray = typename FromEndpoint::Array;
@@ -225,7 +225,7 @@ The format is "Transform<_fromEndpoint_, _toEndpoint_>"
 where _fromEndpoint_ and _toEndpoint_ are the appropriate endpoint printed to the ostream;
 for example "Transform<GenericEndpoint(4), Point3Endpoint()>"
 */
-template <typename FromEndpoint, typename ToEndpoint>
+template <class FromEndpoint, class ToEndpoint>
 std::ostream &operator<<(std::ostream &os, Transform<FromEndpoint, ToEndpoint> const &transform);
 
 }  // geom
--- a/include/lsst/afw/geom/Transform.h
+++ b/include/lsst/afw/geom/Transform.h
@@ -189,6 +189,29 @@ public:
      */
     Eigen::MatrixXd getJacobian(FromPoint const &x) const;
 
+    /**
+     * Concatenate two Transforms.
+     *
+     * @tparam FirstFromEndpoint the starting Endpoint of `first`
+     * @param first the Transform to apply before this one
+     * @returns a Transform that first applies `first` to its input, and then
+     *          this Transform to the result. Its inverse shall first apply the
+     *          inverse of this Transform, then the inverse of `first`.
+     *
+     * @throws InvalidParameterErrror Thrown if `first.getToEndpoint()` and
+     *                                `this->getFromEndpoint()` do not have
+     *                                the same number of axes.
+     * @exceptsafe Provides basic exception safety.
+     *
+     * More than two Transforms can be combined in series. For example:
+     *
+     *     auto skyFromPixels = skyFromPupil.of(pupilFromFp)
+     *                                      .of(fpFromPixels);
+     */
+    template <class FirstFromEndpoint>
+    Transform<FirstFromEndpoint, ToEndpoint> of(
+            Transform<FirstFromEndpoint, FromEndpoint> const &first) const;
+
 private:
     FromEndpoint const _fromEndpoint;
     std::shared_ptr<const ast::FrameSet> _frameSet;
--- a/python/lsst/afw/geom/transform/transform.cc
+++ b/python/lsst/afw/geom/transform/transform.cc
@@ -51,6 +51,17 @@ std::string formatStr(Class const &self, std::string const &pyClassName) {
     return os.str();
 }
 
+template <class ExtraEndpoint, class FromEndpoint, class ToEndpoint, class PyClass>
+void declareMethodTemplates(PyClass &cls) {
+    using FirstTransform = Transform<ExtraEndpoint, FromEndpoint>;
+    using SecondTransform = Transform<FromEndpoint, ToEndpoint>;
+    using FinalTransform = Transform<ExtraEndpoint, ToEndpoint>;
+    // Need Python-specific logic to give sensible errors for mismatched Transform types
+    cls.def("_of", (FinalTransform (SecondTransform::*)(FirstTransform const &) const) &
+                           SecondTransform::template of<ExtraEndpoint>,
+            "first"_a);
+}
+
 // Declare Transform<FromEndpoint, ToEndpoint> using python class name TransformFrom<X>To<Y>
 // where <X> and <Y> are the name of the from endpoint and to endpoint class, respectively,
 // for example TransformFromGenericToPoint3
@@ -81,7 +92,15 @@ void declareTransform(py::module &mod, std::string const &fromName, std::string 
     cls.def("tranInverse", (FromArray (Class::*)(ToArray const &) const) & Class::tranInverse, "array"_a);
     cls.def("tranInverse", (FromPoint (Class::*)(ToPoint const &) const) & Class::tranInverse, "point"_a);
     cls.def("getInverse", &Class::getInverse);
+    /* Need some extra handling of ndarray return type in Python to prevent dimensions
+     * of length 1 from being deleted */
     cls.def("_getJacobian", &Class::getJacobian);
+
+    declareMethodTemplates<GenericEndpoint, FromEndpoint, ToEndpoint>(cls);
+    declareMethodTemplates<Point2Endpoint, FromEndpoint, ToEndpoint>(cls);
+    declareMethodTemplates<Point3Endpoint, FromEndpoint, ToEndpoint>(cls);
+    declareMethodTemplates<SpherePointEndpoint, FromEndpoint, ToEndpoint>(cls);
+
     // str(self) = "<Python class name>[<nIn>-><nOut>]"
     cls.def("__str__", [pyClassName](Class const &self) { return formatStr(self, pyClassName); });
     // repr(self) = "lsst.afw.geom.<Python class name>[<nIn>-><nOut>]"
--- a/python/lsst/afw/geom/transform/transform.cc
+++ b/python/lsst/afw/geom/transform/transform.cc
@@ -42,7 +42,7 @@ namespace {
 
 // Return a string consisting of "_pythonClassName_[_fromNAxes_->_toNAxes_]",
 // for example "TransformGenericToPoint3[4->3]"
-template <typename Class>
+template <class Class>
 std::string formatStr(Class const &self, std::string const &pyClassName) {
     std::ostringstream os;
     os << pyClassName;
@@ -65,7 +65,7 @@ void declareMethodTemplates(PyClass &cls) {
 // Declare Transform<FromEndpoint, ToEndpoint> using python class name TransformFrom<X>To<Y>
 // where <X> and <Y> are the name of the from endpoint and to endpoint class, respectively,
 // for example TransformFromGenericToPoint3
-template <typename FromEndpoint, typename ToEndpoint>
+template <class FromEndpoint, class ToEndpoint>
 void declareTransform(py::module &mod, std::string const &fromName, std::string const &toName) {
     using Class = Transform<FromEndpoint, ToEndpoint>;
     using ToPoint = typename ToEndpoint::Point;
--- a/python/lsst/afw/geom/transform/transformContinued.py
+++ b/python/lsst/afw/geom/transform/transformContinued.py
@@ -21,6 +21,8 @@
 #
 from __future__ import absolute_import, division, print_function
 
+from lsst.pex.exceptions import InvalidParameterError
+
 from . import transform
 
 __all__ = []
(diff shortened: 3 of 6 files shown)


include/lsst/afw/geom/Transform.h
```diff
@@ -189,6 +189,28 @@ class Transform {
...
 
+    /**
+     * Concatenate two Transforms.
+     *
+     * @tparam LastEndpoint the starting Endpoint of the Transform that feeds
+     *                      into this one
+     * @param before the Transform to apply before this one
+     * @returns a Transform that first applies `before` to its input, and then
+     *          this Transform to the result. Its inverse shall first apply the
+     *          inverse of this Transform, then the inverse of `before`.
```
r-owen: I'm finding this hard to follow. The template parameter `LastEndpoint` name seems backwards to me, which makes the rest of it harder for me to follow.

See if something like the following works for you:

I strongly suggest you name the argument `first`, just like ast::Mapping::of. Commonality is a huge help here. (I realize I would naturally prefer `first` since that's what I chose for astshim. If you are very keen for `before` we could consider changing it in both places. But I find `first` clear and concise.)

Then document could look something like this:

Concatenate two transforms to produce a new transform C(x) = this(first(x))

The resulting transform has the "from" endpoint of `first` (FromFirstEndpoint), and the "to" endpoint of this transform (ToEndpoint). The other two endpoints, where the transforms are joined, are required to be identical: thus the "to" transform of `first` must be the same as "from" endpoint of this transform: `FromEndpoint`. The compiler assures that the classes are the same, but a runtime check is required to make sure the number of axes match.

    template<class FirstFromEndpoint>
    Template<class FirstFromEndpoint, ToEndpoint> of(Template<FirstFromEndpoint, FromEndpoint> const & first)

include/lsst/afw/geom/Transform.h
```diff
@@ -60,7 +60,7 @@ since the base and current frames in the FrameSet can be checked against by the
 because data must be copied when converting from LSST data types to the type used by astshim,
 so it didn't seem worth the bother.
 */
-template <typename FromEndpoint, typename ToEndpoint>
+template <class FromEndpoint, class ToEndpoint>
```
r-owen: Good idea.

include/lsst/afw/geom/Transform.h
```diff
@@ -189,6 +189,28 @@ class Transform {
...
+     *
+     * @tparam LastEndpoint the starting Endpoint of the Transform that feeds
+     *                      into this one
+     * @param before the Transform to apply before this one
+     * @returns a Transform that first applies `before` to its input, and then
+     *          this Transform to the result. Its inverse shall first apply the
+     *          inverse of this Transform, then the inverse of `before`.
+     *
+     * @throws InvalidParameterErrror Thrown if `this->getFromEndpoint()` does
+     *                                not match `before.getToEndpoint()`.
```
r-owen: I find this a bit confusing because the compiler makes sure the classes match. But this does not necessarily mean the number of axes match. I tried to say something about that in my suggested documentation above, and that might suffice, but I wonder if this  `@throws` description could also be  clearer.

include/lsst/afw/geom/Transform.h
```diff
@@ -189,6 +189,28 @@ class Transform {
...
+     * @exceptsafe Provides basic exception safety.
+     *
+     * More than two Transforms can be combined in series. For example:
+     *
+     *     auto pixelsToSky = pupilToSky.of(fpToPupil)
+     *                                  .of(pixelsToFp);
```
r-owen: I'm not sure it is worth changing, but I usually name transforms using `from` instead of `to` because `a_values = aFromB(b_values)`. When worded that way this example will read much more naturally.

include/lsst/afw/geom/Transform.h
kfindeisen: I think "number of axes" is a bit too specific. Can we be sure that's the only way that two Endpoints of the same type will differ?

kfindeisen: How about "does not have the same properties (e.g., number of axes) as"?

tests/test_transform.py
```diff
@@ -668,13 +668,104 @@ def checkGetJacobian(self, fromName, toName):
...
+                    assert_allclose(
+                        fromEndpoint.dataFromPoint(inPointMerged),
+                        fromEndpoint.dataFromPoint(inPointSeparate),
+                        err_msg=msg)
+
+        with self.assertRaises(InvalidParameterError):
+            polyMap = makeTwoWayPolyMap(2, 3)
+            transform1 = transform1Class(polyMap)
+            transform2 = transform2Class(polyMap)
+            transform = transform2.of(transform1)
```
r-owen: I suggest rewriting this test so that only the `transform2.of` is expected to fail, and put only that statement into the `with` block. That way you can be sure that a failure occurs in the statement you care about, and that a coding error in the other statements will show up as an unexpected exception.

If I understand correctly you are testing # of axes mismatch, and such a test is only sensible if `midName == "Generic"`. if so I would do something like this:

    # Verify that trying to join the same type of endpoint but mismatched
    # numbers of axes, fails. Note that only GenericEndpoints can have
    # variable numbers of axes.
    if midName == "Generic":
        polyMap = makeTwoWayPolyMap(2, 3)
        transform1 = transform1Class(polyMap)
        transform2 = transform2Class(polyMap)
        with self.assertRaises(InvalidParameterError):
            transform = transform2.of(transform1)

tests/test_transform.py
```diff
@@ -668,13 +668,104 @@ def checkGetJacobian(self, fromName, toName):
...
+
+        if fromName != midName:
+            with self.assertRaises(InvalidParameterError):
+                polyMap = makeTwoWayPolyMap(2, 3)
+                transform1 = transform1Class(polyMap)
+                polyMap = makeTwoWayPolyMap(3, 3)
+                transform2 = transform1Class(polyMap)
+                transform = transform2.of(transform1)
```
r-owen: Again, please rewrite so only the `transform2.of` statement is in the `with` statement.

In this case I think you are trying to test type mismatch (whether or not the # of axes matches, though you already handled mismatched # of axes above)), which is a great idea, but I think to do that properly you should iterate over the # of valid axes for the two transforms you are computing. And again, a code comment explaining what you are testing would be helpful.

tests/test_transform.py
```diff
@@ -668,13 +668,104 @@ def checkGetJacobian(self, fromName, toName):
...
+
+    def checkOfChaining(self):
+        """Test that both conventions for chaining Transform*To*.of give
+        the same result
+        """
+        # Getting order of operations wrong should throw
```
r-owen: Please delete this comment. Or explicitly add a test for it right after the comment, if you prefer.

python/lsst/afw/geom/transform/transform.cc
```diff
@@ -51,10 +51,19 @@ std::string formatStr(Class const &self, std::string const &pyClassName) {
     return os.str();
 }
 
+template <class ExtraEndpoint, class FromEndpoint, class ToEndpoint, class PyClass>
+void declareOf(PyClass &cls) {
+    using PrefixClass = Transform<ExtraEndpoint, FromEndpoint>;
+    using CalledClass = Transform<FromEndpoint, ToEndpoint>;
```
kfindeisen: Is

    (FinalTransform (SecondTransform::*)(FirstTransform const &) const) & 
                     SecondTransform::template of<ExtraEndpoint>
clear?

python/lsst/afw/geom/transform/transformContinued.py
```diff
@@ -33,10 +35,20 @@ def getJacobian(self, x):
                     self.getFromEndpoint().getNAxes())
     return matrix
 
+
+def of(self, before):
+    if isinstance(before.getToEndpoint(), type(self.getFromEndpoint())):
```
kfindeisen: Using the equality test didn't occur to me, that is cleaner.

The original code had `type(...) == type(...)`, but `flake8` didn't like it.

tests/test_transform.py
```diff
@@ -668,13 +668,104 @@ def checkGetJacobian(self, fromName, toName):
...
+        fromName : string
+            the prefix of the starting endpoint (e.g., "Point2" for a
+            Point2Endpoint) for the final, concatenated Transform
+        midName : string
+            the prefix for the shared endpoint where two Transforms will be
+            concatenated
+        toName : string
+            the prefix of the ending endpoint for the final, concatenated
+            Transform
+        """
```
kfindeisen: Ok, I've modernized to numpydoc.

tests/test_transform.py
kfindeisen: Why is checking all numbers of axes relevant here?

python/lsst/afw/geom/transform/transformContinued.py
r-owen: I'm glad equality works.

I'm surprised that flake8 objected to `type(...) == type(...)`. It's perfectly valid, though one rarely should need to do it. For fun I made a small snippet and flake8 had no objections to it:

    a = "string"
    b = "another"
    assert type(b) == type(a)

tests/test_transform.py
r-owen: Great!

python/lsst/afw/geom/transform/transform.cc
r-owen: Much nicer. Thanks!

tests/test_transform.py
r-owen: If you are testing type mismatch then I think the main thing is to test at least one set of valid axes for a given endpoint type. Also, the #s of axes should match if that is allowed for the given types, so you know you are testing type mismatch instead of # of axes mismatch.

However, may be easier and more thorough to cycle through all valid #s of axes for each endpoint type. One could imagine that code testing both type mismatch and axis mismatch, e.g. cycle through all choices and if types differ or #s of axes differ then assert that `of` will throw an exception.

python/lsst/afw/geom/transform/transform.cc
```diff
@@ -51,10 +51,21 @@ std::string formatStr(Class const &self, std::string const &pyClassName) {
...
 
+template <class ExtraEndpoint, class FromEndpoint, class ToEndpoint, class PyClass>
+void declareMethodTemplates(PyClass &cls) {
+    using FirstTransform = Transform<ExtraEndpoint, FromEndpoint>;
+    using SecondTransform = Transform<FromEndpoint, ToEndpoint>;
+    using FinalTransform = Transform<ExtraEndpoint, ToEndpoint>;
+    // Need Python-specific logic to give sensible errors for mismatched Transform types
+    cls.def("_of", (FinalTransform (SecondTransform::*)(FirstTransform const &) const) &
+                           SecondTransform::template of<ExtraEndpoint>,
+            "before"_a);
```
r-owen: Please use `"first"_a`

python/lsst/afw/geom/transform/transform.cc
```diff
@@ -51,10 +51,21 @@ std::string formatStr(Class const &self, std::string const &pyClassName) {
     return os.str();
 }
 
+template <class ExtraEndpoint, class FromEndpoint, class ToEndpoint, class PyClass>
+void declareMethodTemplates(PyClass &cls) {
+    using FirstTransform = Transform<ExtraEndpoint, FromEndpoint>;
+    using SecondTransform = Transform<FromEndpoint, ToEndpoint>;
+    using FinalTransform = Transform<ExtraEndpoint, ToEndpoint>;
+    // Need Python-specific logic to give sensible errors for mismatched Transform types
```
r-owen: Thanks!

python/lsst/afw/geom/transform/transform.cc
```diff
@@ -81,7 +92,14 @@ void declareTransform(py::module &mod, std::string const &fromName, std::string
     cls.def("tranInverse", (FromArray (Class::*)(ToArray const &) const) & Class::tranInverse, "array"_a);
     cls.def("tranInverse", (FromPoint (Class::*)(ToPoint const &) const) & Class::tranInverse, "point"_a);
     cls.def("getInverse", &Class::getInverse);
+    // Need some extra handling of ndarray return type in Python
```
r-owen: This is helpful but specifics would be appreciated, e.g. "Need some extra handling of ndarray return type in Python to avoid dimensions of length 1 from being deleted, leaving a 1-dimensional array or scalar"

(if a 1 x 1 matrix is compressed to numpy scalar? if not, delete the last two words)

python/lsst/afw/geom/transform/transformContinued.py
```diff
@@ -33,10 +35,20 @@ def getJacobian(self, x):
                     self.getFromEndpoint().getNAxes())
     return matrix
 
+
+def of(self, before):
+    if before.getToEndpoint() == self.getFromEndpoint():
+        return self._of(before)
```
r-owen: Please use `first` instead of `before`

tests/test_transform.py
```diff
@@ -668,13 +628,114 @@ def checkGetJacobian(self, fromName, toName):
...
+            with self.assertRaises(InvalidParameterError):
+                transform = transform2.of(transform1)
+
+        # Mismatched types of endpoints should fail
+        if fromName != midName:
+            joinNaxes = set(self.goodNaxes[fromName]).intersection(
+                self.goodNaxes[midName])
+            for nIn in self.goodNaxes[fromName]:
+                for nMid in joinNaxes:
+                    for nOut in self.goodNaxes[midName]:
```
r-owen: Did you mean `self.goodNaxes[toName]` here? On the other hand, the test works so maybe not.

kfindeisen: I'm keeping this subtest simple by making both Transforms the same type, but I agree that it's hard to spot. Any suggestions for how to make it more obvious?

r-owen: Perhaps add `outName = midName  # to keep the test simple` and then use `outName` when referring to the output of the second transform? Or at least add a comment explaining what is going on.

PR #215, opened 2017-04-11
DM-9572: Include instrument model in cameraGeom

jmeyers314:


python/lsst/afw/cameraGeom/camera.py
```diff
@@ -28,24 +28,35 @@
...
         self._nativeCameraSys = self._transformMap.getNativeCoordSys()
+        self._pupilFactoryClass = pupilFactoryClass
         super(Camera, self).__init__(detectorList)
 
     def getName(self):
         """!Return the camera name
         """
         return self._name
 
+    def getPupilFactory(self, *args, **kwargs):
```
SimonKrughoff: I assume you are trying to insulate against changes in the underlying API in the future, but since you call out the positional arguments in the docstring, I think it would be good to put them in the signature.  You can still say `def getPupilFactory(self, visitInfo, pupilSize, npix, *args, **kwargs)` if you want to allow additional positional and keyword arguments.

Another option may be to use the `@wraps` decorator from `functools` to forward the docstring from the factory class.

python/lsst/afw/cameraGeom/cameraFactory.py
```diff
@@ -94,10 +95,11 @@ def makeTransformDict(transformConfigDict):
     if transformConfigDict is not None:
         for key in transformConfigDict:
             transform = transformConfigDict[key].transform.apply()
-            resMap[CameraSys(key)] =  transform
+            resMap[CameraSys(key)] = transform
```
SimonKrughoff: This is not an issue here, but it's encouraged that whitespace only changes go in a separate commit.  Obviously, use common sense, but I wanted to mention it.

tests/testPupilFactory.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(pupilFactory.pupilSize, self.size)
+        self.assertEqual(pupilFactory.pupilScale, self.scale)
+        self.assertEqual(pupilFactory.npix, self.npix)
+        with self.assertRaises(NotImplementedError):
+            pupilFactory.getPupil(point=None)
+
+    def testBasePupilFactoryMethods(self):
+        pupilFactory = afwCameraGeom.PupilFactory(self.visitInfo, self.size, self.npix)
+        pupil = pupilFactory._fullPupil()
+        self.assertFloatsEqual(pupil.illuminated, 1.0)
```
SimonKrughoff: I was a little surprised that this worked since `illuminated` is a bool array, but it obviously does.

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -28,24 +28,35 @@
 class Camera(DetectorCollection):
     """!A collection of Detectors that also supports coordinate transformation
     """
-    def __init__(self, name, detectorList, transformMap):
+    def __init__(self, name, detectorList, transformMap, pupilFactoryClass):
```
jmeyers314: >Can this be defaulted to the ABC?

Sure.  I'll go ahead and eliminate the PRs for the obs packages (and I think pipe_tasks too) that are no longer required.

python/lsst/afw/cameraGeom/camera.py
jmeyers314: tbh I don't recall why I used `*args`, `**kwargs`.  I don't actually envision any additional arguments though, so I just changed this to `visitInfo, pupilSize, npix` full stop.

jmeyers314:
>One comment on the design is that you decide to return new Pupils from the private methods, though you could also have modified in place. The way you did it seems more pythonic, but we do a lot of modifying of array like things in place in other places in afw.

I went ahead and changed this to modify in place.  Should be slightly more efficient, but more importantly, I want to be consistent with the rest of afw.

PR #220, opened 2017-04-20
DM-4639: modernize afw code and reduce doxygen errors

kfindeisen:
This is a large set of style changes to `afw`; see [DM-4639](https://jira.lsstcorp.org/browse/DM-4639) for a discussion.

Changes to documentation and namespaces required significant manual intervention, and are therefore most likely to have errors or missed corrections. The `autopep8` pass also did not make the files flake8-compliant, for unknown reasons; I corrected most remaining issues but found all but the most egregious line length violations too difficult to correct by hand.


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -650,11 +665,13 @@ def getCcdInCamBBoxList(ccdList, binSize, pixelSize_o, origin):
...
         ey = cbbox.getDimensions().getY()//binSize
-        bbox = afwGeom.Box2I(cbbox.getMin(), afwGeom.Extent2I(int(ex), int(ey)))
+        bbox = afwGeom.Box2I(
+            cbbox.getMin(), afwGeom.Extent2I(int(ex), int(ey)))
         bbox = rotateBBoxBy90(bbox, nQuarter, bbox.getDimensions())
         bbox.shift(afwGeom.Extent2I(int(llc.getX()//pixelSize_o.getX()/binSize),
                                     int(llc.getY()//pixelSize_o.getY()/binSize)))
-        bbox.shift(afwGeom.Extent2I(-int(origin.getX()//binSize), -int(origin.getY())//binSize))
+        bbox.shift(afwGeom.Extent2I(-int(origin.getX() //
+                                         binSize), -int(origin.getY())//binSize))
```
timj: This one looks inconsistent. I think it should newline after the `binSize),` and use consistent spacing with the `//` in both.

python/lsst/afw/cameraGeom/cameraGeomEnumDicts.py
```diff
@@ -23,9 +23,10 @@
...
-    cameraGeomLib.FOCUS:     "FOCUS",
-    cameraGeomLib.GUIDER:    "GUIDER",
+    cameraGeomLib.SCIENCE: "SCIENCE",
+    cameraGeomLib.FOCUS: "FOCUS",
+    cameraGeomLib.GUIDER: "GUIDER",
     cameraGeomLib.WAVEFRONT: "WAVEFRONT",
 }
-DetectorTypeNameValDict = {val: key for key, val in DetectorTypeValNameDict.items()}
+DetectorTypeNameValDict = {val: key for key,
+                           val in DetectorTypeValNameDict.items()}
```
timj: I don't like this line break because `key, val` seem like they should be kept together for readability. Can it split after the `in` instead?

python/lsst/afw/table/ampInfo/ampInfoContinued.py
```diff
@@ -35,4 +35,5 @@
     UR: "UR",
     UL: "UL",
 }
-ReadoutCornerNameValDict = {val: key for key, val in ReadoutCornerValNameDict.items()}
+ReadoutCornerNameValDict = {val: key for key,
+                            val in ReadoutCornerValNameDict.items()}
```
timj: `key, val` breakage as above.

python/lsst/gdb/afw/printers.py
```diff
@@ -87,18 +91,20 @@ def __init__(self, val):
...
 
     def getEigenMatrixDimensions(val):
         m_storage = val["m_storage"]
         try:
             nx, ny = m_storage["m_cols"], m_storage["m_rows"]
         except gdb.error:           # only available for dynamic Matrices
             try:
-                nx, ny = val.type.template_argument(1), val.type.template_argument(2)
+                nx, ny = val.type.template_argument(
+                    1), val.type.template_argument(2)
```
timj: I don't think this is a good place to add a new line.

python/lsst/gdb/afw/printers.py
```diff
@@ -752,7 +796,7 @@ def to_string(self):
                                                          self.val["_numIter"],
                                                          self.val["_andMask"])
 
-    #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+    # -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
```
timj: I don't think anyone would object if we removed lines like this. They are no longer part of the coding standard so I'd prefer they disappeared in this older code.

tests/testAmpInfoTable.py
```diff
@@ -78,17 +78,23 @@ def testBasics(self):
             readNoise = -0.523
             linearityCoeffs = (1.1, 2.2, 3.3, 4.4)
             linearityType = "Polynomial"
-            bbox = afwGeom.Box2I(afwGeom.Point2I(3, -2), afwGeom.Extent2I(231, 320))
+            bbox = afwGeom.Box2I(afwGeom.Point2I(
+                3, -2), afwGeom.Extent2I(231, 320))
```
timj: bad place to split.

tests/testApCorrMap.py
```diff
@@ -70,21 +72,27 @@ def compare(self, a, b):
...
         mapping."""
         self.assertEqual(len(self.map), 3)
         self.assertEqual(collections.OrderedDict(self.map),
                          {name: value for (name, value) in list(self.map.items())})
-        self.assertEqual(list(collections.OrderedDict(self.map).keys()), list(self.map.keys()))
-        self.assertEqual(list(collections.OrderedDict(self.map).values()), list(self.map.values()))
-        self.assertEqual(list(collections.OrderedDict(self.map).items()), list(self.map.items()))
-        self.assertEqual(list(collections.OrderedDict(self.map).keys()), list(self.map))
+        self.assertEqual(list(collections.OrderedDict(
+            self.map).keys()), list(self.map.keys()))
```
timj: These entries should split after the `keys()),` I think.

tests/testApproximate.py
```diff
@@ -53,7 +53,8 @@ def makeRamp(self, binsize=1):
         rampCoeffs = (1000, 1, 1)
         for i in range(ramp.getHeight()):
             for j in range(ramp.getWidth()):
-                ramp.set(j, i, (rampCoeffs[0] + rampCoeffs[1]*x[j] + rampCoeffs[2]*y[i], 0x0, var))
+                ramp.set(
+                    j, i, (rampCoeffs[0] + rampCoeffs[1]*x[j] + rampCoeffs[2]*y[i], 0x0, var))
```
timj: This might be better splitting before the `0x0`.

tests/testBackground.py
```diff
@@ -163,7 +168,8 @@ def testRamp(self):
         bctrl.setInterpStyle(afwMath.Interpolate.CUBIC_SPLINE)
         bctrl.setNxSample(6)
         bctrl.setNySample(6)
-        bctrl.getStatisticsControl().setNumSigmaClip(20.0)  # large enough to entirely avoid clipping
+        bctrl.getStatisticsControl().setNumSigmaClip(
+            20.0)  # large enough to entirely avoid clipping
```
timj: Bad split. Move comment.

tests/testBox.py
```diff
@@ -197,41 +204,63 @@ def testRelations(self):
...
-        self.assertFalse(box.contains(geom.Box2I(geom.Point2I(-2, -3), geom.Point2I(3, 1))))
-        self.assertFalse(box.contains(geom.Box2I(geom.Point2I(-3, -3), geom.Point2I(2, 1))))
-        self.assertFalse(box.contains(geom.Box2I(geom.Point2I(-3, -4), geom.Point2I(2, 1))))
-        self.assertTrue(box.overlaps(geom.Box2I(geom.Point2I(-2, -3), geom.Point2I(2, 2))))
-        self.assertTrue(box.overlaps(geom.Box2I(geom.Point2I(-2, -3), geom.Point2I(3, 1))))
-        self.assertTrue(box.overlaps(geom.Box2I(geom.Point2I(-3, -3), geom.Point2I(2, 1))))
-        self.assertTrue(box.overlaps(geom.Box2I(geom.Point2I(-3, -4), geom.Point2I(2, 1))))
-        self.assertTrue(box.overlaps(geom.Box2I(geom.Point2I(-1, -2), geom.Point2I(1, 0))))
+        self.assertFalse(box.contains(geom.Box2I(
+            geom.Point2I(-2, -3), geom.Point2I(2, 2))))
```
timj: These could all break in a more reasable location.

tests/testCamGeomFitsUtils.py
```diff
@@ -118,11 +119,13 @@ def testDetectorBuilder(self):
...
             # test that it raises without setting non-defaulted keywords
-            self.assertRaises(Exception, DetectorBuilder(tmpFile, [tmpFile, tmpFile]))
+            self.assertRaises(Exception, DetectorBuilder(
+                tmpFile, [tmpFile, tmpFile]))
             # ignore non-defaulted keywords
             detBuilder = DetectorBuilder(tmpFile, [tmpFile, ], doRaise=False)
             detBuilder.makeCalib()
-            detBuilder.makeExposure(afwImage.ImageF(10, 10), afwImage.MaskU(10, 10), afwImage.ImageF(10, 10))
+            detBuilder.makeExposure(afwImage.ImageF(
+                10, 10), afwImage.MaskU(10, 10), afwImage.ImageF(10, 10))
```
timj: Bad place to split.

tests/testChebyshevBoundedField.py
```diff
@@ -98,20 +100,26 @@ def testEvaluate(self):
...
-            z2 = np.array([np.dot(ty[:, i], np.dot(coefficients, tx[:, i])) for i in range(nPoints)])
+            z2 = np.array([np.dot(ty[:, i], np.dot(coefficients, tx[:, i]))
+                           for i in range(nPoints)])
             self.assertFloatsAlmostEqual(z1, z2, rtol=1E-12)
 
             scaled = field*factor
-            self.assertFloatsAlmostEqual(scaled.evaluate(x, y), factor*z2, rtol=factor*1E-13)
-            self.assertFloatsEqual(scaled.getCoefficients(), factor*field.getCoefficients())
+            self.assertFloatsAlmostEqual(scaled.evaluate(
+                x, y), factor*z2, rtol=factor*1E-13)
```
timj: split after the `y),`?

tests/testChebyshevBoundedField.py
```diff
@@ -125,7 +133,8 @@ def testIntegrateTrivialBox(self):
         actually (-0.5, 1.5) when converted to a Box2D), and the translation
         doesn't affect the integral.
         """
-        bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(0, 0), lsst.afw.geom.Point2I(1, 1))
+        bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(
+            0, 0), lsst.afw.geom.Point2I(1, 1))
```
timj: bad split.

tests/testChebyshevBoundedField.py
```diff
@@ -300,7 +325,8 @@ def testTruncate(self):
                     if config3.triangular and i + j > max(config3.orderX, config3.orderY):
                         self.assertEqual(field3.getCoefficients()[i, j], 0.0)
                     else:
-                        self.assertEqual(field3.getCoefficients()[i, j], field1.getCoefficients()[i, j])
+                        self.assertEqual(field3.getCoefficients()[
+                                         i, j], field1.getCoefficients()[i, j])
```
timj: split after `j],`

tests/testColor.py
```diff
@@ -58,23 +58,28 @@ def testPhotom(self):
         self.assertEqual(0.0, self.calib.getFluxMag0()[1])
         self.assertEqual(22.5, self.calib.getMagnitude(flux))
         # Error just in flux
-        self.assertAlmostEqual(self.calib.getMagnitude(flux, fluxErr)[1], 2.5/math.log(10)*fluxErr/flux)
+        self.assertAlmostEqual(self.calib.getMagnitude(flux, fluxErr)[
+                               1], 2.5/math.log(10)*fluxErr/flux)
```
timj: Split after `1],`

tests/testColor.py
```diff
@@ -58,23 +58,28 @@ def testPhotom(self):
...
         # Error just in flux0
         self.calib.setFluxMag0(flux0, flux0Err)
         self.assertEqual(flux0Err, self.calib.getFluxMag0()[1])
-        self.assertAlmostEqual(self.calib.getMagnitude(flux, 0)[1], 2.5/math.log(10)*flux0Err/flux0)
+        self.assertAlmostEqual(self.calib.getMagnitude(
+            flux, 0)[1], 2.5/math.log(10)*flux0Err/flux0)
```
timj: Split after `[1],`

tests/testColor.py
```diff
@@ -129,15 +136,17 @@ def testCtorFromMetadata(self):
         # Error just in flux
         self.calib.setFluxMag0(flux0, 0)
 
-        self.assertAlmostEqual(self.calib.getMagnitude(flux, fluxErr)[1], 2.5/math.log(10)*fluxErr/flux)
+        self.assertAlmostEqual(self.calib.getMagnitude(flux, fluxErr)[
+                               1], 2.5/math.log(10)*fluxErr/flux)
```
timj: split after `1],`

tests/testColor.py
```diff
@@ -277,26 +287,31 @@ def testFilterProperty(self):
...
+                self.assertEqual(sorted(f.getAliases()),
+                                 sorted(self.aliases[name0]))
                 self.assertEqual(f.getId(), f0.getId())
                 self.assertEqual(f.getName(), name)
                 self.assertEqual(afwImage.Filter(f.getId()).getName(), name0)
                 self.assertEqual(f.getCanonicalName(), name0)
                 self.assertNotEqual(f.getCanonicalName(), name)
-                self.assertEqual(f.getFilterProperty().getLambdaEff(), f0.getFilterProperty().getLambdaEff())
+                self.assertEqual(f.getFilterProperty().getLambdaEff(
+                ), f0.getFilterProperty().getLambdaEff())
```
timj: split at the comma.

tests/testColor.py
```diff
@@ -305,20 +320,26 @@ def testReset(self):
...
+            self.defineFilterProperty("g", self.g_lambdaEff + 10)
 
-        self.defineFilterProperty("g", self.g_lambdaEff)  # identical redefinition is allowed
+        # identical redefinition is allowed
+        self.defineFilterProperty("g", self.g_lambdaEff)
 
-        afwImage.Filter.define(g, afwImage.Filter("g").getId())  # OK if Id's the same
-        afwImage.Filter.define(g, afwImage.Filter.AUTO)         # AUTO will assign the same ID
+        afwImage.Filter.define(g, afwImage.Filter(
+            "g").getId())  # OK if Id's the same
```
timj: Bad split. Move comment.

tests/testColor.py
```diff
@@ -305,20 +320,26 @@ def testReset(self):
...
+        # AUTO will assign the same ID
+        afwImage.Filter.define(g, afwImage.Filter.AUTO)
 
         with self.assertRaises(pexExcept.RuntimeError):
-            afwImage.Filter.define(g, afwImage.Filter("g").getId() + 10)  # different ID
+            afwImage.Filter.define(g, afwImage.Filter(
+                "g").getId() + 10)  # different ID
```
timj: no need to split if comment is moved.

tests/testConvolve.py
```diff
@@ -97,16 +98,19 @@ def refConvolve(imMaskVar, xy0, kernel, doNormalize, doCopyEdge):
     # transpose of what they are today.  Rather than transpose the logic in this function or put
     # transposes throughout the rest of the file, I have transposed only the inputs and outputs.
     #  - Jim Bosch, 3/4/2011
-    image, mask, variance = (imMaskVar[0].transpose(), imMaskVar[1].transpose(), imMaskVar[2].transpose())
+    image, mask, variance = (imMaskVar[0].transpose(
+    ), imMaskVar[1].transpose(), imMaskVar[2].transpose())
```
timj: split after the comma not the `)`.

tests/testCoord.py
```diff
@@ -271,43 +287,57 @@ def testTicket1761(self):
...
         self.assertAlmostEqual(sep1, known1)
         self.assertAlmostEqual(sep2, known2)
 
         ######################
         # normalize and sep1, sep2 should both equal 54.7356
         norm = True
-        c1 = afwCoord.Coord(afwGeom.Point3D(0.1, 0.1, 0.1), 2000.0, norm, dfltLong)
-        c2 = afwCoord.Coord(afwGeom.Point3D(0.6, 0.6, 0.6), 2000.0, norm, dfltLong)
+        c1 = afwCoord.Coord(afwGeom.Point3D(
+            0.1, 0.1, 0.1), 2000.0, norm, dfltLong)
```
timj: Split at the `2000.0`.

tests/testApproximate.py
kfindeisen: How about splitting before `(rampCoeffs[0]`? Keeping the tuple on one line seems like better style than splitting it in half.

timj: I had not seen that the `0x0` was inside a tuple. Maybe splitting after the `i,`  is better.

r-owen: I think the code is readable enough. Surely we can live with autopep8's choices even if they aren't optimal? Unless the results are truly egregious, hand-formatting code is a waste of developer time. Furthermore autopep8 could be run again at any time, so any hand tuning is likely to be undone at a later point.

TallJimbo: IMO, "could" and "should" are very different things here.  Formatting new code with autopep8 is fine; formatting old code that may have been hand-formatted to make an important piece of logic readable is not okay, and we should neither assume it will happen nor encourage it.

kfindeisen: The question is a bit moot, because I've already made all the formatting changes I intend to.

As far as I can tell, though, autopep8 only intervenes if it sees a PEP8 violation. It does not, for example, remove line breaks that were added by hand.

include/lsst/afw/image/MaskedImage.h
```diff
@@ -79,20 +79,17 @@ class MaskedImage : public lsst::daf::base::Persistable,
...
+    typedef std::shared_ptr<image::Image<ImagePixelT>> ImagePtr;
     /// shared pointer to the Mask
-    typedef typename Mask<MaskPixelT>::Ptr MaskPtr;
+    typedef std::shared_ptr<image::Mask<MaskPixelT>> MaskPtr;
     /// shared pointer to the variance Image
-    typedef typename Image<VariancePixelT>::Ptr VariancePtr;
-    /// shared pointer to a MaskedImage
-    typedef std::shared_ptr<MaskedImage> Ptr;
-    typedef std::shared_ptr<const MaskedImage> ConstPtr;
+    typedef std::shared_ptr<image::Image<VariancePixelT>> VariancePtr;
```
pschella: No objections, but just curious why we are keeping these while dropping `Ptr` elsewhere?

include/lsst/afw/image/Wcs.h
```diff
@@ -503,12 +500,12 @@ PTR(Wcs) makeWcs(PTR(lsst::daf::base::PropertySet) const& fitsMetadata, bool str
...
  */
-PTR(Wcs) makeWcs(coord::Coord const & crval, geom::Point2D const & crpix,
+std::shared_ptr<Wcs> makeWcs(coord::Coord const & crval, geom::Point2D const & crpix,
                  double CD11, double CD12, double CD21, double CD22);
 
 namespace detail {
-    int stripWcsKeywords(PTR(lsst::daf::base::PropertySet) const& metadata, ///< Metadata to be stripped
-                         CONST_PTR(Wcs) const& wcs ///< A Wcs with (implied) keywords
+    int stripWcsKeywords(std::shared_ptr<lsst::daf::base::PropertySet> const& metadata, ///< Metadata to be stripped
+                         std::shared_ptr<Wcs const> const& wcs ///< A Wcs with (implied) keywords
```
pschella: Why do we care about the holder type?

include/lsst/afw/math/Kernel.h
```diff
@@ -129,9 +129,9 @@ using boost::serialization::make_nvp;
 
     public:
         typedef double Pixel;
-        typedef PTR(Kernel) Ptr;
-        typedef CONST_PTR(Kernel) ConstPtr;
-        typedef PTR(lsst::afw::math::Function2<double>) SpatialFunctionPtr;
+        typedef std::shared_ptr<Kernel> Ptr;
+        typedef std::shared_ptr<Kernel const> ConstPtr;
```
pschella: I'm confused. Where are we keeping and where are we removing `Ptr` member types? At first I thought we would remove `PTR` and `CONST_PTR` but keep the rest, but looking at the other PRs it seems like a bit of a mixed bag.

tests/testTableArchives.cc
```diff
@@ -606,16 +606,16 @@ BOOST_AUTO_TEST_CASE(LinearCombinationKernel2) {
...
     for (int i = 0; i < nComponents; ++i) {
         kernelList[i].reset(new afwMath::DeltaFunctionKernel(width, height, afwGeom::Point2I(i, i)));
         kernelParams[i] *= kernelParams[i]; // want positive amplitudes
     }
-    PTR(afwMath::LinearCombinationKernel) p1(
+    std::shared_ptr<afwMath::LinearCombinationKernel> p1(
         new afwMath::LinearCombinationKernel(kernelList, kernelParams)
     );
-    PTR(afwMath::LinearCombinationKernel) p2 = roundtrip(p1.get());
+    std::shared_ptr<afwMath::LinearCombinationKernel> p2 = roundtrip(p1.get());
```
pschella: I assume `roundtrip` does not take ownership here.

include/lsst/afw/image/MaskedImage.h
kfindeisen: I kept anything not strictly matching `Ptr` or `ConstPtr`, because these more complex typedefs seem useful for expressing class-dependent concepts (e.g., do you really want to remember the exact type for a `MaskedImage`'s variance component?)

Do you think we should eliminate all typedefs for `shared_ptr`, no matter what they're called or what their purpose seems to be?

include/lsst/afw/math/Kernel.h
kfindeisen: This one was a mistake, fixed now.

include/lsst/afw/image/MaskedImage.h
r-owen: That seems reasonable to me. We aren't using them widely, but are making some use of them, and it's simpler than having to use a pixel type to construct a pointer type.

src/math/ConvolveImage.cc
```diff
@@ -329,19 +240,14 @@ void afwMath::convolve(
...
-        KernelT const& kernel,      ///< convolution kernel
-        bool doNormalize,           ///< if true, normalize the kernel, else use "as is"
-        bool doCopyEdge)            ///< if false (default), set edge pixels to the standard edge pixel;
-                                    ///< if true, copy edge pixels from input and set EDGE bit of mask
+        OutImageT& convolvedImage,
+        InImageT const& inImage,
+        KernelT const& kernel,
+        bool doNormalize,
+        bool doCopyEdge)
+
```
TallJimbo: Extra newline (hopefully fixed by clang-format in a later commit).

src/math/Statistics.cc
```diff
@@ -398,25 +384,34 @@ namespace {
...
+     * @param msk mask
+     * @param var variance
+     * @param weights weights to apply to each pixel
      * @param flags  an integer (bit field indicating which statistics are to be computed
+     * @param weightsAreMultiplicative weights are multiplicative (not inverse)
+     * @param andMask mask of bad pixels
+     * @param calcErrorFromInputVariance estimate errors from variance
+     * @param doCheckFinite check for NaN/Inf
+     * @param doGetWeighted use the weights
+     * @param maskPropagationThresholds
```
TallJimbo: Should not be a Doxygen comment.

src/math/Statistics.cc
```diff
@@ -473,22 +468,35 @@ namespace {
...
+     *   @param var variance
+     *   @param weights weights to apply to each pixel
+     *   @param flags what to measure
+     *   @param clipinfo the center and cliplimit for the first clip iteration
+     *   @param weightsAreMultiplicative weights are multiplicative (not inverse)
+     *   @param andMask mask of bad pixels
+     *   @param calcErrorFromInputVariance estimate errors from variance
+     *   @param doCheckFinite check for NaN/Inf
+     *   @param doGetWeighted use the weights,
+     *   @param maskPropagationThresholds
```
TallJimbo: Should not be a Doxygen comment.

src/math/Statistics.cc
```diff
@@ -523,12 +531,11 @@ namespace {
...
 
-    /** percentile()
-     *
-     * @brief A wrapper using the nth_element() built-in to compute percentiles for an image
+    /**
+     * A wrapper using the nth_element() built-in to compute percentiles for an image
      *
      * @param img       an afw::Image
-     * @param quartile  the desired percentile.
+     * @param fraction the desired percentile.
```
TallJimbo: Should not be a Doxygen comment.

src/math/Statistics.cc
```diff
@@ -576,16 +583,13 @@ namespace {
...
- * @param quartile  the desired percentile.
- *
- */
     typedef std::tuple<double, double, double> MedianQuartileReturn;
 
+    /**
+     * A wrapper using the nth_element() built-in to compute median and Quartiles for an image
+     *
+     * @param img       an afw::Image
+     */
```
TallJimbo: Should not be a Doxygen comment.

src/math/Statistics.cc
```diff
@@ -1029,45 +1013,42 @@ std::pair<double, double> afwMath::Statistics::getResult(
...
 
+/**
+ * Specialisation for Masks; just calculate the "Sum" as the bitwise OR of all pixels
+ *
+ * @param msk Mask whose properties we want
+ * @param msk2 A mask to control which pixels
+ * @param var A variance
+ * @param flags Describe what we want to calculate
+ * @param sctrl Control how things are calculated
+ */
```
TallJimbo: Should this have been moved to a header?  Is it redundant with docs for the template?

src/math/detail/BasicConvolve.cc
```diff
@@ -53,12 +49,12 @@ namespace mathDetail = lsst::afw::math::detail;
...
-    /*
+    /**
      * Assert that the dimensions of convolvedImage, inImage and kernel are compatible with convolution.
      *
-     * @throw lsst::pex::exceptions::InvalidParameterError if convolvedImage dimensions != inImage dim.
-     * @throw lsst::pex::exceptions::InvalidParameterError if inImage smaller than kernel in width or h.
-     * @throw lsst::pex::exceptions::InvalidParameterError if kernel width or height < 1
+     * @throws lsst::pex::exceptions::InvalidParameterError if convolvedImage dimensions != inImage dim.
+     * @throws lsst::pex::exceptions::InvalidParameterError if inImage smaller than kernel in width or h.
+     * @throws lsst::pex::exceptions::InvalidParameterError if kernel width or height < 1
```
TallJimbo: Should not be a Doxygen comment.

src/math/detail/KernelImagesForRegion.cc
```diff
@@ -78,27 +69,17 @@ mathDetail::KernelImagesForRegion::KernelImagesForRegion(
...
-        lsst::afw::geom::Point2I const &xy0,    ///< xy0 of image
-        bool doNormalize,                       ///< normalize the kernel images?
-        ImagePtr bottomLeftImagePtr,            ///< kernel image and sum at bottom left of region
-        ImagePtr bottomRightImagePtr,           ///< kernel image and sum at bottom right of region
-        ImagePtr topLeftImagePtr,               ///< kernel image and sum at top left of region
-        ImagePtr topRightImagePtr)              ///< kernel image and sum at top right of region
+        KernelConstPtr const kernelPtr,
+        lsst::afw::geom::Box2I const &bbox,
+
+
```
TallJimbo: Unnecessary newlines.

src/math/detail/KernelImagesForRegion.cc
```diff
@@ -50,18 +46,13 @@ namespace afwGeom = lsst::afw::geom;
...
-        KernelConstPtr kernelPtr,               ///< kernel
-        lsst::afw::geom::Box2I const &bbox,      ///< bounding box of region of an image
-                                                ///< for which we want to compute kernel images
-                                                ///< (inclusive and relative to parent image)
-        lsst::afw::geom::Point2I const &xy0,    ///< xy0 of image for which we want to compute kernel images
-        bool doNormalize)                       ///< normalize the kernel images?
+        KernelConstPtr kernelPtr,
+        lsst::afw::geom::Box2I const &bbox,
+
+
```
TallJimbo: Unnecessary newlines.

src/table/Match.cc
```diff
@@ -59,14 +59,14 @@ struct CmpRecordPtr {
...
+ * Extract source positions from `set`, convert them to cartesian coordinates
+ * (for faster distance checks) and sort the resulting array of `RecordPos`
  * instances by declination. Records with positions containing a NaN are skipped.
  *
  * @param[in] set          set of sources to process
- * @param[out] positions   pointer to an array of at least @c set.size()
+ * @param[out] positions   pointer to an array of at least `set.size()`
  *                         RecordPos instances
- * @return                 The number of sources with positions not containing a NaN.
+ * @returns                 The number of sources with positions not containing a NaN.
```
TallJimbo: Should not be a Doxygen comment.

include/lsst/afw/geom/Transform.h
```diff
@@ -198,9 +198,9 @@ class Transform {
      *          this Transform to the result. Its inverse shall first apply the
      *          inverse of this Transform, then the inverse of `first`.
      *
-     * @throws InvalidParameterErrror Thrown if `first.getToEndpoint()` and
-     *                                `this->getFromEndpoint()` do not have
-     *                                the same number of axes.
+     * @throws pex::exceptions::InvalidParameterErrror Thrown if
+     *         `first.getToEndpoint()` and `this->getFromEndpoint()` do not
+     *         have the same number of axes.
```
kfindeisen: Good catch. I unfortunately can't test this change locally because Doxygen doesn't have any `pex::exceptions` docs to link to.

include/lsst/afw/detection/FootprintSet.h
```diff
@@ -119,14 +151,41 @@ class FootprintSet : public lsst::daf::base::Citizen {
...
+     * @param region the desired region
+     */
     FootprintSet(geom::Box2I region);
+    /**
+     * Copy constructor
+     *
+     * @param rhs the input FootprintSet
+     */
     FootprintSet(FootprintSet const&);
     FootprintSet(FootprintSet const& set, int rGrow, FootprintControl const& ctrl);
```
kfindeisen: Not sure I understand this comment. Doxygen reorders class members anyway, so order matters only on the summary list; in addition, detailed docs are displayed for only one overload at a time. So I'm not sure what this change would accomplish.

src/coord/Coord.cc
```diff
@@ -68,14 +66,20 @@ CoordSystemMap const getCoordSystemMap() {
...
+ *
+ * @param dAlpha Difference between RAs (or longitudes), alpha1-alpha2
+ * @param dDelta Difference between Decs (or latitudes), delta1-delta2
+ * @param cosDelta1 Cosine of the first Dec, cos(delta1)
+ * @param cosDelta2 Cosine of the second Dec, cos(delta2)
+ */
+afwGeom::Angle haversine(afwGeom::Angle const dAlpha,
+                         afwGeom::Angle const dDelta,
+                         double const cosDelta1,
+                         double const cosDelta2
```
kfindeisen: That would make the process of documenting code more complex than it needs to be, so I'd rather keep things consistent.

Currently, the [Doxygen settings](https://github.com/lsst/base/blob/master/doc/base.inc#L435) disallow documentation of anonymous namespace members. I think that's the right place to make decisions about what is/isn't included in the API documentation.

TallJimbo: I suppose it's less about anonymous namespaces and more about whether the documentation should be limited to public functions.  I'd be okay with documenting these with a `@internal` tag or something similar, but I don't like the idea of functions that can't be called outside a source file appearing in the documentation.

src/math/Statistics.cc
kfindeisen: This template specialization is not mentioned in the header, so I have nowhere to move it.

TallJimbo: That means we're documenting an implementation detail as if it was a public interface, right?

I suppose it's probably out of scope to fix this here, though, and it should be made moot by @pschella eventually anyway.

src/coord/Coord.cc
kfindeisen: Per the current style guide, public *and protected* functions must be documented. :wink:

I'm willing to add `@internal` for documentation of private/local code, but I do think trusting in the centralized settings is the better long-term solution. Perhaps @jonathansick can weigh in on how this decision affects DMTN-030-type plans?

include/lsst/afw/image/ImageExceptions.h.m4
```diff
@@ -9,13 +9,11 @@ dnl comment to go into the output file
...
-  *
-  * @ingroup afw
-  */
+/*
+ * LSST exception handler classes for Image exceptions
+ *
+ * There are currently no afw::image exceptions.  If there ever are, we
+ * need to refactor the lsst::lex::exceptions code (maybe the m4 file)
+ * to allow us to define new Exceptions without repeating that boilerplate
+ */
```
kfindeisen: What about the `Runtime.h.m4` the comment mentions? Is that something we no longer have?

PR #224, opened 2017-04-27
DM-8440: Create new Wcs class

r-owen:


python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+        Parameters
+        ----------
+        nIn, nOut : integers
+            the dimensions of the input and output data; see makeTwoWayPolyMap
+        inArray : ndarray
+            an array of size `nIn` representing the point at which the Jacobian
+            is measured
+
+        Returns
+        ----------
```
PaulPrice: Too much underlining (prolly not important, but looks funny).

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+
+    @staticmethod
+    def makeJacobian(nIn, nOut, inArray):
+        """Make a Jacobian matrix for the equation described by makeTwoWayPolyMap.
+
+        Parameters
+        ----------
+        nIn, nOut : integers
+            the dimensions of the input and output data; see makeTwoWayPolyMap
+        inArray : ndarray
```
PaulPrice: `np.ndarray`

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+        baseCoeff = 2.0 * 0.001
+        coeffs = np.empty((nOut, nIn))
+        for iOut in range(nOut):
+            coeffOffset = baseCoeff * iOut
+            for iIn in range(nIn):
+                coeffs[iOut, iIn] = baseCoeff * (iIn + 1) + coeffOffset
+                coeffs[iOut, iIn] *= inArray[iIn]
+        assert coeffs.ndim == 2
+        assert coeffs.shape == (nOut, nIn)
+        # Avoid spurious errors when comparing to a simplified array
```
PaulPrice: Does this comment apply to the preceding `assert`s?

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+        return coeffs
+
+    def checkTransformation(self, transform, mapping, msg=""):
+        """Check tranForward and tranInverse for a transform
+
+        Parameters
+        ----------
+        transform : Transform
```
PaulPrice: `lsst.afw.geom.Transform`

python/lsst/afw/cameraGeom/cameraSys.py
```diff
@@ -21,7 +21,7 @@
 #
 from __future__ import absolute_import, division, print_function
 
-from lsst.afw.geom import addTransformMapMethods
+from lsst.afw.geom.python import addTransformMapMethods
```
r-owen: The previous commit added a new function `lsst.afw.geom.python.addTransformMethods` but left the pre-existing function `lsst.afw.geom.addTransformMapMethods` where it was. This commit moves that second function down into the `python` sub-namespace.

python/lsst/afw/geom/python/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import
```
r-owen: It doesn't really matter because `__init__.py` files never contain code, only imports. I agree it would be simpler to do the full import everywhere, but our convention seems to be to only import this much in `__init__.py` files and it's done pretty consistently throughout the packages I know of, so I just went with along with it.

python/lsst/afw/geom/python/transform.py
```diff
@@ -0,0 +1,58 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+"""pybind11 utilities for Transform classes
+
+@anchor pybind11_transform Transforms are instantiations of
```
r-owen: Good point. I think Doxygen is still being run on Python but new Python code should use numpydoc and I'll switch. I hope it has linking.

python/lsst/afw/geom/python/transform.py
```diff
@@ -46,13 +49,34 @@ def of(self, first):
...
 
 
+def saveToFile(self, path):
+    """Save this @ref pybind11_transform "Transform" to the specified file
+    """
+    className = type(self).__name__
+    bodyText = self.getFrameSet().show()
+    with open(path, "w") as outFile:
+        outFile.write(className + "\n")
+        outFile.write(bodyText)
```
r-owen: The format *is* human-readable and is an AST standard, and very flexible. Pickle support is a good idea and something I'd like to add later. I suspect pickle support will consist of pickling the AST string (without this extra header).

src/geom/Transform.cc
```diff
@@ -51,28 +51,29 @@ Transform<FromEndpoint, ToEndpoint>::Transform(ast::Mapping const &mapping, bool
 
 template <class FromEndpoint, class ToEndpoint>
 Transform<FromEndpoint, ToEndpoint>::Transform(ast::FrameSet const &frameSet, bool simplify)
-        : _fromEndpoint(frameSet.getNin()), _frameSet(), _toEndpoint(frameSet.getNout()) {
+        : Transform(simplify ? std::dynamic_pointer_cast<ast::FrameSet>(frameSet.simplify())
+                             : frameSet.copy()) {}
```
r-owen: That's what clang-format did. I don't always like clang-format's choices, but was really pleased with this one.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+    auto oldActualPixel0Index = ast::FrameSet::NOFRAME;
+    std::shared_ptr<ast::Frame> actualPixel0Frame;
+    auto pixelToActualPixelMap = frameSet->findFrame(ast::Frame(2, "Domain=ACTUAL_PIXEL0"));
+    if (pixelToActualPixelMap) {
+        actualPixel0Frame = frameSet->getFrame(ast::FrameSet::CURRENT);
+        oldActualPixel0Index = frameSet->getCurrent();
+    }
+
+    // Compute new mapping from GRID to PIXEL0
+    std::vector<double> shiftVec = {dx, dy};
```
PaulPrice: `shift`.  We know it's a vector from the type declaration.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+/**
+Compute the pixel position from the sky position
+*/
+std::pair<double, double> SkyWcs::skyToPixel(Angle const& ra, Angle const& dec) const {
+    auto pixel = tranInverse(SpherePoint(ra, dec));
+    return std::pair<double, double>(pixel[0], pixel[1]);
+};
+
+SkyWcs::SkyWcs(std::shared_ptr<ast::FrameSet> &&frameSet)
+        : Transform<Point2Endpoint, SpherePointEndpoint>(std::move(frameSet)){};
```
PaulPrice: Space before `{}`.
Is there a reason this isn't defined in the header, since it's so simple?

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+
+std::shared_ptr<ast::FrameSet> SkyWcs::_checkFrameSet(ast::FrameSet const & frameSet) const {
+    // checking alters the frameSet current pointer, so use a copy
```
PaulPrice: This is a recurring theme, and it's going to get someone into trouble eventually. I think you want an API for `FrameSet` that will check but not alter (or check and reset).

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+    auto frameSetCopy = frameSet.copy();
+    int baseIndex = frameSetCopy->getBase();
+    int currentIndex = frameSetCopy->getCurrent();
+
+    auto baseFrame = frameSetCopy->getFrame(baseIndex, false);
+    auto baseDomain = baseFrame->getDomain();
+    if (baseDomain != "ACTUAL_PIXEL0" && baseDomain != "PIXEL0") {
```
PaulPrice: Just checking: these are `std::string` comparisons, right?

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class TanSkyWcsTestCase(TransformTestBaseClass):
+    """Test pure TAN constructor of SkyWcs
+    """
+
+    def setUp(self):
+        TransformTestBaseClass.setUp(self)
+        self.crpix = Point2D(100, 100)
+        self.crval = SpherePoint(30 * degrees, 45 * degrees)
```
PaulPrice: You don't want to test something at RA=0 and at the poles?

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+        """Construct a pure TAN SkyWcs and check that it operates as specified
+
+        Parameters
+        ----------
+        orientation : lsst.afw.geom.Angle
+            Position angle of focal plane +Y, measured from N through E.
+            At 0 degrees, +Y is along N and +X is along E/W if flipX false/true
+            At 90 degrees, +Y is along E and +X is along S/N if flipX false/true
+        flipX : bool
+            Fip x axis? See `orientation` for details.
```
PaulPrice: "Fip"

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Returns
+        -------
+        wcs : SkyWcs
```
PaulPrice: `lsst.afw.geom.SkyWcs`

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+            The generated pure TAN SkyWcs
+        """
+        wcs = SkyWcs(
+            crpix = self.crpix,
+            crval = self.crval,
+            cdMatrix = makeCdMatrix(scale=self.scale, orientation=orientation, flipX=flipX),
+        )
+        self.checkPersistence(wcs)
```
PaulPrice: We'll need it to pickle too.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+        skyList = wcs.tranForward(pixelList)
+
+        # check pixels to sky
+        predSkyList = [
+            self.crval,
+            self.crval.offset(xoffAng - orientation, self.scale),
+            self.crval.offset(90*degrees - orientation, self.scale)
+        ]
+        self.assertSpherePointListsAlmostEqual(predSkyList, skyList)
+        self.assertSpherePointListsAlmostEqual(predSkyList, wcs.pixelToSky(pixelList))
```
PaulPrice: What sort of precision does this allow?  The points are pretty close, so is it possible you're making a huge relative error but small absolute error?

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+            self.assertPairsAlmostEqual(pixel, Point2D(*xyPair))
+
+        crval = wcs.getSkyOrigin()
+        self.assertSpherePointsAlmostEqual(crval, self.crval)
+
+        crpix = wcs.getPixelOrigin()
+        self.assertPairsAlmostEqual(crpix, self.crpix)
+
+        cdMatrix = wcs.getCdMatrix()
+        predCdMatrix = makeCdMatrix(scale=self.scale, orientation=orientation, flipX=flipX)
```
PaulPrice: Just save the one you made earlier.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertFloatsAlmostEqual(cdMatrix, predCdMatrix)
```
PaulPrice: Shouldn't they be exactly the same?  You're just getting out what you put in.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+
+        pixelScale = wcs.getPixelScale(self.crpix)
+        self.assertAnglesAlmostEqual(self.scale, pixelScale)
```
PaulPrice: Shouldn't they be exact?  At least set the allowed difference down to machine precision rather than leaving the default.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Compute a WCS with the pixel origin shifted by an arbitrary amount
+        # The resulting sky origin should not change
+        offset = Extent2D(500, -322)  # arbitrary
+        shiftedWcs = wcs.shiftedPixelOrigin(*offset)
```
PaulPrice: Use the `Extent2D` version of this method.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+        predShiftedPixelOrigin = self.crpix + offset
+        self.assertPairsAlmostEqual(shiftedWcs.getPixelOrigin(), predShiftedPixelOrigin)
```
PaulPrice: Again, the tolerance should be pretty tight.

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertSpherePointsAlmostEqual(shiftedWcs.getSkyOrigin(), self.crval)
```
PaulPrice: This should be exact, right?

tests/test_skyWcs.py
```diff
@@ -0,0 +1,123 @@
...
+
+        shiftedPixelList = [p + offset for p in pixelList]
+        shiftedSkyList = shiftedWcs.tranForward(shiftedPixelList)
+        self.assertSpherePointListsAlmostEqual(skyList, shiftedSkyList)
+
+        return wcs
+
+    def testOrient0FlipXFalse(self):
```
PaulPrice: Why multiple methods instead of a `for` loop?
You should iterate over the CRVAL too.  Be sure to include ra,dec = 0,0 and 0,+90 and 0,-90.  Each with and without flips and a few rotations.

include/lsst/afw/formatters/Utils.h
```diff
@@ -139,7 +141,24 @@ int extractVisitId(std::shared_ptr<lsst::daf::base::PropertySet const> const& pr
...
 std::string formatFitsProperties(lsst::daf::base::PropertySet const& prop);
+/**
+Format a PropertyList into a FITS header string
+
+@param[in] prop  Properties to format
+@param[in] excludeNames  Names of properties to exclude from the returned string
+@return a FITS header string (exactly 80 characters per "card", no line terminators)
+*/
+std::string formatFitsProperties(lsst::daf::base::PropertyList const& prop,
+                                 std::set<std::string> const& excludeNames = {});
```
PaulPrice: I think you use this in a previous commit, so this commit should precede that one.

include/lsst/afw/formatters/Utils.h
```diff
@@ -139,7 +141,24 @@ int extractVisitId(std::shared_ptr<lsst::daf::base::PropertySet const> const& pr
 int extractCcdId(std::shared_ptr<lsst::daf::base::PropertySet const> const& properties);
 int extractAmpId(std::shared_ptr<lsst::daf::base::PropertySet const> const& properties);
 
+/**
+Format a PropertySet into a single FITS header string
+
+@param[in] prop  Properties to format
+@return a FITS header string (exactly 80 characters per "card", no line terminators)
```
PaulPrice: I'd prefer getting a list of cards, so I can look at each individually or print it and have it make sense if my terminal isn't set to width=80. Concatenating them is trivial (`sum(cards, "")`), but splitting a single string is less trivial (not difficult, but annoying).

include/lsst/afw/formatters/Utils.h
```diff
@@ -139,7 +141,24 @@ int extractVisitId(std::shared_ptr<lsst::daf::base::PropertySet const> const& pr
...
+
+@warning PropertySet is unordered, so the order of entries in the returned string is unpredictable.
+*/
 std::string formatFitsProperties(lsst::daf::base::PropertySet const& prop);
```
PaulPrice: Add the `excludeNames` feature to this one?

src/formatters/Utils.cc
```diff
@@ -48,6 +50,54 @@ using lsst::daf::persistence::LogicalLocation;
...
+namespace {
+
+/**
+Format a PropertySet into a FITS header string (exactly 80 characters per "card", no line terminator)
+
+@param[in] paramNames  Names of properties to format
+@param[in] prop  Properties to format
+
+@todo Once we stop using the old WCS class we can remove the PropertySet version of formatFitsProperties
+and simplify this function to not worry about dotted names.
```
PaulPrice: Jira ticket.

src/formatters/Utils.cc
```diff
@@ -48,6 +50,54 @@ using lsst::daf::persistence::LogicalLocation;
...
+*/
+std::string formatFitsPropertiesImpl(std::vector<std::string> const& paramNames,
+                                     daf::base::PropertySet const& prop) {
+    std::ostringstream result;
+    for (auto const & fullName : paramNames) {
+        std::size_t lastPeriod = fullName.rfind(char('.'));
+        auto name = (lastPeriod == std::string::npos) ? fullName : fullName.substr(lastPeriod + 1);
+        std::type_info const& type = prop.typeOf(name);
+
+        std::string out = "";
```
PaulPrice: `out.reserve(80);`

src/formatters/Utils.cc
```diff
@@ -48,6 +50,54 @@ using lsst::daf::persistence::LogicalLocation;
...
+        } else {
+            out = (boost::format("%-8s= ") % name).str();
+        }
+
+        if (type == typeid(int)) {
+            out += (boost::format("%20d") % prop.get<int>(name)).str();
+        } else if (type == typeid(double)) {
+            out += (boost::format("%20.15g") % prop.get<double>(name)).str();
+        } else if (type == typeid(std::string)) {
+            out += (boost::format("'%-67s' ") % prop.get<std::string>(name)).str();
```
PaulPrice: Are we just chopping off long strings? I'd be very sad to lose the `CONTINUE` convention we've been using through cfitsio.
Which prompts the question: why aren't you using cfitiso to do this header wrangling? The less we directly interact with FITS, the better.

src/formatters/Utils.cc
```diff
@@ -48,6 +50,54 @@ using lsst::daf::persistence::LogicalLocation;
...
+        std::string out = "";
+        if (name.size() > 8) {  // Oh dear; too long for a FITS keyword
+            out += "HIERARCH = " + name;
+        } else {
+            out = (boost::format("%-8s= ") % name).str();
+        }
+
+        if (type == typeid(int)) {
+            out += (boost::format("%20d") % prop.get<int>(name)).str();
+        } else if (type == typeid(double)) {
```
PaulPrice: What about single-precision `float`s?

src/formatters/Utils.cc
```diff
@@ -48,6 +50,54 @@ using lsst::daf::persistence::LogicalLocation;
...
+            out += (boost::format("%20.15g") % prop.get<double>(name)).str();
+        } else if (type == typeid(std::string)) {
+            out += (boost::format("'%-67s' ") % prop.get<std::string>(name)).str();
+        }
```
PaulPrice: Need to trap types not in the above set that you deal with.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+                               invFrameSet.tranForward(rawOutArray),
+                               err_msg=msg)
+
+    def checkTransformFromMapping(self, fromName, toName):
+        """Check a Transform_<fromName>_<toName> using the Mapping constructor
+
+        Parameters
+        ----------
+        fromName, toName : string
+            one of self.endpointPrefixes
```
jonathansick: So it's technically allowed (e.g., https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections) but in the Developer Guide I don't talk about it because, yeah, you can't go wrong spelling out _exactly_ what each parameter is.

tests/test_transform.py
```diff
@@ -728,6 +733,45 @@ def checkOfChaining(self):
...
+        """Check persistence of a transform
+        """
+        className = type(transform).__name__
+        fileName = "persisted_{}.dat".format(className)
+        filePath = os.path.join(TestDir, fileName)
+        transform.saveToFile(filePath)
+        transformRoundTrip = afwGeom.readTransform(filePath)
+        self.assertEqual(type(transform), type(transformRoundTrip))
+        self.assertEqual(transform.getFrameSet().show(),
+                         transformRoundTrip.getFrameSet().show())
```
r-owen: Not at this time, not. It might be worth adding.

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+        }
+
+        # BadAxes is dict of endpoint class name prefix:
+        #    tuple containing 0 or more invalid numbers of axes
+        self.badNaxes = {
+            "Generic": (),  # all numbers of axes are valid for GenericEndpoint
+            "Point2": (1, 3, 4),
+            "Point3": (1, 2, 4),
+            "SpherePoint": (1, 3, 4),
+        }
```
r-owen: Since they are, technically, mutable I would rather not risk that.

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+    def makeEndpoint(name, nAxes):
+        """Make an endpoint
+
+        Parameters
+        ----------
+        name : string
+            one of "Generic", "Point2", "Point3" or "SpherePoint"
+        nAxes : integer
+            number of axes; ignored if the name is not "Generic"
+        """
```
r-owen: I think it's obvious from the method names, but I'll add it.

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+    @staticmethod
+    def makeEndpoint(name, nAxes):
+        """Make an endpoint
+
+        Parameters
+        ----------
+        name : string
+            one of "Generic", "Point2", "Point3" or "SpherePoint"
+        nAxes : integer
+            number of axes; ignored if the name is not "Generic"
```
r-owen: I'll use None. I also found that 0 was acceptable and added code to Endpoint to prohibit non-positive values.

python/lsst/afw/geom/testUtils.py
```diff
@@ -90,3 +99,757 @@ def __iter__(self):
...
+        """
+        endpointClassName = name + "Endpoint"
+        endpointClass = getattr(afwGeom, endpointClassName)
+        if name == "Generic":
+            return endpointClass(nAxes)
+        return endpointClass()
+
+    @classmethod
+    def makeGoodFrame(cls, name, nAxes):
+        """Return the appropriate frame for the given name and nAxes
```
r-owen: I've standardized on the description "Endpoint prefix"

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        inArray : ndarray
+            an array of size `nIn` representing the point at which the Jacobian
+            is measured
+
+        Returns
+        ----------
+        J : numpy.ndarray
+            an nOut x nIn array of first derivatives
+        """
+        baseCoeff = 2.0 * 0.001
```
r-owen: I'm not sure where the `2.0` comes from (I didn't write this code) but 0.001 comes from `makeTwoWayPolyMap` so I modified the code to make that clearer.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        mapping : astshim.Mapping
+            The mapping the transform should use. This mapping
+            must contain valid forward or inverse transformations,
+            but they need not match if both present. Hence the
+            mappings returned by make*PolyMap are acceptable.
+        msg : string
+            Error message suffix describing test parameters
+        """
+        fromEndpoint = transform.getFromEndpoint()
+        toEndpoint = transform.getToEndpoint()
```
r-owen: I'm not sure I understand the scope or intent of your remark. It is true that users will rarely get or manipulate an endpoint, but it seems appropriate for unit tests to use them to transform data from generic form to the specific form taken and returned by a Transform. It allows writing generic tests that can test many types of transforms without worrying about the specific data type of data that a given transform manipulates.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+            outPoint = toEndpoint.pointFromData(rawOutPoint)
+            rawOutArray = self.makeRawArrayData(nPoints, nOut)
+            outArray = toEndpoint.arrayFromData(rawOutArray)
+
+            self.assertFalse(transform.hasForward())
+
+        if mapping.hasInverse():
+            self.assertTrue(transform.hasInverse())
+            # inverse transformation of one point;
+            # remember that the inverse will not give the original values!
```
r-owen: It is described in the description of the `mapping` parameter of this method. I added a pointer

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+
+    def checkInverseTransformation(self, forward, inverse, msg=""):
+        """Check that two Transforms are each others' inverses.
+
+        Parameters
+        ----------
+        forward : Transform
+            the reference Transform to test
+        inverse : Transform
+            the transform that should be the inverse of `forward`
```
r-owen: References to parameter names are supposed to have backticks

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        # transformations of arrays of points
+        nPoints = 7  # arbitrary
+        rawInArray = self.makeRawArrayData(nPoints, fromEndpoint.getNAxes())
+        inArray = fromEndpoint.arrayFromData(rawInArray)
+        rawOutArray = self.makeRawArrayData(nPoints, toEndpoint.getNAxes())
+        outArray = toEndpoint.arrayFromData(rawOutArray)
+
+        if forward.hasForward():
+            self.assertEqual(forward.tranForward(inPoint),
+                             inverse.tranInverse(inPoint), msg=msg)
```
r-owen: That hasn't been implemented yet. Sounds like a good idea.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+
+        Parameters
+        ----------
+        fromName, toName : string
+            one of self.endpointPrefixes
+        fromAxes, toAxes : integer
+            number of axes in fromFrame and toFrame, respectively
+        """
+        transformClassName = "Transform{}To{}".format(fromName, toName)
+        transformClass = getattr(afwGeom, transformClassName)
```
r-owen: I'm not keen on that convention. It's a variable that holds a pointer to a class.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        baseMsg = "transformClass={}".format(transformClass.__name__)
+        for nIn in self.goodNaxes[fromName]:
+            for nOut in self.goodNaxes[toName]:
+                msg = "{}, nIn={}, nOut={}".format(baseMsg, nIn, nOut)
+
+                baseFrame = self.makeGoodFrame(fromName, nIn)
+                currFrame = self.makeGoodFrame(toName, nOut)
+                frameSet = self.makeFrameSet(baseFrame, currFrame)
+                self.assertEqual(frameSet.getNframe(), 4)
```
r-owen: I agree the name isn't wonderful, but I have chosen to expose AST conventions to avoid nuisance differences.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+                        with self.assertRaises(InvalidParameterError):
+                            transformClass(reallyBadFrameSet)
+                for badCurrFrame in self.makeBadFrames(toName):
+                    badFrameSet = self.makeFrameSet(baseFrame, badCurrFrame)
+                    with self.assertRaises(InvalidParameterError):
+                        transformClass(badFrameSet)
+
+                transform = transformClass(frameSet)
+
+                desStr = "{}[{}->{}]".format(transformClassName, nIn, nOut)
```
r-owen: It is the desired output from `str(frameSet)`.

python/lsst/afw/geom/testUtils.py
PaulPrice: I haven't looked over the `Transform` API in detail, and I haven't tried to use them so it may be that my comments aren't relevant.  However, I thought it was strange to bundle an `Endpoint` with the `Transform`, especially if as you say that "users will rarely get or manipulate an endpoint".  I would have thought for tests they would come instead from a helper function, e.g., `getFormatHelperForTransform(transform)`.  But maybe what the `Endpoint` (or `FormatHelper`) does is needed by the `Transform`, in which case it makes sense for it to live in the `Transform`.

OK, I've now looked at the `Transform` API, and I disagree with it.  I think a single `Transform` should operate on a variety of `Endpoint`s using overloads, rather than being templated over them.

python/lsst/afw/geom/testUtils.py
PaulPrice: I was wrong.

python/lsst/afw/geom/testUtils.py
RobertLuptonTheGood: Who did write the code?  It seems scary to have code we don't understand at this point in the package's history.

python/lsst/afw/geom/testUtils.py
PaulPrice: It's an actual class.  We used that convention with `CmdLineTask.RunnerClass` and `CmdLineTask.ConfigClass`.  We use this convention every time we write `class Foo`....

python/lsst/afw/geom/testUtils.py
PaulPrice: I think that's the wrong choice.  afw should be self-consistent.

RobertLuptonTheGood: I am confused.  We are wrapping AST, not adapting, it.  If there are potential "nuisance differences" don't we need to document them, not hide them?

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+
+        if frameSet.hasInverse():
+            nPoints = 7  # arbitrary
+            rawOutArray = self.makeRawArrayData(nPoints, nOut)
+            outArray = toEndpoint.arrayFromData(rawOutArray)
+            inArray = transform.tranInverse(outArray)
+            inData = fromEndpoint.dataFromArray(inArray)
+            inArrayRoundTrip = transformRoundTrip.tranInverse(outArray)
+            inDataRoundTrip = fromEndpoint.dataFromArray(inArrayRoundTrip)
+            assert_allclose(inData, inDataRoundTrip)
```
r-owen: I only see two places this is used, right next to each other, so I'll leave it alone for now.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        os.remove(filePath)
```
r-owen: I intentionally made it not `finally` so the file would be left behind on test failure. I'll add a code comment.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+Make FITS metadata for a pure tangent WCS
+
+@param[in] crpix  Center of projection on the CCD using the LSST convention:
+                        0, 0 is the lower left pixel of the image
+@param[in] crval  Center of projection on the sky
+@param[in] cdMatrix  CD matrix where element (i-1, j-1) corresponds to FITS keyword CDi_j
+                        and i, j have range [1, 2]
+*/
+std::shared_ptr<daf::base::PropertyList> makeTanWcsMetadata(Point2D const& crpix, SpherePoint const& crval,
+                                                            Eigen::Matrix2d const& cdMatrix);
```
r-owen: I am trying to replace the existing Wcs classes with SkyWcs.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+Read a FITS convention WCS FrameSet from FITS metadata
+
+The resulting FrameSet may be any kind of WCS supported by FITS;
+if it is a celestial WCS then 1,1 will be the lower left corner of the image
+(the FITS convention, not the LSST convention).
+
+@todo Add the ability to purge read fits cards.
```
r-owen: Holdover. I added the functionality but forgot to remove the TODO.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+
+@param[in,out] metadata  FITS header cards
+@param[in] strip  If true strip items from `metadata` that are used to create the WCS
+*/
+std::shared_ptr<ast::FrameSet> readFitsWcs(daf::base::PropertyList& metadata, bool strip=true);
+
+/**
+Read an LSST sky WCS FrameSet from a FITS header: 0,0 is the lower left corner of the parent image
+
+Read standard FITS WCS header cards plus optional LSST-specific keywords "LTV1" and "LTV2",
```
r-owen: Right. It's an IRAF standard. I have updated the text accordingly.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+which, if found, specify the position of the subimage in the parent image.
+
+Set the output SkyFrame's SkyRef to CRVAL and SkyRefIs="Ignore" so the SkyRef is ignored
+instead of being treated as an offset.
```
r-owen: The SkyRef frame already includes the effects of CRVAL. This is a way of recording that value for future reference without modifying the behavior of the already-correct WCS FrameSet.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+
+@warning the general WCS generated by LSST software cannot be exactly represented using
```
r-owen: I've updated the comment.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+for two purposes:
+- Read the standard FITS WCS found in raw data and other images from non-LSST observatories
+    and convert it to the LSST pixel convention.
+- Read the approximate FITS WCS that LSST writes to FITS images (for use by non-LSST code).
+
+The frames of the returned WCS will be as follows:
+- base: an ast::Frame with domain "PIXELS0": pixels with 0,0 the lower left corner of the parent image
+- current: an ast::SkyFrame: ICRS RA, Dec
+
+@todo Add the ability to purge read fits cards.
```
r-owen: I removed it.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+@param[in,out] metadata  FITS header cards
+@param[in] strip  If true strip items from `metadata` that are used to create the WCS,
+    excluding `LTV1` and `LTV2`, which are always retained.
+*/
+std::shared_ptr<ast::FrameSet> readLsstSkyWcs(daf::base::PropertyList& metadata, bool strip=false);
+
+}  // namespace detail
+}  // namespace geom
+}  // namespace afw
+}  // namespace lsst
```
r-owen: Technically yes, but clang-format will undo it. I prefer to live with clang-format.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+- base: an ast::Frame with domain "PIXELS0": pixels with 0,0 the lower left corner of the parent image
+- current: an ast::SkyFrame: ICRS RA, Dec
+
+@todo Add the ability to purge read fits cards.
+
+@param[in,out] metadata  FITS header cards
+@param[in] strip  If true strip items from `metadata` that are used to create the WCS,
+    excluding `LTV1` and `LTV2`, which are always retained.
+*/
+std::shared_ptr<ast::FrameSet> readLsstSkyWcs(daf::base::PropertyList& metadata, bool strip=false);
```
r-owen: The LSST convention is that we strip out the keywords when reading an Exposure. This is a more general function. Since we disagree on the default, this suggests that having a default is risky. I'll remove it.

python/lsst/afw/geom/detail/frameSetUtils.cc
```diff
@@ -0,0 +1,71 @@
+/*
+ * LSST Data Management System
+ * See COPYRIGHT file at the top of the source tree.
```
r-owen: It is now allowed, and even encouraged, but afw has no such file so I changed the above.

src/geom/detail/frameSetUtils.cc
```diff
@@ -0,0 +1,246 @@
...
+    pl->add("CRVAL1", crval[0].asDegrees());
+    pl->add("CRVAL2", crval[1].asDegrees());
+    pl->add("CRPIX1", crpix[0] + 1);
+    pl->add("CRPIX2", crpix[1] + 1);
+    pl->add("CD1_1", cdMatrix(0, 0));
+    pl->add("CD1_2", cdMatrix(0, 1));
+    pl->add("CD2_1", cdMatrix(1, 0));
+    pl->add("CD2_2", cdMatrix(1, 1));
+    pl->add("CUNIT1", "deg");
+    pl->add("CUNIT2", "deg");
```
r-owen: (Answered earlier. This new code is intended to eventually replace the old code.)

include/lsst/afw/geom/detail/frameSetUtils.h
PaulPrice: Actually, precedent supports `false`....

https://github.com/lsst/afw/blob/master/include/lsst/afw/image/Wcs.h#L472

src/geom/detail/frameSetUtils.cc
```diff
@@ -0,0 +1,246 @@
...
+    // Compute a frame set that maps from the original sky frame to our desired sky frame;
+    // this is what we will insert into the frame set.
+    // (Temporarily set the base frame to the sky frame, because findFrame
+    // produces a mapping from base to the found frame).
+    auto initialBaseIndex = frameSet->getBase();
+    frameSet->setBase(initialSkyIndex);
+    auto stdSkyFrameSet = frameSet->findFrame(stdSkyFrame);
+    if (!stdSkyFrameSet) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError,
+                          "Bug: found a SkyFrame the first time, but not the second time");
```
r-owen: I feel I've done what I can in the code comments before both blocks of code that call findFrame. I agree it's weird.

tests/test_frameSetUtils.py
```diff
@@ -0,0 +1,92 @@
+from __future__ import absolute_import, division, print_function
+import unittest
+
+from lsst.afw.geom import arcseconds, degrees, makeCdMatrix, Point2D, SpherePoint
+from lsst.afw.geom.detail import makeTanWcsMetadata, readFitsWcs, readLsstSkyWcs
+import lsst.utils.tests
+
+PrintStrippedNames = False
```
r-owen: You may be right but I'm going to resist this one. I hate uppercase with underscores.

tests/test_frameSetUtils.py
```diff
@@ -0,0 +1,92 @@
...
+        metadata.add("SIMPLE", True)
+        metadata.add("BITPIX", 16)
+        metadata.add("NAXIS", 2)
+        metadata.add("NAXIS1", 500)
+        metadata.add("NAXIS2", 200)
+        metadata.add("BZERO", 32768)
+        metadata.add("BSCALE", 1)
+        metadata.add("TIMESYS", "UTC")
+        metadata.add("UTC-OBS", "12:04:45.73")
+        metadata.add("DATE-OBS", "2006-05-20")
```
r-owen: I'd never do that. I guess I should state it's arbitrary.

tests/test_frameSetUtils.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def testReadFitsWcsStripMetadata(self):
+        metadata = self.makeMetadata()
+        self.assertEqual(len(metadata.toList()), 32)
+        readFitsWcs(metadata, strip=False)
+        self.assertEqual(len(metadata.toList()), 32)
+        readFitsWcs(metadata, strip=True)
+        self.assertEqual(len(metadata.toList()), 18)
+
+    def xtestReadLsstSkyWcsStripMetadata(self):
```
r-owen: Good catch. Leftover.

tests/test_frameSetUtils.py
PaulPrice: Well, leading capital letter is reserved for class names.

tests/test_frameSetUtils.py
PaulPrice: No need.  I was just curious.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+@return the CD matrix, where element (i-1, j-1) corresponds to FITS keyword CDi_j
+                        and i, j have range [1, 2]
+
+*/
+Eigen::Matrix2d makeCdMatrix(Angle const &scale, Angle const &orientation = 0 * degrees, bool flipX = false);
+
+/**
+A WCS that transform pixels to ICRS RA/Dec using the LSST standard for pixels
+
+The LSST standard is: 0,0 is the lower left pixel of the parent image
```
r-owen: OK. I hope RHL won't object.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+All are of class ast::Frame except the sky frame. The domain is listed first, if set:
+- ACTUAL_PIXEL0 (optional): actual pixels in LSST convention: 0,0 is lower left corner of parent image.
+    The 0 is to remind users that this frame is 0-based.
+- PIXEL0: nominal pixels in LSST convention: 0,0 is lower left corner of parent image
+- GRID: nominal pixels in FITS convention: 1,1 is lower left corner of subimage
+- IWC: intermediate world coordinates (the FITS WCS concept)
+- An ast::SkyFrame with System=ICRS and standard axis order RA, Dec.
+
+If ACTUAL_PIXEL0 is present then it will be the base frame, else PIXEL0 will be the base frame.
+The SkyFrame is always the current frame.
```
r-owen: Good point. I put in a bit more text. I am hoping it will suffice, and we can find a better place for if more is needed.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+
+Constructors must set the following properties of the contained ast::FrameSet:
+- Include the frames listed above, with specified domains.
+    Note that AST will provide the IWC frame when reading standard WCS FITS
+    data using an ast::FitsChan if you set "IWC=1" as a property of ast::FitsChan.
+    Other constructors may have to work harder.
+- For the output ast::SkyFrame set SkyRef=CRVAL (in radians) and SkyRefIs="Ignored".
+    This makes CRVAL easy to look up, and "Ignored" prevents CRVAL from being used as an offset.
+*/
+class SkyWcs : public Transform<Point2Endpoint, SpherePointEndpoint> {
```
r-owen: I'm willing to discuss this (preferably after merging this ticket). I think Wcs has some useful connotations but I agree it comes with a lot of baggage.

PaulPrice: I know you're feeling time pressure, but I fear that if the name doesn't get "fixed" before the merge it'll never happen --- especially if it's not RFC-ed or ticketed or an explicit element of an existing ticket.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+    */
+    explicit SkyWcs(Point2D const &crpix, SpherePoint const &crval, Eigen::Matrix2d const &cdMatrix);
+
+    /**
+    Construct a WCS from FITS keywords
+
+    In addition to standard FITS WCS keywords uses these LSST-specific keywords:
+    LTV1, LTV2: offset of subimage with respect to parent image
+
+    TODO implement this using readFrameSet and adding full standardization
```
r-owen: Good catch. I just added `strip` and that completes the TODO.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+    */
+    Point2D getPixelOrigin() const;
+
+    /**
+    Get CRVAL, the sky origin or celestial fiducial point
+    */
+    SpherePoint getSkyOrigin() const;
+
+    /**
+    Get the CD matrix at the pixel origin
```
r-owen: Perhaps you can clarify what sort of text you would like to see? I though the CD matrix was always a local linear approximation. In any case I'll say "Get the 2x2 CD matrix the pixel origin" since then it's clear it must be linear.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+    Get a local TAN WCS approximation to this WCS at the specified pixel position
+    */
+    SkyWcs getTanWcs(Point2D const &pixel) const;
+
+    /**
+    Return a copy of this SkyWcs with the pixel origin by the specified amount
+    */
+    SkyWcs shiftedPixelOrigin(double dx, double dy) const;
```
r-owen: I prefer two doubles because that's how it's usually called (which is rarely).

I agree the name could be better. I'll use `copyAtShiftedPixelOrigin`. It's verbose, but explicit.

include/lsst/afw/geom/SkyWcs.h
PaulPrice: I guess this is one of those places where I think we're carrying around FITS WCS baggage: who would care about a "CD matrix" but for FITS?  What one really cares about is a local linear approximation, which you could choose to make at the `getPixelOrigin()`.

How about:

    This is a locally linear approximation to the transformation between the focal plane and the projected tangent plane at the origin point.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+    Compute sky position(s) from pixel position(s)
+
+    This is another name for tranForward, plus an overload that takes a pair of doubles.
```
r-owen: I agree that the names are less than ideal, and we are going to have to have this discussion. But I'd rather keep it out of this ticket.

Transform is already in the stack and already has tranForward and tranInverse. This is a subclass of that. I don't think I can responsibly hide those methods, whatever we end up calling them.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+
+Eigen::Matrix2d makeCdMatrix(Angle const& scale, Angle const& orientation, bool flipX) {
+    Eigen::Matrix2d cdMatrix;
+    double orientRad = orientation.asRadians();
+    double scaleDeg = scale.asDegrees();
+    double xmult = flipX ? -1.0 : 1.0;
+    cdMatrix(0, 0) = std::cos(orientRad) * scaleDeg * xmult;
+    cdMatrix(0, 1) = std::sin(orientRad) * scaleDeg;
+    cdMatrix(1, 0) = -std::sin(orientRad) * scaleDeg * xmult;
+    cdMatrix(1, 1) = std::cos(orientRad) * scaleDeg;
```
r-owen: I expect the compiler to do that already.

include/lsst/afw/geom/SkyWcs.h
PaulPrice: We have a type specifically to carry two doubles with the meaning of an offset.  Please use it, or at least support it.  Even the old `Wcs` class supports it.  And `Box2D.shift` (and friends) use an `Extent2D` exclusively.  In the name of consistency of interfaces across afw, please at the very least support it.

include/lsst/afw/geom/SkyWcs.h
PaulPrice: I think that argues for composition instead of inheritance.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+    auto skyDy = skyVec[2].getVector() - skyLL;
+
+    // Compute |cross product| = area of parallelogram with sides dx,dy, in radians^2
+    // This is slightly incorrect -- it's making the small-angle
+    // approximation, taking the distance *through* the unit sphere
+    // rather than over its surface, but the difference should be negligible for pixels of normal size
+    double pixAreaRadSq = std::sqrt(square(skyDx[1] * skyDy[2] - skyDx[2] * skyDy[1]) +
+                                    square(skyDx[2] * skyDy[0] - skyDx[0] * skyDy[2]) +
+                                    square(skyDx[0] * skyDy[1] - skyDx[1] * skyDy[0]));
+    return std::sqrt(pixAreaRadSq) * radians;
```
r-owen: Probably. I'll switch.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+SpherePoint SkyWcs::getSkyOrigin() const {
+    // CRVAL is stored as the SkyRef property of the sky frame (the current frame of the SkyWcs)
+    auto const skyFrame = detail::getSkyFrame(*_frameSet, ast::FrameSet::CURRENT, false);
+    auto const crvalRad = skyFrame->getSkyRef();
+    return SpherePoint(crvalRad.data());
+}
+
+Eigen::Matrix2d SkyWcs::getCdMatrix() const {
+    // Use a copy because looking up the IWC frame alters the FrameSet
+    auto frameSet = getFrameSet()->copy();
```
r-owen: FrameSet is const so it is necessary. I agree that a "lookup and reset", or better yet, a const "lookup and give me the index" would be nice to have.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+    auto prototype = ast::Frame(2, "Domain=IWC");
+    auto iwcFrameSet = frameSet->findFrame(prototype);
+    if (!iwcFrameSet) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, "IWC frame not found");
```
r-owen: I disagree. I prefer to use `NotFoundError` for situations where the user expect a lookup to be done. This is not one of those. This problem can only occur if the `SkyWcs` is malformed and I don't expect that to ever happen.

src/geom/SkyWcs.cc
r-owen: A better solution is probably to cache these indices when the SkyWcs is created. That saves future failure to find them and avoids repeated lookup.

src/geom/SkyWcs.cc
r-owen: clang-format. As to complexity, I consider it messy enough that I'd rather leave it out of the header.

src/geom/SkyWcs.cc
r-owen: Yes

tests/test_skyWcs.py
r-owen: Agreed. But I'd rather hold off for another ticket.

tests/test_skyWcs.py
r-owen: SpherePoint tests look at angular separation on the sky and are only absolute, not relative. The default separation allowed is 0.001 arc seconds. It might be nice to expand the asserts to handle relative errors as well (feel free to file a ticket if you feel strongly about this).

tests/test_skyWcs.py
r-owen: No. The CD matrix is not stored directly as such in a FrameSet.

tests/test_skyWcs.py
r-owen: As above, but I'll make it epsilon radians * 10

tests/test_skyWcs.py
r-owen: No, but tight.

include/lsst/afw/formatters/Utils.h
r-owen: This is a pre-existing method that exists as an aid to code that needs FITS style header strings. I added a variant that preserves order (and am quite surprised the version we've been using does not).

src/formatters/Utils.cc
r-owen: This is pre-existing code, but I believe they'll come out as double here.

src/formatters/Utils.cc
r-owen: This is pre-existing code (adapted to an order-preserving variant). For its purposes we don't care about CONTINUE cards or COMMENT cards or any of that. We're trying to get WCS numbers out.

src/formatters/Utils.cc
r-owen: See above. The function can ignore anything that doesn't match one of these.

src/formatters/Utils.cc
r-owen: I added documentation to this effect to the two overloads and the implementation.

include/lsst/afw/formatters/Utils.h
PaulPrice: So is the order of operations:

1. FITS headers --> `PropertyList`
2. `PropertyList` --> FITS headers
3. FITS headers --> `Wcs`

?!?

include/lsst/afw/formatters/Utils.h
r-owen: No thanks. It isn't needed and I want to get rid of this version.

src/geom/SkyWcs.cc
PaulPrice: In that case, an `abort` would be appropriate.

tests/test_skyWcs.py
r-owen: I'm afraid I accidentally edited Paul's comment instead of my own. He said that Transform and SkyWcs should be pickleable. I agree and filed https://jira.lsstcorp.org/browse/DM-10416 to record that.

src/geom/SkyWcs.cc
PaulPrice: As I've said, it doesn't really matter here, but I was curious so I investigated.  The answer: gcc will, clang will not.
More details at http://stackoverflow.com/a/37877352/834250
Verified myself using https://godbolt.org .

tests/test_skyWcs.py
r-owen: Good point. Testing at the poles will require some extra work, but testing near the poles and at or near RA=0 can use the existing code.

include/lsst/afw/geom/detail/frameSetUtils.h
RobertLuptonTheGood: It is indeed an IRAF standard, but we use it to track {{XY0}} (so it isn't really optional), and I'm not quite sure that that's the IRAF usage.  I think we should probably document this.

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+#include "lsst/afw/geom/SkyWcs.h"
+#include "lsst/daf/base/PropertyList.h"
+#include "lsst/pex/exceptions.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+namespace {
+
+inline double square(double x) { return x * x; }
```
RobertLuptonTheGood: I think that the overload {{pow(float, int)}} was removed so it may not inline.

PaulPrice: `std::pow(x, 2)` translates to a single multiply in gcc 6.3 and clang 4.0: https://godbolt.org/g/6qLAky

include/lsst/afw/geom/SkyWcs.h
RobertLuptonTheGood: I agree that `(0.0, 0.0)` is the centre of the lower-left pixel.

include/lsst/afw/formatters/Utils.h
RobertLuptonTheGood: I'm not following this.  Is this a user-visible change, or purely internal?

include/lsst/afw/geom/detail/frameSetUtils.h
RobertLuptonTheGood: When the new code replaces the current code it *must* be called `Wcs`.  Is this a temporary name so you can merge to master and keep two versions of coordinate transform alive simultaneously?  If so, the ticket to rename `SkyWcs` back to `Wcs` must be filed before this ticket can be merged to master.

tests/test_frameSetUtils.py
TallJimbo: > You may be right but I'm going to resist this one. I hate uppercase with underscores.

And I hate camelcase acronyms like `Lsst`, `Wcs`, or `Psf`, but neither of these choices are developer prerogative.

include/lsst/afw/formatters/Utils.h
PaulPrice: It seems to me that it's intended to be an internal function for feeding a FITS header into AST to create a `SkyWcs`.  It's declared in a header file, but that header file isn't included anywhere that a user will pick it up (unless they explicitly include it themselves).

include/lsst/afw/geom/SkyWcs.h
TallJimbo: I'm okay with deferring these naming (and other design-related) questions to another ticket, but we should make sure we tackle them before we start to use `Transform` more broadly.

src/geom/SkyWcs.cc
TallJimbo: I strongly agree that we should not use `NotFoundError` for a malformed class problem.  That should be either an abort or `LogicError`, since it indicates a problem with the class's own code, not user inputs.

PaulPrice: But the general comment is that this submission tends to call everything a `RuntimeError`.  It's important to use the appropriate exception type, because the *type* of the exception is important for categorizing errors, both manually and in the code.

include/lsst/afw/geom/detail/frameSetUtils.h
timj: @RobertLuptonTheGood so is your intent not to use any of this WCS machinery for the spectrograph? `Wcs` is more generic name than what you are actually getting here. Maybe the spectrometer is also generating a `Wcs` even though it has a spectral frame in it?

SimonKrughoff: @RobertLuptonTheGood I tend to agree with @r-owen and @timj.  This naming makes it very clear that this class is specifically for transforming to/from sky based coordinates.  This is typically what we mean by WCS, but in reality WCS is much more general than that.

TallJimbo: I don't think there is an objective reality for what WCS means to all people.  Dave Monet regularly assumes it means "FITS WCS" and hence that it's not sufficient.  Everyone uses it to mean both "transform" and "coordinate system".  We've always used it in afw to mean "pixel to celestial" coordinate transformations (or the reverse), and that history is not irrelevant.

timj: What is the spectrograph plan then? No WCS at all for that? WCS means "the world" not "the sky". WCS meaning solely "pixel to sky" is a real optical imaging bias that we'd benefit from dropping.

TallJimbo: > WCS meaning solely "pixel to sky" is a real optical imaging bias that we'd benefit from dropping.

That sounds like an argument for calling this "CartesianToCelestialTransform", not "SkyWcs", which also has the exact same real optical imaging bias.  But we are at our core a real optical imaging project, and it seems reasonable to occasionally use shortcuts in nomenclature that reflect that.  It's also probably not reasonable to use such a shortcut in *every* context where we could.

In any case, if this is controversial, code review doesn't provide any way resolving that conflict; only RFC does.  And since "Wcs" is taken by an existing class right now, it's not going to be renamed to that in this ticket anyway, so it's not urgent.

python/lsst/afw/geom/python/transform.py
```diff
@@ -29,9 +29,12 @@
 
 from __future__ import absolute_import, division, print_function
 
-__all__ = ["addTransformMethods"]
+__all__ = ["addTransformMethods", "transformRegistry"]
 
-from lsst.pex.exceptions import InvalidParameterError
+import lsst.pex.exceptions
```
r-owen: Agreed. I may just undo it.

include/lsst/afw/formatters/Utils.h
r-owen: It is purely internal. The existing function is *only* used by Wcs and I have a ticket filed to get rid of it once we get rid of that class.

tests/test_skyWcs.py
r-owen: I don't think a TanWcs with CRVAL at the north and south poles is very well defined in the sense that there's no way to specify the position angle (which way is "N and E" the CCD). I'm going to leave off that test and just test very near the pole for now. If you think it important I will gladly look into this again when I understand better what to expect at the poles. Note: I realize it is possible to map X,Y <-> RA/Dec  even with CRVAL at a pole (and Wcs and SkyWcs will presumably both do so without complaining); I'm just saying that I think the code has to make an arbitrary decision to do so, which makes it harder to test.

include/lsst/afw/geom/detail/frameSetUtils.h
RobertLuptonTheGood: I don't want to perturb the design of the imaging pixel <--> sky mappings by considering the auxiliary spectrograph.

But to answer the question... it's a slitless spectrograph so the spatial part will be represented by {{Wcs}} like any other imaging data.  Having extracted the spectra we'll use some standard representation (e.g. Lagrange polynomials);  no need for something like AST.

timj: Why wouldn't you take advantage of AST's ability to handle spectroscopy WCS? Pixel to wavelength is just another mapping.

include/lsst/afw/geom/Transform.h
```diff
@@ -220,7 +220,6 @@ class Transform {
     */
     explicit Transform(std::shared_ptr<ast::FrameSet> &&frameSet);
 
-private:
```
r-owen: I want efficient access without allowing users to mess with the internals. Thus I can't expose the pointers with public accessors. I could add protected accessors but don't see enough advantage to that over just making the instance variables protected.

python/lsst/afw/geom/transform/transformContinued.py
```diff
@@ -21,17 +21,11 @@
 #
 from __future__ import absolute_import, division, print_function
 
-import astshim
```
r-owen: I think this ended up getting fixed. I can't find it anymore.

python/lsst/afw/geom/testUtils.py
```diff
@@ -91,3 +100,759 @@ def __iter__(self):
...
+        # tell unittest to use the msg argument of asserts as a supplement
+        # to the error message, rather than as the whole error message
+        self.longMessage = True
+
+        # list of endpoint class name prefixes; the full name is prefix + "Endpoint"
+        self.endpointPrefixes = ("Generic", "Point2", "Point3", "SpherePoint")
+
+        # GoodNaxes is dict of endpoint class name prefix:
+        #    tuple containing 0 or more valid numbers of axes
+        self.goodNaxes = {
```
r-owen: I went with `goodNAxes` for consistency with the getter.

python/lsst/afw/geom/testUtils.py
r-owen: @kfindeisen wrote it.

python/lsst/afw/geom/testUtils.py
r-owen: I'll be interested to learn more about your suggestion, but I hope you agree it is out of scope for this ticket.

PaulPrice: Yes, out of scope for this ticket.

python/lsst/afw/geom/testUtils.py
r-owen: OK. I'll change it.

python/lsst/afw/geom/testUtils.py
kfindeisen: The explicit product was to make it easier to keep in sync with Russell's code.

The 2 is because d/dx (x^2) = 2x.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -0,0 +1,113 @@
...
+#include "ndarray.h"
+
+#include "lsst/afw/geom/Endpoint.h"
+#include "lsst/afw/geom/Transform.h"
+#include "lsst/daf/base/PropertyList.h"
+
+namespace lsst {
+namespace afw {
+namespace geom {
+namespace detail {
```
r-owen: I hope these will not be widely used. I view these as I/O details users should never see.

include/lsst/afw/geom/detail/frameSetUtils.h
r-owen: Drat! I would prefer to take your original suggestion for now, since I think you make a good case.

PaulPrice: Do it, then.  No objection from me.

include/lsst/afw/geom/SkyWcs.h
r-owen: I don't think you need to worry about that. We will have some sort of design review before switching to this new class. I'm just hoping something of this code will remain afterwards.

include/lsst/afw/geom/SkyWcs.h
r-owen: You can get the local linear approximation of pixels to sky using `getJacobian` or `getTanWcs` (depending if you want a matrix or a WCS). I think `getCdMatrix` is a linear approximation to the mapping from pixels to `IWC`. It is mostly used in the SIP fitter at present. I propose to leave it in for now, but I hope we can get rid of it.

include/lsst/afw/geom/SkyWcs.h
r-owen: OK. I'll change it.

src/geom/SkyWcs.cc
r-owen: I think the code is more readable using `square`.

src/geom/SkyWcs.cc
r-owen: I'll make it `LogicError`

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+
+SkyWcs SkyWcs::shiftedPixelOrigin(double dx, double dy) const {
+    auto frameSet = getFrameSet()->copy();
+
+    // Save the SkyFrame so we can search for it at the end, to make it current again
+    auto skyFrameIndex = frameSet->getCurrent();
+
+    // Find the GRID frame
+    if (!frameSet->findFrame(ast::Frame(2, "Domain=GRID"))) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, "GRID frame not found");
```
r-owen: LogicError

src/geom/SkyWcs.cc
```diff
@@ -0,0 +1,249 @@
...
+    }
+    auto gridFrameIndex = frameSet->getCurrent();
+
+    // Find the PIXEL0 frame and the old mapping from GRID to PIXEL0
+    frameSet->setBase(gridFrameIndex);
+    auto oldGridToPixel0Map = frameSet->findFrame(ast::Frame(2, "Domain=PIXEL0"));
+    if (!oldGridToPixel0Map) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError, "PIXEL0 frame not found");
```
r-owen: LogicError

src/geom/SkyWcs.cc
r-owen: But `shift` is now the name of the `Extent2D` argument, so I'm retaining `shiftVec`. I know you hate Hungarian notation but I find that in very small doses it can sometimes be helpful for disambiguating variable names, and I think this is precisely that sort of situation.

src/geom/SkyWcs.cc
r-owen: Agreed, but not this ticket, please.

include/lsst/afw/formatters/Utils.h
r-owen: To answer @PaulPrice's earlier question, yes. We unpersist several kinds of objects based on FITS headers (`Wcs`, `Calib`, `VisitInfo`) and it all works the same way: we read the FITS header into a `PropertyList`, then use the `PropertyList` to create whatever kind of object we want. This is not ideal, especially since `PropertyList` cannot handle all kinds of FITS headers (for instance those with a name but no value, meaning "unknown value"). However, it seems to have sufficed, so far.

src/geom/SkyWcs.cc
PaulPrice: It's not inappropriate in this case.  (:

PR #228, opened 2017-05-15
DM-10429: Write a version of the warper that uses SkyWcs and compare performance

r-owen:


src/math/warpExposure.cc
```diff
@@ -494,6 +495,232 @@ int warpImage(DestImageT &destImage, SrcImageT const &srcImage, geom::XYTransfor
...
+
+    // Set each pixel of destExposure's MaskedImage
+    LOGL_DEBUG("TRACE3.afw.math.warp", "Remapping masked image");
+
+    int const maxCol = destWidth - 1;
+    int const maxRow = destHeight - 1;
+
+    detail::WarpAtOnePoint<DestImageT, SrcImageT> warpAtOnePoint(srcImage, control, padValue);
+
+    if (interpLength > 0) {
```
pschella: The part in this branch can definitely benefit from some refactoring.
It looks ok, but there is way too much going on, I would suggest factoring some steps out into helper functions.

examples/timeWarpExposureUsingTransform.py
```diff
@@ -0,0 +1,156 @@
...
+    afwdataDir, "ImSim/calexp/v85408556-fr/R23/S11.fits")
+
+
+def timeWarp(destMaskedImage, srcMaskedImage, srcFromDest, warpingControl):
+    """Time warpImage
+
+    int warpImage(
+        DestMaskedImageT &destMaskedImage,
+        SrcMaskedImageT const &srcMaskedImage,
+        SeparableKernel &warpingKernel, int const interpLength=0);
```
r-owen: Left in by accident

examples/timeWarpExposureUsingTransform.py
```diff
@@ -0,0 +1,156 @@
...
+        if endTime - startTime > MaxTime:
+            break
+
+    return (endTime - startTime, nIter, goodPix)
+
+
+def run():
+    if len(sys.argv) < 2:
+        srcExposure = afwImage.ExposureF(InputExposurePath)
+        if True:
```
r-owen: It intended to allow warping the whole image (slower) instead of a subregion. I added a global constant to this and `timeWarpExposure.py` to clarify this.

include/lsst/afw/math/warpExposure.h
```diff
@@ -461,6 +461,30 @@ int warpImage(DestImageT &destImage,                            ///< remapped %i
               );
 
 /**
+ * @brief A variant of warpImage that uses a Transform<Point2Endpoint, Point2Endpoint>
+ * instead of a pair of WCS to describe the transformation.
+ *
+ * @param[in,out] destImage  Destination image; all pixels are set
+ * @param[in] srcImage  Source image
+ * @param[in] srcFromDest  Transformation from destination to source parent pixel position
+ *    This can be computed as srcWcs.getInverse().of(destWcs)
```
r-owen: I'm afraid that's how it's done. Warping requires a transformation from destination pixel position to source. I could name it destToSrc but the same math is still required to derive it. Once our old `image.Wcs` class is replaced with `geom.SkyWcs` you can call the override of warpImage that takes the source and destination WCS directly, or call `warpExposure`, and not have to think about it at all.

include/lsst/afw/math/warpExposure.h
```diff
@@ -461,6 +461,30 @@ int warpImage(DestImageT &destImage,                            ///< remapped %i
...
+ *   the region of pixels that can be computed from the source image
+ * @return the number of good pixels
+ */
+template <typename DestImageT, typename SrcImageT>
+int warpImage(DestImageT &destImage,
+              SrcImageT const &srcImage,
+              geom::Transform<geom::Point2Endpoint, geom::Point2Endpoint> const & srcFromDest,
+              WarpingControl const &control,
+              typename DestImageT::SinglePixel padValue = lsst::afw::math::edgePixel<DestImageT>(
+                      typename lsst::afw::image::detail::image_traits<DestImageT>::image_category())
```
r-owen: I'm afraid I don't know. @RobertLuptonTheGood might be able to answer this question.

src/math/warpExposure.cc
r-owen: I agree, but this situation is temporary and I couldn't see an easy way to share code between the two blocks. Once SkyWcs and Transform are adopted, the similar code that handles XYTransform will disappear, leaving only the new code.

tests/testWarpExposure.py
```diff
@@ -432,6 +432,47 @@ def testNonIcrs(self):
...
+
+            # original image is source, warped image is destination
+            srcFromDest = originalSkyWcs.getInverse().of(warpedSkyWcs)
+
+            afwWarpedMaskedImage = afwImage.MaskedImageF(swarpedImage.getDimensions())
+            originalMaskedImage = originalExposure.getMaskedImage()
+
+            numGoodPix = afwMath.warpImage(afwWarpedMaskedImage, originalMaskedImage,
+                                           srcFromDest, warpingControl)
+            self.assertGreater(numGoodPix, 50)
```
r-owen: To make sure that some pixels were warped.

tests/testWarpExposure.py
```diff
@@ -432,6 +432,47 @@ def testNonIcrs(self):
...
+            swarpedImageName = "medswarp1%s.fits" % (kernelName,)
+            swarpedImagePath = os.path.join(dataDir, swarpedImageName)
+            swarpedDecoratedImage = afwImage.DecoratedImageF(swarpedImagePath)
+            swarpedImage = swarpedDecoratedImage.getImage()
+
+            swarpedMetadata = swarpedDecoratedImage.getMetadata()
+            warpedSkyWcs = afwGeom.SkyWcs(swarpedMetadata)
+
+            # original image is source, warped image is destination
+            srcFromDest = originalSkyWcs.getInverse().of(warpedSkyWcs)
```
r-owen: If we want it to be less confusing I fear some syntactic sugar may be required. I'm reluctant to fatten the API of `SkyWcs` until we are further along, preferring to document what's going on, especially since (as I say above) few users will ever need to do this.

include/lsst/afw/math/warpExposure.h
RobertLuptonTheGood: No good reason that I know of.  Is there any advantage over clarity?

src/math/warpExposure.cc
pschella: I wasn't thinking about sharing code between the blocks, but rather separating part of the one long block out a bit. It really is quite long. But there is no pressing need for it. Just feel like it would make it easier to grok.

src/math/warpExposure.cc
```diff
@@ -494,6 +496,228 @@ int warpImage(DestImageT &destImage, SrcImageT const &srcImage, geom::XYTransfor
...
+                        computeRelativeArea(srcPos, prevSrcPosList[col], prevSrcPosList[col + 1]);
+
+                if (warpAtOnePoint(destXIter, srcPos, relativeArea,
+                                   typename image::detail::image_traits<DestImageT>::image_category())) {
+                    ++numGoodPixels;
+                }
+            }  // for col
+            // move points from srcPosList to prevSrcPosList (we don't care about what ends up in srcPosList
+            // because it will be reallocated anyway)
+            std::swap(srcPosList, prevSrcPosList);
```
pschella: Generally prefer `using std::swap` to support user defined `swap` operations. In this case it is a `std::vector` so it doesn't matter, but we might change the type later :)

r-owen: OK. I hate how that makes it hard to tell where `swap` is defined, but it does appear to be standard.

src/math/warpExposure.cc
r-owen: I'll take a stab at it.

r-owen: @pschella I pushed an additional commit with some refactoring of the interpolated warping code. It's not a big change, but it cleans up some code and avoids messing with tight loops where performance is a concern.

r-owen: I withdraw my most recent change. It needlessly caches row information that can be computed on the fly. A clever implementation could compute the end row/column of each band and the fractional width on the fly (storing almost nothing) because all bands except the last are the same width. But that's more cleverness than I want to code.

PR #229, opened 2017-05-15
Add documentation to Footprints

rearmstr:
Add doxygen documentation to the constructors for the Footprint
object.

### DIFF
--- a/include/lsst/afw/detection/Footprint.h
+++ b/include/lsst/afw/detection/Footprint.h
@@ -71,11 +71,28 @@ public:
      * that to construct the Footprint.
      */
 
-    explicit Footprint(std::shared_ptr<geom::SpanSet> inputSpans, geom::Box2I const &region = geom::Box2I());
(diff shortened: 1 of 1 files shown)


include/lsst/afw/detection/Footprint.h
```diff
@@ -70,13 +70,30 @@ class Footprint : public lsst::daf::base::Citizen,
      * that to construct the Footprint.
      */
 
+    /** @brief Constructor for Footprint object
+     *
+     * @param inputSpans Shared pointer to a SpanSet defining the pixels included in
+                         the Footprint.
+     * @param region Bounding box of the MaskedImage in which the Footprint was created,
```
rearmstr: Why does this refer to a MaskedImage?  This could be any type of image correct?  Same question for the other constructor.

natelust: oh good point, I just pulled over from old footprints. I maybe should just change that to lowercase image to just give idea of generic image.

PR #231, opened 2017-05-17
DM-9939: Stop transposing data in Mapping.tranForward and tranInverse

r-owen:
astshim point data is now in standard AST order:
x0, x1, ...xnPts, y0, y1, ..., ynPts, ...
which is the transposition of the previous order.
Update Transform and Endpoint accordingly.
Drop support for Point3Endpoint because it will require extra code
and we have no immediate use case.

### DIFF
--- a/src/geom/Endpoint.cc
+++ b/src/geom/Endpoint.cc
@@ -87,13 +87,13 @@ template <typename Point>
 ndarray::Array<double, 2, 2> BaseVectorEndpoint<Point>::dataFromArray(Array const& arr) const {
     const int nAxes = this->getNAxes();
     const int nPoints = this->getNPoints(arr);
-    ndarray::Array<double, 2, 2> data = ndarray::allocate(ndarray::makeVector(nPoints, nAxes));
-    auto dataRowIter = data.begin();
+    ndarray::Array<double, 2, 2> data = ndarray::allocate(ndarray::makeVector(nAxes, nPoints));
+    auto dataColIter = data.transpose().begin();
     for (auto const& point : arr) {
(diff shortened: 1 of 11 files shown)


kfindeisen:
Why does Point3Endpoint require extra code? I thought `PointEndpoint<N>` took care of that.

r-owen:
@kfindeisen because with the data transposed point-like objects must be constructed by passing each argument separately (e.g. `Point2D(dataCol[0], dataCol[1])`. Formerly, when the data was transposed, we could construct each object with a vector of the appropriate length.

src/geom/Endpoint.cc
```diff
@@ -136,17 +124,26 @@ ndarray::Array<double, 2, 2> GenericEndpoint::arrayFromData(ndarray::Array<doubl
...
 
-template <int N>
-void PointEndpoint<N>::normalizeFrame(std::shared_ptr<ast::Frame> framePtr) const {
+std::vector<Point2D> Point2Endpoint::arrayFromData(ndarray::Array<double, 2, 2> const& data) const {
+    this->_assertNAxes(this->_getNAxes(data));
+    int const nPoints = this->_getNPoints(data);
+    Array array;
+    array.reserve(nPoints);
+    for (auto const& dataCol : data.transpose()) {
+        array.emplace_back(dataCol[0], dataCol[1]);
```
kfindeisen: @r-owen Isn't `dataCol` a 1×2 `ndarray::Array`? What's to stop you from declaring an array of length N, filling it from `dataCol`, and passing that in?

r-owen: That would work, but it makes an extra copy, which I want to avoid.

PR #232, opened 2017-05-17
DM-10541: Add properties to image classes

TallJimbo:
This adds:
 - `array` properties to `ImageBase` (and hence both `Image` and `Mask`).  The setters for these properties delegate to NumPy, so they can accept any value `NumPy` accepts when assigning to an array (including arrays with different types and scalars).
 - `image`, `mask`, and `variance` properties for `MaskedImage`.  The setters for these require `Image` and `Mask` objects of the appropriate (so you have to say `.image.array = value` to coerce types or assign scalars).
 - `setImage`, `setMask`, and `setVariance` to `MaskedImage`, for consistency with `Exposure.setMaskedImage` and the properties.
 - a `maskedImage` property for `Exposure`.  The setter requires a `MaskedImage` of the appropriate type.
 - Getters and setters for the image, mask, and variance of `Exposure` that just delegate to the getters and setters for `MaskedImage`.
 - `image`, `mask`, and `variance` properties for `Exposure`, using the just-added getters and setters.

I've also removed a useless and almost unused argument in the `MaskedImage` getters; it was intended to change the error-handling behavior when null pointers are encountered, but the `MaskedImage` constructors guarantee (and indeed, most of our code assumes) that those pointers can never be null.

### DIFF
--- a/tests/testExposure.py
+++ b/tests/testExposure.py
@@ -154,6 +154,17 @@ class ExposureTestCase(lsst.utils.tests.TestCase):
         self.assertEqual(crOnlyWidth, self.exposureCrOnly.getWidth())
         self.assertEqual(crOnlyHeight, self.exposureCrOnly.getHeight())
 
+
+    def testProperties(self):
+        self.assertMaskedImagesEqual(self.exposureMiOnly.maskedImage,
+                                     self.exposureMiOnly.getMaskedImage())
+        mi2 = afwImage.MaskedImageF(self.exposureMiOnly.getDimensions())
+        mi2.image.array = 5.0
+        mi2.variance.array = 3.0
+        mi2.mask.array = 0x1
+        self.exposureMiOnly.maskedImage = mi2
+        self.assertMaskedImagesEqual(self.exposureMiOnly.maskedImage, mi2)
+
     def testGetWcs(self):
         """
         Test if a WCS can be obtained from each Exposure created with
--- a/tests/testExposure.py
+++ b/tests/testExposure.py
@@ -158,26 +158,26 @@ class ExposureTestCase(lsst.utils.tests.TestCase):
         self.assertMaskedImagesEqual(self.exposureMiOnly.maskedImage,
                                      self.exposureMiOnly.getMaskedImage())
         mi2 = afwImage.MaskedImageF(self.exposureMiOnly.getDimensions())
-        mi2.image.array = 5.0
-        mi2.variance.array = 3.0
-        mi2.mask.array = 0x1
+        mi2.image.array[:] = 5.0
+        mi2.variance.array[:] = 3.0
+        mi2.mask.array[:] = 0x1
         self.exposureMiOnly.maskedImage = mi2
         self.assertMaskedImagesEqual(self.exposureMiOnly.maskedImage, mi2)
         self.assertImagesEqual(self.exposureMiOnly.image,
                                self.exposureMiOnly.maskedImage.image)
 
         image3 = afwImage.ImageF(self.exposureMiOnly.getDimensions())
-        image3.array = 3.0
+        image3.array[:] = 3.0
         self.exposureMiOnly.image = image3
         self.assertImagesEqual(self.exposureMiOnly.image, image3)
 
         mask3 = afwImage.MaskU(self.exposureMiOnly.getDimensions())
-        mask3.array = 0x2
+        mask3.array[:] = 0x2
         self.exposureMiOnly.mask = mask3
         self.assertMasksEqual(self.exposureMiOnly.mask, mask3)
 
         var3 = afwImage.ImageF(self.exposureMiOnly.getDimensions())
-        var3.array = 2.0
+        var3.array[:] = 2.0
         self.exposureMiOnly.variance = var3
         self.assertImagesEqual(self.exposureMiOnly.variance, var3)
 
--- a/tests/testExposure.py
+++ b/tests/testExposure.py
@@ -154,7 +154,6 @@ class ExposureTestCase(lsst.utils.tests.TestCase):
         self.assertEqual(crOnlyWidth, self.exposureCrOnly.getWidth())
         self.assertEqual(crOnlyHeight, self.exposureCrOnly.getHeight())
 
-
     def testProperties(self):
         self.assertMaskedImagesEqual(self.exposureMiOnly.maskedImage,
                                      self.exposureMiOnly.getMaskedImage())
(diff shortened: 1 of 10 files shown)


tests/testExposure.py
```diff
@@ -154,6 +154,35 @@ def testGetMaskedImage(self):
...
+        mask3.array = 0x2
+        self.exposureMiOnly.mask = mask3
+        self.assertMasksEqual(self.exposureMiOnly.mask, mask3)
+
+        var3 = afwImage.ImageF(self.exposureMiOnly.getDimensions())
+        var3.array = 2.0
+        self.exposureMiOnly.variance = var3
+        self.assertImagesEqual(self.exposureMiOnly.variance, var3)
+
+
```
r-owen: I'm a bit disappointed that `x.array = scalar` works, in that it would do something else (undesirable) if this was a standard Python object; the standard Python technique would be `x.array[:, :] = scalar-or-array`). In any case please also test subscripted setting.

Also you have an extra blank line here. flake8 should have warned; if you haven't run that on the the Python code that you modified, please do so before merging.

tests/testImage.py
```diff
@@ -106,11 +106,17 @@ def testArrays(self):
             self.assertEqual(array1.shape[0], image2.getHeight())
             self.assertEqual(array1.shape[1], image2.getWidth())
             self.assertEqual(type(image3), cls)
+            array2 = image1.array
+            self.assertEqual(array1.shape, array2.shape)
```
r-owen: Why this test? If you want to assert these are identical I suggest using `numpy.testing.assert_array_equal` or `assertFloatsEqual`. I admit to a strong preference for the former because it is standard (thus any user can quickly look it up) and because it compares NaNs as equal.

tests/testMask.py
```diff
@@ -132,6 +132,12 @@ def testArrays(self):
         self.assertEqual(type(image3), afwImage.MaskU)
         array1[:, :] = np.random.uniform(low=0, high=10, size=array1.shape)
         self.assertMasksEqual(image1, array1)
+        array2 = image1.array
+        self.assertFloatsEqual(image1.array, array2)
```
r-owen: This is confusing because the values aren't floats. How about `numpy.testing.assert_array_equal` or `assertMasksEqual` (which does work on arrays)  instead?

tests/testMaskedImage.py
```diff
@@ -127,6 +127,23 @@ def testArrays(self):
         mimage2 = afwImage.makeMaskedImageFromArrays(image, mask, variance)
         self.assertEqual(type(mimage2), type(self.mimage))
 
+    def testProperties(self):
+        self.assertImagesEqual(self.mimage.image, self.mimage.getImage())
+        self.assertImagesEqual(self.mimage.mask, self.mimage.getMask())
```
r-owen: I suggest `assertMasksEqual` instead. It makes the code clearer and there may come a time when it compares the mask plane dicts.

tests/testMaskedImage.py
```diff
@@ -127,6 +127,23 @@ def testArrays(self):
...
+        self.assertImagesEqual(self.mimage.image, image2)
+        mask2 = self.mimage.mask.Factory(self.mimage.getDimensions())
+        mask2.array = 0x4
+        self.mimage.mask = mask2
+        self.assertMasksEqual(self.mimage.mask, mask2)
+        var2 = self.mimage.image.Factory(self.mimage.getDimensions())
+        var2.array = 3.0
+        self.mimage.variance = var2
+        self.assertImagesEqual(self.mimage.variance, var2)
+
```
r-owen: A test that modifies a subset of elements using the `array` property would be appreciated.

tests/testExposure.py
TallJimbo: I had considered the ability to assign a scalar without extra brackets a feature, not a bug, but it's not something I'm wedded to.  I wonder what other Python style experts think (@timj, @parejkoj, @natelust?).

However, I do think it's hard to get coercion of arrays with other types without supporting scalars; I think it's either both or neither.

Subscripted setting should just delegate to our `__getitem__` and NumPy's `__setitem__`, but I'm happy to test it as well.

It occurs to me that what really needs to be tested is in-place binary operators.  I'm not sure if those will work unless I do implement `__setitem__` via NumPy and hence support assigning scalars, but I'll see.

TallJimbo: One more thought: if we do disable `x.array = scalar`, we should probably disable `x.array = array2` as well; regular attribute assignment is shallow, not deep, so array assignment doesn't behave like a regular attribute either.

TallJimbo: Some good news: if I support `x.array = array2` and not `x.array = scalar`, we do get support for `x.array += array2` and `x.array += scalar`.

I personally think supporting `x.array += scalar` is quite important, and supporting `x.array += scalar` without supporting `x.array = scalar` is mildly surprising, but as I said above, my views are not too strong on this.  I'm going to leave this PR open for a day or so before merging to see if it gathers any more opinions.

timj: `x.array = scalar` seems to me to imply that `x.array` would be the scalar `x`. If I assign a scalar to a python thing I would expect the python thing to be set to that scalar. `+= scalar` should definitely offset the array matching numpy behavior.

RobertLuptonTheGood: I assume that `x.foo = 1` still throws an error; this is important as otherwise a typo appears to set something in the C++ layer.  Given this, I don't think that python semantics are sufficiently important to stop the natural `x.image = 666` working.

parejkoj: I would strongly suggest numpy-like behavior, so `x.array = scalar` means `x.array` is a scalar (or, in this case, perhaps raises, since we should probably enforce `image`, `mask`, and `variance` having the same shape?), but `x.array[:] = scalar` sets the whole array to `scalar`, and similarly `x.array += scalar` should work as expected.

tests/testMaskedImage.py
TallJimbo: I've now done that in `testImage.py`, but I don't think we need to do it here as well, as `MaskedImage` doesn't actually have a `.array` interface (it's all delegated to `Image` and `Mask`).

tests/testExposure.py
TallJimbo: I've changed the array setter to only accept arrays of the correct type, to reflect @parejkoj's, @r-owen's, and @timj's recommendations; setting an `array` property with a scalar or an array of a different type will now raise an exception.

`x.foo = 1` (where foo is a nonexistent attribute) does indeed still raise an exception.

PR #233, opened 2017-05-19
DM-10274: Transform is not assignable

r-owen:


TallJimbo:
Assignment means mutable; is Transform already mutable in other respects?

r-owen:
Yes Transform is immutable. I'm about to redo the last commit to enable the default copy operator (the main point) and delete the assigment operator and move assigment operators (the latter of which has been enabled since the beginning).

PR #234, opened 2017-05-19
DM-8606: Improve call signature for makeCameraPoint

r-owen:


include/lsst/afw/cameraGeom/Detector.h
```diff
@@ -163,9 +163,9 @@ class Detector {
     std::shared_ptr<lsst::afw::table::AmpInfoRecord const> _get(int i) const;
 
     /**
-     * Get the amplifier specified by name as a shared_ptr
+     * Get the amplifier specified by name, returning a shared pointer to an AmpInfo record
      *
-     * @warning Intended only for internal use by pybind11. This exists because
+     * @warning Intended only for internal use, and use by pybind11. This exists because
```
natelust: Consider: "Intended only for internal and pybind11 use"

include/lsst/afw/cameraGeom/Detector.h
```diff
@@ -210,17 +210,17 @@ class Detector {
...
         return CameraPoint(point, cameraSys);
     }
 
     /**
      * Make a CameraPoint from a point and a camera system prefix
      */
-    CameraPoint makeCameraPoint(geom::Point2D point,             ///< 2-d point
-                                CameraSysPrefix cameraSysPrefix  ///< coordinate system prefix
+    CameraPoint makeCameraPoint(geom::Point2D const &point,             ///< 2-d point
+                                CameraSysPrefix const &cameraSysPrefix  ///< coordinate system prefix
```
natelust: Our coding standards don't specify a right way to do this, so feel free to leave this as it is, but my personal preference is for the syntax to be `geom::Point2D const & point` with a space after &. This is more in line with the rest of our code base, and is visually distinctive from address of a variable.

r-owen: It's the result of running clang-format, which I prefer to hand-formatting. I don't like all its choices, but feel they're usually good enough, and our standards accept it.

PR #237, opened 2017-05-24
DM-9765: Suspicious numerical precision code in Angle

r-owen:


include/lsst/afw/geom/Angle.h
```diff
@@ -407,9 +406,8 @@ inline Angle Angle::wrapNear(Angle const& refAng) const noexcept {
     if (wrapped - refAngRad >= PI) {
         wrapped -= TWOPI;
     }
-    // maximum relative roundoff error for subtraction is 2 epsilon
     if (wrapped - refAngRad < -PI) {
-        wrapped -= wrapped * 2.0 * std::numeric_limits<double>::epsilon();
+        wrapped += TWOPI;
```
kfindeisen: While this is certainly cleaner, I do wonder why the original code didn't use `TWOPI` here when it did everywhere else.

kfindeisen: While this is certainly cleaner, I do wonder why the original code didn't use `TWOPI` here when it did everywhere else.

r-owen: Because the original code tried to keep within bounds at both ends in radians. By adding `2 pi` the result may put `wrapped - refAngRad >= 2 pi`. Still, I'm inclined to live with that. The limits are squishy for other units anyway and we have no use case for precise limits.

PR #239, opened 2017-05-26
DM-10610: Optimise SpanSet components used by CModel

pschella:


src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+                    // A span was started but the current pixel is not to be included in the new SpanSet,
+                    // the current SpanSet should be close and added to the vector
+                    newVec.push_back(Span(y, minX, maxX));
+                    started = false;
+                }
+            }
+        }
+    }
+    return std::make_shared<SpanSet>(std::move(newVec));
+}
```
pschella: Seems overly complicated. Why not something like (weird, lazy and quite possibly broken pseudo code):

```
start = max(span.min, bbox.min)
end = min(span.max, bbox.max)

smin = -1 // or some other sentinel value
smax = -1
for x between start, end {
    if (mask & x) {
        smin = min(x, smin)
        smax = max(x, smax)
    } else if (min != -1) {
         v.push_back(smin, smax)
         smin = -1
         smax = -1
    }
}
if (min != -1) {
         v.push_back(smin, x) // close last unclosed span
}
```

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
                    : false;
 }
 
+/* Determine the intersection with a mask or its logical inverse
+ *
+ * spanSet - SpanSet object with which to intersect the mask
+ * mask - Mask object to be intersectedNot
+ * bitmask - bitpattern to used when deciding pixel membership in the mask
+ *
```
natelust: This is not actual mark-up at all, this function is in an anonymous namespace, the comment block was provided strictly for future readers

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+ * Templates:
+ * T - pixel type of the mask
+ * invert - boolean indicating if the intersection is with the mask or the inverse of the mask
+            templated parameter to ensure the check is compiled away if it can be
```
natelust: I don't feel too strongly about that as this function can only be called w/n this file, and I like not having a default in that case, as it provides symmetry in readability at the point it is called in code.

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+template <typename T, bool invert>
+std::shared_ptr<SpanSet> maskIntersect(SpanSet const & spanSet, image::Mask<T> const & mask, T bitmask) {
+    // This vector will store our output spans
+    std::vector<Span> newVec;
+    // Variable to mark if a new span has been started
+    bool started = false;
+    auto maskBBox = mask.getBBox();
+    for (auto const & spn : spanSet) {
+        // Reset the started flag for each different span
+        started = false;
```
natelust: done

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+        int y = spn.getY();
+        // Reset the min and max variables
```
natelust: removed

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+        int minX = 0;
+        int maxX = 0;
+        for (int x = spn.getMinX(); x <= spn.getMaxX(); ++x){
+            // Verify that the point actually falls within the bounds of the mask
+            if (maskBBox.contains(Point2I(x, y))) {
```
natelust: Changed for more efficient looping

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+                }
+                // If the pixel is to be included in the operation, start or append recording values
+                // needed to later construct a Span
+                if (pixelCompare) {
+                    if (!started) {
+                        started = true;
+                        minX = x;
+                        maxX = x;
+                    } else {
+                        if (minX > x) {
```
natelust: no

src/geom/SpanSet.cc
```diff
@@ -99,6 +99,72 @@ bool spansContiguous(Span const& a, Span const& b, bool compareY = true) {
...
+                            minX = x;
+                        } else if (maxX < x) {
```
natelust: no

src/geom/SpanSet.cc
natelust: I would need to think about it, but this may not be dealing with disconnected spans well. I think this is basically the first implementation I had but needed to expand things a bit. I do not favor getting rid of the boolean just to be cleaver as I think the bool helps with readability of the algorithm

natelust: also, you can't use the sentinel value to be -1 and then take the minimum, because in most all cases smin would always be -1

src/geom/SpanSet.cc
```diff
@@ -147,24 +213,33 @@ SpanSet::SpanSet(std::vector<Span>&& vec, bool normalize) : _spanVector(std::mov
...
+        if (spansContiguous(newSpansEnd, *iter)) {
+            newSpansEnd = Span(newSpansEnd.getY(), std::min(newSpansEnd.getMinX(), (*iter).getMinX()),
+                                 std::max(newSpansEnd.getMaxX(), (*iter).getMaxX()));
+        } else {
+            newSpans.push_back(Span(*iter));
         }
     }
+
+    // Set the new normalized vector of spans to the internal span vector container
+    _spanVector = newSpans;
```
PaulPrice: `std::vector::erase` is evil, or at least it is when it's used in a loop because you can get a huge number of moves. It's better to take this approach of creating a new vector containing the desired contents.

src/geom/SpanSet.cc
pschella: like I said, possibly broken. The point was more the general idea.

pschella: You could for instance use `smin = std::numeric_limits<T>::max();`. But you can also keep it as is.

src/geom/SpanSet.cc
```diff
@@ -635,32 +717,69 @@ std::shared_ptr<SpanSet> SpanSet::intersectNot(SpanSet const& other) const {
...
     for (auto const& spn : _spanVector) {
-        added = false;
-        for (auto const& otherSpn : other) {
-            if (spansOverlap(spn, otherSpn)) {
+        bool added = false;
+        bool spanStarted = false;
+        int spanBottom = 0;
+        while (otherIter != other.end() && otherIter->getY() <= spn.getY()) {
+            if (spansOverlap(spn, *otherIter)) {
                 added = true;
```
pschella: So this will go wrong in case `other` fully contains `spn` (nothing is added).
Perhaps it is best in the unit test to generate many cases at random (once) and compare them with a brute force comparison.
It doesn't seem possible to device a full set of test cases.

src/geom/SpanSet.cc
```diff
@@ -734,6 +734,7 @@ std::shared_ptr<SpanSet> SpanSet::intersectNot(SpanSet const& other) const {
         bool spanStarted = false;
         int spanBottom = 0;
         while (otherIter != other.end() && otherIter->getY() <= spn.getY()) {
+        //for(auto otherIter = other.begin(); otherIter != other.end(); ++otherIter) {
```
pschella: Please remove commented line.

tests/testSpanSets.py
```diff
@@ -292,7 +292,30 @@ def testIntersectNot(self):
...
+            # to be the negated SpanSet
+            randChoice = np.random.randint(0, 2)
+            negatedRandChoice = int(not randChoice)
+            sourceSpanSet = listOfRandomSpanSets[randChoice]
+            targetSpanSet = listOfRandomSpanSets[negatedRandChoice]
+            resultSpanSet = sourceSpanSet.intersectNot(targetSpanSet)
+            for span in resultSpanSet:
+                for point in span:
+                    self.assertTrue(sourceSpanSet.contains(point))
+                    self.assertFalse(targetSpanSet.contains(point))
```
pschella: Almost, but not complete. This checks that every point in the result should be there. But it does not check that every point that should be part of the result is (for instance, this will always work if the result is empty, even when it shouldn't). Thus it is better to do the result check on the pixel level.

src/geom/SpanSet.cc
```diff
@@ -774,6 +775,9 @@ std::shared_ptr<SpanSet> SpanSet::intersectNot(SpanSet const& other) const {
                     }
                 }
             }
+        if (otherIter->getMaxX() > spn.getMaxX()) {
+            break;
+        }
```
pschella: Please reindent this file.

natelust: check

src/geom/SpanSet.cc
natelust: check

tests/testSpanSets.py
natelust: added the converse check as well, as requested

PR #240, opened 2017-06-01
DM-10764: Rename Transform::of and Mapping::of to ::then

r-owen:
Change `Transform::of(first)` to `Transform::then(next)` to improve readability.

### DIFF
--- a/include/lsst/afw/geom/Transform.h
+++ b/include/lsst/afw/geom/Transform.h
@@ -200,25 +200,25 @@ public:
     /**
      * Concatenate two Transforms.
      *
-     * @tparam FirstFromEndpoint the starting Endpoint of `first`
-     * @param first the Transform to apply before this one
-     * @returns a Transform that first applies `first` to its input, and then
-     *          this Transform to the result. Its inverse shall first apply the
-     *          inverse of this Transform, then the inverse of `first`.
+     * @tparam NextToEndpoint the "to" Endpoint of `next`
+     * @tparam FromEndpoint the 
+     * @param next the Transform to apply after this one
+     * @returns a Transform that first applies this transform to its input, and then
+     *          `next` to the result. Its inverse shall first apply the
+     *          inverse of `next`, and then the inverse of this transform.
      *
      * @throws pex::exceptions::InvalidParameterError Thrown if
-     *         `first.getToEndpoint()` and `this->getFromEndpoint()` do not
+     *         `getToEndpoint()` and `next.getFromEndpoint()` do not
      *         have the same number of axes.
      * @exceptsafe Provides basic exception safety.
      *
      * More than two Transforms can be combined in series. For example:
      *
-     *     auto skyFromPixels = skyFromPupil.of(pupilFromFp)
-     *                                      .of(fpFromPixels);
+     *     auto pixelsToSky = pixelsToFP.then(fpToPupil.then(pupilToSky));
      */
-    template <class FirstFromEndpoint>
-    Transform<FirstFromEndpoint, ToEndpoint> of(
-            Transform<FirstFromEndpoint, FromEndpoint> const &first) const;
+    template <class NextToEndpoint>
+    Transform<FromEndpoint, NextToEndpoint> then(
+            Transform<ToEndpoint, NextToEndpoint> const &next) const;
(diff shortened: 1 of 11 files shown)


python/lsst/afw/geom/transform/transform.cc
```diff
@@ -99,9 +99,9 @@ void declareTransform(py::module &mod, std::string const &fromName, std::string
      * of length 1 from being deleted */
     cls.def("_getJacobian", &Class::getJacobian);
 
-    declareMethodTemplates<GenericEndpoint, FromEndpoint, ToEndpoint>(cls);
-    declareMethodTemplates<Point2Endpoint, FromEndpoint, ToEndpoint>(cls);
-    declareMethodTemplates<SpherePointEndpoint, FromEndpoint, ToEndpoint>(cls);
+    declareMethodTemplates<FromEndpoint, ToEndpoint, GenericEndpoint>(cls);
```
kfindeisen: I find this form slightly less readable because it puts the most important information at the end, though I can see why you changed it.

src/math/warpExposure.cc
```diff
@@ -535,7 +535,8 @@ int warpImage(DestImageT &destImage, SrcImageT const &srcImage,
     std::vector<double> const srcParentToLocalVec = {static_cast<double>(-srcImage.getX0()),
                                                      static_cast<double>(-srcImage.getY0())};
     auto const srcParentToLocalMap = ast::ShiftMap(srcParentToLocalVec);
-    auto const localDestToSrcMap = srcParentToLocalMap.of(destToSrc.getFrameSet()->of(destLocalToParentMap));
+    auto const localDestToSrcMap =
+            destLocalToParentMap.then(*(destToSrc.getFrameSet())).then(srcParentToLocalMap);
```
kfindeisen: Oops -- should `Transform`'s constructors be `explicit`?

include/lsst/afw/geom/Transform.h
```diff
@@ -200,25 +200,25 @@ class Transform {
...
-     *         `first.getToEndpoint()` and `this->getFromEndpoint()` do not
+     *         `getToEndpoint()` and `next.getFromEndpoint()` do not
      *         have the same number of axes.
      * @exceptsafe Provides basic exception safety.
      *
      * More than two Transforms can be combined in series. For example:
      *
-     *     auto skyFromPixels = skyFromPupil.of(pupilFromFp)
-     *                                      .of(fpFromPixels);
+     *     auto pixelsToSky = pixelsToFP.then(fpToPupil.then(pupilToSky));
```
r-owen: Agreed. I'll fix it on a new ticket.

python/lsst/afw/geom/transform/transform.cc
r-owen: Yes, I thought about that. I'm still happier with it because I can instantly see that it matches the header file and other code.

src/math/warpExposure.cc
r-owen: `Transform`'s constructors are `explicit` because I usually make all constructors explicit under the argument that it's harmless and can avoid trouble (e.g. if a 2 argument ctor is changed to be a 1 argument ctor). But why do you ask?

As to why not `just use destToSrc`: it maps parent coords to parent coords, but C++ indexing of images is done using local coords. The above is needed for the code to work when xy0 is non-zero.

kfindeisen: Hmm, so they are. Then I have no idea how `then` can take a `FrameSet`. :confused:

r-owen: If you mean "why not use destToSrc directly instead of getting the FrameSet from it?"....that's because everything else is a Mapping and I don't want to mix Mapping and Transform in my concatenation. And I don't want to use Transforms when Mappings will do just fine. I suppose I could change that.

kfindeisen: Aha, I missed the part where `destLocalToParentMap` was not a `Transform`. It all makes sense now. :slightly_smiling_face:

r-owen: That's why I ended the name with "Map". A bit of Hungarian notation that was clearly not as helpful as I'd hoped.

PR #245, opened 2017-06-15
DM-10542: Replace XYTransform::linearizeTransform

kfindeisen:
The first two commits on this ticket implement testing-related RFCs, as recommended at the end of [this Community post](https://community.lsst.org/t/python-unit-test-module-migrations-rfc-215-and-rfc-229/). They are not directly related to `linearizeTransform`.

The third commit adds a free function that takes a Transform and linearizes it. Per Effective C++, it's been placed in a different header from Transform itself; I expect the factory functions mentioned in [RFC-346](https://jira.lsstcorp.org/browse/RFC-346) will go in the same file once they're implemented.


src/geom/transformFactory.cc
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Convert a Matrix to the equivalent ndarray.
+ *
+ * @param matrix The matrix to convert.
+ * @returns an ndarray containing a copy of the data in `matrix`
+ */
+template <class T, int Rows, int Cols>
+ndarray::Array<T, 2, 2> toNdArray(Eigen::Matrix<T, Rows, Cols> const &matrix) {
+    ndarray::Array<T, 2, 2> array = ndarray::allocate(ndarray::makeVector(matrix.rows(), matrix.cols()));
```
kfindeisen: No, actually. `Rows` and `Cols` may have the dummy value -1 to mean "determined at run-time", and in fact they do have this value for the matrix returned by `getJacobian`. (Also, I'm pretty sure that in fixed-size matrices the implementations for `rows()` and `cols()` are optimized for inlining.)

PR #248, opened 2017-06-20
Review for DM-10965: FootprintSet setter unable to accept results from getter

jdswinbank:


tests/test_footprint1.py
```diff
@@ -682,6 +682,16 @@ def testSetFromFootprint(self):
...
 
+        # Verify the FootprintSet footprint list setter can accept inputs from
+        # the footprint list getter
+        getterSetterTestOK = True
+        try:
+            ds.setFootprints(ds.getFootprints())
+        except:
+            getterSetterTestOK = False
+        self.assertTrue(getterSetterTestOK, "The FootprintSet footprint "
+                        "setter can't accept getter output as an input")
```
jdswinbank: As we discussed, I find the above a little convoluted: I think it would be more straightforward to simply let the exception propagate rather than catch it and `assertTrue`. While I realise you're trying to print a helpful message, I don't think expecting somebody debugging a test failure to look at a comment in the source to understand what went wrong is unreasonable.

I also think we could be a bit more ambitious with the test. Demonstrating that the setter doesn't throw when called with the output of the getter is necessary, but it's not sufficient behaviour for the code: we really want to see that `setFootprints` really does set the footprints when it's called. As far as I can see, this function isn't tested anywhere else (if it was, we'd probably never have hit this error), so dropping in a real test for it now seems worthwhile.

Off the top of my head, I would:

- Retrieve a list of footprints from `ds.getFootprints()`;
- Perform some transformation on that list (e.g. just delete the last one, assuming there are multiple footprints in the list);
- Feed that to `setFootprints()`;
- Check that calling `getFootprints()` gets you back the modified list.

That probably takes more characters to write in English than it does in Python, so I shouldn't think it'd be a big effort. Am I wrong?

python/lsst/afw/detection/footprintSet.cc
```diff
@@ -102,6 +102,18 @@ PYBIND11_PLUGIN(_footprintSet) {
...
+    // The pybind11 wrapped getFootprints dereferences the shared pointer to the std::vector of Footprints
+    // held by the FootprintSet when returning to python to work around a stl caster issue. However this
+    // creates a problem for the setter that now expects a shared pointer, making the setter unable to
+    // operate on the output form the getter. Below defines two overrides for the setter function, one
+    // that can handle a shared pointer (in case somewhere else in the stack has one somehow) and one
+    // that will accept a standard vector of Footprints, which will be internally placed in a shared
+    // pointer to match the underlying C++ function signature
+    clsFootprintSet.def("setFootprints",
+        [](FootprintSet &self, std::shared_ptr<FootprintSet::FootprintList> footListPtr)
+        {self.setFootprints(footListPtr);});
```
natelust: To the former point, there is no difference to the standard &FootprintSet::setFootprints. Where the difference lies is that if you have two function signatures, you must do a function pointer cast to disambiguate. So I could have written (void (FootprintSet::*)(std::shared_ptr<FootprintSet::FootprintList>)) &FootprintSet::setFootprints but for the sake of symmetry with the other argument (and easier readability) I opted to have them both as lambda calls.

As for the circumstances for which someone could hold a pointer to a FootprintList? I am not sure, I don't actually think there is any (at least that I can think up) but I worried that there is somewhere and somebody will exercise the code at some point and we will have a new breakage. So I hedged against that and included it, as it doesn't really add much overhead to the bindings anyway.

jdswinbank: > ... there is no difference to the standard `&FootprintSet::setFootprints`. Where the difference lies is that if you have two function signatures, you must do a function pointer cast to disambiguate

Does this mean we can just delete the `clsFootprintSet.def("setFootprints", &FootprintSet::setFootprints);` line entirely? If I'm understanding the above, it's now unused.

To the second point, I'd tend to only include code that we explicitly are using and (vitally!) testing. However, I don't feel strongly enough about this to argue that you should take it out.

natelust: Yes, I had forgotten to take out that line

PR #250, opened 2017-06-23
DM-10257: Fix compiler warnings in afw

kfindeisen:
This PR eliminates all but one compiler warning (unused `fits::{}::getFormatCode(int8_t)`) when building `afw` with Ubuntu+GCC and MacOS+Clang. Justifications for individual changes are included in the commit messages.

During the MacOS port a few unrelated clang-format changes were made because of the mismatch in clang-format versions between Ubuntu and MacOS. @parejkoj and I allowed them to get committed on the grounds that we will probably standardize on clang-format 5 anyway.

### DIFF
--- a/src/math/Interpolate.cc
+++ b/src/math/Interpolate.cc
@@ -301,7 +301,7 @@ std::vector<double> Interpolate::interpolate(std::vector<double> const &x) const
 ndarray::Array<double, 1> Interpolate::interpolate(ndarray::Array<double const, 1> const &x) const {
     int const num = x.getShape()[0];
     ndarray::Array<double, 1> out = ndarray::allocate(ndarray::makeVector(num));
-    for (size_t i = 0; i < num; ++i) {
+    for (int i = 0; i < num; ++i) {
         std::cout << "Interpolating " << x[i] << std::endl;
         out[i] = interpolate(x[i]);
     }
--- a/src/math/Interpolate.cc
+++ b/src/math/Interpolate.cc
@@ -161,6 +161,9 @@ namespace {
             throw LSST_EXCEPT(pex::exceptions::LogicError,
                               str(boost::format("You can't get here: style == %") % style));
     }
+    // don't use default statement to let compiler check switch coverage
+    throw LSST_EXCEPT(pex::exceptions::LogicError,
+                      str(boost::format("You can't get here: style == %") % style));
 }
 }
 
--- a/src/table/io/FitsSchemaInputMapper.cc
+++ b/src/table/io/FitsSchemaInputMapper.cc
@@ -38,7 +38,7 @@ private:
     std::string const &_v;
 };
 
-}  // anonymous
+}  // namespace
 
 class FitsSchemaInputMapper::Impl {
 public:
@@ -136,13 +136,10 @@ FitsSchemaInputMapper::FitsSchemaInputMapper(daf::base::PropertyList &metadata, 
         // Read slots saved using an old mechanism in as aliases, since the new slot mechanism delegates
         // slot definition to the AliasMap.
         static std::array<std::pair<std::string, std::string>, 7> oldSlotKeys = {
-                std::make_pair("PSF_FLUX", "slot_PsfFlux"),
-                std::make_pair("AP_FLUX", "slot_ApFlux"),
-                std::make_pair("INST_FLUX", "slot_InstFlux"),
-                std::make_pair("MODEL_FLUX", "slot_ModelFlux"),
-                std::make_pair("CALIB_FLUX", "slot_CalibFlux"),
-                std::make_pair("CENTROID", "slot_Centroid"),
-                std::make_pair("SHAPE", "slot_Shape")};
+                {std::make_pair("PSF_FLUX", "slot_PsfFlux"), std::make_pair("AP_FLUX", "slot_ApFlux"),
+                 std::make_pair("INST_FLUX", "slot_InstFlux"), std::make_pair("MODEL_FLUX", "slot_ModelFlux"),
+                 std::make_pair("CALIB_FLUX", "slot_CalibFlux"), std::make_pair("CENTROID", "slot_Centroid"),
+                 std::make_pair("SHAPE", "slot_Shape")}};
         for (std::size_t i = 0; i < oldSlotKeys.size(); ++i) {
             std::string target = metadata.get(oldSlotKeys[i].first + "_SLOT", std::string(""));
             if (!target.empty()) {
@@ -693,7 +690,7 @@ std::string replaceSuffix(std::string const &s, std::size_t n, std::string const
     return s.substr(0, s.size() - n) + suffix;
 }
 
-}  // anonymous
+}  // namespace
 
 Schema FitsSchemaInputMapper::finalize() {
     if (_impl->version == 0) {
@@ -780,7 +777,7 @@ void FitsSchemaInputMapper::readRecord(BaseRecord &record, afw::fits::Fits &fits
         (**iter).readCell(record, row, fits, _impl->archive);
     }
 }
-}
-}
-}
-}  // namespace lsst::afw::table::io
+}  // namespace io
+}  // namespace table
+}  // namespace afw
+}  // namespace lsst
--- a/examples/wcsTest.cc
+++ b/examples/wcsTest.cc
@@ -68,7 +68,6 @@ int main(int argc, char **argv) {
     }
     std::cout << "Opening exposure " << inImagePath << std::endl;
 
-    auto miMetadata(new PropertySet);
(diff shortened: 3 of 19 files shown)


src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -136,13 +136,10 @@ FitsSchemaInputMapper::FitsSchemaInputMapper(daf::base::PropertyList &metadata,
...
-                std::make_pair("AP_FLUX", "slot_ApFlux"),
-                std::make_pair("INST_FLUX", "slot_InstFlux"),
-                std::make_pair("MODEL_FLUX", "slot_ModelFlux"),
-                std::make_pair("CALIB_FLUX", "slot_CalibFlux"),
-                std::make_pair("CENTROID", "slot_Centroid"),
-                std::make_pair("SHAPE", "slot_Shape")};
+                {std::make_pair("PSF_FLUX", "slot_PsfFlux"), std::make_pair("AP_FLUX", "slot_ApFlux"),
+                 std::make_pair("INST_FLUX", "slot_InstFlux"), std::make_pair("MODEL_FLUX", "slot_ModelFlux"),
+                 std::make_pair("CALIB_FLUX", "slot_CalibFlux"), std::make_pair("CENTROID", "slot_Centroid"),
+                 std::make_pair("SHAPE", "slot_Shape")}};
```
r-owen: Please run clang-format on this file. I found it results in minor but helpful improvements.

src/math/Interpolate.cc
```diff
@@ -157,7 +157,7 @@ ::gsl_interp_type const *styleToGslInterpType(Interpolate::Style const style) {
         case Interpolate::UNKNOWN:
             throw LSST_EXCEPT(pex::exceptions::InvalidParameterError,
                               "I am unable to make an interpolator of type UNKNOWN");
-        case Interpolate::NUM_STYLES:
+        default:
```
r-owen: Adding a default clause is good, but removing the NUM_STYLES clause worries me. Some compilers warn if your case statements do not cover all cases (and I'm surprised clang did not).

src/table/io/FitsSchemaInputMapper.cc
kfindeisen: I thought the weird brace alignment was from clang-format 5... (@parejkoj can confirm)

src/math/Interpolate.cc
kfindeisen: I would expect a compiler to warn you if you have redundant cases in front of your default, but I guess I could add it back in and see what happens.

src/table/io/FitsSchemaInputMapper.cc
parejkoj: Yes, that came from saving it in sublimetext with `clang-format version 5.0.0 (tags/google/stable/2017-03-17)`. What version do you have, @r-owen ?

src/math/Interpolate.cc
kfindeisen: Ok, neither compiler cares. I'll leave it in.

src/table/io/FitsSchemaInputMapper.cc
r-owen: I agree the braces are weird. That's why I ran clang-format on the file (I guessed the braces were clang-format's fault and I wanted to check that). That confirmed by suspicions, but I was surprised when some other part of the file got reformatted (for the better, I think, though that's not very relevant).

I am running 5.0.0 installed using Homebrew

src/math/Interpolate.cc
RobertLuptonTheGood: I cares.   If you use default the compiler can't check for missing enums.

r-owen: My understanding is that `default` is considered an important part of defensive programming, because even in C++ it is possible to provide illegal values (e.g. by casting).

I'm disappointed that compilers don't warn about missing values when a `default` clauses is included.

kfindeisen: Without a `default` clause GCC gives a warning about falling through to the end of the function even when every constant is mentioned (despite also doing the check you mentioned).

I think I have an idea that will satisfy both rules, let me check.

r-owen: I turns out recent gcc can warn in this situation, but it requires a flag: https://stackoverflow.com/questions/5402745/gcc-switch-on-enum-retain-missing-warning-but-use-default

PaulPrice: What about:

         case Interpolate::NUM_STYLES:
              throw LSST_EXCEPT(pex::exceptions::LogicError,
                                str(boost::format("You can't get here: style == %") % style));
         default:
              abort();

kfindeisen: As a workaround, I've removed the `default` and added an exception after the switch. That lets GCC check that the cases are complete (and yes, I tested it) without complaining about reaching the end of the function. It's clunky, but I think it's the only way we can satisfy both constraints at the moment.

kfindeisen: Russell's suggestion relies on `#pragma GCC diagnostic`, which is broken at the moment. It would certainly be my preferred solution otherwise. :disappointed:

src/table/io/FitsSchemaInputMapper.cc
parejkoj: My format file looks the same as yours, except with a `---` as the first line, and `...` as the last line (copied directly from the [dev-guide](https://developer.lsst.io/tools/clang_format.html)).

I just re-ran clang-format outside of sublimeText, and it didn't change the file as checked-in. My version is also installed via homebrew.

RobertLuptonTheGood:
Does switching to a new-style enum solve this?  They don't accept ints, and I don't think we ever OR these ones together.

						R

P.S.  the abort() is better than nowt, but I'd much prefer static testing.

kfindeisen:
Good idea, but no, GCC still doesn't like the original function structure. (Also, `Interpolate::Style` is used as an integer quite frequently.)

The new current version (as of 30 minutes ago) allows static testing, so I propose we live with it.

src/math/Interpolate.cc
```diff
@@ -161,6 +161,9 @@ ::gsl_interp_type const *styleToGslInterpType(Interpolate::Style const style) {
             throw LSST_EXCEPT(pex::exceptions::LogicError,
                               str(boost::format("You can't get here: style == %") % style));
     }
+    // don't use default statement to let compiler check switch coverage
+    throw LSST_EXCEPT(pex::exceptions::LogicError,
+                      str(boost::format("You can't get here: style == %") % style));
```
PaulPrice: If you can't ever get there, this is a programming error, and therefore an `abort()` is more appropriate than an exception.

kfindeisen: An abort would cause the entire Stack to crash, with no explanation, in the middle of processing. There is no justification for calling `abort()` in low-level functions in modern code.

PaulPrice: That is _exactly_ the point! You _want_ the entire stack to crash (with a core dump) because you're in a completely bad situation that cannot be recovered from. If you throw an exception it might be caught and ignored, which makes it so much harder to know there's a problem at all, let alone debug it.

src/table/io/FitsSchemaInputMapper.cc
r-owen: Must be user error on my part. I just reformatted again and although Sublime Text claims the file changed, this time there are no visible changes (last time there were, and I'm not sure why).

src/math/Interpolate.cc
r-owen: I will say that I've had this same argument with folks before. I'm firmly on @kfindeisen's side -- I want a proper stack trace when something goes wrong, so I always thrown an exception. But there are certainly those in the project who are prefer C++ `assert` for this situation. I don't think it'll get resolved at our level, and rather than bring in the big guns for such a small and helpful ticket, I suggest we keep what @kfindeisen has done and move on.

r-owen: ...If for no other reason than because it is the minimal change from what we had already!

PaulPrice: You can get "a proper stack trace" from the core dump, or from re-running under a debugger. Throwing an exception does _not_ guarantee that you're going to get a stack trace at all, or ever even notice that there's something horribly wrong.

It's absolutely essential when we _know_ that the system is in an inconsistent state (e.g., smashed stack, memory corruption) that we not attempt to push forward and make things worse and undebuggable.

r-owen: Agreed for memory corruption. That is not the concern here. A simple cast can get you this issue.

PaulPrice: I think it's worth doing even for "a simple cast" error, but I give up.

PR #252, opened 2017-07-03
DM-11052: variable-length string fields in tables

TallJimbo:


src/table/FieldBase.cc
```diff
@@ -71,20 +71,35 @@ FieldBase<std::string>::FieldBase(int size) : _size(size) {
...
                                       std::string const &value) const {
-    if (value.size() > std::size_t(_size)) {
-        throw LSST_EXCEPT(
-                lsst::pex::exceptions::LengthError,
-                (boost::format("String (%d) is too large for field (%d).") % value.size() % _size).str());
+    if (isVariableLength()) {
+        // p is a pointer to a std::string; replace its contents with a copy of `value`
+        reinterpret_cast<std::string *>(p)->clear();
+        reinterpret_cast<std::string *>(p)->reserve(value.size());
+        reinterpret_cast<std::string *>(p)->insert(0, value);
```
TallJimbo: I think this block should be written as:
```
*reinterpret_cast<std::string *>(p) = value;
```
This will frequently have the same behavior, but it won't require a memory allocation if one isn't needed (and more importantly, it leaves that decisions up to `std::string`).

include/lsst/afw/table/FieldBase.h
```diff
@@ -237,30 +241,50 @@ struct FieldBase<std::string> {
...
 
     /// Defines how Fields are printed.
     void stream(std::ostream &os) const { os << ", size=" << _size; }
 
-    /// Used to implement RecordBase::operator[] (non-const).
-    Reference getReference(Element *p, ndarray::Manager::Ptr const &m) const { return p; }
+    /// Used to implement BaseRecord::operator[] (non-const).
+    Reference getReference(Element *p, ndarray::Manager::Ptr const &m) const {
+        if (isVariableLength()) {
+            // can't be done until C++17; who would manage the memory?
```
kfindeisen: Sorry, I don't. I don't even understand why the C++ version is relevant. Any modifiable reference to the internals of an object can potentially corrupt that object, unless it has no invariants in the first place.

src/fits.cc
```diff
@@ -848,10 +848,10 @@ void Fits::readTableArray(std::size_t row, int col, int nElements, T *value) {
     }
 }
 
-void Fits::readTableScalar(std::size_t row, int col, std::string &value) {
+void Fits::readTableScalar(std::size_t row, int col, std::string &value, bool isVariableLength) {
     int anynul = false;
-    long size = getTableArraySize(col);
-    // We can't directly write into a std::string (may be changing in C++11).
+    long size = isVariableLength ? getTableArraySize(row, col) : getTableArraySize(col);
```
r-owen: The `(row, col)` version calls `fits_read_descript`, which is documented to "Return the descriptor for a variable length column in a binary table..." so I don't think it is safe to use on fixed-width columns.

include/lsst/afw/table/FieldBase.h
r-owen: Good point. I've changed the comment to `// Can't be done; who would manage the memory?`

src/table/BaseRecord.cc
```diff
@@ -22,8 +22,12 @@ struct CopyValue {
 
     template <typename U>
     void operator()(Key<Array<U> > const& inputKey, Key<Array<U> > const& outputKey) const {
+        if (inputKey.isVariableLength() != outputKey.isVariableLength()) {
+            throw LSST_EXCEPT(lsst::pex::exceptions::InvalidParameterError,
+                              "At least one input array field is variable-length"
+                              " and the correponding output is not, or vice-versa");
+        }
```
r-owen: Agreed.

include/lsst/afw/table/FieldBase.h
TallJimbo: I'd have changed it the opposite way: it's clear this could be done with C++17 (at least as it's imagined today), but it's not about memory management.

src/fits.cc
r-owen: To expand on this: fits.h and fits.cc is low-level code and I'd rather keep it simple. Trying to do better in an efficient way would probably require adding some sort of cache to the `Fits` object for metadata about columns. I'm not even sure if such work is a good idea and it is more than I feel comfortable doing. So I propose to keep my solution.

include/lsst/afw/table/FieldBase.h
r-owen: I'll change it to `// Can't be done until C++17, which allows read/write access to the internal buffer`

PR #253, opened 2017-07-03
DM-10728: miscellaneous changes in support of PhotoCalib outputs in meas_mosaic

TallJimbo:


parejkoj:
My first concern is your change of `_instFlux[Err]` -> `_flux[Sigma]`; I tried very hard to get PhotoCalib self-consistent in its use of `instFlux` and `Err`.

Talking with @SimonKrughoff about this just now, the question is whether you can hold off a month or two for us to make the big `instFlux` and `Sigma`->`Err` changes, or whether we make this change (making PhotoCalib no longer self-consistent), and revert it later? Thoughts?

python/lsst/afw/math/chebyshevBoundedField.cc
```diff
@@ -72,15 +70,8 @@ PYBIND11_PLUGIN(_chebyshevBoundedField) {
...
-    clsChebyshevBoundedField.def("__rmul__",
-                                 [](ChebyshevBoundedField &bf, double const scale) { return bf * scale; },
-                                 py::is_operator());
-    clsChebyshevBoundedField.def("__mul__",
-                                 [](ChebyshevBoundedField &bf, double const scale) { return bf * scale; },
-                                 py::is_operator());
-    clsChebyshevBoundedField.def("__eq__", &BoundedField::operator==, py::is_operator());
-    clsChebyshevBoundedField.def("__ne__", &BoundedField::operator!=, py::is_operator());
+    /* Operators are defined only in the BoundedField base class;
+       we let Python inheritance provide them here. */
```
parejkoj: Nice improvement.

src/image/Calib.cc
```diff
@@ -285,7 +284,7 @@ std::pair<ndarray::Array<double, 1>, ndarray::Array<double, 1>> Calib::getMagnit
             continue;
         }
         double f, df;
-        convertToMagWithErr(&f, &df, _fluxMag0, fluxMag0InvSNR, *fluxIter, *fluxErrIter);
```
parejkoj: Wow, it was even more internally-inconsistent than I'd previously thought?

src/math/BoundedField.cc
```diff
@@ -83,19 +83,110 @@ struct Divide {
...
+        };
+
+        int y0 = region.getBeginY();
+        int y1 = y0 + yStep;
+        while (y1 < region.getEndY()) {
+            processRowBlock(y0, y1, y1);
+            y0 = y1;
+            y1 += yStep;
+        }
+        {
```
parejkoj: What's this second bracket-block for?

src/math/BoundedField.cc
```diff
@@ -83,19 +83,110 @@ struct Divide {
...
+            while (x1 < region.getEndX()) {
+                z10 = field.evaluate(x1, y0);
+                z11 = field.evaluate(x1, y1);
+                interpolate(x0, x1, x1, y0, y1, yEnd, z00, z01, z10, z11);
+                x0 = x1;
+                x1 += xStep;
+                z00 = z10;
+                z01 = z11;
+            }
+            {
```
parejkoj: What's this second bracket-block for?

src/math/BoundedField.cc
```diff
@@ -83,19 +83,110 @@ struct Divide {
...
+        // the box (in y, then x) to fill interpolated values.
+        // This means we evaluate the function multiple times at some points
+        // (when we move to a new interval in y, to be specific).
+        // Could probably be optimized for the cases where either step is 1,
+        // but that seems premature.
+
+        // Local function to do bilinear interpolation using points
+        // [(x0, y0), (x0, y1), (x1, y0), (x1, y1)] with values
+        // [z00, z01, z10, z11], in pixels [x0:xEnd, y0:yEnd]
+        auto interpolate = [&img, &functor](
```
parejkoj: I feel like `interpolate` and `processRowBlock` are complicated enough that they should be anonymous functions instead of lambdas. I'd strongly make that case in python (or that they at least be inner functions), but I don't know what the standard is in C++, and our dev guide doesn't even mention lambdas.

tests/test_chebyshevBoundedField.py
```diff
@@ -89,6 +89,23 @@ def setUp(self):
...
+        field = lsst.afw.math.ChebyshevBoundedField(bbox, coefficients)
+        image1 = lsst.afw.image.ImageF(bbox)
+        image2 = lsst.afw.image.ImageF(bbox)
+        image3 = lsst.afw.image.ImageF(bbox)
+        image4 = lsst.afw.image.ImageF(bbox)
+        field.fillImage(image1)
+        field.fillImage(image2, xStep=3)
+        field.fillImage(image3, yStep=4)
+        field.fillImage(image4, xStep=3, yStep=4)
+        self.assertFloatsAlmostEqual(image1.array, image2.array, rtol=2E-2, atol=2E-2)
```
parejkoj: Shouldn't the tolerances for these each be different, since the step sizes are different?

src/math/BoundedField.cc
TallJimbo: I'm still learning when use C++ lambdas myself, but I like the way this encapsulates the local functions inside the only scope where they're used.  Would be interested in @pschella's and @kfindeisen's opinions on best-practices here.

src/math/BoundedField.cc
TallJimbo: It's conceptually a special-cased last iteration of the loop, and I like to do that at the same/scope indentation level as the rest of the loop body.  I'll add a comment to that effect.

TallJimbo:
The change from `instFlux`->`flux` (etc) is a single commit that can and should be reverted easily later.  Until the `SourceCatalogs` we produce adopt the new conventions, I think it's much more valuable to have usable code than self-consistent naming.

And as much as I'm in favor of switching to the new conventions, I'm worried that backwards compatibility requirements may drag that conversion out longer than we'd all like.

src/math/BoundedField.cc
kfindeisen: Hmm... dejà vu.

I agree with @parejkoj that these lambdas are way too long and that they should be proper functions (C++ doesn't have anonymous functions, BTW). Having this code in the middle of `applyToImage` just makes it hard to read (I already mistook the end of `interpolate` for the end of `applyToImage`, for example). Lambdas are designed for the strategy pattern and, to a lesser degree, functional programming, and neither seems to be happening here.

kfindeisen: Also, this is a very long parameter list with identical types. Is there some way to factor it (XYZ triplets, BoundingBoxes, etc.)?

TallJimbo: I've refactored the interpolation code along both of these lines (no more lambdas, factored parameter lists).  Please feel welcome (but not compelled) to take another look.

PR #254, opened 2017-07-06
DM-10777: Create TransformBoundedField

r-owen:


tests/test_transformBoundedField.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def setUp(self):
+        self.longMessage = True
+
+        # an arbitrary bounding box (not that this kind of field cares)
+        self.bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(-3, 4),
+                                        lsst.afw.geom.Extent2I(5, 30))
+
+        # a list of points contained in the bbox
+        self.pointList = lsst.afw.geom.Box2D(self.bbox).getCorners()
```
pschella: Are corners still considered to be in the box?

tests/test_transformBoundedField.py
```diff
@@ -0,0 +1,171 @@
...
+
+            x, y = point
+            res2 = self.boundedField.evaluate(x, y)
+            self.assertFloatsAlmostEqual(res2, predRes)
+
+        resArr = self.boundedField.evaluate(self.xList, self.yList)
+        # applylForward returns an array with one row of values per axis
+        # and in this case there is just one axis
+        predResArr = self.transform.applyForward(self.pointList)[0]
+        assert_allclose(resArr, predResArr)
```
pschella: What about things outside of the box?

tests/test_transformBoundedField.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        multFactor = 5.3
+
+        predResult = self.transform.applyForward(self.pointList)[0] * multFactor
+
+        scaledField1 = self.boundedField * multFactor
+        assert_allclose(scaledField1.evaluate(self.xList, self.yList), predResult)
+
+        scaledField2 = multFactor * self.boundedField
+        assert_allclose(scaledField2.evaluate(self.xList, self.yList), predResult)
```
pschella: What about very large, very small, negative or otherwise problematic or invalid multiplication factors?

tests/test_transformBoundedField.py
```diff
@@ -0,0 +1,171 @@
...
+        readField = TransformBoundedField.readFits(filename)
+
+        resArr = self.boundedField.evaluate(self.xList, self.yList)
+        readResArr = readField.evaluate(self.xList, self.yList)
+        assert_allclose(resArr, readResArr)
+        self.assertEqual(readField.getBBox(), self.bbox)
+
+        self.assertTrue(self.boundedField == readField)
+        self.assertFalse(self.boundedField != readField)
+        self.assertEqual(self.boundedField, readField)
```
pschella: Move equality tests up before the previous block.

include/lsst/afw/math/TransformBoundedField.h
```diff
@@ -0,0 +1,94 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2017 LSST Corporation.
```
r-owen: It starts out as the year of authorship. That is changed to a range for future edits in a later year.

include/lsst/afw/math/TransformBoundedField.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+class TransformBoundedField : public table::io::PersistableFacade<TransformBoundedField>,
+                              public BoundedField {
+public:
+    using Transform = geom::Transform<geom::Point2Endpoint, geom::GenericEndpoint>;
+
+    /**
+     *  Create a TransformBoundedField from a bounding box and transform.
+     */
+    TransformBoundedField(geom::Box2I const& bbox, Transform const& transform);
```
r-owen: Good point. I'll emulate Transform, which allows copying but not assignment.

include/lsst/afw/math/TransformBoundedField.h
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Get the contained Transform
+    Transform getTransform() const { return _transform; }    
+
+    /// @copydoc BoundedField::evaluate
+    virtual double evaluate(geom::Point2D const & position) const override;
+
+    /// @copydoc BoundedField::evaluate
+    virtual ndarray::Array<double, 1, 1> evaluate(ndarray::Array<double const, 1> const& x,
+                                                  ndarray::Array<double const, 1> const& y) const override;
```
r-owen: That was true before this ticket. I changed it to virtual on this ticket, for this reason.

include/lsst/afw/math/TransformBoundedField.h
```diff
@@ -0,0 +1,94 @@
...
+
+    using BoundedField::evaluate;
+
+    /// TransformBoundedField is always persistable.
+    virtual bool isPersistable() const override { return true; }
```
r-owen: I think there are only these three.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include <memory>
+#include <string>
+
+#include "ndarray/eigen.h"
+#include "astshim.h"
```
r-owen: I consider astshim to be 3rd party, in that it is meant to be standalone and does not use the lsst namespace

include/lsst/afw/math/TransformBoundedField.h
```diff
@@ -0,0 +1,94 @@
...
+
+    /// @copydoc BoundedField::operator*
+    virtual std::shared_ptr<BoundedField> operator*(double const scale) const;
+
+    /// @copydoc BoundedField::operator==
+    virtual bool operator==(BoundedField const& rhs) const;
+
+protected:
```
r-owen: I suspect protected is required because these methods aren't called by code in TransformBoundedField, but I'm not sure. I copied ChebyshevBoundedField from @TallJimbo without thinking very hard about it. I propose to leave it for now and consider changing both later, if needed.

python/lsst/afw/math/transformBoundedField.cc
```diff
@@ -0,0 +1,86 @@
...
+namespace lsst {
+namespace afw {
+namespace math {
+namespace {
+
+using ClsField = py::class_<TransformBoundedField, std::shared_ptr<TransformBoundedField>, BoundedField,
+                            lsst::afw::table::io::PersistableFacade<TransformBoundedField>>;
+
+PYBIND11_PLUGIN(_transformBoundedField) {
+    py::module mod("_transformBoundedField", "Python wrapper for afw _transformBoundedField library");
```
r-owen: It looks like several sub-packages escaped cleanup, including cameraGeom, fits and possibly others. I can't remove the leading underscore without fixing up the whole package and using addUnderscore=False in SConscript, which is out of scope of this ticket. So I'll remove the comment for now.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+        : BoundedField(bbox), _transform(transform) {
+    if (transform.getToEndpoint().getNAxes() != 1) {
+        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError,
+                          "To EndPoint of transform must have 1 axis but has " +
+                                  std::to_string(transform.getToEndpoint().getNAxes()));
+    }
+}
+
+double TransformBoundedField::evaluate(geom::Point2D const& position) const {
+    return _transform.applyForward(position)[0];
```
r-owen: _transform is of a type `Transform<Point2Endpoint, GenericEndpoint>` and the generic endpoint has a single axis. As such, the single-point version of `applyForward` takes a Point2D and returns a vector with one element (for the one axis). So from that perspective it is certainly safe. Does that answer your question, or were you thinking of something else?

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+        xy[0][col] = x[col];
+        xy[1][col] = y[col];
+    }
+
+    auto res2D = _transform.getFrameSet()->applyForward(xy);
+
+    // res2D has shape 1 x N; return a 1-D view with the extra dimension stripped
+    auto resShape = ndarray::makeVector(nPoints);
+    auto resStrides = ndarray::makeVector(1);
+    return ndarray::external(res2D.getData(), resShape, resStrides, res2D);
```
r-owen: Yes, that is why `res2D` is provided as the final argument. However, I'll copy @TallJimbo on this, just to be absolutely sure.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+
+struct PersistenceHelper {
+    table::Schema schema;
+    table::Key<int> orderX;
+    table::PointKey<int> bboxMin;
+    table::PointKey<int> bboxMax;
+    table::Key<std::string> frameSet;
+
+    PersistenceHelper()
+            : schema(),
```
r-owen: `orderX` is not supposed to exist; it was a holdover from `ChebyshevBoundedField`. Thanks for catching that.

src/math/TransformBoundedField.cc
TallJimbo: Yes, exactly.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+        // NOTE: needed invert=false in case min=-1, max=0 (empty bbox). See RFC-324 and DM-10200
+        geom::Box2I bbox(record.get(keys.bboxMin), record.get(keys.bboxMax), false);
+        auto frameSetStr = record.get(keys.frameSet);
+        auto stream = ast::StringStream(frameSetStr);
+        auto astObjectPtr = ast::Channel(stream).read();
+        auto frameSet = std::dynamic_pointer_cast<ast::FrameSet>(astObjectPtr);
+        if (!frameSet) {
+            throw LSST_EXCEPT(pex::exceptions::RuntimeError,
+                              "The contents of the frameSet field could not be parsed as an ast::FrameSet");
+        }
```
r-owen: Agreed. I'll think about this more before merging. Adding `Transform::toString` and `fromString` would be a big help here, though it may be a bit messy.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+        auto transform = geom::Transform<geom::Point2Endpoint, geom::GenericEndpoint>(*frameSet);
+        return std::make_shared<TransformBoundedField>(bbox, transform);
+    }
+
+    TransformBoundedFieldFactory(std::string const& name) : afw::table::io::PersistableFactory(name) {}
```
r-owen: OK. I'll change ChebyshevBoundedField as well, to keep them in synch

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+};
+
+std::string getTransformBoundedFieldPersistenceName() { return "TransformBoundedField"; }
+
+TransformBoundedFieldFactory registration(getTransformBoundedFieldPersistenceName());
```
r-owen: I am very reluctant to depart from the way ChebyshevBoundedField is doing persistence without understanding it a lot better.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+
+std::string TransformBoundedField::getPythonModule() const { return "lsst.afw.math"; }
+
+void TransformBoundedField::write(OutputArchiveHandle& handle) const {
+    PersistenceHelper const keys;
+    table::BaseCatalog catalog = handle.makeCatalog(keys.schema);
+    std::shared_ptr<table::BaseRecord> record = catalog.addNew();
+    record->set(keys.bboxMin, getBBox().getMin());
+    record->set(keys.bboxMax, getBBox().getMax());
+    record->set(keys.frameSet, getTransform().getFrameSet()->show());
```
r-owen: Agreed. If I write `Transform::toString` it'll probably look a lot cleaner.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+    auto rhsCasted = dynamic_cast<TransformBoundedField const*>(&rhs);
+    if (!rhsCasted) return false;
+
+    return getBBox() == rhsCasted->getBBox() &&
+           *(getTransform().getFrameSet()) == *(rhsCasted->getTransform().getFrameSet());
+}
+
+std::string TransformBoundedField::toString() const {
+    std::ostringstream os;
+    os << "TransformBoundedField (containing " << _transform.getFrameSet()->getNFrame() << " frames)";
```
r-owen: The internal string representation is many lines of text -- not at all suitable for a basic representation. Unfortunately I don't know of a good way to summarize a frame set. @parejkoj requested the # of frames as better than nothing, but I disagree; I think it is meaningless (and will usually be 2). I'm going to drop it for now.

tests/test_transformBoundedField.py
r-owen: I believe so. In this case it doesn't matter -- the functional form of the transform has no intrinsic limits; the bbox is arbitrary.

tests/test_transformBoundedField.py
r-owen: `BoundedField.evaluate` does not checks limits, so such a test is not helpful.

tests/test_transformBoundedField.py
r-owen: I'll add a few values. I don't think we need to torture-test.

src/math/TransformBoundedField.cc
```diff
@@ -0,0 +1,172 @@
...
+                                                    false)) {}
+
+    PersistenceHelper(table::Schema const& s)
+            : schema(s), bboxMin(s["bbox_min"]), bboxMax(s["bbox_max"]), frameSet(s["frameSet"]) {}
+};
+
+class TransformBoundedFieldFactory : public table::io::PersistableFactory {
+public:
+    virtual std::shared_ptr<table::io::Persistable> read(InputArchive const& archive,
+                                                         CatalogVector const& catalogs) const {
```
r-owen: override

src/math/TransformBoundedField.cc
r-owen: Based on Slack discussion on the software-dev, we now have a standard: `writeString` will write a serialization to a string and `readString` will deserialize from a string.

include/lsst/afw/math/TransformBoundedField.h
pschella: The template used to be different. But no problem, just curious.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -208,6 +208,35 @@ class SkyWcs : public Transform<Point2Endpoint, IcrsCoordEndpoint> {
...
+     * - The short class name, as obtained from getShortClassName
+     * - A space
+     * - The contained ast::FrameSet written using FrameSet.show(os, false)
+     *
+     * @param[out] os  output stream to which to serialize this SkyWcs
+     */
+    void writeStream(std::ostream & os) const override;
+
+    /// Serialize this SkyWcs to a string, using the same format as writeStream
+    virtual std::string writeString() const;
```
pschella: `override` instead of `virtual`?

include/lsst/afw/geom/detail/transformUtils.h
```diff
@@ -0,0 +1,116 @@
...
+    auto frameSet = std::dynamic_pointer_cast<ast::FrameSet>(astObjectPtr);
+    if (!frameSet) {
+        std::ostringstream os;
+        os << "The AST serialization was read as a " << astObjectPtr->getClassName()
+           << " instead of a FrameSet";
+        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError, os.str());
+    }
+
+    return Transform(*frameSet);
+}
```
pschella: Nice

include/lsst/afw/geom/detail/transformUtils.h
```diff
@@ -0,0 +1,116 @@
...
+
+template<class Transform>
+void writeStream(Transform const & transform, std::ostream & os) {
+    os << "1 " << Transform::getShortClassName();
```
pschella: Please add a `serializationVersion` (or other name, perhaps just `version`) `constexpr` member for the version magic number.

include/lsst/afw/math/TransformBoundedField.h
jdswinbank: It should be different — not because of the date, but because the copyright isn't held by LSST Corporation. Our usual convention is to use "Copyright NNNN AURA/LSST" (which is likely also incorrect, but at least it's in the template...).

include/lsst/afw/geom/detail/transformUtils.h
r-owen: OK. Given that I the unpersistence code is in the  file, and has to use hard-coded numbers, I don't see a lot of gain, but I guess it beats writing an int as a string. I added `currentVersion` as a local constant to the .cc file.

PR #256, opened 2017-07-06
DM-6718 Add str and repr for afw catalog/record

parejkoj:


PaulPrice:
Great addition John, thanks!

python/lsst/afw/table/base/baseContinued.py
```diff
@@ -315,5 +315,15 @@ def __getattr__(self, name):
         except AttributeError:
             return getattr(self.columns, name)
 
+    def __str__(self):
+        if self.isContiguous():
+            return str(self.asAstropy())
+        else:
+            fields = ' '.join(x.field.getName() for x in self.schema)
```
natelust: Why does the Astropy implementation print out rows (so you get something that looks like a table) and ours does not?

python/lsst/afw/table/base/baseContinued.py
```diff
@@ -21,15 +21,15 @@
...
-
 from future.utils import with_metaclass
 
 import numpy as np
 
 from lsst.utils import continueClass, TemplateMeta
 from .base import BaseRecord, BaseCatalog
 
+__all__ = ["Catalog"]
+
```
kfindeisen: Is this something that's been announced to the LSST community? The last I'd heard about this issue I'd gotten the impression that it was the guidelines that were not PEP8-compliant.

jdswinbank: It was announced: https://community.lsst.org/t/--all---should-go-before-imports-in-python-code/1769/3

parejkoj: Got it! Yes, my flake8 needed to be upgraded.

python/lsst/afw/table/base/baseContinued.py
parejkoj: We discussed this offline, but I'll make a response here, after talking it over with @SimonKrughoff :

We don't want to be constantly patching over the annoyances we have with our non-contiguous tables. We should instead fix them to behave better in general. So, instead of adding extra special casing to make non-contiguity "nicer" in each individual case, we'll do a "minimum functionality" thing and fix the non-contiguous tables themselves, by e.g. making numpy views into them that behave correctly.

PR #257, opened 2017-07-08
Tickets/dm 11236

RobertLuptonTheGood:
See also obs_decam (changed a test)

### DIFF
--- a/include/lsst/afw/image/fits/fits_io.h
+++ b/include/lsst/afw/image/fits/fits_io.h
@@ -56,8 +56,14 @@ inline void fits_read_array(fits::Fits& fitsfile, ndarray::Array<PixelT, 2, 2>& 
         }
         shape = shape3.last<2>();
     }
-
-    fitsfile.readMetadata(metadata, true);
+    /*
+     * TODO: DM-11235 We cannot call fitsfile.readMetadata(metadata, true)
+     * as it doesn't handle INHERIT correctly
+     */
+    const int hdu = fitsfile.getHdu();
+    std::shared_ptr<daf::base::PropertyList> header(readMetadata(fitsfile, true));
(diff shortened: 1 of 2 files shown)


include/lsst/afw/image/fits/fits_io.h
```diff
@@ -56,8 +56,14 @@ inline void fits_read_array(fits::Fits& fitsfile, ndarray::Array<PixelT, 2, 2>&
         }
         shape = shape3.last<2>();
     }
-
-    fitsfile.readMetadata(metadata, true);
+    //
+    // Note that we cannot call fitsfile.readMetadata(metadata, true)
```
parejkoj: Should this note be a TODO?

Also, no need for the extra `//` comment lines above and below.

include/lsst/afw/image/fits/fits_io.h
```diff
@@ -56,8 +56,14 @@ inline void fits_read_array(fits::Fits& fitsfile, ndarray::Array<PixelT, 2, 2>&
...
+    // as it doesn't handle INHERIT correctly (DM-11235)
+    //
+    const int hdu = fitsfile.getHdu();
+    std::shared_ptr<daf::base::PropertyList> header(readMetadata(fitsfile, true));
```
parejkoj: Use `make_shared` instead?

RobertLuptonTheGood: It won't compile; I think that the problem is that there's some PropertyList copy/move operator that doesn't exist.

parejkoj: Interesting. Because it looks like you did just that in fits.cc:1047 below. What's the difference between them?

PR #258, opened 2017-07-11
DM-7477: Increase mask plane size to 32 bits

natelust:


src/image/Mask.cc
```diff
@@ -461,7 +461,7 @@ Mask<MaskPixelT>::Mask(fits::Fits& fitsfile, std::shared_ptr<daf::base::Property
                        afw::geom::Box2I const& bbox, ImageOrigin const origin, bool const conformMasks)
         : ImageBase<MaskPixelT>(), _maskDict(detail::MaskDict::makeMaskDict()) {
     // These are the permitted input file types
-    typedef boost::mpl::vector<unsigned char, unsigned short, short> fits_mask_types;
+    typedef boost::mpl::vector<unsigned char, unsigned short, short, std::int32_t> fits_mask_types;
```
PaulPrice: `MaskPixel` instead of `std::int32_t`?

PaulPrice:
I think you need to add a test for the legacy `HeavyFootprint` support.

PR #259, opened 2017-07-11
Tickets/dm 10953

yalsayyad:


tests/test_kernel.py
```diff
@@ -112,6 +116,53 @@ def testAnalyticKernel(self):
 
         self.verifyCache(kernel, hasCache=False)
 
+    def verifyResized(self, kernel, testClip=True, oddPadRaises=False):
+        """Verify kernels can be resized properly.
+
+        @param testClip bool     Test whether a kernel can be safely
+                                 resized such that one dimension is shorter than original.
+                                 Kernels that start out smaller than 5x5 cannot in this test.
+        @param oddPadRaises bool Test that attempting to resize by an odd number of pixel raises
```
pschella: pixels

tests/test_kernel.py
```diff
@@ -112,6 +116,53 @@ def testAnalyticKernel(self):
...
+
+        kWidth, kHeight = kernel.getDimensions()
+
+        badDims = [(0, 0), (-1, 1), (3, -3)]  # impossible kernel dimensions
+        goodPads = [(0, 0), (4, 2), (2, 4)]  # all kernels can resize by even, non-negative number of pixels
+        oddPads = [(3, 3), (1, 3), (3, 2), (0, 3)]  # at least one padding dimension is odd
+        clippingPads = [(-2, -2), (2, -2), (-2, 2)]  # at least one dimension is smaller than orig
+
+        if oddPadRaises:
+            badDims += [(kWidth + w, kHeight + h) for (w, h) in oddPads]
```
pschella: Personally I find `.extend` clearer than `+=` on a list.

tests/test_kernel.py
```diff
@@ -112,6 +116,53 @@ def testAnalyticKernel(self):
...
+        @param oddPadRaises bool Test that attempting to resize by an odd number of pixel raises
+                                 an InvalidParameterError.
+                                 Fixed/Delta Kernels cannot be resized by an odd amount because
+                                 they are padded/clipped rather than reevalutated like Analytic Kernels.
+
+        """
+
+        kWidth, kHeight = kernel.getDimensions()
+
+        badDims = [(0, 0), (-1, 1), (3, -3)]  # impossible kernel dimensions
```
pschella: What about `(0, 1)` or `(1,0)`?

tests/test_kernel.py
```diff
@@ -112,6 +116,53 @@ def testAnalyticKernel(self):
...
+        # Check that error is raised when bad dimensions specified (and optionally odd padding amounts)
+        for badWidth, badHeight in badDims:
+            with self.assertRaises(pexExcept.InvalidParameterError):
+                kernelResized = kernel.resized(badWidth, badHeight)
+
+        # Check that resizing is successful
+        for padX, padY in goodPads:
+            # Do not test if padding will make resized kernel smaller than 0
+            if kWidth <= -padX or kHeight <= -padY:
+                continue
```
pschella: Should still be tested right?

tests/test_kernel.py
```diff
@@ -807,6 +873,52 @@ def basicTestComputeImageRaise(self, kernel, doRaise, kernelDescr=""):
...
+                           (kernel1.getSpatialParameters(), kernel2.getSpatialParameters()))
+        if kernel1.getNSpatialParameters() != kernel2.getNSpatialParameters():
+            retStrs.append("# spatial parameters differs: %s != %s" %
+                           (kernel1.getNSpatialParameters(), kernel2.getNSpatialParameters()))
+        if not kernel1.isSpatiallyVarying() and hasattr(kernel1, "getKernelParameters"):
+            if kernel1.getKernelParameters() != kernel2.getKernelParameters():
+                retStrs.append("kernel parameters differs: %s != %s" %
+                               (kernel1.getKernelParameters(), kernel2.getKernelParameters()))
+        if retStrs:
+            return "; ".join(retStrs)
```
pschella: Why not short circuit the return in each of the cases?

tests/test_gaussianPsf.py
```diff
@@ -125,6 +125,20 @@ def testBBox(self):
...
+            resizedPsf = self.psf.resized(newLen, newLen)
+            newBBox = resizedPsf.computeBBox()
+            self.assertEqual(newBBox.getWidth(), newLen)
+            self.assertEqual(newBBox.getHeight(), newLen)
+            self.assertEqual(resizedPsf.getSigma(), self.psf.getSigma())
+
+            image = resizedPsf.computeKernelImage()
+            check = makeGaussianImage(newBBox, self.psf.getSigma())
+            self.assertFloatsAlmostEqual(image.getArray(), check.getArray())
+            self.assertFloatsAlmostEqual(image.getArray().sum(), 1.0, atol=1E-14)
```
pschella: Why this value?

src/math/FixedKernel.cc
```diff
@@ -59,6 +59,45 @@ std::shared_ptr<Kernel> FixedKernel::clone() const {
...
+        << "), because at least one dimension would change by an odd value.";
+        throw LSST_EXCEPT(pexExcept::InvalidParameterError, os.str());
+    }
+
+    lsst::afw::geom::Box2I bboxNew(lsst::afw::geom::Point2I((1 - width)/2, (1 - height)/2),
+                                   lsst::afw::geom::Extent2I(width, height));
+    std::shared_ptr<image::Image<Pixel>> imNew(new image::Image<Pixel>(bboxNew));
+
+    lsst::afw::geom::Box2I bboxIntersect = this->getBBox();
+    bboxIntersect.clip(bboxNew);
```
yalsayyad: I added the following comment explaining why modifying the BBox here is OK:
```
    // getBBox() instantiates a new BBox from member data _width, _height, _ctrX, _ctrY
    // modifying it is OK
```

src/math/FixedKernel.cc
```diff
@@ -59,6 +59,45 @@ std::shared_ptr<Kernel> FixedKernel::clone() const {
...
+    lsst::afw::geom::Box2I bboxIntersectShifted = lsst::afw::geom::Box2I(
+        lsst::afw::geom::Point2I(bboxIntersect.getMinX() + offsetX,
+                                 bboxIntersect.getMinY() + offsetY),
+        bboxIntersect.getDimensions());
+    image::Image<Pixel> imIntersect = image::Image<Pixel>(_image, bboxIntersectShifted);
+
+    imNew->assign(imIntersect, bboxIntersect);
+
+    std::shared_ptr<image::Image<Pixel> const> imConst = imNew;
+    std::shared_ptr<Kernel> retPtr(new FixedKernel(*imConst));
```
yalsayyad: Thank you.  I needed imNew because I'm only assigning  pixels on the new image where the old image and new image overlap.  I don't think we have `Image` constructor that can do that.  I did remove the imConst though (make_shared fixed the casting issue I was having before).

include/lsst/afw/detection/GaussianPsf.h
```diff
@@ -62,6 +62,9 @@ class GaussianPsf : public afw::table::io::PersistableFacade<GaussianPsf>, publi
     /// Polymorphic deep copy; should usually be unnecessary because Psfs are immutable.
     virtual std::shared_ptr<afw::detection::Psf> clone() const;
 
+    /// Return a clone with specified kernel dimensions
+    virtual std::shared_ptr<afw::detection::Psf> resized(int width, int height) const;
```
pschella: See previous comments (on related PRs) about using `override` instead of `virtual`.

src/math/LinearCombinationKernel.cc
```diff
@@ -107,6 +107,24 @@ std::shared_ptr<Kernel> LinearCombinationKernel::clone() const {
     return retPtr;
 }
 
+std::shared_ptr<Kernel> LinearCombinationKernel::resized(int width, int height) const {
+    KernelList kernelList;
+
+    for (KernelList::const_iterator kIter = _kernelList.begin(), kEnd = _kernelList.end(); kIter != kEnd;
+         ++kIter) {
+        std::shared_ptr<Kernel> basisKernelPtr = (*kIter)->resized(width, height);
+        kernelList.push_back(basisKernelPtr);
```
yalsayyad: Takes more characters and is harder to read (in my opinion) than

```
    for (const std::shared_ptr<Kernel>& kIter : getKernelList()) {
        kernelList.push_back(kIter->resized(width, height));
    }
```
Performance gains worth it?

tests/test_kernel.py
yalsayyad: Sure, why not. I replaced continue with another:
```
                with self.assertRaises(pexExcept.InvalidParameterError):
                    kernelResized = kernel.resized(kWidth + padX, kHeight + padY)
```

src/math/LinearCombinationKernel.cc
natelust: why not
`for (auto const & kIter : getKernelList()){`

yalsayyad: I'm surprised my personal opinion on `auto` hasn't come up in conversation before: it is easier to write, but harder to read.  Consider my typing out the type as a gift, so you don't need to go read the declaration of getKernelList().

natelust: I don't have a feeling either way, I was just wondering if that is why you chose to write it all out.

tests/test_kernel.py
yalsayyad: I'm guessing original author of `compareKernels` wanted all the kernel differences to be printed out rather than the first to fail.  I agree It is a little unexpected in a unit test. I pulled out some of the duplication between the kernel comparison functions, but left in the string concatenation.

PR #264, opened 2017-07-27
DM-11268: Detector: add crosstalk matrix

PaulPrice:
Contains coefficients for intra-CCD crosstalk correction.

### DIFF
--- a/python/lsst/afw/cameraGeom/cameraConfig.py
+++ b/python/lsst/afw/cameraGeom/cameraConfig.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import numpy as np
 import lsst.pex.config as pexConfig
 from .transformConfig import TransformMapConfig
 
@@ -44,6 +45,22 @@ class DetectorConfig(pexConfig.Config):
     # to be transposed before rotation to put it in camera coordinates.
     transposeDetector = pexConfig.Field(
         "Transpose the pixel grid before orienting in focal plane?", bool)
+
+    crosstalk = pexConfig.ListField(
+        dtype=float,
+        doc=("Flattened crosstalk coefficient matrix; should have nAmps x nAmps entries. "
+             "Once 'reshape'-ed, ``coeffs[i][j]`` is the fraction of the j-th amp present on the i-th amp."),
+        optional=True
+    )
+
+    def getCrosstalk(self, numAmps):
+        """Return a 2-D numpy array of crosstalk coefficients of the proper shape"""
+        if not self.crosstalk:
+            return None
+        try:
+            return np.array(self.crosstalk, dtype=np.float32).reshape((numAmps, numAmps))
+        except:
+            raise RuntimeError("Cannot reshape 'crosstalk' coefficients to square matrix")
 
 
 class CameraConfig(pexConfig.Config):
--- a/python/lsst/afw/cameraGeom/testUtils.py
+++ b/python/lsst/afw/cameraGeom/testUtils.py
@@ -39,6 +39,7 @@ class DetectorWrapper(object):
                  orientation=Orientation(),
                  plateScale=20.0,
                  radialDistortion=0.925,
+                 crosstalk=None,
                  modFunc=None,
                  ):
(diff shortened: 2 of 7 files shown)


python/lsst/afw/cameraGeom/testUtils.py
```diff
@@ -104,6 +106,9 @@ def __init__(self,
             CameraSys(TAN_PIXELS, self.name): pixelToTanPixel,
             CameraSys(ACTUAL_PIXELS, self.name): afwGeom.RadialXYTransform([0, 0.95, 0.01]),
         }
+        if crosstalk is None:
+            crosstalk = [[0.0 for _ in range(numAmps)] for _ in range(numAmps)]
```
parejkoj: Shouldn't this be `[0,0]` (or maybe `[]`) based on the logic in `afw/cameraGeom/Detector.h:136`? Otherwise, it looks like there is a crosstalk matrix that just happens to be all zero.

src/cameraGeom/Detector.cc
```diff
@@ -117,6 +119,22 @@ void Detector::_init() {
             throw LSST_EXCEPT(pexExcept::InvalidParameterError, os.str());
         }
     }
+
+    // ensure crosstalk coefficients matrix is square
+    if (hasCrosstalk()) {
+        auto shape = _crosstalk.getShape();
+        assert(shape.size() == 2);  // we've declared this as a 2D array
+        if (shape[0] != shape[1]) {
```
parejkoj: These all seem like good tests. :+1:

tests/test_detector.py
```diff
@@ -91,6 +93,16 @@ def addBadCameraSys(dw):
         with self.assertRaises(lsst.pex.exceptions.Exception):
             DetectorWrapper(modFunc=addBadCameraSys)
 
+        # These break in the pybind layer
+        for crosstalk in ([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],  # 1D and not numpy
```
parejkoj: You have failed crosstalk matrices, but are there tests for ok ones?

tests/test_detector.py
```diff
@@ -91,6 +93,16 @@ def addBadCameraSys(dw):
...
+                          np.array([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0]),  # 1D, wrong numpy type
+                          np.array([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0], dtype=np.float32),  # 1D
+                          ):
+            self.assertRaises(TypeError, DetectorWrapper, crosstalk=crosstalk)
+        # This breaks in the Detector ctor: wrong shape
+        self.assertRaises(lsst.pex.exceptions.InvalidParameterError,
```
parejkoj: Since there are two different `throw`s in `detector.cc`, you should probably have one test for each of those conditions.

python/lsst/afw/cameraGeom/cameraConfig.py
```diff
@@ -45,6 +46,19 @@ class DetectorConfig(pexConfig.Config):
     transposeDetector = pexConfig.Field(
         "Transpose the pixel grid before orienting in focal plane?", bool)
 
+    crosstalk = pexConfig.ListField(
+        dtype=float,
+        doc=("Flattened crosstalk coefficient matrix; should have nAmps x nAmps entries. "
```
PaulPrice: Yes. We don't have a `MatrixField` in pex_config.

tests/test_detector.py
PaulPrice: Yes, the default in testUtils.py:110.

python/lsst/afw/cameraGeom/testUtils.py
PaulPrice: This is a crosstalk matrix that is all zero. I could put other dummy values instead of zeros, but it wouldn't add anything to the test.

PaulPrice:
Changes made in response to review are on the fixup commit.

PR #262, opened 2017-07-24
DM-10729: Implement PhotoCalib operations that output to catalogs.

TallJimbo:
NOTE: this PR is a comparison against tickets/DM-10728, not master; tickets/DM-10729 must be rebased on master after tickets/DM-10728 is merged.

This implements two `PhotoCalib` methods that had previously been disabled pending RFC-322.  I have implemented them using pre-RFC-322 conventions so they're usable now; the strings involved can easily be changed after RFC-322 has been implemented elsewhere.

PR #265, opened 2017-07-31
Implement PhotoCalib ops that output to catalogs.

TallJimbo:
Names used here will have to be updated when RFC-322 is adopted.

This PR is a new one (against master) that supercedes #262.

### DIFF
--- a/src/image/PhotoCalib.cc
+++ b/src/image/PhotoCalib.cc
@@ -102,8 +102,16 @@ ndarray::Array<double, 2, 2> PhotoCalib::instFluxToMaggies(afw::table::SourceCat
 
 void PhotoCalib::instFluxToMaggies(afw::table::SourceCatalog &sourceCatalog, std::string const &instFluxField,
                                    std::string const &outField) const {
-    throw LSST_EXCEPT(pex::exceptions::LogicError,
-                      "Will not be implemented until RFC-322 is implemented. See DM-10155");
+    auto instFluxKey = sourceCatalog.getSchema().find<double>(instFluxField + "_flux").key;
+    auto instFluxErrKey = sourceCatalog.getSchema().find<double>(instFluxField + "_fluxSigma").key;
+    auto maggiesKey = sourceCatalog.getSchema().find<double>(outField + "_calFlux").key;
+    auto maggiesErrKey = sourceCatalog.getSchema().find<double>(outField + "_calFluxErr").key;
+    for (auto & record : sourceCatalog) {
+        auto result = instFluxToMaggies(record.get(instFluxKey), record.get(instFluxErrKey),
+                                        record.getCentroid());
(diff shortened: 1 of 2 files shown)


src/image/PhotoCalib.cc
```diff
@@ -102,8 +102,16 @@ ndarray::Array<double, 2, 2> PhotoCalib::instFluxToMaggies(afw::table::SourceCat
...
 void PhotoCalib::instFluxToMaggies(afw::table::SourceCatalog &sourceCatalog, std::string const &instFluxField,
                                    std::string const &outField) const {
-    throw LSST_EXCEPT(pex::exceptions::LogicError,
-                      "Will not be implemented until RFC-322 is implemented. See DM-10155");
+    auto instFluxKey = sourceCatalog.getSchema().find<double>(instFluxField + "_flux").key;
+    auto instFluxErrKey = sourceCatalog.getSchema().find<double>(instFluxField + "_fluxSigma").key;
+    auto maggiesKey = sourceCatalog.getSchema().find<double>(outField + "_calFlux").key;
+    auto maggiesErrKey = sourceCatalog.getSchema().find<double>(outField + "_calFluxErr").key;
+    for (auto & record : sourceCatalog) {
+        auto result = instFluxToMaggies(record.get(instFluxKey), record.get(instFluxErrKey),
```
r-owen: Consider `auto const`.

I'm also curious if `auto const &` would be give better performance, but I hope the answer is "no".

src/image/PhotoCalib.cc
```diff
@@ -153,8 +161,16 @@ ndarray::Array<double, 2, 2> PhotoCalib::instFluxToMagnitude(afw::table::SourceC
...
 void PhotoCalib::instFluxToMagnitude(afw::table::SourceCatalog &sourceCatalog,
                                      std::string const &instFluxField, std::string const &outField) const {
-    throw LSST_EXCEPT(pex::exceptions::LogicError,
-                      "Will not be implemented until RFC-322 is implemented. See DM-10155.");
+    auto instFluxKey = sourceCatalog.getSchema().find<double>(instFluxField + "_flux").key;
+    auto instFluxErrKey = sourceCatalog.getSchema().find<double>(instFluxField + "_fluxSigma").key;
+    auto magKey = sourceCatalog.getSchema().find<double>(outField + "_mag").key;
+    auto magErrKey = sourceCatalog.getSchema().find<double>(outField + "_magErr").key;
+    for (auto & record : sourceCatalog) {
+        auto result = instFluxToMagnitude(record.get(instFluxKey), record.get(instFluxErrKey),
```
r-owen: Likewise

src/image/PhotoCalib.cc
TallJimbo: It shouldn't change the performance, but we can't use it here anyway because we also call `set`.

PR #268, opened 2017-08-01
DM-6718 Add str and repr for afw catalog/record

parejkoj:
Re-merge, after fixing the variable-length arrays problem in DM-11429.

### DIFF
--- a/python/lsst/afw/table/base/baseContinued.py
+++ b/python/lsst/afw/table/base/baseContinued.py
@@ -21,14 +21,14 @@
 #
 from __future__ import absolute_import, division, print_function
 
-__all__ = ["Catalog"]
-
 from future.utils import with_metaclass
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


timj:
there is no merge commit indicating that 2 of these commits were distinct.

parejkoj:
I'm confused by your comment, @timj ? Why would there be such?

timj:
Oh. I was confused by the base commit of this PR. Are you merging this branch directly to master or are you merging this branch to 11429 and then merging 11429 to master? I expected the latter.

parejkoj:
I think my plan is to merge 11429 to master, then rebase this onto master (which shouldn't change anything), then merge this. But I'm open to suggestions.

timj:
That is one option. Doing it in two stages like that is fine with me (so this PR is not itself being merged).

PR #269, opened 2017-08-08
DM-11514: Fix some pytest problems

timj:
* Use correct name of pytest initialization function.
* Rename a test file.
* Cleanup logic in SConscript to ignore pybind11 test files.


kfindeisen:
I'm a bit surprised by the "Ignore..." and "Fix name of..." commits, because I've run `pytest` with no arguments from the command line with no problems. Is this an OS- or otherwise environment-specific issue?

timj:
@kfindeisen if you run `pytest` without arguments it will discover most of the tests but not the one test that was named incorrectly. Also, the `ignore` is for SCons' benefit so that it won't try to test that file. If you ask pytest to test a file without any tests in it, pytest returns non-zero exit status as it assumes you've made a mistake somewhere.

kfindeisen:
Ok, I didn't realize `sconsUtils` did its own file search.

timj:
Should have mentioned this PR depends on lsst/sconsUtils#34.

timj:
@kfindeisen does this count as a review?

kfindeisen:
No, I was asking for personal edification.

tests/SConscript
```diff
@@ -10,6 +10,7 @@ afwdataDir = env.ProductDir("afwdata")
 if afwdataDir:
     env["ENV"]["AFWDATA_DIR"] = afwdataDir
 
-scripts.BasicSConscript.tests(noBuildList=[name + '.cc' for name in pybind11_test_modules])
+scripts.BasicSConscript.tests(noBuildList=[name + '.cc' for name in pybind11_test_modules],
+                              ignoreList=[name + '.py' for name in pybind11_test_modules])
```
r-owen: I'm curious: why is this change needed? I suppose if you use `pytest tests/*.py` instead of `pytest`, as we may have to do initially, that would explain it.

timj: This is needed to stop SCons from thinking that the pybind11 .py files are things that should be tested. This is important when we are in a mode where pytest runs each test file separately because pytest exits with bad status if no tests are found. See my comment above replying to @kfindeisen. With these changes `pytest` without arguments would work but `sconsUtils` is not yet set up to do that (that's a change we can make later once the basics are done).

timj:
The tweaking of the {{afwdata}} stuff in the SCons file is a left over of DM-609. We don't need to filter out any tests because the C++ tests trap the situation themselves. I have tested with and without afwdata.

PR #270, opened 2017-08-11
DM-10843: Implement new statistics framework

pschella:
Will not be merged to master! Only for intermediate review purposes.


include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+    static const bool value = std::is_same<T, U>::value;
+};
+
+template <typename T, typename U, typename... Ts>
+struct IsIn<T, U, Ts...> {
+    static const bool value = std::is_same<T, U>::value || IsIn<T, Ts...>::value;
+};
+
+/**
+ * Type trait. If `T` is part of `ProductSet<Ts...>`, provides the member constant value equal to `true`.
```
r-owen: Please put "value" in backticks, e.g. "provides the member constant `value`..."

include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+};
+
+/**
+ * Type function to add types `Ts...` to `ProductSet` `PS`.
+ *
+ * Keeps the invariant that if `PS` is a set of unique (and not `None`) types,
+ * so is `ProductSetInsertMultiple<PS, Ts...>::type`.
+ */
+template <typename PS, typename... Ts>
+struct ProductSetInsertMultiple;
```
r-owen: I worry that users will find it confusing to have `ProductSetInsert` work so differently than `ProductSetInsertMultiple`. `ProductSetInsert ` is templated on the contained product types, with the new type first, whereas `ProductSetInsertMultiple` is templated on the product set followed by the types to be inserted.

Could you manage with just `ProductSetInsertMultiple` and rename it to `ProductSetInsert` (or `AddProduct` or `ProductSetAdd`, as suggested above), which inserts one or more products into a `ProductSet`? If so, I think it would be less confusing.

include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+};
+
+/**
+ * Type function to add type `T` to `ProductSet<Ts...>`.
+ *
+ * If `T` is not `None` and not already in `Ts...` then `type` is `ProductSet<T, Ts...>`,
+ * otherwise it is just `ProductSet<Ts...>`.
+ *
+ * Keeps the invariant that if `ProductSet<Ts...>` is a set of unique (and not `None`) types,
+ * so is `InsertProductSet<T, ProductSet<Ts...>>::type`.
```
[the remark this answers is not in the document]
r-owen: Isn't it important to be able to add a product to a `ProductSet` that contains a single `None`? If so, I would expect that to result in a `ProductSet` that only contains the inserted type (dropping the `None`)? I suppose adding `None` to `ProductSet<None>` results in `ProductSet<None>`, sigh.

This is one of the reasons I suggested dropping None above, using an empty ProductSet instead.

Also, a picky point, but to me "Insert" implies ordering, with the new `Product` type being inserted before some other product, or at the beginning. Technically `ProductSet` is ordered, but if my understanding is correct, the order doesn't matter (making `ProductSet` a good name). So I wonder if the name `add` would be clearer: e.g. `AddProduct` or `ProductSetAdd`.

include/lsst/afw/math/statistics/Statistics.h
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param img Image whose properties we want
+ * @param flags Describes what we want to calculate
+ * @param sctrl Control how things are calculated
+ *
+ * @returns statistics that were calculated.
+ */
+template <typename Pixel>
+StatisticsResult standardStatistics(lsst::afw::image::Image<Pixel> const& img, int flags,
+                                    statistics::StatisticsControl const& sctrl);
```
r-owen: Nice! This offers simple, direct access to the results..

It's a pity that C++ makes it so difficult to get just the statistics that were computed. Please consider returning a `PropertyList` with just one named entry for each computed statistic.  That makes access is slightly uglier. But it eliminates the ambiguity between "not computed because not requested" and "requested, but computed to be Nan". Also, it makes it easy to list all computed statistics.

On the Python side I see you return a `dict`. Please consider returning a `pipe_base` `Struct` instead. That gives direct access, e.g. `result.mean` while still allowing the user to see which fields are present (`dir(result)` or `result.getDict().keys()`) and even allowing a dict view (`result.getDict()`). I admit this will require `pipe_base` to finally be a dependency of `afw` but that seems a small price to pay to have access to `Struct`. Or we could move `Struct` somewhere else, such as `daf_base`.

include/lsst/afw/math/statistics/Validators.h
```diff
@@ -0,0 +1,87 @@
...
+#ifndef LSST_AFW_MATH_STATISTICS_VALIDATORS_H
+#define LSST_AFW_MATH_STATISTICS_VALIDATORS_H
+
+#include "lsst/afw/image/MaskedImage.h"
+
+namespace lsst {
+namespace afw {
+namespace math {
+namespace statistics {
+
```
r-owen: It appears that `Validator.operator()` should take 4 arguments: value, mask, variance, and weight. Please document that clearly somewhere in this file. (Initially I thought the variadic arguments were related to `PropertySet`!). Also, I suggest listing them explicitly; clarity at the expense of verbosity.  Thus, for example, I think `ValidateRange` would be clearer as:
```
    template <typename Value, typename Mask, typename Variance, typename Weight>
    bool operator()(Value val, Mask, Variance, Weight) const {
        Value tmp = fabs(val - _center);
        return (tmp <= _limit);
    }
```
It's tricky having 4 arguments and getting the order right. Listing them ought to help (though of course there is no way to prevent listing them in the wrong order).

r-owen: Why are validators listed as internal (the `@internal` tag in the doc string)? They appear to be part of the public API. For example `SinglePassStatistics` accepts a user-specified validator, and I assume it is public.

include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+/**
+ * Base class for statistics products (e.g. Sum, Count, Variance, etc.).
+ *
+ * A Product is templated on the intermediate Products it requires (or None),
+ * and defines an `accumulate` member function (default noop).
+ */
+template <typename... Ts>
+class Product {
+public:
+    using Intermediate = typename FlattenOne<ProductSet<Ts...>>::type;
```
r-owen: I'm surprised that you flatten only one level. Is this easy to explain? Does this fully avoid computing the same statistic multiple times for different products?

include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+
+    /**
+     * Combine two products
+     */
+    template <typename T>
+    void combine(T const&) { /* noop for abstract products */
```
r-owen: why not set all the no-op methods = 0 in the abstract base class?

include/lsst/afw/math/statistics/Framework.h
```diff
@@ -0,0 +1,615 @@
...
+};
+
+/**
+ * Class that represents a concrete Product.
+ *
+ * Creates an optimal packing of unique intermediate products and provides
+ * an `accumulate` member function that accumulates each one.
+ *
+ * Results may be extracted from such objects using the non-member `extract` function.
+ */
```
r-owen: Please explain the difference between `ConcreteProduct` and `Product`. When might you subclass one vs. the other?

jdswinbank:
This work can't be merged as is. Withdrawing the PR until somebody has time to pick up DM-10843 again.

PR #271, opened 2017-08-11
DM-5922: Rework camera geometry to use the replacement for XYTransform

r-owen:


include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * TransformMap supports transforming between any two supported CameraSys using
+ * the @ref transform methods, even if no explicit transform was provided.
+ *
+ * @exceptsafe Unless otherwise specified, all methods guarantee only basic
+ *             exception safety.
+ */
+class TransformMap final {
+public:
+    using Transform = afw::geom::Transform<geom::Point2Endpoint, geom::Point2Endpoint>;
```
TallJimbo: Defining a `typedef` with the same name as a template can cause a lot of problems (I don't recall exactly what the problems are, but I've always regretted doing it).

I'd also strongly recommend that you have a global typedef for `Transform<Point2Endpoint,Point2Endpoint>`, and just use that throughout (much as we do with e.g. `Point2D`).

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -0,0 +1,213 @@
...
+    using CameraSysIterator = boost::transform_iterator<GetKey, CameraSysFrameIdMap::const_iterator>;
+
+    /**
+     * Define a set of camera transforms.
+     *
+     * @param reference Coordinate system to which `transforms` converts.
+     * @param transforms A map whose keys are coordinate systems, and whose
+     *                   values point to Transforms that convert from
+     *                   `reference` to the corresponding key. All Transforms
+     *                   must be invertible.
```
TallJimbo: The docs for the parameters are inconsistent: docs for `reference` say that all elements in `transforms` convert *to* `reference`, while docs for `transforms` say all elements in `transforms` convert *from* `reference`.

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -0,0 +1,213 @@
...
+    TransformMap(CameraSys const &reference,
+                 std::map<CameraSys, std::shared_ptr<Transform>> const &transforms);
+
+    /**
+     * Define a set of camera transforms.
+     *
+     * @overload
+     */
+    TransformMap(CameraSys const &reference,
+                 std::unordered_map<CameraSys, std::shared_ptr<Transform>> const &transforms);
```
TallJimbo: Do we really expect to use both `std::map` and `std::unordered_map`?  I suggest we just pick one.

src/cameraGeom/TransformMap.cc
```diff
@@ -0,0 +1,171 @@
...
+ * @param root Common coordinate system serving as root of FrameSet's tree.
+ * @param transforms A map whose keys are coordinate systems, and whose
+ *                   values point to Transforms that convert from `root` to the
+ *                   corresponding key. All Transforms must be invertible.
+ * @return a newly allocated FrameSet corresponding to the method arguments
+ *
+ * @throws lsst::pex::exceptions::InvalidParameterError Thrown if
+ *         `transforms` contains `root`, or if any Transform is not
+ *         invertible.
+ */
```
TallJimbo: Make this a regular comment; it's private.  Same for other private functions in this file.

src/cameraGeom/TransformMap.cc
```diff
@@ -0,0 +1,171 @@
...
+template <class Map>
+std::unique_ptr<ast::FrameSet> makeTransforms(CameraSys const &root, Map const &transforms) {
+    ast::Frame rootFrame(2, "Ident=" + root.getSysName());
+    auto result = std::unique_ptr<ast::FrameSet>(new ast::FrameSet(rootFrame));
+
+    for (auto keyValue : transforms) {
```
TallJimbo: `auto const &` here would save an unnecessary (but probably unimportant) copy.

src/cameraGeom/TransformMap.cc
```diff
@@ -0,0 +1,171 @@
...
+ *         corresponding frame number in `makeTransforms(reference, transforms)`.
+ *
+ * @warning Does not perform input validation.
+ */
+template <class Map>
+std::unordered_map<CameraSys, int> makeTranslator(CameraSys const &reference, Map const &transforms) {
+    std::unordered_map<CameraSys, int> result({std::make_pair(reference, 1)});
+    int nFrames = 1;
+
+    for (auto keyValue : transforms) {
```
TallJimbo: Recommend `auto const &`.

src/cameraGeom/TransformMap.cc
```diff
@@ -0,0 +1,171 @@
...
+
+        auto toFrame = value->getFrameSet()->getFrame(ast::FrameSet::CURRENT);
+        toFrame->setIdent(key.getSysName());
+        result->addFrame(ast::FrameSet::BASE, *(value->getFrameSet()->getMapping()), *toFrame);
+    }
+    return result;
+}
+
+/**
+ * Identify the frames in to a FrameSet constructed by makeTransforms.
```
TallJimbo: Missing or superfluous word somewhere.

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -0,0 +1,213 @@
...
+     * @return an invertible Mapping that converts from `fromCoordSys` to `toCoordSys`
+     *
+     * @throws lsst::pex::exceptions::InvalidParameterError Thrown if either
+     *         `fromCoordSys` or `toCoordSys` is not supported.
+     */
+    std::shared_ptr<ast::Mapping const> _getMapping(CameraSys const &fromCoordSys,
+                                                    CameraSys const &toCoordSys) const;
+
+    /// Allows conversions between LSST and AST data formats
+    static geom::Point2Endpoint _pointConverter;
```
TallJimbo: Could this be in an anonymous namespace in the .cc file instead?  I suppose it's a bit of a judgement call which is better, but I like minimizing what goes into header files.

include/lsst/afw/image/DistortedTanWcs.h
```diff
@@ -48,25 +48,27 @@ namespace image {
  */
 class DistortedTanWcs : public TanWcs {
 public:
+    using Transform = geom::Transform<geom::Point2Endpoint, geom::Point2Endpoint>;
+
```
TallJimbo: Be wary of typedefs that shadow templates.

python/lsst/afw/cameraGeom/cameraConfig.py
```diff
@@ -61,7 +61,7 @@ class CameraConfig(pexConfig.Config):
     # scaled by the plate scale in appropriate units
     radialCoeffs = pexConfig.ListField(
         "Coefficients for radial distortion", float)
-    # The following is commented until radialXYTransform supports an offset (ticket/3155)
+    # The following is commented until radialTransform supports an offset (ticket/3155)
```
TallJimbo: This looks frighteningly like a Trac ticket number.  Given that we haven't needed this so far, perhaps we should just delete the commented-out code?  The alternative would be creating a JIRA ticket, but I bet it'd either linger or get marked Won't Fix.

python/lsst/afw/cameraGeom/detector.cc
```diff
@@ -61,15 +60,15 @@ PYBIND11_PLUGIN(_detector) {
...
                      table::AmpInfoCatalog const &, Orientation const &, geom::Extent2D const &,
-                     CameraTransformMap::Transforms const &>(),
+                     TransformMap::Transforms const &>(),
             "name"_a, "id"_a, "type"_a, "serial"_a, "bbox"_a, "ampInfoCatalog"_a, "orientation"_a,
             "pixelSize"_a, "transforms"_a);
 
     /* Operators */
     cls.def("__getitem__",
-            (std::shared_ptr<table::AmpInfoRecord const>(Detector::*)(int) const) & Detector::_get, "i"_a);
+            (std::shared_ptr<table::AmpInfoRecord const> (Detector::*)(int) const) & Detector::_get, "i"_a);
```
TallJimbo: Lots of (probably clang-format) spurious whitespace changes on this commit in this file.  May be past the point where it's worth separating them, though.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -246,6 +246,22 @@ class SkyWcs : public Transform<Point2Endpoint, IcrsCoordEndpoint> {
...
+ *          pixels to `dst` pixels, and whose inverse transformation converts
+ *          in the opposite direction.
+ *
+ * @exceptsafe Provides basic exception safety.
+ *
+ * @note Parameters are provided in the order `dst`, `src` for consistency with
+ *       XYTransformFromWcsPair.
+ */
+Transform<Point2Endpoint, Point2Endpoint> makeWcsPairTransform(SkyWcs const &dst, SkyWcs const &src);
+
```
r-owen: Thank you!

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -0,0 +1,213 @@
...
+namespace lsst {
+namespace afw {
+namespace cameraGeom {
+
+/**
+ * Functor for boost::transform_iterator: given an entry in a std::map or unordered_map, return the key
+ */
+struct GetKey {
+    CameraSys const &operator()(std::pair<const CameraSys, int> const &p) const { return p.first; };
+};
```
r-owen: Sweet. I'll try that.

include/lsst/afw/cameraGeom/TransformMap.h
r-owen: I agree. I intended to add a few typedefs for a the most common Transform template instantiations in any case and this ticket seems like a good time to do it. I will, of course, use the same names as the Python classes (even though they are a bit long and clumsy).

include/lsst/afw/cameraGeom/TransformMap.h
r-owen: I agree but @kfindeisen wrote this, so I'd like to give him a chance to weigh in.

Some thoughts:
- We never rely on ordering.
- It is easier to write `operator<` than `hash`, though as long as the hash does not have to be perfect, offering one seems reasonable.
- We presently use the hash for Python dict as well; that would be easy to change, but it is convenient.

include/lsst/afw/geom/SeparableXYTransform.h
```diff
@@ -41,6 +41,11 @@ class Functor;
 /** @brief A 2D transform for which the pixel distortions in the in
  *  the x- and y-directions are separable.
  *
+ * @deprecated Deprecated in 14.0. Will be removed in 15.0. It is partially
+ *             replaced by ast::ParallelMap or ast::Mapping::under, but the
+ *             replacement of Functors with Mappings must be done on a
+ *             case-by-case basis.
+ *
```
kfindeisen: Isn't "one release" the shortest deprecation period possible? :confused: 

(Hopes DM-8969 is implemented soon...)

include/lsst/afw/cameraGeom/TransformMap.h
kfindeisen: Oops! For the record, the description for `transforms` is the correct one.

include/lsst/afw/cameraGeom/TransformMap.h
kfindeisen: My motivation was to prevent users from having to commit to a particular map implementation if they want to use `TransformMap` (since STL is terrible about container interoperability). IIRC @r-owen's counterargument was that we never create the map in C++ anyway.

If we have to pick one I'd prefer `unordered_map`, on the grounds that it would have fewer surprises converting to and from `dict`.

src/cameraGeom/TransformMap.cc
kfindeisen: I think "to" should be removed? I don't actually remember what I meant to say. :fearful:

include/lsst/afw/geom/SeparableXYTransform.h
TallJimbo: I suppose I just don't consider the cycle releases to be terribly meaningful from a code/interface content standpoint right now; the only release unit that's actually meaningful to developers are the weeklies.  (I'd say that weeklies are also the only unit that most external users should pay attention to right now, though that's not universal).  While we ought to have a release cadence in-between weekly and semiannually that actually gets backports so it can stay genuinely usable for more than a couple of weeks, we simply don't have that at present.

include/lsst/afw/geom/transformFactory.h
```diff
@@ -71,6 +72,72 @@ DISABLE(IcrsCoordEndpoint, Point2Endpoint);
...
+ *               represent the identity transformation; otherwise must have
+ *               `size` > 1, `coeffs[0]` = 0, and `coeffs[1]` &ne; 0.
+ * @returns the radial distortion represented by `coeffs`. The Transform shall
+ *          have an inverse, which may be approximate.
+ *
+ * @throws pex::exceptions::InvalidParameterError Thrown if `coeffs` does not
+ *         have the required format.
+ * @exceptsafe Provides basic exception safety.
+ */
+Transform<Point2Endpoint, Point2Endpoint> makeRadialTransform(std::vector<double> const &coeffs);
```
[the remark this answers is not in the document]
r-owen: Renaming sounds like a good idea. I'm less convinced about the benefits of using ndarray; it's strictly 1-d and the arrays are always short. Most existing code uses a list of float in Python to initialize these.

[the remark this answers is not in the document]
TallJimbo: If the arrays are always short, I agree that ndarray is unnecessary.

include/lsst/afw/cameraGeom/TransformMap.h
TallJimbo: Could we shorten the Python class names, then?  The simple 2-d to 2-d case is so much more important than all the others that I wouldn't object to calling it `Transform` and using a different name for the template, though e.g. `Transform22` would be acceptable.

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -39,30 +39,30 @@ namespace afw {
...
  * @exceptsafe Unless otherwise specified, all methods guarantee only basic
  *             exception safety.
  */
 class TransformMap final {
+private:
+     // Functor for boost::transform_iterator: given an entry in a std::map or unordered_map, return the key
+    struct GetKey {
+        CameraSys const &operator()(std::pair<const CameraSys, int> const &p) const { return p.first; };
+    };
+
```
kfindeisen: According to the [style guide](https://developer.lsst.io/coding/cpp_style_guide.html#the-parts-of-a-class-must-be-sorted-public-protected-and-private), the definition of `GetKey` is out of order. I don't see a way to declare but not define `CameraSysIterator`, though.

TallJimbo: IMO, that section of the style guide is naive, and should not apply to the rare situations (like this one) where a certain ordering (and possible repetition) of private/protected/public sections is required for code correctness.

kfindeisen: Just pointing out that you may catch some flak over this.

TallJimbo: Yeah, maybe I'll just RFC a tiny change to the style guide to add an escape clause.  I wouldn't let that RFC block this ticket.

TallJimbo: As per discussion on the above RFC, @ktlim, could you sign off on the above violation of C++ 5.2?

ktlim: I approve as long as `GetKey` has no general usefulness as a public interface.

r-owen: Thank you. And not `GetKey` is not used outside the class.

src/cameraGeom/TransformMap.cc
r-owen: I rewrote the documentation (and changed the leading `/**` to `/*` so Doxygen would ignore it)

include/lsst/afw/image/DistortedTanWcs.h
r-owen: Removed in favor of the new typedef `geom::TransformPoint2ToPoint2`

r-owen:
I updated cameraGeom.dox (it needed many changes) and it now builds without warnings.

python/lsst/afw/cameraGeom/detector.cc
r-owen: I'm afraid so.

include/lsst/afw/image/DistortedTanWcs.h
TallJimbo: Sounds good.  I'd prefer something even shorter, but if that's the name we've been using in Python I can at least accept that changing it is out of scope for this ticket.

r-owen: Yes. Trying to come up with even shorter names is more than I want to try to do on this ticket. I'm in favor of shorter names if they are sufficiently descriptive, but so far have not come up with anything.

PR #272, opened 2017-08-12
Make it possible to specify an origin (e.g. PARENT) with e.g. im[bbox]

RobertLuptonTheGood:
That is, when using the syntactic sugar which makes using bboxes easier,
allow an extra argument such as
    im[bbox, afwImage.PARENT]
or
	 im[1000:1010, -1:, afwImage.PARENT]

In the second case we'll use PARENT coordinates when interpreting the
slices (i.e. the "1000" includes x0)

### DIFF
--- a/python/lsst/afw/image/slicing.py
+++ b/python/lsst/afw/image/slicing.py
@@ -1,13 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
 import lsst.afw.geom as afwGeom
-from .image import LOCAL
+from .image import LOCAL, PARENT, ImageOrigin
 
 __all__ = ["supportSlicing"]
 
 
 def _getBBoxFromSliceTuple(img, imageSlice):
-    """Given a slice specification return the proper Box2I
+    """Given a slice specification return the proper Box2I and origin
     This is the worker routine behind __getitem__ and __setitem__
 
     The imageSlice may be:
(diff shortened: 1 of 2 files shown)


python/lsst/afw/image/slicing.py
```diff
@@ -31,9 +31,30 @@ def _getBBoxFromSliceTuple(img, imageSlice):
...
+     im[1001:1004, 2006:2010, afwImage.PARENT]
+    with xy0 = (1000, 2000) refers to the same pixels as
+     im[1:4, 6:10, afwImage.LOCAL]
+    )
     """
+    origin = LOCAL                      # this sets the default value
+
+    try:
+        _origin = imageSlice[-1]
+    except TypeError:
```
SimonKrughoff: Where would you get a type error here.  It would happen if `imageSlice` isn't listy, but then you've got bigger problems.  Are you thinking about the case where `imageSlice` is a `slice` object?

python/lsst/afw/image/slicing.py
```diff
@@ -31,9 +31,30 @@ def _getBBoxFromSliceTuple(img, imageSlice):
...
+        _origin = None
+
+    if isinstance(_origin, ImageOrigin):
+        origin = _origin
+        imageSlice = imageSlice[0] if len(imageSlice) <= 2 else imageSlice[:-1]
```
SimonKrughoff: I must be missing something here too.  It seems like imageSlice would only ever be 2 or 3 elements long.  If it's 3, it has an origin, if not you want both the elements.  Why only pass on the first element?

python/lsst/afw/image/slicing.py
RobertLuptonTheGood: You can get `bbox` or `bbox, origin` as well as list-like things

I don't guarantee that I have the minimal logic to cover all the cases, so I should take another look.

python/lsst/afw/image/slicing.py
RobertLuptonTheGood: The problem is that you can get passed lots of different things:
 - bbox
 - bbox, origin
 - : 
 - :, origin
 - n:m, :
 - n:m, :, origin
 - n:m, i:j
 - n:m, i:j, origin

You don't want to convert `bbox` to `[bbox]`

python/lsst/afw/image/slicing.py
SimonKrughoff: I hadn't appreciated that `imageSlice` could be a bbox.

python/lsst/afw/image/slicing.py
SimonKrughoff: I see now.  It may be worth adding a few more of these as examples in the docstring.

PR #273, opened 2017-08-18
DM-11514: Fix a race condition in tests under pytest-xdist

timj:
Also fix some deprecation warnings and use of AFW_DIR.

### DIFF
--- a/tests/test_apCorrMap.py
+++ b/tests/test_apCorrMap.py
@@ -134,8 +134,8 @@ class ApCorrMapTestCase(lsst.utils.tests.TestCase):
 
     def testExposureCatalogBackwardsCompatibility(self):
         """Test that we can read an ExposureCatalog written with an old version of the code."""
-        filename = os.path.join(
-            os.environ["AFW_DIR"], "tests", "data", "version-0-ExposureCatalog.fits")
+        testPath = os.path.abspath(os.path.dirname(__file__))
+        filename = os.path.join(testPath, "data", "version-0-ExposureCatalog.fits")
         cat = lsst.afw.table.ExposureCatalog.readFits(filename)
         record = cat[0]
         self.assertIsNone(record.getApCorrMap())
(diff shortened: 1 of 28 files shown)


timj:
@PaulPrice can you take a look at this for me please? Should be fairly non-controversial.

tests/test_apCorrMap.py
```diff
@@ -109,7 +109,7 @@ def testPersistence(self):
 
     def testExposurePersistence(self):
         """Test that the ApCorrMap is saved with an Exposure"""
-        filename = "testApCorrMap.fits"
+        filename = "testApCorrMap-exposure.fits"
```
PaulPrice: I suggest using:

    with lsst.utils.tests.getTempFilePath(".fits") as filename:

That gets you automatic cleanup, and is more immune to cargo culting because it chooses a name based on the test.

Ditto in `testExposureRecordPersistence`.

tests/test_sourceMatch.py
```diff
@@ -181,6 +181,7 @@ def testPhotometricCalib(self):
             s.setDec(dec * afwGeom.degrees)
             s.set(self.table.getPsfFluxKey(), psfMags[band])
 
+        ifd.close()
```
PaulPrice: Is it worth using the context manager? This is exactly what it's for.

    with open(os.path.join(afwdataDir, "CFHT", "D2", "sdss.dat"), "r") as ifd:
        allLines = ifd.readlines()
    ....
    for line in allLines:

Ditto below.

timj: I'm happy to do that. The little Robert voice in my head was wondering if the big indenting diff was worth it.

tests/test_apCorrMap.py
timj: good idea.

tests/test_sourceMatch.py
PaulPrice: He's in my head too, which is why it was just a suggestion.
I slightly lean towards using the context manager, but it's up to you.

timj: I made that change.

tests/test_apCorrMap.py
timj: Implemented temp files.

PR #277, opened 2017-09-23
DM-11957: Cannot round-trip >7th degree Chebyshev photometry models

r-owen:


include/lsst/afw/table/types.h
```diff
@@ -25,20 +25,20 @@
...
-#define AFW_TABLE_ARRAY_FIELD_TYPES std::uint16_t, int, float, double
+#define AFW_TABLE_ARRAY_FIELD_TYPE_N 5
+#define AFW_TABLE_ARRAY_FIELD_TYPES std::uint16_t, int, float, double, std::uint8_t
 #define AFW_TABLE_ARRAY_FIELD_TYPE_TUPLE BOOST_PP_LPAREN() AFW_TABLE_ARRAY_FIELD_TYPES BOOST_PP_RPAREN()
 
 // Field types: all the types we allow for fields.
-#define AFW_TABLE_FIELD_TYPE_N 12
+#define AFW_TABLE_FIELD_TYPE_N 14
 #define AFW_TABLE_FIELD_TYPES                                                                        \
     AFW_TABLE_SCALAR_FIELD_TYPES, Flag, std::string, Array<std::uint16_t>, Array<int>, Array<float>, \
```
pschella: Unrelated, but is there any reason why we use `Array<int>` while the scalar is specifically qualified as `std::int32_t`?

python/lsst/afw/table/base/base.cc
```diff
@@ -103,11 +103,13 @@ PyBaseRecord declareBaseRecord(py::module &mod) {
     declareBaseRecordOverloads<double>(cls, "D");
     declareBaseRecordOverloads<float>(cls, "F");
     declareBaseRecordOverloads<lsst::afw::table::Flag>(cls, "Flag");
+    declareBaseRecordOverloads<std::uint8_t>(cls, "B");
```
pschella: I really think we should add some kind of type trait at some point to standardize the suffixes. Such that we can say `lsst::utils::default_suffix<std::uint8_t>::value` and always get the same one.
Maybe not now, since it leads to the slightly tricky problem of compile-time string concatenation, but perhaps file a new ticket for it? If you agree that is.

src/fits.cc
```diff
@@ -87,6 +87,10 @@ struct FitsType<char> {
     static int const CONSTANT = TSTRING;
 };
 template <>
+struct FitsType<signed char> {
+    static int const CONSTANT = TSBYTE;
+};
+template <>
 struct FitsType<unsigned char> {
     static int const CONSTANT = TBYTE;
 };
```
pschella: Any reason we don't specifically also specialize for `std::uint8_t`? I know it maps on most platforms, but it doesn't have to.

src/table/FieldBase.cc
```diff
@@ -20,6 +20,10 @@ template <typename T>
 struct TypeTraits;
 
 template <>
+struct TypeTraits<std::uint8_t> {
+    static char const *getName() { return "B"; }
```
pschella: Ok, so we have one. Then why don't we use it everywhere? Preferably pulling it up to `utils`?
In that case I would still modify it slightly such that it can be used in `constexpr` contexts (e.g. one can use it in a `switch` or something).

include/lsst/afw/formatters/Utils.h
```diff
@@ -167,6 +171,16 @@ std::string formatFitsProperties(lsst::daf::base::PropertyList const& prop,
...
 
+/**
+ * Encode a std::string as a vector of uint8
+ */
+ndarray::Array<std::uint8_t, 1, 1> stringToBytes(std::string const &str);
+
+/**
+ * Decode a std::string from a vector of uint8 returned by stringToBytes
+ */
+std::string bytesToString(ndarray::Array<std::uint8_t const, 1, 1> bytes);
```
pschella: Why pass by value?

src/formatters/Utils.cc
```diff
@@ -286,6 +286,24 @@ std::string formatFitsProperties(daf::base::PropertyList const& prop,
 int countFitsHeaderCards(lsst::daf::base::PropertySet const& prop) {
     return prop.paramNames(false).size();
 }
+
+ndarray::Array<std::uint8_t, 1, 1> stringToBytes(std::string const& str) {
+    auto nbytes = str.size() * sizeof(char) / sizeof(std::uint8_t);
```
pschella: `sizeof(typename std::string::value_type)`? Or do we trust the user to update this if they switch the input to some other string type?

src/formatters/Utils.cc
```diff
@@ -286,6 +286,24 @@ std::string formatFitsProperties(daf::base::PropertyList const& prop,
...
+    std::uint8_t const* byteCArr = reinterpret_cast<std::uint8_t const*>(str.data());
```
pschella: So we don't do any formal encoding? Hmm, I'm very worried that this will create problems when the data is read on a different platform then the one it was written.

include/lsst/afw/table/types.h
r-owen: I have no idea. @TallJimbo?

src/fits.cc
r-owen: I'm not sure. I am copying existing code which may be based on this entry in the CFITSIO documentation: https://heasarc.gsfc.nasa.gov/docs/software/fitsio/quick/node9.html. It is true that the entry refers to images (it was the only such mention I found).

(begin quote)
The datatype parameter specifies the datatype of the C array in the program, which need not be the same as the datatype of the FITS image itself. If the datatypes differ then CFITSIO will convert the data as it is read or written. The following symbolic constants are allowed for the value of datatype:
  TBYTE     unsigned char
  TSBYTE    signed char
  TSHORT    signed short
  TUSHORT   unsigned short
  TINT      signed int
  TUINT     unsigned int
  TLONG     signed long
  TLONGLONG signed 8-byte integer
  TULONG    unsigned long
  TFLOAT    float
  TDOUBLE   double
(end quote)

include/lsst/afw/formatters/Utils.h
r-owen: I'll change it. It shouldn't matter because ndarray uses a pointer to the data internally, but it will be more obviously safe if I change it.

src/formatters/Utils.cc
r-owen: In my code if somebody changes the argument type then the code will continue to operate correctly (though I hate to think what will happen when reading old persisted data). In your suggested change the type appears in two different places which increases the risk that they will go out of synch if somebody changes the type of the `str` argument. So I think my code is safer than yours. What am I missing?

src/formatters/Utils.cc
r-owen: Discussed on the JIRA ticket with the resolution that we will file another ticket to look at this and use this code for now.

src/table/FieldBase.cc
r-owen: These characters are FITS table characters. I'm not sure I'd want to use it anywhere else.

I don't understand your suggested modification but respectfully suggest that any significant refactoring should go into a new ticket.

python/lsst/afw/table/base/base.cc
r-owen: I agree this code would benefit from a cleanup pass. But I'm not comfortable enough with type traits to have an opinion on this exact issue.

src/formatters/Utils.cc
r-owen: I don't see why your suggested change is any better -- my code and yours both seem to need updating if a user changes the type of the `str` argument. For the length computation to be safe if the type of `str` changes perhaps we want `sizeof(str[0])`?

Of course if the type of the `str` argument changes then old persisted data will probably not unpersist correctly.

r-owen: `std::string::data` is documented to return `const char *` so I think the current code is safest.

include/lsst/afw/table/types.h
TallJimbo: `std::int32_t` would be better, but changing it would require concerted changes across a lot of the codebase.  Should probably be spun off into another ticket.

python/lsst/afw/table/base/base.cc
TallJimbo: This is another issue that is worth doing as a separate ticket.

src/fits.cc
TallJimbo: Yes, as @r-owen said, this is limited by what cfitsio supports.

PR #278, opened 2017-09-26
DM-12025: make Transform pickleable

PaulPrice:
The key functionality was already present (readString, writeString
methods) and just needed to be hooked up to the pickle machinery
and tested.


python/lsst/afw/math/interpolate.cc
```diff
@@ -82,6 +82,8 @@ PYBIND11_PLUGIN(_interpolate) {
             "x"_a, "y"_a, "style"_a = Interpolate::AKIMA_SPLINE);
 
     mod.def("stringToInterpStyle", stringToInterpStyle, "style"_a);
+    mod.def("lookupMaxInterpStyle", lookupMaxInterpStyle, "n"_a);
+    mod.def("lookupMinInterpPoints", lookupMinInterpPoints, "style"_a);
```
r-owen: It would be much appreciated if you could add a few simple unit tests for these (including `stringToInterpStyle`).

PR #279, opened 2017-09-29
DM-12061: Eliminate test warnings in test_methods.py

r-owen:
Use `unittest.assertWarns` if available to check that deprecated
asserts raise `DeprecationWarning`. If the version of Python
is too old for unittest to have `assertWarngs` then use a
null context manager instead (and pytest will issue warnings).

### DIFF
--- a/tests/test_maskedImageIO.py
+++ b/tests/test_maskedImageIO.py
@@ -39,7 +39,7 @@ import tempfile
 
 from builtins import object
 import numpy as np
-import pyfits
+import astropy.io
 
 import lsst.utils
 import lsst.utils.tests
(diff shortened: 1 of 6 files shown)


r-owen:
Good question. The unittest documentation says of `assertWarns`: "This method works regardless of the warning filters in place when it is called". I double checked that by running the test as follows:
```
python -W ignore::DeprecationWarning tests/test_methods.py
```

tests/test_maskedImageIO.py
```diff
@@ -39,7 +39,7 @@
 
 from builtins import object
 import numpy as np
-import pyfits
+import astropy.io
```
timj: Many people do this as `import astropy.io.fits as pyfits` to minimize code impact but that is probably more confusing in the long term since it leaves the `pyfits` string in the code.

r-owen: Agreed. I considered `from astropy.io import fits` but felt it was too ambiguous, and in at least one case `fits` was already used as a variable.

timj: Looking at other AFW code this should be plain `import astropy.io.fits` (we don't want to import all the other astropy IO plugins).

timj:
In case my comment got lost when I attached it to a line; I think we need `import astropy.io.fits` and not `import astropy.io` since I don't think we want to import *all* the I/O plugins.

r-owen:
Sorry, I missed that. I'll fix in on DM-10765

r-owen:
I also found and fixed another instance in test_wcsFitsTable.py on that same ticket.

PR #280, opened 2017-09-30
DM-10765: Replace existing WCS classes with SkyWcs

r-owen:


r-owen:
@TallJimbo I am happy to undo hoisting ellipses, but wanted to at least ask about hoisting a few symbol, e.g. Quadrupole and Ellipse?

TallJimbo:
Yup, those two at least sound fine (in fact, I think `Ellipse` was hoisted into `afw.geom` in the past, or maybe just in C++).

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -170,18 +170,14 @@ class SkyWcs final : public table::io::PersistableFacade<SkyWcs>, public table::
...
      * Return the WCS as FITS WCS metadata
      *
-     * @param[in] precise  If True then a "fairly precise" representation is required
-     *   (one with no significant deviation from linear over 100x100 pixels).
-     *   If false then write a local TAN approximation if a precise representation is not possible.
-     *
-     * @warning AST cannot yet write out SIP terms as FITS WCS metadata (though we are trying to get
-     * that capability added). Thus if precise=true, this method this will fail for a TAN-SIP WCS
-     * that has significant distortion.
+     * @param[in] precise  Fail if the WCS WCS cannot be represented to sufficient precision as a FITS WCS?
```
rearmstr: duplicate "WCS"

src/geom/SkyWcs.cc
```diff
@@ -220,7 +219,9 @@ std::shared_ptr<daf::base::PropertyList> SkyWcs::getFitsMetadata(bool precise) c
             throw LSST_EXCEPT(lsst::pex::exceptions::RuntimeError,
                               "Could not represent this SkyWcs using FITS-WCS metadata");
         } else {
-            // A large FitsTol was not sufficient; write a local TAN WCS approximation
+            // A large FitsTol was not sufficient; get a local TAN WCS approximation
```
rearmstr: I don't understand this comment here.  What is FitsTol?

src/geom/SkyWcs.cc
```diff
@@ -220,7 +219,9 @@ std::shared_ptr<daf::base::PropertyList> SkyWcs::getFitsMetadata(bool precise) c
...
+            // to the PIXELS to TANwrite a local TAN WCS approximation
```
rearmstr: typo "TANwrite"

tests/test_skyWcs.py
```diff
@@ -304,6 +304,19 @@ def safeOffset(crval, direction, amount):
...
+        """Check SkyWcs.getFitsMetadata for a WCS that cannot be represented as a FITS-WCS
+        """
+        # the modified WCS should not be representable as pure FITS-WCS
+        self.assertFalse(wcs.isFits)
+        with self.assertRaises(RuntimeError):
+            wcs.getFitsMetadata(True)
+
+        # the approximation returned by getFitsMetadata is poor (pure TAN) until DM-13170
+        wcsFromMetadata = makeSkyWcs(wcs.getFitsMetadata(False))
+        bbox = Box2I(Point2I(0, 0), Extent2I(2000, 2000))  # arbitrary but reasonable
```
rearmstr: elsewhere you have changed the (0,0) point to something else.  Do you want to use that here as well?

tests/test_exposure.py
```diff
@@ -70,7 +70,7 @@ class ExposureTestCase(lsst.utils.tests.TestCase):
 
     def setUp(self):
         maskedImage = afwImage.MaskedImageF(inFilePathSmall)
-        maskedImageMD = afwImage.readMetadata(inFilePathSmall)
+        maskedImageMD = lsst.afw.fits.readMetadata(inFilePathSmall)
```
rearmstr: In other files you've imported readMetadata itself.  I don't have a strong opinion on which one to use, but I think it should be consistent.

src/formatters/Utils.cc
```diff
@@ -55,45 +55,56 @@ namespace {
 /**
 Format a PropertySet into a FITS header string (exactly 80 characters per "card", no line terminator)
 
-This function is designed to format data for creating a WCS. It truncates long string
-values and skips properties whose type it cannot handle.
+This function is designed to format data for creating a WCS. As such, it is quite limited:
```
rearmstr: Is there a need to repeat this comment in the "cc" file since it is already in the header?

tests/data/makeOldWcsFitsFiles.py
```diff
@@ -0,0 +1,154 @@
...
+    tanWcs.writeFits("oldTanWcs.fits")
+
+    tanSipWcsMetata = getTanSipWcsMetata()
+    tanSipWcs = makeWcs(tanSipWcsMetata)
+    tanSipWcs.writeFits("oldTanSipWcs.fits")
+
+
+def getTanWcsMetata():
+    metadata = PropertyList()
+    # the following was fit using CreateWcsWithSip from meas_astrom
```
rearmstr: Was this really created with "CreateWcsWithSip" since there are no Sip terms?

include/lsst/afw/math/warpExposure.h
```diff
@@ -461,7 +462,7 @@ int warpImage(DestImageT &destImage,                 ///< remapped %image
 template <typename DestImageT, typename SrcImageT>
 int warpImage(DestImageT &destImage,
               SrcImageT const &srcImage,
-              geom::TransformPoint2ToPoint2 const & destToSrc,
```
rearmstr: I'm not following why the directions are swapped for all these transforms.  Were they incorrect before?

include/lsst/afw/geom/SpanSet.h
```diff
@@ -216,6 +217,12 @@ class SpanSet : public afw::table::io::PersistableFacade<lsst::afw::geom::SpanSe
      */
     std::shared_ptr<SpanSet> transformedBy(XYTransform const &t) const;
```
rearmstr: Is the XYTransform version still being kept?

src/formatters/ExposureFormatter.cc
```diff
@@ -452,8 +449,8 @@ void ExposureFormatter<ImagePixelT, MaskPixelT, VariancePixelT>::delegateSeriali
     if (ip == 0) {
         throw LSST_EXCEPT(pex::exceptions::RuntimeError, "Serializing non-Exposure");
     }
-    std::shared_ptr<image::Wcs> wcs = ip->getWcs();
-    ar& * ip->getMetadata() & ip->_maskedImage& wcs;
+    std::shared_ptr<geom::SkyWcs const> wcs = ip->getWcs();
+    ar& * ip->getMetadata() & ip->_maskedImage;  // & wcs; // TODO: replace this with what?
```
rearmstr: Why is the wcs removed here?  Do we no longer serialize it as before?

src/geom/SkyWcs.cc
r-owen: Good catch. The comment was out of date. For the record: FitsTol is something AST uses to decide if it can write a WCS out as FITS header cards; code a few lines above sets it very small and that's the only time it's ever set. At one time the code below this comment tried setting it larger to see if AST could write a useful approximation, but I decided the current code was simpler and produced better results.

tests/test_exposure.py
r-owen: The only time I do not use "from lsst.afw.fits import readMetadata" is if the file uses several other symbols from the same library (as for any library). In this case there's only one other symbol so I'll import both. But 2 or 3 is my preferred limit.

src/formatters/Utils.cc
r-owen: The reason I did it is because it applies to formatFitsProperties in the .h file and formatFitsPropertiesImpl in the .cc file. I'll replace the detailed description with `See @ref formatFitsProperties for details.`

tests/data/makeOldWcsFitsFiles.py
r-owen: Good catch. The comment belonged to `getTanSipWcsMetata`. I've cleaned that up.

include/lsst/afw/geom/SpanSet.h
r-owen: Yes. It is likely that we can remove the XYTransform versions of most or all code once this ticket is merged, but that should be a different ticket and likely an RFC as well.

src/formatters/ExposureFormatter.cc
r-owen: SkyWcs is serialized in an entirely different way. Note that ExposureFormatter.cc is probably dead code.

include/lsst/afw/math/warpExposure.h
r-owen: The version of warpImage that takes a Transform was modified to use the same direction as the one that takes an XYTransform because the Transform version was not yet in wide use and the XYTransform version was heavily used. This is mentioned in my community post about the changes https://community.lsst.org/t/replacing-wcs-and-tanwcs-with-skywcs/2531/2

src/formatters/ExposureFormatter.cc
PaulPrice: I could be wrong, but I think the butler still uses the formatter stuff.

PR #281, opened 2017-10-04
DM-11473: Add SpherePoint(long, lat, unit) constructor

r-owen:


tests/test_spherePoint.py
```diff
@@ -128,6 +128,14 @@ def testCopyConstructor(self):
         spcopy = SpherePoint(sp)
         self.assertEqual(sp, spcopy)
 
+    def testDoubleConstructor(self):
```
kfindeisen: `DoubleConstructor` is potentially ambiguous. Maybe `SharedUnitConstructor`?

src/geom/SpherePoint.cc
```diff
@@ -63,7 +63,8 @@ Angle haversine(Angle const& deltaLon, Angle const& deltaLat, double cosLat1, do
 }
 }  // end namespace
 
-SpherePoint::SpherePoint(double const lonLatRad[2]) : _longitude(lonLatRad[0]), _latitude(lonLatRad[1]) {}
+SpherePoint::SpherePoint(double longitude, double latitude, AngleUnit units)
+        : _longitude(longitude * units), _latitude(latitude * units) {}
```
kfindeisen: I can't believe I overlooked this: please add input validation for `latitude` (as implied by the documentation) and test it. The easiest way to implement input validation is to delegate to the `(Angle, Angle)` constructor.

r-owen: Sorry I missed this the first time. I'll add shortly it in a new ticket.

PR #283, opened 2017-10-04
DM-12102: add input validation for SpherePoint(double, double, AngleUnits) constructor

r-owen:


tests/test_spherePoint.py
```diff
@@ -135,6 +135,10 @@ def testDoubleDoubleUnitsConstructor(self):
         self.assertEqual(sp2, SpherePoint(1.1 * radians, -0.6 * radians))
         with self.assertRaises(TypeError):
             SpherePoint(-42.0, 45.0)  # units must be specified
+        with self.assertRaises(pexEx.InvalidParameterError):
+            SpherePoint(-42.0, 91.0, degrees)  # latitude out of range
+        with self.assertRaises(pexEx.InvalidParameterError):
+            SpherePoint(-42.0, -91.0, degrees)  # latitude out of range
```
kfindeisen: Could you also add a test that e.g. negative longitudes get wrapped correctly? (Possibly as part of `testGetLongitudeValue` rather than here, since that's where the tests are done for the original constructor).

kfindeisen: I don't know if you want the other tests from `testInit2ArgErrors`, like infinite latitude.

r-owen:
I greatly expanded the testing of the new `SpherePoint(double, double, AngleUnit)` constructor and integrated it into testing of the `SpherePoint(Angle, Angle)` constructor. I also took the liberty of switching to the new constructor in some places where it made the test easier to read. Please have another look.

tests/test_spherePoint.py
```diff
@@ -92,26 +92,37 @@ def positions():
...
-        """Test if 2-argument form of __init__ handles invalid input.
+    def testLongLatConstructors(self):
+        """Test that the longitude, latitude constructors handle invalid input.
         """
         # Latitude should be checked for out-of-range.
         for lat in self._poleLatitudes:
             with self.assertRaises(pexEx.InvalidParameterError):
                 SpherePoint(0.0*degrees, self.nextUp(lat))
             with self.assertRaises(pexEx.InvalidParameterError):
+                SpherePoint(0.0, self.nextUp(lat).asDegrees(), degrees)
```
kfindeisen: Might be cleaner to call `np.nextafter` directly, or modify `nextUp` and `nextDown` to work on floating-point numbers.

tests/test_spherePoint.py
```diff
@@ -155,8 +154,7 @@ def testInitNArgFail(self):
     def testGetLongitudeValue(self):
         """Test if getLongitude() returns the expected value.
         """
-        for lon, lat in self._dataset:
-            point = SpherePoint(lon, lat)
+        def checkGetLongitudeValue(point, lon, lat):
```
kfindeisen: I think this is a bit too long to treat as a lambda. Consider making a private member of `SpherePointTestSuite` for readability.

tests/test_spherePoint.py
```diff
@@ -191,8 +193,7 @@ def testTicket1394(self):
     def testGetLatitudeValue(self):
         """Test if getLatitude() returns the expected value.
         """
-        for lon, lat in self._dataset:
-            point = SpherePoint(lon, lat)
+        def checkGetLatitudeValue(point, lon, lat):
```
kfindeisen: Again, having this in `checkGetLatitudeValue` breaks up the flow of the code.

tests/test_spherePoint.py
```diff
@@ -212,21 +217,29 @@ def testGetVectorValue(self):
...
 
             # Convert back to spherical.
             newLon, newLat = SpherePoint(newVector)
-            self.assertAlmostEqual(newLon.asDegrees(), lonLat[0])
-            self.assertAlmostEqual(newLat.asDegrees(), lonLat[1])
+            self.assertAlmostEqual(newLon.asDegrees(), lon.asDegrees())
+            self.assertAlmostEqual(newLat.asDegrees(), lat.asDegrees())
+
+        for lonLatDeg, vector in pointList:
+            # Convert to Point3D.
```
kfindeisen: Does this comment belong inside `checkGetVector`?

tests/test_spherePoint.py
```diff
@@ -212,21 +217,29 @@ def testGetVectorValue(self):
             ((0.0, 90.0), afwGeom.Point3D(0.0, 0.0, 1.0)),
         ]
 
-        for lonLat, vector in pointList:
-            # Convert to Point3D.
-            point = SpherePoint(lonLat[0]*degrees, lonLat[1]*degrees)
+        def checkGetVector(point, lon, lat, vector):
```
kfindeisen: Long lambda.

tests/test_spherePoint.py
```diff
@@ -275,17 +288,26 @@ def testTicket1761(self):
...
-            [SpherePoint(42.0*degrees, self.nextUp(0.0*radians - lat)) for lat in self._poleLatitudes] + [
-            SpherePoint(afwGeom.Point3D(9.9e-7, 0.0, 1.0)),
-            SpherePoint(afwGeom.Point3D(9.9e-7, 0.0, -1.0)),
-            SpherePoint(0.0*degrees, nan*degrees),
-        ]
+        poleList = \
+            [SpherePoint(42.0*degrees, lat) for lat in self._poleLatitudes] + \
+            [SpherePoint(42.0, lat.asDegrees(), degrees) for lat in self._poleLatitudes] + \
+            [SpherePoint(42.0*degrees, 0.0*radians - lat) for lat in self._poleLatitudes] + \
+            [SpherePoint(42.0, (0.0*radians - lat).asDegrees(), degrees) for lat in self._poleLatitudes] + \
```
kfindeisen: `-lat.asDegrees()` would be more concise.

Actually, since I added a negation operator to `Angle`, all the references to `0.0*radians` can probably be removed.

tests/test_spherePoint.py
```diff
@@ -333,8 +360,7 @@ def testGetItemError(self):
     def testGetItemValue(self):
         """Test if indexing returns the expected value.
         """
-        for lon, lat in self._dataset:
-            point = SpherePoint(lon, lat)
+        def checkGetItemValue(point, lon, lat):
```
kfindeisen: Long lambda.

tests/test_spherePoint.py
r-owen: The current code is a bit clumsy but is obviously the same test, which I think is worth the clumsiness. I like the fact that nextUp works on Angles and am reluctant to change that.

kfindeisen: I meant "in addition to Angles", but ok.

tests/test_spherePoint.py
r-owen: Technically that this is not a "lambda", since Python uses the keyword `lambda` for those. But I understand and will make them methods.

r-owen: I reverted all the new in-line functions and am using better iteration instead. This really cleans up the code.

tests/test_spherePoint.py
r-owen: No idea. It's gone now.

tests/test_spherePoint.py
```diff
@@ -188,38 +207,43 @@ def testGetLatitudeValue(self):
...
+                SpherePoint(lon, lat),
+                SpherePoint(lon.asDegrees(), lat.asDegrees(), degrees),
+                SpherePoint(lon.asRadians(), lat.asRadians(), radians),
+            ):
+                newVector = point.getVector()
+                self.assertIsInstance(newVector, afwGeom.Point3D)
+                for oldElement, newElement in zip(vector, newVector):
+                    self.assertAlmostEqual(oldElement, newElement)
+
+                # Convert back to spherical.
```
kfindeisen: If you're removing the "Convert to Point3D" comment, you might remove this one as well

tests/test_spherePoint.py
```diff
@@ -61,6 +61,16 @@ def setUp(self):
...
         ]
 
+    @property
+    def pointSet(self):
+        for lon, lat in self._dataset:
+            for point in (
+                SpherePoint(lon, lat),
+                SpherePoint(lon.asDegrees(), lat.asDegrees(), degrees),
+                SpherePoint(lon.asRadians(), lat.asRadians(), radians),
+            ):
```
kfindeisen: Very thorough!

tests/test_spherePoint.py
```diff
@@ -147,23 +160,29 @@ def testInitNArgFail(self):
...
-
-                # Longitude not guaranteed to match input at pole
-                if not point.atPole():
-                    # assertAnglesAlmostEqual handles angle wrapping internally
-                    self.assertAnglesAlmostEqual(lon, point.getLongitude())
+            for point in (
+                SpherePoint(lon, lat),
+                SpherePoint(lon.asDegrees(), lat.asDegrees(), degrees),
+                SpherePoint(lon.asRadians(), lat.asRadians(), radians),
+            ):
```
kfindeisen: Why not just use `self.pointSet`?

tests/test_spherePoint.py
```diff
@@ -188,38 +207,43 @@ def testGetLatitudeValue(self):
...
-            # data representation
-            if point.isFinite():
-                self.assertGreaterEqual(point.getLatitude().asDegrees(), -90.0)
-                self.assertLessEqual(point.getLatitude().asDegrees(), 90.0)
-                self.assertAnglesAlmostEqual(lat, point.getLatitude())
+            for point in (
+                SpherePoint(lon, lat),
+                SpherePoint(lon.asDegrees(), lat.asDegrees(), degrees),
+                SpherePoint(lon.asRadians(), lat.asRadians(), radians),
+            ):
```
kfindeisen: Again, use `self.pointSet` instead of a double loop.

tests/test_spherePoint.py
r-owen: I do this in two places because in those places the test needs the input longitude or latitude and pointSet does not provide them. I considered a property pointLonLatSet but I felt it was excessive clutter for just two tests.

tests/test_spherePoint.py
r-owen: Sure. I thought I had reverted the other comment to its original place but maybe not. I'll remove both of them as I don't think they add much.

tests/test_spherePoint.py
kfindeisen: Good point. A comment explaining that might be helpful, though.

PR #285, opened 2017-10-16
Tickets/dm 12176

RobertLuptonTheGood:


python/lsst/afw/display/interface.py
```diff
@@ -219,6 +219,7 @@ def __getattr__(self, name):
 
     def close(self):
         if hasattr(self, "_impl") and self._impl:
```
parejkoj: For this line, would `if getattr(self, "_impl", None) is not None` be a bit better?

And relatedly, why do you `del self._impl` and then set it to None? Why not just `del` it?

python/lsst/afw/display/interface.py
```diff
@@ -57,7 +57,7 @@
 MAGENTA = "magenta"
 YELLOW = "yellow"
 ORANGE = "orange"
-
+IGNORE = "ignore"
 
 def _makeDisplayImpl(display, backend, *args, **kwargs):
```
parejkoj: flake8 won't like the single line before a method here.

python/lsst/afw/display/interface.py
RobertLuptonTheGood: I don't see the `getattr` as being clearer than the `hasattr`, but sure; I can change it.

This function closes the display but doesn't destroy it, so it's better to set it to None.  Looking at the rest of the code this would still cause problems as there are almost no checks for None, but still it's a start and I think I'll leave it this way.  I could just set it to None, but that relies on the gc deleting `_impl` promptly.

parejkoj: My point with the `getattr` is that it replaces both parts of the `if`.

Note that calling `del` does not cause the object to be garbage collected immediately, it just decrements the refcount, which means it may be garbage collected at some future time. No guarantees.

PR #286, opened 2017-10-24
DM-12207: Add color selection for photocal

PaulPrice:


include/lsst/afw/image/Calib.h
```diff
@@ -69,6 +70,55 @@ inline double fluxErrFromABMagErr(double magErr, double mag) {
...
+    }
+    return out;
+}
+
+/// Compute AB magnitude error from flux and flux error in Janskys
+template <typename T>
+ndarray::Array<T, 1> abMagErrFromFluxErr(ndarray::Array<T, 1> const& fluxErr,
+                                         ndarray::Array<T, 1> const& flux) {
+    if (flux.getNumElements() != fluxErr.getNumElements()) {
+        throw LSST_EXCEPT(pex::exceptions::LengthError, "Length mismatch");
```
parejkoj: A more explicit "Length of flux and fluxErr arrays are mismatched: XX vs. YY" might be nice.

python/lsst/afw/table/base/baseContinued.py
```diff
@@ -127,8 +127,18 @@ def __getitem__(self, key):
                 stop = len(self)
             return self.subset(start, stop, step)
         elif isinstance(key, np.ndarray):
+            if not self.isContiguous():
+                cat = self.__class__(self.table)
+                for rr, ok in zip(self, key):
```
parejkoj: This block feels somewhat magical: can we have a comment explaining how it's supposed to work?

python/lsst/afw/table/base/baseContinued.py
```diff
@@ -127,8 +127,18 @@ def __getitem__(self, key):
...
+                    if ok:
+                        cat.append(rr)
+                return cat
             return self.subset(key)
         elif isinstance(key, Key) or isinstance(key, basestring):
+            if not self.isContiguous():
+                if isinstance(key, basestring):
+                    key = self.schema[key].asKey()
+                return np.array([rr.get(key) for rr in self])
```
parejkoj: This is returning a numpy array instead of an afw table. That seems bad.

tests/test_fluxFromABMag.py
```diff
@@ -26,16 +26,18 @@
 import unittest
 import math
```
parejkoj: Do we need `math` anymore, now that you've imported numpy?

tests/test_fluxFromABMag.py
```diff
@@ -56,6 +58,20 @@ def testBasics(self):
                     abMagErr, abMag)
                 self.assertAlmostEqual(fluxErr, fluxErrRoundTrip)
 
+    def testVector(self):
+        flux = np.array([1.0, 210.0, 3210.0, 43210.0, 543210.0])
+        abMag = afwImage.abMagFromFlux(flux)
+        np.testing.assert_array_almost_equal(abMag, refABMagFromFlux(flux))
```
parejkoj: I would suggest using `self.assertFloatsAlmostEqual` for your comparisons here: it's more strict that `np.testing`.

https://developer.lsst.io/coding/python_testing.html#special-asserts

python/lsst/afw/table/base/baseContinued.py
PaulPrice: But isn't the whole idea that when you index a `Catalog` with a `Key` you do in fact get a numpy array? Or are you objecting that mutating the returned array would not mutate the `Catalog`?

PaulPrice: I would appreciate @TallJimbo's view on these changes to afw table too.

PaulPrice: If you're worried about the mutation problem I mentioned, what if we [made the returned array read only](https://stackoverflow.com/a/5541452/834250)?

parejkoj: Oh! I didn't know that these returned a numpy array instead of an afw.Table. That's somewhat surprising to me: I would expect `Thing.__getitem__[slice]` to return another `Thing`.

(thinking out loud here) But I guess `Thing[Key]` gets a column, which is like an array. That's not obvious from the alternative `return self.subset(key)`, which I had trouble finding because it's defined in C++. Maybe we should make the docstring for `__getitem__` more clear about the types of the possible return values? (says the python person who's clearly been staring at too much C++ lately...)

You are correct though that the mutation question is important: `subset` is explicitly a shallow copy, so this would cause a different behavior.

include/lsst/afw/image/Calib.h
PaulPrice: This was in a header file, and I was both being lazy and wanting to avoid the extra includes that stringifying the lengths requires. I've now moved the definitions out of the header and expanded the error string to include the lengths.

PaulPrice:
Revised according to comments.

include/lsst/afw/image/Calib.h
```diff
@@ -69,6 +70,25 @@ inline double fluxErrFromABMagErr(double magErr, double mag) {
     return std::abs(-0.4 * magErr * fluxFromABMag(mag) * std::log(10.0));
 }
 
+
+/// Compute AB magnitude from flux in Janskys
+template <typename T>
+ndarray::Array<T, 1> abMagFromFlux(ndarray::Array<T const, 1> const& flux);
```
parejkoj: As I think about it, I think these functions should be `abMagFromJansky`, to make the units very explicit (especially given RFC-356).

PaulPrice: That's an API change that would require an RFC. I would support such an RFC, but do not want to lead one, at least not now.

TallJimbo:
There two reasons why single-column array access hasn't been supported for noncontiguous catalogs thus far:
 - I was worried about sometimes returning a view and sometimes not returning a view being surprising behavior to users.  The actual usage patterns have convinced me this fear was unfounded, and returning const arrays (as you have done here) mitigates that further.
 - I didn't want users to think they were doing something efficient when they weren't.  I was originally worried about having a lot of unexpected memory allocations (because people were getting deep copies instead of views), but I think that fear was also unfounded.  However, this is definitely a problem with the Python implementation here - people will think they're doing NumPy vectorization, but they're actually writing a lot of pure-Python for loops.

So, I'm hesitant but willing to sign off on this as is, with the provision that we open an issue to move the implementation to C++.  If you have the time, I'd prefer to do that on this ticket, even if it just goes into a templated free function in the pybind11 wrappers (there isn't really a need for this functionality at all in C++, because `for` loops over records there are just as fast).

include/lsst/afw/image/Calib.h
parejkoj: Ok, that's fair. I guess we can just leave it until we throw out all of `Calib` for `PhotoCalib`.

PaulPrice:
Moved the iteration into the pybind layer.

PR #288, opened 2017-11-02
Point2I __repr__ reported wrong type.

SimonKrughoff:


tests/test_box.py
```diff
@@ -106,6 +106,22 @@ def testOverflowDetection(self):
...
+        repr_str = box.__repr__()
+        self.assertTrue(repr_str.startswith('Box2D'))
+
+    def testPointExtent(self):
+        box = geom.Box2I()
+        self.assertTrue(type(box.Point()) is geom.Point2I)
+        self.assertTrue(type(box.Extent()) is geom.Extent2I)
+        box = geom.Box2D()
+        self.assertTrue(type(box.Point()) is geom.Point2D)
+        self.assertTrue(type(box.Extent()) is geom.Extent2D)
```
TallJimbo: Might be slightly better to write this as e.g. `assertIs(box.Point, geom.Point2I)`.

SimonKrughoff: So the problem is that `box.Point` is some sort of pybind11 wrapper type.  I could say `self.assertIs(box.Point(), geom.Point2I)`.  Is that acceptable?

kfindeisen: Can you elaborate? Type comparisons should Just Work for pybind11 wrappers.

timj: Is the point that we always prefer a specific test method over `assertTrue`?

SimonKrughoff: @timj I think we should prefer specific tests if we can.

@kfindeisen This is what I'm seeing.  I didn't know if I should expect this or not.  The fact that the `Point` on a `Box2I` is a `Point2D` is the bug I'm trying to fix in this ticket.
```>>> import lsst.afw.geom as afwGeom
>>> box = afwGeom.Box2I()
>>> box.Point
<class 'lsst.afw.geom.coordinates.coordinates.Point2D'>
>>> type(box.Point)
<class 'pybind11_builtins.pybind11_type'>
>>> type(box.Point) is afwGeom.Point2D
False
```

kfindeisen: Very strange, since the `__init__` files import `lsst.afw.geom.coordinates.coordinates.Point2D` as `lsst.afw.geom.Point2D`. I'd suggest asking about it on Slack.

kfindeisen: `box.Point` should be a class object, so the last line makes sense. Is `box.Point is afwGeom.Point2D` also false?

SimonKrughoff: In python2
```>>> import lsst.afw.geom as afwGeom
>>> box = afwGeom.Box2I()
>>> box.Point
<class 'lsst.afw.geom.coordinates.coordinates.Point2D'>
>>> type(box.Point)
<class 'pybind11_type'>
```

SimonKrughoff: Very interesting:
```>>> import lsst.afw.geom as afwGeom
>>> box = afwGeom.Box2I()
>>> box.Point is afwGeom.Point2D
True
>>> box.Point == afwGeom.Point2D
True
```
Looks like I can say:
`self.assertIs(box.Point, afwGeom.Point2D)`

PR #289, opened 2017-11-06
DM-12103: Investigate centering for vignetted pupils

jmeyers314:
Adds _centerPupil function and **kwargs to getPupilFactory to
optionally recenter the illuminated portion of a pupil within
a pupil array, similar to what is done in Zemax.


mfisherlevine:
Looks good. Just change this commit message to have the same first line as the other one (first word in the imperative and all that), and move this one to be after a line break to make the bulk of the message.

Other than that, and my OOB comment on the indentation about the in the `_centerPupil()` method I think you're good to go.

PR #290, opened 2017-11-06
DM-12513 some optimizing of multiMatch.add()

parejkoj:


python/lsst/afw/table/multiMatch.py
```diff
@@ -80,37 +80,37 @@ def add(self, catalog, dataId):
...
+        matchedCatIds = set()
         for refRecord, newRecord, distance in matches:
             objId = refRecord.get(self.objectKey)
-            if objById.pop(objId, None) is None:
+            if objId in matchedRefIds:
                 # We've already matched this object against another new source,
                 # mark it as ambiguous.
                 self.ambiguous.add(objId)
-            if newById.pop(newRecord.get(self.idKey), None) is None:
+            matchedRefIds.add(objId)
```
TallJimbo: Probably irrelevant: Python is making me sad here; a method like `set.add` should really return [whether it succeeded or not](http://en.cppreference.com/w/cpp/container/set/insert), so you don't have to have the separate if statement above (which involves an identical search through the same data structure).  I've been trying to think of a way to do that with what the interface does provide, and I just don't think it's doable.  Somebody should PEP that in their spare time.

parejkoj: Took me a bit to figure out what you meant here, but I agree. Although a bit of googling found this python list thread where exactly this proposal was discussed:

https://mail.python.org/pipermail/python-ideas/2009-February/002877.html

Some of the comments are rather infuriating, but the consensus seems to be that there's a much smaller speed gain to be had there (because of caching and set operations being O(1)). In our case here, those set operations are about 1-2% of the runtime, so pretty minuscule. But if you do want to write up a PEP on it (maybe a new method, so as not change the existing API?) I'd support it.

PR #297, opened 2017-12-15
DM-12473: Add getParallacticAngle() to visitInfo

isullivan:
Adds a simple function to VisitInfo, that calculates the parallactic angle at the boresight of the exposure.

### DIFF
--- a/include/lsst/afw/image/VisitInfo.h
+++ b/include/lsst/afw/image/VisitInfo.h
@@ -174,6 +174,17 @@ public:
     // get hour angle at the boresight
     geom::Angle getBoresightHourAngle() const;
 
+    /**
+     * Get parallactic angle at the boresight
+     *
+     * Equal to the angle between the North celestial pole and Zenith at the boresight.
+     * Or, the angular separation between two great circle arcs that meet at the object:
+     *   One passing through the North celestial pole, and the other through zenith.
+     * For an object on the meridian the angle is zero if it is South of zenith and pi if it is North of zenith
+     * The angle is positive for objects East of the meridian, and negative for objects to the West.
+     */
+    geom::Angle getBoresightParAngle() const;
+
 protected:
     virtual std::string getPersistenceName() const;
 
--- a/src/image/VisitInfo.cc
+++ b/src/image/VisitInfo.cc
@@ -414,6 +414,19 @@ geom::Angle VisitInfo::getLocalEra() const { return getEra() + getObservatory().
 
 geom::Angle VisitInfo::getBoresightHourAngle() const { return getLocalEra() - getBoresightRaDec()[0]; }
 
+geom::Angle VisitInfo::getBoresightParAngle() const {
+  /**
+   * Compute the parallactic angle.
+   * Defined as the angle between the North celestial pole and Zenith at the boresight.
+   */
+  double _parallactic_y, _parallactic_x, result;
+  _parallactic_y = sin(getBoresightHourAngle().asRadians());
+  _parallactic_x = cos((getBoresightRaDec()[1]).asRadians())*tan(getObservatory().getLatitude().asRadians()) -
+    sin((getBoresightRaDec()[1]).asRadians())*cos(getBoresightHourAngle().asRadians());
+  result = atan2(_parallactic_y, _parallactic_x);
+  return result * geom::radians;
+}
+
 std::ostream& operator<<(std::ostream& os, VisitInfo const& visitInfo) {
     os << "VisitInfo(";
     os << "exposureId=" << visitInfo.getExposureId() << ", ";
--- a/tests/test_visitInfo.py
+++ b/tests/test_visitInfo.py
@@ -24,10 +24,12 @@ import math
 import os
 import unittest
 import collections
+import numpy as np
 
 import lsst.utils.tests
 import lsst.pex.exceptions
 from lsst.daf.base import DateTime, PropertySet, PropertyList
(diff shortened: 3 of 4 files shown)


include/lsst/afw/image/VisitInfo.h
```diff
@@ -174,6 +174,9 @@ class VisitInfo : public table::io::PersistableFacade<VisitInfo>, public table::
     // get hour angle at the boresight
     geom::Angle getBoresightHourAngle() const;
 
+    // get parallactic angle at the boresight
```
r-owen: Please expand this comment so define the returned angle -- it is the angle from what to what?

src/image/VisitInfo.cc
```diff
@@ -414,6 +414,15 @@ geom::Angle VisitInfo::getLocalEra() const { return getEra() + getObservatory().
 
 geom::Angle VisitInfo::getBoresightHourAngle() const { return getLocalEra() - getBoresightRaDec()[0]; }
 
+geom::Angle VisitInfo::getBoresightParAngle() const {
+  double _parallactic_y, _parallactic_x, result;
+  _parallactic_y = sin(getBoresightHourAngle());
```
r-owen: Please make the conversion to radians explicit, e.g. `getBoresightHourAngle.asRadians()`

src/image/VisitInfo.cc
```diff
@@ -425,6 +434,7 @@ std::ostream& operator<<(std::ostream& os, VisitInfo const& visitInfo) {
     os << "boresightRaDec=" << visitInfo.getBoresightRaDec() << ", ";
     os << "boresightAzAlt=" << visitInfo.getBoresightAzAlt() << ", ";
     os << "boresightAirmass=" << visitInfo.getBoresightAirmass() << ", ";
+    os << "boresightParAngle=" << visitInfo.getBoresightParAngle() << ", ";
```
r-owen: I suggest omitting this because it is not a constructor argument

isullivan:
I have added three new unit tests and edited the comments to define parallactic angle more precisely.

include/lsst/afw/image/VisitInfo.h
```diff
@@ -174,6 +174,10 @@ class VisitInfo : public table::io::PersistableFacade<VisitInfo>, public table::
...
+    // Equal to the angle measured at zenith between the North celestial pole and the boresight.
```
r-owen: This still does not specify the sign of the angle. Please add more words or reword. Here is one long, if clumsy, possibility: "This is the angular separation between two great circle arcs that meet at the zenith: the meridian and the arc from the zenith to the target. The angle is 0 if the target is on the meridian north of the zenith, positive if the object is east of the meridian, and pi if the object is on the meridian south of the zenith."

Also, use a /** style comment block when the doc is more than one line (as it now is, and justifiably so).

tests/test_visitInfo.py
```diff
@@ -486,6 +488,67 @@ def test_str(self):
...
+                                           item.boresightAirmass,
+                                           item.boresightRotAngle,
+                                           item.rotType,
+                                           item.observatory,
+                                           item.weather,
+                                           )
+            self.assertAnglesAlmostEqual(visitInfo.getBoresightParAngle(), parAngle)
+
+    def test_parallacticAngleNorthMeridian(self):
+        """An observation on the Meridian in the Northern hemisphere has a parallactic angle of pi radians."""
```
r-owen: In the office you said it will be pi if the target is south of the telescope, 0 if north of the telescope on the meridian. I think that is clearer than "in the Norther hemisphere", especially since the hemisphere should not matter.

Also please add a test for positive or negative not on the meridian. The two tests you have now will pass regardless of the sign of the parallactic angle! I realize it may be a bit trickier to predict the correct angle, but I think it is well worth testing at least one non-trivial case, since trivial cases can all too easily hide math errors.

Also I suggest you simplify this test by only setting the VisitInfo fields that matter. VisitInfo can be constructed using keyword arguments and any of them can be omitted.

isullivan: Good points, you are correct. I will update the two meridian tests to be north and south of zenith, not in the northern or southern hemisphere.

For the test of the sign of the parallactic angle, doesn't the first test `test_parallacticAngle` catch that? There it checks the results for the coordinates of data1 and data2 against precomputed values.

isullivan:
Please take another look. I have revised the documentation and the tests, and fixed a discrepancy in the definition while working it over. Please also check if the test `test_parallacticAngle` satisfies your request for a non-trivial test.
I will plan to rebase to clean up all of the commits before merging.

tests/test_visitInfo.py
```diff
@@ -48,7 +50,7 @@ def propertySetFromDict(keyValDict):
     return metadata
 
 
-class VisitInfoTestCase(unittest.TestCase):
+class VisitInfoTestCase(lsst.utils.tests.TestCase):
```
r-owen: Thank you this change

tests/test_visitInfo.py
PaulPrice: Shouldn't have underscores in Science Pipelines code.

isullivan: Okay. I was copying the style of the test immediately before the ones I added, but I suppose that one was an exception.

r-owen:
Regarding @PaulPrice's comment about the test case names: I think we usually prefer to keep things consistent within a file. You could update all the names, I suppose.

PR #299, opened 2017-12-21
DM-13096: Add refraction calculation

isullivan:
Adds a detailed calculation of refraction to `afw.coord` that can account for local atmospheric conditions in `Weather`.
Needed for generating Differential Chromatic Refraction-corrected template images.

### DIFF
--- a/python/lsst/afw/coord/refraction.py
+++ b/python/lsst/afw/coord/refraction.py
@@ -65,11 +65,11 @@ def refraction(wavelength, elevation, weather, observatory):
         raise ValueError("Refraction calculation is valid for wavelengths between 230.2 and 2058.6 nm.")
     latitude = observatory.getLatitude()
     altitude = observatory.getElevation()
-    atmosScaleheightRatio = float(4.5908E-6*temperature/units.Kelvin)
     if weather is None:
         weather = defaultWeather(altitude*units.meter)
     reducedN = deltaN(wavelength, weather)/deltaRefractScale
     temperature = extractTemperature(weather, useKelvin=True)
+    atmosScaleheightRatio = 4.5908E-6*temperature/units.Kelvin
 
     # Account for oblate Earth
     # This replicates equation 10 of Stone 1996
@@ -78,10 +78,9 @@ def refraction(wavelength, elevation, weather, observatory):
 
     # Calculate the tangent of the zenith angle.
     tanZ = np.tan(np.pi/2. - elevation.asRadians())
-
     atmosTerm1 = reducedN*relativeGravity*(1. - atmosScaleheightRatio)
     atmosTerm2 = reducedN*relativeGravity*(atmosScaleheightRatio - reducedN/2.)
-    result = Angle(float(atmosTerm1*tanZ + atmosTerm2*tanZ**3.))
+    result = Angle(atmosTerm1*tanZ + atmosTerm2*tanZ**3.)
     return result
 
 
@@ -137,11 +136,8 @@ def deltaN(wavelength, weather):
         calculated as (n_air - 1)*10^8
     """
     waveNum = 1E3/wavelength  # want wave number in units 1/micron
-
     dryAirTerm = 2371.34 + (683939.7/(130. - waveNum**2.)) + (4547.3/(38.9 - waveNum**2.))
-
     wetAirTerm = 6487.31 + 58.058*waveNum**2. - 0.71150*waveNum**4. + 0.08851*waveNum**6.
-
     return (dryAirTerm*densityFactorDry(weather) +
             wetAirTerm*densityFactorWater(weather))
 
@@ -167,12 +163,9 @@ def densityFactorDry(weather):
     waterVaporPressure = humidityToPressure(weather)
     airPressure = weather.getAirPressure()*units.pascal
     dryPressure = airPressure - waterVaporPressure
-
     eqn = (dryPressure/cds.mbar)*(57.90E-8 - 9.3250E-4*units.Kelvin/temperature +
                                   0.25844*units.Kelvin**2/temperature**2.)
-
-    densityFactor = float((1. + eqn)*(dryPressure/cds.mbar)/(temperature/units.Kelvin))
-
+    densityFactor = (1. + eqn)*(dryPressure/cds.mbar)/(temperature/units.Kelvin)
     return densityFactor
 
 
@@ -195,14 +188,11 @@ def densityFactorWater(weather):
     """
     temperature = extractTemperature(weather, useKelvin=True)
     waterVaporPressure = humidityToPressure(weather)
-
-    densityEqn1 = float(-2.37321E-3 + 2.23366*units.Kelvin/temperature -
-                        710.792*units.Kelvin**2/temperature**2. +
-                        7.75141E-4*units.Kelvin**3/temperature**3.)
-
-    densityEqn2 = float(waterVaporPressure/cds.mbar)*(1. + 3.7E-4*waterVaporPressure/cds.mbar)
-
-    relativeDensity = float(waterVaporPressure*units.Kelvin/(temperature*cds.mbar))
+    densityEqn1 = (-2.37321E-3 + 2.23366*units.Kelvin/temperature -
+                   710.792*units.Kelvin**2/temperature**2. +
+                   7.75141E-4*units.Kelvin**3/temperature**3.)
+    densityEqn2 = (waterVaporPressure/cds.mbar)*(1. + 3.7E-4*waterVaporPressure/cds.mbar)
+    relativeDensity = waterVaporPressure*units.Kelvin/(temperature*cds.mbar)
     densityFactor = (1 + densityEqn2*densityEqn1)*relativeDensity
 
     return densityFactor
@@ -230,8 +220,7 @@ def humidityToPressure(weather):
     temperature = extractTemperature(weather)
     temperatureEqn1 = (temperature + 238.3*units.Celsius)*x + 17.2694*temperature
     temperatureEqn2 = (temperature + 238.3*units.Celsius)*(17.2694 - x) - 17.2694*temperature
-    dewPoint = 238.3*float(temperatureEqn1/temperatureEqn2)
-
+    dewPoint = 238.3*temperatureEqn1/temperatureEqn2
     waterVaporPressure = (4.50874 + 0.341724*dewPoint + 0.0106778*dewPoint**2 + 0.184889E-3*dewPoint**3 +
                           0.238294E-5*dewPoint**4 + 0.203447E-7*dewPoint**5)*133.32239*units.pascal
 
--- a/python/lsst/afw/coord/refraction.py
+++ b/python/lsst/afw/coord/refraction.py
@@ -21,56 +21,59 @@ from __future__ import absolute_import, division, print_function
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
 
-from astropy import units as u
+from astropy import units
 from astropy.units import cds
 import numpy as np
 
 from lsst.afw.geom import Angle
-from lsst.afw.coord import Observatory, Weather
-from lsst.afw.geom import degrees
+from lsst.afw.coord import Weather
 
 __all__ = ["refraction", "differentialRefraction"]
 
-lsstLat = -30.244639*degrees
-lsstLon = -70.749417*degrees
-lsstAlt = 2663.
-lsstTemperature = 20.*u.Celsius  # in degrees Celcius
-lsstHumidity = 10.  # in percent
-lsstPressure = 101325.*u.pascal  # 1 atmosphere.
-
-lsstWeather = Weather(lsstTemperature.value, lsstPressure.value, lsstHumidity)
-lsstObservatory = Observatory(lsstLon, lsstLat, lsstAlt)
+# The differential refractive index is the difference of the refractive index from 1.,
+#    multiplied by 1E8 to simplify the notation and equations.
+deltaRefractScale = 1.0E8
 
 
-def refraction(wavelength, elevation, weather=lsstWeather, observatory=lsstObservatory):
+def refraction(wavelength, elevation, weather, observatory):
     """Calculate overall refraction under atmospheric and observing conditions.
 
+    The calculation is taken from Stone 1996
+    "An Accurate Method for Computing Atmospheric Refraction"
     Parameters
     ----------
-    wavelength : float
+    wavelength : `float`
         wavelength is in nm (valid for 230.2 < wavelength < 2058.6)
-    elevation : lsst.afw.geom Angle
+    elevation : `lsst.afw.geom.Angle`
         Elevation of the observation, as an Angle.
-    weather : lsst.afw.coord Weather, optional
+    weather : `lsst.afw.coord.Weather`
         Class containing the measured temperature, pressure, and humidity
         at the observatory during an observation
-    observatory : lsst.afw.coord Observatory, optional
-        Class containing the longitude, latitude, and altitude of the observatory.
+    observatory : `lsst.afw.coord.Observatory`
+        Class containing the longitude, latitude,
+        and altitude of the observatory.
 
     Returns
     -------
-    lsst.afw.geom Angle
-        The angular refraction for light of the given wavelength, under the given observing conditions.
+    `lsst.afw.geom.Angle`
+        The angular refraction for light of the given wavelength,
+        under the given observing conditions.
     """
+    if wavelength < 230.2:
+        raise ValueError("Refraction calculation is valid for wavelengths between 230.2 and 2058.6 nm.")
+    if wavelength > 2058.6:
+        raise ValueError("Refraction calculation is valid for wavelengths between 230.2 and 2058.6 nm.")
     latitude = observatory.getLatitude()
     altitude = observatory.getElevation()
+    weatherNanSafe = defaultWeather(weather, altitude=altitude*units.meter)
 
-    reducedN = deltaN(wavelength, weather)*1E-8
+    reducedN = deltaN(wavelength, weatherNanSafe)/deltaRefractScale
 
-    temperature = _extractTemperature(weather, useKelvin=True)
-    atmosScaleheightRatio = float(4.5908E-6*temperature/u.Kelvin)
+    temperature = extractTemperature(weatherNanSafe, useKelvin=True)
+    atmosScaleheightRatio = float(4.5908E-6*temperature/units.Kelvin)
 
     # Account for oblate Earth
+    # This replicates equation 10 of Stone 1996
     relativeGravity = (1. + 0.005302*np.sin(latitude.asRadians())**2. -
                        0.00000583*np.sin(2.*latitude.asRadians())**2. - 0.000000315*altitude)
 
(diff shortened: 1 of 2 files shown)


python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+        Elevation of the observation, as an Angle.
+    weather : lsst.afw.coord Weather, optional
+        Class containing the measured temperature, pressure, and humidity
+        at the observatory during an observation
+    observatory : lsst.afw.coord Observatory, optional
+        Class containing the longitude, latitude, and altitude of the observatory.
+
+    Returns
+    -------
+    lsst.afw.geom Angle
```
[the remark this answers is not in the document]
PaulPrice: I think this is a magnitude. There's also a direction (the parallactic angle). Do you want to calculate that here as well?

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    refractionStart = refraction(wavelength, elevation, weather=weather, observatory=observatory)
+    refractionEnd = refraction(wavelengthRef, elevation, weather=weather, observatory=observatory)
+    return refractionStart - refractionEnd
+
+
+def deltaN(wavelength, weather):
+    """Calculate the differential refractive index of air.
+
+    The differential refractive index is the difference of the refractive index from 1.,
```
PaulPrice: Too long for a docstring line: limit is 80 chars (though I can't find the specific item in the dev guide).

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    multiplied by 1E8 to simplify the notation and equations.
```
PaulPrice: This magic number (`1.0e8`) should be a module-level constant.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    float
+        The difference of the refractive index of air from 1., calculated as (n_air - 1)*10^8
+    """
+    waveNum = 1E3/wavelength  # want wave number in units 1/micron
+
+    dryAirTerm = 2371.34 + (683939.7/(130. - waveNum**2.)) + (4547.3/(38.9 - waveNum**2.))
+
+    wetAirTerm = 6487.31 + 58.058*waveNum**2. - 0.71150*waveNum**4. + 0.08851*waveNum**6.
+
+    return (dryAirTerm*densityFactorDry(weather) +
```
PaulPrice: No need for all those empty lines in this function.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+def deltaN(wavelength, weather):
+    """Calculate the differential refractive index of air.
+
+    The differential refractive index is the difference of the refractive index from 1.,
+    multiplied by 1E8 to simplify the notation and equations.
+    Calculated as (n_air - 1)*10^8
+
+    This replicates equation 14 of Stone 1996 "An Accurate Method for Computing Atmospheric Refraction"
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    wetAirTerm = 6487.31 + 58.058*waveNum**2. - 0.71150*waveNum**4. + 0.08851*waveNum**6.
+
+    return (dryAirTerm*densityFactorDry(weather) +
+            wetAirTerm*densityFactorWater(weather))
+
+
+def densityFactorDry(weather):
+    """Calculate dry air pressure term to refractive index calculation.
+
+    This replicates equation 15 of Stone 1996 "An Accurate Method for Computing Atmospheric Refraction"
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    Parameters
+    ----------
+    weather : lsst.afw.coord Weather
+        Class containing the measured temperature, pressure, and humidity
+        at the observatory during an observation
+
+    Returns
+    -------
+    float
+        Returns the relative density of dry air at the given pressure and temperature.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+
+    densityFactor = float((1. + eqn)*(dryPressure/cds.mbar)/(temperature/u.Kelvin))
+
+    return densityFactor
+
+
+def densityFactorWater(weather):
+    """Calculate water vapor pressure term to refractive index calculation.
+
+    This replicates equation 16 of Stone 1996 "An Accurate Method for Computing Atmospheric Refraction"
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+
+    Returns
+    -------
+    float
+        Returns the relative density of water vapor at the given pressure and temperature.
+    """
+    temperature = _extractTemperature(weather, useKelvin=True)
+    waterVaporPressure = humidityToPressure(weather)
+
+    densityEqn1 = float(-2.37321E-3 + 2.23366*u.Kelvin/temperature -
```
PaulPrice: Why the explicit cast? Here and below.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    densityEqn2 = float(waterVaporPressure/cds.mbar)*(1. + 3.7E-4*waterVaporPressure/cds.mbar)
+
+    relativeDensity = float(waterVaporPressure*u.Kelvin/(temperature*cds.mbar))
+    densityFactor = (1 + densityEqn2*densityEqn1)*relativeDensity
+
+    return densityFactor
+
+
+def humidityToPressure(weather):
+    """Simple function that converts humidity and temperature to water vapor pressure.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    Parameters
+    ----------
+    weather : lsst.afw.coord Weather
+        Class containing the measured temperature, pressure, and humidity
+        at the observatory during an observation
+
+    Returns
+    -------
+    float
+        The water vapor pressure in millibar, calculated from the given humidity and temperature.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    if np.isnan(weather.getHumidity()):
+        humidity = lsstWeather.getHumidity()
```
PaulPrice: As argued above, please don't hard-code LSST specifics.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    temperatureEqn2 = (temperature + 238.3*u.Celsius)*(17.2694 - x) - 17.2694*temperature
+    dewPoint = 238.3*float(temperatureEqn1/temperatureEqn2)
+
+    waterVaporPressure = (4.50874 + 0.341724*dewPoint + 0.0106778*dewPoint**2 + 0.184889E-3*dewPoint**3 +
+                          0.238294E-5*dewPoint**4 + 0.203447E-7*dewPoint**5)*133.32239*u.pascal
+
+    return waterVaporPressure
+
+
+def _extractTemperature(weather, useKelvin=False):
```
PaulPrice: Why does this function get a leading underscore when the others don't? You're declaring, "you should never have to use this", but if it's useful why wouldn't you want it to be used? There are no implementation-specific details in here. You've already made this less-than-public by not including it in `__all__`.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+        This is needed because Astropy can't easily convert between Kelvin and Celsius.
+
+    Returns
+    -------
+    astropy.units
+        The temperature in Celsius, unless `useKelvin` is set.
+    """
+    temperature = weather.getAirTemperature()
+    if np.isnan(temperature):
+        temperature = lsstWeather.getAirTemperature()*u.Celsius
```
PaulPrice: Flagging another LSST default.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    else:
+        temperature *= u.Celsius
+    if useKelvin:
+        temperature = temperature.to(u.Kelvin, equivalencies=u.temperature())
+    return temperature
+
+
+def _extractPressure(weather):
```
PaulPrice: Ditto on the leading underscore.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    """Thin wrapper to return the measured pressure from an observation with astropy units attached.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    """Thin wrapper to return the measured temperature from an observation with astropy units attached.
+
+    Parameters
+    ----------
+    weather : lsst.afw.coord Weather
+        Class containing the measured temperature, pressure, and humidity
+        at the observatory during an observation
+    useKelvin : bool, optional
+        Set to True to return the temperature in Kelvin instead of Celsius
+        This is needed because Astropy can't easily convert between Kelvin and Celsius.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    dewPoint = 238.3*float(temperatureEqn1/temperatureEqn2)
+
+    waterVaporPressure = (4.50874 + 0.341724*dewPoint + 0.0106778*dewPoint**2 + 0.184889E-3*dewPoint**3 +
+                          0.238294E-5*dewPoint**4 + 0.203447E-7*dewPoint**5)*133.32239*u.pascal
+
+    return waterVaporPressure
+
+
+def _extractTemperature(weather, useKelvin=False):
+    """Thin wrapper to return the measured temperature from an observation with astropy units attached.
```
PaulPrice: Too long.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+
+    Parameters
+    ----------
+    weather : lsst.afw.coord Weather
+        Class containing the measured temperature, pressure, and humidity
+        at the observatory during an observation
+
+    Returns
+    -------
+    astropy.units
```
PaulPrice: Are you sure about the type?

    >>> type(1.2345*astropy.units.pascal)
    <class 'astropy.units.quantity.Quantity'>

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+    waterVaporPressure = humidityToPressure(weather)
+    airPressure = _extractPressure(weather)
+    dryPressure = airPressure - waterVaporPressure
+
+    eqn = (dryPressure/cds.mbar)*(57.90E-8 - 9.3250E-4*u.Kelvin/temperature +
+                                  0.25844*u.Kelvin**2/temperature**2.)
+
+    densityFactor = float((1. + eqn)*(dryPressure/cds.mbar)/(temperature/u.Kelvin))
+
+    return densityFactor
```
PaulPrice: There's nothing terribly wrong with it, but why do you like to do:

    value = calculate(something)
    
    return value

instead of:

    return calculate(something)

?

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from astropy import units as u
+from astropy.units import cds
```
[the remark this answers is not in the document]
isullivan: It seems odd to me, but I get the error: `AttributeError: module 'astropy.units' has no attribute 'cds'`. That's why I imported `cds` on it's own line.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+from lsst.afw.geom import degrees
+
+__all__ = ["refraction", "differentialRefraction"]
+
+lsstLat = -30.244639*degrees
+lsstLon = -70.749417*degrees
+lsstAlt = 2663.
+lsstTemperature = 20.*u.Celsius  # in degrees Celcius
+lsstHumidity = 10.  # in percent
+lsstPressure = 101325.*u.pascal  # 1 atmosphere.
```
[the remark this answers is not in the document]
RobertLuptonTheGood: I totally agree with Paul here, but he doesn't make a suggestion.  I think that the only sensible solution is to add an observatory-specific object (singleton?) to the obs packages, which can  be imported or possibly be returned by the butler (cf. `camera`).  You'd then need to either pass this to the objects or import it via some camera-agnostic name.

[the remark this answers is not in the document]
isullivan: I agree in principle, and am comfortable removing these defaults. Both are typically stored in an exposures `VisitInfo`, but the problem I have been having is that the `weather` entry is often all nan's. The weather is only a second order effect for the DCR calculations I care about, so I do want the code to work even if that isn't present.

python/lsst/afw/coord/refraction.py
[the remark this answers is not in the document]
isullivan: Yes, this is only the magnitude. The parallactic angle is now computed separately in `VisitInfo`. Should I provide an example in the documentation of how to get that?

RobertLuptonTheGood:
The whole VisitInfo design could do with a revisit.  In this case, we'd use better defaults if we had a flag saying that they were only defaults.

P.S.  Do you really want a default of one atmosphere on Cerro Pachon?

isullivan:
@RobertLuptonTheGood : you're suggesting that we edit the obs packages, so that if the `weather` parameters are `NAN`, they get replaced by sensible defaults?

python/lsst/afw/coord/refraction.py
isullivan: I simply removed the entire function as part of eliminating the LSST-specific defaults.

RobertLuptonTheGood:
I think it'd take more than that.  We'd need to add a flag that says if the value is a correct or a guess.  I'd also want to make VisitInfo mutable -- it's a pain not being able to set values in python after the object is created.  Probably not in this ticket, and I'm not sure that we wouldn't want a bigger review of VisitInfo at the same time.

isullivan:
I've attempted to put in more generic defaults, by calculating the typical temperature and pressure based on the observatory's altitude. I know that won't be exactly correct, but it is plenty good enough for my purposes. I am willing to change that if there is a strong objection, of course.

python/lsst/afw/coord/refraction.py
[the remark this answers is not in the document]
isullivan: @RobertLuptonTheGood @PaulPrice I've added a function that calculates reasonable default values if the weather parameters are `NAN`. Will this satisfy your concerns? One problem I see is that it won't catch parameters that are merely bad, and not `NAN`. In the Decam data I have, for example, the pressure is set to 1 atmosphere.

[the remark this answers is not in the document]
PaulPrice: I think the obs package should take responsibility for setting the correct values or reasonable defaults. That's the approach taken with `VisitInfo`.

python/lsst/afw/coord/refraction.py
isullivan: At one point my prototype code required the cast, but it doesn't any more. I'll remove these.

python/lsst/afw/coord/refraction.py
isullivan: I find it helpful for debugging since I can easily add a breakpoint or print statement, and that it makes the code more readable. But it's mostly a habit, and one I'm happy to change when there are performance or memory issues.

PaulPrice: Fair enough. No performance or memory issues I'm aware of so do whatever you're comfortable with.

python/lsst/afw/coord/refraction.py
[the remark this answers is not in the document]
isullivan: That sounds reasonable, but I would like the option of calculating expected atmospheric parameters. For example, the Decam data I mentioned above has non-`NAN` but bogus values in the `VisitInfo`, and anything read with `obs_lsstSim` has `NAN` values for humidity. What if I calculate default values if `weather` is not passed in at all (maybe log a warning), but raise an error if `weather` is passed in with invalid parameters?
My concern is that the weather conditions are not otherwise depended on, and it doesn't seem like it's usually a high priority to make sure they are present and correct.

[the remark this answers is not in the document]
PaulPrice: Couldn't you fix the `VisitInfo`s in those obs packages?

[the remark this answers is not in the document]
isullivan: What is there to fix in the case of `obs_decam`? It correctly reads the value from the input data correctly, but that was stored as 1 atmosphere instead of the real value. I don't want to have to manually fix the data, but if I use that value, then the refraction I calculate will be ~20% too high.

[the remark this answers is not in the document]
PaulPrice: It's the obs package's job to provide the correct values. If obs_decam isn't providing correct values, then I think that's the place to plug in your approximation.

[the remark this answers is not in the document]
isullivan: I think that would require a substantial change to the `obs_*` packages, and likely involves a lengthy discussion. For now, I would like to use the approximation in *all* cases, even when `weather` information is present. I would also create a new ticket to sanitize `weather` when it is read in, and hook it in to the refraction code at that point.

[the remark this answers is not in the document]
PaulPrice: My preference would be to do it properly now, but perhaps there are good reasons to cut corners. @jdswinbank can advise.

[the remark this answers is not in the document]
jdswinbank: Can we be more specific about what's actually being proposed here?

If it's simply adding some sort of `if` statement around [this](https://github.com/lsst/obs_decam/blob/921a630251e9741fb10e1cc7b1d5110117e4afa7/python/lsst/obs/decam/makeDecamRawVisitInfo.py#L68) to make it set an appropriate pressure, then it seems like fixing it on this ticket would be trivial and we should just do it.

If it's redesigning `VisitInfo`, updating obs packages with new singletons, holding extensive design discussions, etc, then I agree with @isullivan that it belongs on another ticket.

RobertLuptonTheGood:
I don't think we can put an if here without having a ticket to remove it soon.  It's an observatory-specific if in afw

							R

jdswinbank:
@RobertLuptonTheGood Are you replying to my comment https://github.com/lsst/afw/pull/299#discussion_r158565333? If so, note that it was explicitly a link to where the `if` should go *in obs_decam*, not in afw.

RobertLuptonTheGood:
Mea culpa.  We would have to remove these defaults here at the same time, or at the absolute least label them as strongly-deprecated with a ticket number.

jdswinbank:
I agree — if this can be implemented in a low-overhead way in obs packages, it seems like a no-brainer to strip out the defaults here. If obs-package implementation is harder than it naively seemed in my three minutes of squinting at them (either for technical reasons I didn't twig, or because it gets caught up in an effort to redesign `VisitInfo`), then I'll need to think harder.

isullivan:
I'm afraid I do not understand everyone's objections, or what you are proposing I do to finish this. My proposal (added in the commit above) was to calculate reasonable values for temperature, pressure, and humidity as part of `refraction.py`. This does not hard-code any telescope-specific values. What I am hearing is that I should instead edit each `obs_*` package to test if the `weather` parameters are reasonable, and replace them with hard-coded defaults if they are `NAN` or outside of a reasonable range. I can do that, though I think it's messier than everyone is suggesting. 

However, I do still want to keep my function for calculating reasonable `weather` parameters in the refraction calculation if `weather` is not supplied at all. There are many cases (outside of production) where someone wants to calculate expected refraction for a given observatory, and they shouldn't have to have an exposure handy to do that.

jdswinbank:
Hi @isullivan — I have no objection to the existence of the `defaultWeather` function. I suggest the following:

Obs packages should be responsible for providing valid weather. Pipeline could should be able to rely on this. That means there doesn't need to be a call to `defaultWeather` within `refraction`.

If you know that obs_decam is broken (in that it doesn't always supply valid weather), the right place to fix that is in obs_decam. This seems to me like it would be as simple as calling `defaultWeather` [here](https://github.com/lsst/obs_decam/blob/921a630251e9741fb10e1cc7b1d5110117e4afa7/python/lsst/obs/decam/makeDecamRawVisitInfo.py#L68) (or close by) under some circumstances, rather than in the body of `refraction`.

You are not responsible for ensuring that all obs packages return good values under all conditions. If you know obs_decam is wrong, we should fix it now; if we later find out that other obs packages have similar issues, we can fix them on different tickets.

python/lsst/afw/coord/refraction.py
```diff
@@ -0,0 +1,284 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from astropy import units as u
```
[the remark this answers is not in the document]
parejkoj: @PaulPrice I see @isullivan changed it, but the above is the standard astropy way. See http://docs.astropy.org/en/stable/units/

isullivan:
As discussed above, I have modified the `obs_decam` and `obs_lsstSim` packages to always supply valid temperature, pressure, and humidity values. This involved adding a method to `obs_base` as well, so that it is available to all `obs` packages. The pull requests for these changes can be found below:
https://github.com/lsst/obs_decam/pull/73
https://github.com/lsst/obs_base/pull/78
https://github.com/lsst/obs_lsstSim/pull/51

PaulPrice:
Good approach!

PR #302, opened 2018-01-02
DM-12924: SpherePoint.offset should work at the poles and for negative offsets

r-owen:


include/lsst/afw/geom/SpherePoint.h
```diff
@@ -98,6 +98,9 @@ class SpherePoint final {
      * @throws pex::exceptions::InvalidParameterError
      *         Thrown if `vector` is the zero vector.
      *
+     * @note If the SpherePoint is at a pole then longitude is set to 0.
+     * That provides predictable behavior for methods such as bearingTo and offset.
+     *
```
kfindeisen: Are there any guarantees for what `northPole.getLongitude()` or `southPole.getLongitude()` returns, or does the pole's longitude depend on the arguments to `SpherePoint(lon, lat)`?

kfindeisen: IIRC `bearingTo` will auto-generate a link, but `offset` will not. Have you checked?

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -315,11 +315,6 @@ class SpherePoint final {
      *               circle defined by `bearing`
      * @returns a new point created by shifting this point
      *
-     * @throws pex::exceptions::DomainError
-     *         Thrown if `this.atPole()`.
-     * @throws pex::exceptions::InvalidParameterError
-     *         Thrown if `amount` is negative.
-     *
```
kfindeisen: If we're making all operations well-defined at the poles, do we still need `atPole`? IIRC it was originally created as a way to guard against these exceptions.

src/geom/SpherePoint.cc
```diff
@@ -134,12 +138,6 @@ bool SpherePoint::operator==(SpherePoint const& other) const noexcept {
 bool SpherePoint::operator!=(SpherePoint const& other) const noexcept { return !(*this == other); }
 
 Angle SpherePoint::bearingTo(SpherePoint const& other) const {
-    if (atPole()) {
-        stringstream buffer;
-        buffer << "Cannot calculate offset from pole " << *this << ".";
-        throw pexExcept::DomainError(buffer.str());
-    }
-
```
kfindeisen: As I understand it, `bearingTo` can give different results for the same point (e.g., the north pole) depending on how the longitude has been initialized. If this is the desired behavior, consider adding a warning to the documentation to make this explicit.

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -265,9 +268,6 @@ class SpherePoint final {
      * @param other the point to which to measure the bearing
      * @returns the direction, as defined above, in the interval [0, 2&pi;).
      *
-     * @throws pex::exceptions::DomainError
-     *         Thrown if `this.atPole()`.
-     *
```
kfindeisen: Please add a paragraph documenting the behavior at the poles. The current description implies that the bearing must always be -π/2 at the north pole and π/2 at the south pole, which I suspect is no longer what happens.

include/lsst/afw/geom/SpherePoint.h
kfindeisen: Please update the description to give the behavior at the poles (depending on how the new code relates to `bearingTo`, this may be just a matter of removing the phrase "not at a coordinate pole").

tests/test_spherePoint.py
```diff
@@ -134,15 +134,36 @@ def testLonLatConstructorErrors(self):
         with self.assertRaises(pexEx.InvalidParameterError):
             SpherePoint(-42.0, -inf, degrees)
 
-    def testInit1ArgErrors(self):
-        """Test if 1-argument form of __init__ handles invalid input.
-        """
+    def testPoint3DConstructor(self):
+        # test poles
```
kfindeisen: Did you mean to remove the docstring?

tests/test_spherePoint.py
```diff
@@ -481,18 +502,6 @@ def testInquality(self):
...
-        safe = SpherePoint(0.0, 0.0, degrees)
-
-        with self.assertRaises(pexEx.DomainError):
-            northPole.bearingTo(safe)
-        with self.assertRaises(pexEx.DomainError):
-            southPole.bearingTo(safe)
-
     def testBearingToValue(self):
         """Test if bearingTo() returns the expected value.
         """
```
kfindeisen: I would like to see some test coverage at the poles. Does `bearingTo` produce the output it should (especially as a function of the pole's longitude)?

tests/test_spherePoint.py
```diff
@@ -762,28 +771,12 @@ def testRotatedAlias(self):
...
-            nonSingularity.offset(90.0*degrees, -0.1*degrees)
-
     def testOffsetValue(self):
         """Test if offset() returns the expected value.
         """
         # This should cover arcs over the meridian, across the pole, etc.
         for lon1, lat1 in self._dataset:
             point1 = SpherePoint(lon1, lat1)
-            if point1.atPole():
-                continue
```
kfindeisen: Again, please add some test coverage at the poles (including testing for consistency between `offset` and `bearingTo`/`separation`).

tests/test_spherePoint.py
r-owen: I meant to not include one. I think the name of the method suffices.

include/lsst/afw/geom/SpherePoint.h
r-owen: OK. I rewrote the description to one whose behavior is clearer at the poles.

include/lsst/afw/geom/SpherePoint.h
r-owen: I believe `atPole` may still be useful, as some operations are not practical at the pole. One example: a user may query it to find out if longitude was set to 0 by the Point3D constructor; this could help avoid discontinuities in some situations.

src/geom/SpherePoint.cc
r-owen: I think the documentation is sufficiently clear, now that I have reworked it.

tests/test_spherePoint.py
r-owen: Fair enough. I had not noticed until now that the existing test are all on the equator.

tests/test_spherePoint.py
r-owen: Done in testOffsetValues

include/lsst/afw/geom/SpherePoint.h
r-owen: I'll add `@ref`

r-owen:
Good points. Can you please have another look? I improved the documentation (including finding a few more items that needed updating). I also changed `SpherePoint.operator==` to compare longitude even at poles, because it makes a difference for `bearingTo` and `offset`, and updated the documentation and unit test accordingly.

r-owen:
A question: should we document methods that return a new SpherePoint as possibly resulting in longitude 0 at the poles? I suspect so. There are only two, and both return a SpherePoint constructed from a Point3D.

tests/test_spherePoint.py
```diff
@@ -544,6 +527,25 @@ def testBearingToValue(self):
...
+                    sp1 = SpherePoint(longDeg, lat1Deg, degrees)
+                    if sp0.atPole() and sp1.atPole():
+                        # the points are at opposite poles; any bearing may be returned
+                        continue
+                    bearing = sp0.bearingTo(sp1)
+                    if lat1Deg > lat0Deg:
+                        self.assertAnglesAlmostEqual(bearing, 90 * degrees)
+                    else:
+                        self.assertAnglesAlmostEqual(bearing, -90 * degrees)
+
```
kfindeisen: Test bearing from a pole to a point not on the same longitude? As I understand it, the answer should be the difference in longitudes minus 90 degrees for the north pole, and the difference plus 90 degrees for the south pole.

tests/test_spherePoint.py
```diff
@@ -762,48 +764,60 @@ def testRotatedAlias(self):
...
                 if point1.isFinite() and point2.isFinite():
                     if not point2.atPole():
                         self.assertAnglesAlmostEqual(
                             point2.getLongitude(), newPoint.getLongitude())
+                        self.assertAnglesAlmostEqual(
+                            point2.getLongitude(), newPoint2.getLongitude())
                     self.assertAnglesAlmostEqual(
                         point2.getLatitude(), newPoint.getLatitude())
+                    self.assertAnglesAlmostEqual(
+                        point2.getLatitude(), newPoint2.getLatitude())
```
kfindeisen: These assertions are redundant with your use of `assertSpherePointsAlmostEqual`, above.

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -296,7 +301,8 @@ class SpherePoint final {
      * @param axis a point defining the north pole of the rotation axis.
      * @param amount the amount of rotation, where positive values
      *               represent right-handed rotations around `axis`.
-     * @returns a new point created by rotating this point
+     * @returns a new point created by rotating this point. If the point is at the pole
+     * then its longitude will be 0.
```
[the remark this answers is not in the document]
r-owen: I like your suggested wording change and will adopt it.

Your suggestion that the longitude be the direction of entry is very interesting. Another way to state it is that it is the limiting value as the amount of rotation causes the new point to approach the pole. From that perspective it is useful in eliminating a discontinuity. However, I don't know how to compute it simply and robustly. One obvious technique is to offset by a slightly smaller offset (using an epsilon computed to be enough to get away from the pole. If that results in a point with invalid longitude (going too far, over the pole) try a slightly larger offset. But even if that works, computing the size of the required epsilon may be tricky (consider a small circle near the pole).

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -305,20 +311,17 @@ class SpherePoint final {
     /**
      * Return a point offset from this one along a great circle.
      *
-     * For any point `A` not at a coordinate pole, and any two angles `b`
-     * and `delta`, `A.bearingTo(A.offset(b, delta))` = `b` and
-     * `A.separationTo(A.offset(b, delta))` = `delta`.
+     * For any point `A` not at a coordinate pole, any angle `bearing` and any non-negative angle `amount`,
+     * if `B` = A.offset(bearing, amount)` then `A.bearingTo(B) = amount` and `A.separationTo(B) = amount`.
+     * Negative values of `amount` are supported in the obvious manner:
+     * `A.offset(bearing, delta) = A.offset(bearing + 180*degrees, -delta)`
```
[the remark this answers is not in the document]
r-owen: The new tests check all the points in the standard point list, which include points at the pole. Is that what you are asking?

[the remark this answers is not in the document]
kfindeisen: The text says "For any point `A` *not at a coordinate pole*" (emphasis added). I don't understand why we still need that disclaimer.

tests/test_spherePoint.py
```diff
@@ -449,52 +476,8 @@ def checkTransitive(self, delta):
...
                             point1 or point2 == point1)
 
-    def testEqualityAlias(self):
-        """Test if == handles coordinate degeneracies correctly.
-        """
-        # Longitude wrapping
-        self.assertEqual(SpherePoint(360.0, -42.0, degrees),
-                         SpherePoint(0.0, -42.0, degrees))
-        self.assertEqual(SpherePoint(-90.0, -42.0, degrees),
-                         SpherePoint(270.0, -42.0, degrees))
```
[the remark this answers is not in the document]
r-owen: Good point. Most of the tests in this test method were obsoleted by my other changes but we still want to test wrapping of longitude. I'll restore this under a new test name.

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -85,6 +89,9 @@ class SpherePoint final {
      *         Thrown if `latitude` isout of range.
      *
      * @exceptsafe Provides strong exception guarantee.
+     *
+     * @warning If longitude is +/- inf then the resulting longitude is `NaN`
+     * since infinity cannot be wrapped into the range
```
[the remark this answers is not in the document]
r-owen: I meant to remove the warning entirely, and have now done so. The same information is contained in the description of the `longitude` argument.

tests/test_spherePoint.py
r-owen: I disagree. assertSpherePointsAlmostEqual tests angular separation, and thus cannot tell the difference between two SpherePoints at (or very near) the pole that have different longitudes.

kfindeisen: Then can you add a comment to that effect?

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -63,6 +63,8 @@ class SpherePoint final {
      * Construct a SpherePoint from a longitude and latitude.
      *
      * @param longitude The longitude of the point.
+     *                  It will be wrapped into the range [0, 2 &pi;) radians,
+     *                  and `+/-inf` is wrapped to `nan`
```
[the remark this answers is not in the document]
r-owen: Stating that the angle is wrapped here makes it easier to understand why infinity is converted to nan, but I'll try a gentler rewording.

r-owen:
I'm done. Please have another look at the docs. I updated the lon/lat constructors and operator== (a brief note that wrap is ignored when comparing longitude)

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -225,12 +223,12 @@ class SpherePoint final {
      */
 
     /**
-     * `true` if two points represent the same position.
+     * `true` if two points have the same longitude (ignoring wrap) and latitude.
```
kfindeisen: Sorry, but "ignoring wrap" is kind of confusing (does "ignore" mean that 90 degrees and 450 degrees are considered the same, or not?). It also seems redundant, given that `getLongitude()` is guaranteed to return a wrapped angle or NaN.

r-owen: I'm feel that I am in a Catch-22. The way I see it is:
- If we assume that the constructor wraps longitude on construction (which you complained about when I documented that), then it is unambiguous what equal longitude means in `operator==`.
- If we do not assume that the constructor wraps longitude on construction, then `operator==` is ambiguous, in that it is not clear if it compares the actual longitude or the wrapped version returned by `getLongitude()`

I personally prefer to document that longitude is wrapped on construction. That neatly explains why `+/-inf` is turned into `nan` and gives an the user a clear model of what is going on which can be extended to methods such as `operator==`. But I can see why you felt that exposed the implementation too much. If you still feel that way then please suggest clear and unambiguous documentation for `operator==`

kfindeisen: How about removing the parenthetical note from the title, but writing:

> `@returns true if this point has exactly the same values of `getLongitude()` and `getLatitude()` as `other`, false otherwise`

(Also, I just noticed while looking at this that `@note` is in front of the params/returns sections; it should go next to the warning).

r-owen: OK. I moved one other note as well

include/lsst/afw/geom/SpherePoint.h
```diff
@@ -225,17 +223,18 @@ class SpherePoint final {
...
-     * Points are considered equal if and only if they represent the same
-     * location, regardless of how they were constructed. In particular,
-     * representations of the north or south pole with different longitudes
-     * are considered equal.
+     * `true` if two points have the same longitude and latitude.
      *
      * @param other the point to test for equality
-     * @returns true if this point matches `other` exactly, false otherwise
+     * @returns true if this point has exactly the same values ofgetLongitude() and getLatitude()
+     * asother, false otherwise
```
kfindeisen: Couple of missing spaces (last comment, I promise!)

PR #303, opened 2018-01-02
DM-13065: SipApproximation

TallJimbo:
Best to start with the docs for the afw/geom/SipApproximation class and the afw/math/polynomials namespace (in afw/math/polynomials.h) to get the big picture.

### DIFF
--- a/include/lsst/afw/geom/SipApproximation.h
+++ b/include/lsst/afw/geom/SipApproximation.h
@@ -1,0 +1,370 @@
+// -*- lsst-c++ -*-
+/*
+ * Developed for the LSST Data Management System.
+ * This product includes software developed by the LSST Project
+ * (https://www.lsst.org).
+ * See the COPYRIGHT file at the top-level directory of this distribution
+ * for details of code ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#ifndef LSST_AFW_GEOM_SipApproximation_h_INCLUDED
+#define LSST_AFW_GEOM_SipApproximation_h_INCLUDED
+
+#include <memory>
+#include <vector>
+
+#include "Eigen/Core"
+
+#include "lsst/afw/geom/Transform.h"
+#include "lsst/afw/geom/Box.h"
+#include "lsst/afw/geom/LinearTransform.h"
+
+namespace lsst { namespace afw { namespace geom {
+
+/**
+ *  A fitter and results class for approximating a general Transform in a form compatible with
+ *  FITS WCS persistence.
+ *
+ *  The Simple Imaging Polynomial (SIP) convention (Shupe et al 2005) adds
+ *  forward and reverse polynomial mappings to a standard projection FITS WCS
+ *  projection (e.g. "TAN" for gnomonic) that relate Intermediate World
+ *  Coordinates (see Calabretta & Greisen 2002) to image pixel coordinates.
+ *  The SIP "forward" transform is defined by polynomial coeffients @f$A@f$
+ *  and @f$B@f$ that map pixel coordinates @f$(u, v)@f$ to Intermediate World
+ *  Coordinates @f$(x, y)@f$ via
+ *  @f[
+ *     \boldsymbol{S}\left[\begin{array}{c}
+ *        x \\
+ *        y
+ *     \end{array}\right]
+ *     \equiv
+ *     \left[\begin{array}{c}
+ *        x_s \\
+ *        y_s
+ *     \end{array}\right]
+ *     =
+ *     \left[\begin{array}{c}
+ *        (u - u_0) + \displaystyle\sum_{p,q}^{0 \le p + q \le N} \mathrm{A}_{p,q} (u - u_0)^p (v - v_0)^q \\
+ *        (v - v_0) + \displaystyle\sum_{p,q}^{0 \le p + q \le N} \mathrm{B}_{p,q} (u - u_0)^p (v - v_0)^q
+ *     \end{array}\right]
+ *  @f]
+ *  The reverse transform has essentially the same form:
+ *  @f[
+ *     \left[\begin{array}{c}
+ *        u - u_0 \\
+ *        v - v_0
+ *     \end{array}\right]
+ *     =
+ *     \left[\begin{array}{c}
+ *        x_s + \displaystyle\sum_{p,q}^{0 \le p + q \le N} \mathrm{AP}_{p,q} x_s^p y_s^q \\
+ *        y_s + \displaystyle\sum_{p,q}^{0 \le p + q \le N} \mathrm{BP}_{p,q} x_s^p y_s^q
+ *     \end{array}\right]
+ *  @f]
+ *  In both cases, @f$(u_0, v_0)@f$ is the pixel origin (CRPIX in FITS WCS) and
+ *  @f$\boldsymbol{S}@f$ is the *inverse* of the Jacobian "CD" matrix.  Both
+ *  CRPIX and CD are considered fixed inputs, and we do not attempt to null
+ *  the zeroth- and first-order terms of @f$A@f$ and @f$B@f$ (as some SIP
+ *  fitters do); together, these conventions make solving for the coefficients
+ *  a much simpler linear problem.
+ *
+ *  @note In the implementation, we typically refer to @f$(u-u_0, v-v_0)@f$ as
+ *  @c dpix (for "pixel delta"), and @f$(x_s, y_s)@f$ as @c siwc (for "scaled
+ *  intermediate world coordinates").
+ *
+ *  While LSST WCSs are in general too complex to be described exactly in FITS
+ *  WCS, they can generally be closely approximated by standard FITS WCS
+ *  projection with additional SIP distortions.  This class fits such an
+ *  approximation, given a TransformPoint2ToPoint2 object that represents the
+ *  exact mapping from pixels to Intermediate World Coordinates with a SIP
+ *  distortion.
+ */
+class SipApproximation {
+public:
+
+    /**
+     *  Construct a new approximation by fitting on a grid of points.
+     *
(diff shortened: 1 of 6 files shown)


include/lsst/afw/math/polynomials/Basis1d.h
```diff
@@ -0,0 +1,107 @@
...
+#define LSST_AFW_MATH_POLYNOMIALS_Basis1d_h_INCLUDED
+#ifdef DOXYGEN
+
+namespace lsst { namespace afw { namespace math { namespace polynomials {
+
+/**
+ *  A basis concept for 1-d series expansions.
+ *
+ *  @note This class is only present in the documentation, as it represents an
+ *        abstract concept.
```
r-owen: I have never seen this done before. Clearly you could make this a real class by using template parameters. I assume you are trying to avoid inheritance, as per your documentation in polynomials.h ("the classes... are not polymorphic (no virtual functions)"). A few more words of explanation here would be appreciated.

include/lsst/afw/math/polynomials.h
```diff
@@ -0,0 +1,96 @@
...
+ *
+ *  The library also includes a few utility classes and functions:
+ *
+ *   - BinomialMatrix, which provides an efficient and stable way to get
+ *     binomial coefficients.
+ *
+ *   - PackedIndexIterator and PackedIndexRange, which handle the flattening
+ *     of pairs of 1-d polynomials coefficients and basis functions into 1-d
+ *     arrays.
+ */
```
r-owen: I strongly suspect you should put this code (including useful functions based on it) in a new package, rather than `afw`, since it seems potentially generally useful and has a minimum of dependencies. I'm thinking along the lines of your RFC for using sphgeom and I'd like to avoid a change like that in the future.

include/lsst/afw/math/polynomials/Basis1d.h
TallJimbo: Concepts are very much a part of C++ - they're not in the language yet, but they've been a part of how people think about the language since the inception of the STL (see, e.g. http://en.cppreference.com/w/cpp/concept).  Sometimes people try to *implement* those with a templated base class (i.e. CRTP, as in the Eigen class hierarchies), and I tried that a bit here, but I ultimately rejected that approach because I felt the associated boilerplate obfuscated things more than it helped.

But this way of documenting a Concept as by telling Doxygen that it's a class is something new, here, and it seemed the best way to provide a separate documentation page that had more-or-less the right relationship to the concrete classes that implement the Concept.  I'm pretty happy with the result, but documentation is best judged by someone other than the author and I'm open to other ways of documenting this.  Given how focused it is on C++, it's a bit strange that Doxygen doesn't include something native for Concepts.

include/lsst/afw/math/polynomials.h
TallJimbo: Working on this was part of the inspiration for that RFC, but because some of this code *does* depend on `afw.geom` it can't be moved out of afw yet, and I'd rather not split it up based on that dependency problem; I'd rather address the dependency problem by moving both the geometry primitives and this code to another package.  But since that's a lot more work, I can't make a good case that it's what LSST should spend its time on, and hence the RFC.

include/lsst/afw/math/polynomials/Basis1d.h
r-owen: A few words in the documentation would be helpful -- at least a link to the cppreference page you site. I had no idea the word "concept" was formal in this note.

TallJimbo: Good idea.  Will do.  I'll also capitalize Concept in the docs, which I should have done before.

include/lsst/afw/math/polynomials/PackedBasis2d.h
```diff
@@ -0,0 +1,234 @@
...
+template <typename Basis>
+class Function2d;
+
+/**
+ *  A Basis2d formed from the product of a Basis1d for each of x and y, truncated
+ *  at the sum of their orders.
+ *
+ *  If @f$B_n(x)@f$ are the basis functions for the nested Basis1d, the basis
+ *  functions of a PackedBasis2d with order @f$N@f$ are @f$B_m(x)B_n(y)@f$ for all combinations
+ *  with @f$m + n \le N@f$.
```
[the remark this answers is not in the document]
TallJimbo: This is documented in the classes in `PackedIndex.h`; I've added a reference to that documentation here.  The convention is the latter of the two examples you gave.  I'm not sure what one FunctionLibrary uses, but this is the one used by `lsst.shapelet` and (internally) by `ChebyshevBoundedField`.

include/lsst/afw/math/polynomials/PackedIndex.h
```diff
@@ -0,0 +1,238 @@
...
+            _index.x = 0;
+        } else {
+            --_index.y;
+            ++_index.x;
+        }
+        return *this;
+    }
+
+    /// Move to the next element in the packed array and return a copy of the iterator before the move.
+    PackedIndexIterator operator++(int) noexcept {
```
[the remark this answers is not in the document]
TallJimbo: It is necessary to meet the formal requirements of being a conformant STL [`InputIterator`](https://en.cppreference.com/w/cpp/named_req/InputIterator).

TallJimbo:
@r-owen, I think this is ready for another look.

I actually took some time to look more closely into why `unscaled` involved a loss of precision, including adding a "safe sum" algorithm that should avoid any loss of precision beyond the truncation imposed by the actual result type (you still can't compute more digits than fit in a `double`, of course).  The results were a bit surprising: the loss of precision in `unscaled` really isn't that bad, even without the safe algorithm.  The safe algorithm is slightly better, so I've kept it around (especially as I'm much more confident in its worst-case performance), but the bottom line is that all of the loss of precision in the high-level test in afw happens in the fit of the unscaled polynomials (which we can't use the safe sum on, because it's a matrix inversion, not a simple sum), rather than the conversion to unscaled polynomials.  The `50*DEFAULT_RTOL` seen in the tests is scarier than it looks - note that this is still an accuracy of `~1E-15`.  As a result, I don't actually see a problem with renaming `unscaled` to `simplified`, and that's what I've done here.

Also, note that there is an afw PR for this, with mostly obsolete comments dating back to when the polynomial was in afw.  Please take a look at that, too.

include/lsst/afw/geom/SipApproximation.h
```diff
@@ -0,0 +1,297 @@
...
+    double getA(int p, int q) const;
+
+    /// Return a coefficient of the forward transform polynomial.
+    double getB(int p, int q) const;
+
+    /// Return a coefficient of the reverse transform polynomial.
+    double getAP(int p, int q) const;
+
+    /// Return a coefficient of the reverse transform polynomial.
+    double getBP(int p, int q) const;
```
r-owen: Please add methods to return the coefficients as Eigen matrices

src/geom/SipApproximation.cc
```diff
@@ -0,0 +1,340 @@
...
+    for (int iy = 0; iy < dimensions.getY(); ++iy) {
+        double const y = bbox.getMinY() + iy*dy;
+        for (int ix = 0; ix < dimensions.getX(); ++ix) {
+            points.emplace_back(bbox.getMinX() + ix*dx, y);
+        }
+    }
+    return points;
+}
+
+poly::PolynomialFunction2dYX readFunctionFromCoeffMatrix(ndarray::Array<double const, 2> const & coeffs) {
```
r-owen: Please add brief documentation of what this does. Why this name instead of, say, "makePolynomial"?

include/lsst/afw/geom/SipApproximation.h
```diff
@@ -0,0 +1,297 @@
...
+     *                              pixelToIwc.applyForward.
+     *  @param[in]  svdThreshold    Fraction of the largest singular value at which to
+     *                              declare smaller singular values zero in the least
+     *                              squares solution.  Negative values use Eigen's
+     *                              internal default.
+     */
+    SipApproximation(
+        std::shared_ptr<TransformPoint2ToPoint2> pixelToIwc,
+        Point2D const & crpix,
+        LinearTransform const & cd,
```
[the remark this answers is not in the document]
TallJimbo: FWIW, I think we should be probably using `LinearTransform` in all of those places, but I'll bow to local precedent here.

include/lsst/afw/geom/SipApproximation.h
```diff
@@ -0,0 +1,297 @@
...
+     *                              corresponding to powers of @f$x_s@f$ and the
+     *                              second corresponding to powers of @f$y_s@f$.
+     *  @param[in]  bp              Matrix of BP coefficients, with the first dimension
+     *                              corresponding to powers of @f$x_s@f$ and the
+     *                              second corresponding to powers of @f$y_s@f$.
+     *  @param[in]  useInverse      If true, the inverse SIP transform will be compared
+     *                              to data points generated by calls to
+     *                              pixelToIwc.applyInverse instead of
+     *                              pixelToIwc.applyForward.
+     */
```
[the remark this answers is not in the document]
TallJimbo: The purpose of this is to be able to test that `SipApproximation` internal evaluate identical transformations the same way that `SkyWcs` does (which is something that needs to be verified before testing that it evaluates approximation transforms similarly), so I can't really use `SkyWcs` as a replacement for this.

include/lsst/afw/geom/SipApproximation.h
```diff
@@ -0,0 +1,297 @@
...
+     *  Return the maximum deviation of the solution from the exact transform
+     *  on the current grid.
+     *
+     *  The deviations are in scaled intermediate world coordinates
+     *  @f$\sqrt{\delta x_s^2 \delta y_s^2}@f$ for the forward transform and in
+     *  pixels @f$(\delta u^2, \delta v^2)@f$ for the reverse transform
+     *  (respectively).  Note that in the common case where the CD matrix
+     *  includes the scaling from angle units to pixel units, the *scaled*
+     *  intermediate world coordinate values are also in (nominal) pixel
+     *  units.
```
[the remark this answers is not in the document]
TallJimbo: Yup, that's the natural result of the computation, though putting all of the deviations in pixel units was actually the motivation.

python/lsst/afw/geom/coordinates/__init__.py
```diff
@@ -1,2 +0,0 @@
-# provide backwards compatibility for pickle
-from .coordinates import *
```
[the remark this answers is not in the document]
TallJimbo: Huh.  It's been long enough that I'm not sure what I was thinking.  Will revert and see what happens.

TallJimbo:
afw changes from the last round of review comments are now ready to.

> Do you anticipate any syntactic sugar for making a WCS from one of these (e.g. a new function)?

Not immediately; I'm not entirely certain how we'll want to use this class to construct a new WCS in detail - after initializing a `SipApproximation`, you might fit multiple times and change the grid size or order a few times before deciding you're happy with the results.  We'll need to find some reasonable default logic for that, of course, but it's not something I have time to do on this ticket.

PR #306, opened 2018-01-09
DM-13189: add FunctorKey for Boxes

TallJimbo:
This adds `FunctorKeys` for `Box2I` and `Box2D`, which should start us towards recording boxes in tables in a consistent way.

I've also updated three other classes in afw to use the new `FunctorKey` because they were already using the same convention and hence the change could be done in a backwards-compatible way.  Two others (`AmpInfoRecord` and `Chebyshev1Function2d`) use a different convention and cannot be converted without changing APIs and/or persistence formats.

### DIFF
--- a/include/lsst/afw/table/aggregates.h
+++ b/include/lsst/afw/table/aggregates.h
@@ -34,6 +34,9 @@ namespace geom {
 template <typename T, int N>
 class Point;
 
+class Box2I;
+class Box2D;
+
 namespace ellipses {
 
 class Quadrupole;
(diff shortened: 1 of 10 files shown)


python/lsst/afw/table/aggregates/aggregatesContinued.py
```diff
@@ -40,6 +40,11 @@ class PointKey(with_metaclass(TemplateMeta, object)):
 PointKey.register((np.float64, 2), aggregates.Point2DKey)
 
 
+# Because Boxes aren't templates themselves, we don't expose the fact that
```
parejkoj: Should this be in a docstring?

include/lsst/afw/table/aggregates.h
```diff
@@ -106,6 +109,77 @@ class PointKey : public FunctorKey<lsst::afw::geom::Point<T, 2> > {
 typedef PointKey<int> Point2IKey;
 typedef PointKey<double> Point2DKey;
 
+
+/**
+ *  A FunctorKey used to get or set a geom::Box2I or Box2D from a (min, max) pair of PointKeys.
+ *
+ *  the Box2IKey and Box2DKey typedefs should be preferred to using the template name directly.
```
parejkoj: capitalize "The"

include/lsst/afw/table/aggregates.h
```diff
@@ -106,6 +109,77 @@ class PointKey : public FunctorKey<lsst::afw::geom::Point<T, 2> > {
...
+
+    /// Type of coordinate elements (i.e. int or double).
+    using Element = typename Box::Element;
+
+    /**
+     *  Add _min_x, _min_y, _max_x, _max_y fields to a Schema, and return a BoxKey that points to them.
+     *
+     *  @param[in,out] schema  Schema to add fields to.
+     *  @param[in]     name    Name prefix for all fields; suffixes above will be appended to this
+     *                         to form the full field names.
```
parejkoj: Please give an example, `e.g. "somename_min_x"`

include/lsst/afw/table/aggregates.h
```diff
@@ -106,6 +109,77 @@ class PointKey : public FunctorKey<lsst::afw::geom::Point<T, 2> > {
...
+
+/**
+ *  A FunctorKey used to get or set a geom::Box2I or Box2D from a (min, max) pair of PointKeys.
+ *
+ *  the Box2IKey and Box2DKey typedefs should be preferred to using the template name directly.
+ */
+template <typename Box>
+class BoxKey : public FunctorKey<Box> {
+public:
+
```
parejkoj: Specify default/delete copy/move (RFC-209)?

tests/test_functorKeys.py
```diff
@@ -112,6 +112,52 @@ def testPointKey(self):
         self.doTestPointKey("D", lsst.afw.table.Point2DKey,
                             lsst.afw.geom.Point2D)
 
+    def doTestBoxKey(self, pointFieldType, functorKeyType, valueType):
```
parejkoj: Docstring, please, at least for the args.

tests/test_functorKeys.py
```diff
@@ -112,6 +112,52 @@ def testPointKey(self):
...
+        schema = lsst.afw.table.Schema()
+        fKey0 = functorKeyType.addFields(schema, "a", "box", "pixel")
+        minKey = pointFieldType(schema["a_min"])
+        maxKey = pointFieldType(schema["a_max"])
+        # we create two equivalent functor keys, using the two different
+        # constructors
```
parejkoj: don't have to do 80 chars, I don't think...

python/lsst/afw/table/aggregates/aggregatesContinued.py
TallJimbo: I don't think so; the potential confusion this comment is trying to avoid is from seeing a change in the pattern established in the rest of the file, and hence I don't think there's much to explain for someone who isn't reading the code.

tests/test_functorKeys.py
TallJimbo: Correct, this is just what my editor does when I tell it to wrap (since it treats comments like docs, which do have to by 80 chars, I think).  Anyhow, I agree it looks funny here; will fix.

TallJimbo:
Also, thanks for the reminder on RFC-209; I just had another reminder about that from another angle (DM-9932), and the implementation of the RFC is now in review.

PR #309, opened 2018-01-16
DM-12995/RFC-436: Random: ensure seed is non-zero

PaulPrice:
Random has accepted a seed value of zero and then thrown an exception.
This is not necessary, since we can construct a seed value that is
non-zero from whatever we're given.


PaulPrice:
Correct, I have abandoned my original (stupid) plan and embraced @ktlim's sensible method.

PR #310, opened 2018-01-17
DM-10819: Define Endpoint equality

kfindeisen:
The C++ operator only handles Endpoints with the same `BaseEndpoint` parent (a more general operator is hard to implement because of how templates and inheritance interact, and would not be useful in C++ anyway). The Pybind11 wrapper uses conditional templates to provide a universal comparison operator for the Python API.

### DIFF
--- a/tests/test_endpoint.cc
+++ b/tests/test_endpoint.cc
@@ -1,0 +1,148 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * See COPYRIGHT file at the top of the source tree.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program. If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#define BOOST_TEST_DYN_LINK
+#define BOOST_TEST_MODULE EndpointCpp
+
+#include "boost/test/unit_test.hpp"
+
+#include <string>
+#include <vector>
+
+#include "lsst/afw/geom/Endpoint.h"
+
+using namespace std;
+
+/* C++ unit tests for Endpoint.
+ *
+ * In addition to tests of C++-only functionality, this test suite includes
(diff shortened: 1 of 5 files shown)


tests/test_endpoint.cc
```diff
@@ -0,0 +1,137 @@
...
+
+    for (auto const endpoint1 : endpoints) {
+        for (auto const endpoint2 : endpoints) {
+            for (auto const endpoint3 : endpoints) {
+                bool const equals12 = *endpoint1 == *endpoint2;
+                bool const equals13 = *endpoint1 == *endpoint3;
+                bool const equals23 = *endpoint2 == *endpoint3;
+
+                // Iteration covers all permutations, so no need to be redundant
+                bool const transitive = !equals12 || !equals13 || equals23;
```
r-owen: Can you please try to make the test easier to follow?  For a given combination, it isn't sufficient. For instance 1!=2 is enough to make it true, so even if 1==3 and 2==3 this will pass. I suspect it works by the time you iterate over all possible combinations, but it's not obvious to me that is so.

tests/test_endpoint.cc
```diff
@@ -0,0 +1,137 @@
...
+
+/// Test whether Endpoint equality treats subclasses as non-substitutable.
+BOOST_AUTO_TEST_CASE(EndpointEqualsNotPolymorphic) {
+    IcrsCoordEndpoint superclass;
+    GratuitousIcrsCoordEndpoint subclass;
+
+    // Test both IcrsCoordEndpoint::operator== and GratuitousIcrsCoordEndpoint::operator==
+    BOOST_TEST(superclass != subclass);
+    BOOST_TEST(subclass != superclass);
+}
```
r-owen: I suggest also testing `!(superlass == subclass)` and vice-versa.

tests/test_endpoint.cc
```diff
@@ -0,0 +1,137 @@
...
+        make_shared<IcrsCoordEndpoint>()
+    };
+
+    BOOST_TEST(*(endpoints[0]) != *(endpoints[1]));
+    BOOST_TEST(*(endpoints[1]) != *(endpoints[0]));
+    BOOST_TEST(*(endpoints[0]) == *(endpoints[2]));
+    BOOST_TEST(*(endpoints[2]) == *(endpoints[0]));
+
+    for (auto const endpoint1 : endpoints) {
+        for (auto const endpoint2 : endpoints) {
```
r-owen: I suggest adding a test here that != gives the opposite answer to ==

tests/test_endpoint.cc
kfindeisen: I'm not sure I can. Transitivity is a fundamentally difficult assertion to test because it's an implication: the only way to verify it is by trying lots of combinations of objects (arguably I don't have enough) and seeing that the equalities and inequalities always line up.

I can add the missing permutations, but will that make the test easier or harder to follow?

r-owen: Perhaps some comments might help. I am not even sure what this particular line is testing.

kfindeisen: // Test endpoint1 == endpoint2 => endpoint2 == endpoint3
                // Iteration covers all permutations, so no need to be redundant
                bool const transitive = !equals12 || !equals13 || equals23;

PR #318, opened 2018-02-01
DM-13437: makeImageFromCamera: catch exceptions from retrieving image

PaulPrice:
This allows the 'imageSource' to throw an exception rather than returning
a nonsense input.

### DIFF
--- a/python/lsst/afw/cameraGeom/utils.py
+++ b/python/lsst/afw/cameraGeom/utils.py
@@ -730,7 +730,11 @@ def makeImageFromCamera(camera, detectorNameList=None, background=numpy.nan, buf
 
     boxList = getCcdInCamBBoxList(ccdList, binSize, pixelSize_o, origin)
     for det, bbox in zip(ccdList, boxList):
-        im = imageSource.getCcdImage(det, imageFactory, binSize)[0]
+        try:
+            im = imageSource.getCcdImage(det, imageFactory, binSize)[0]
+        except Exception as e:
+            print("Unable to get image for detector %s: %s" % (det.getName(), e))
+            continue
 
         nQuarter = det.getOrientation().getNQuarter()
         im = afwMath.rotateImageBy90(im, nQuarter)


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -730,7 +730,11 @@ def makeImageFromCamera(camera, detectorNameList=None, background=numpy.nan, buf
 
     boxList = getCcdInCamBBoxList(ccdList, binSize, pixelSize_o, origin)
     for det, bbox in zip(ccdList, boxList):
-        im = imageSource.getCcdImage(det, imageFactory, binSize)[0]
+        try:
+            im = imageSource.getCcdImage(det, imageFactory, binSize)[0]
+        except Exception as exc:
```
RobertLuptonTheGood: Don't we usually write this as `Exception as e:`?  If not, no need to change it.

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -730,7 +730,11 @@ def makeImageFromCamera(camera, detectorNameList=None, background=numpy.nan, buf
...
+            print("Unable to get image for detector %s: %s" % (det.getName(), exc))
```
RobertLuptonTheGood: I assume no logger is available, but shouldn't this go to `sys.stderr`?

python/lsst/afw/cameraGeom/utils.py
PaulPrice: This is often writen as `except Exception as e:`, but using a single-character `e` for a variable name bugs me, and I have used `exc` instead of `e` for quite a while. I believe this is a common pattern in the python world, e.g.:
* https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement
* https://www.python.org/dev/peps/pep-0008/#id51

python/lsst/afw/cameraGeom/utils.py
PaulPrice: This follows the existing pattern. Would you like me to change that here?

PR #319, opened 2018-02-09
DM-13511: fits: extend list of header keywords to ignore

PaulPrice:
There are some keywords that, if we write, it will corrupt the
FITS file because cfitsio takes care of them; these typically
have to do with core FITS concepts (size, column definitions,
compression) that we handle explicitly through cfitsio
interfaces. And the user shouldn't care about them either.
However, there are some keywords that need to leak through
our keyword stripping because afw::table::io needs to see
them in order to build a Schema; we just won't allow those
to be written.

This includes refactoring the 'isKeyIgnored' function to make
the list of keywords more easily extensible.

### DIFF
--- a/src/fits.cc
+++ b/src/fits.cc
@@ -5,6 +5,8 @@
 #include <complex>
 #include <cmath>
 #include <sstream>
+#include <unordered_set>
+#include <unordered_map>
 
 #include "fitsio.h"
 extern "C" {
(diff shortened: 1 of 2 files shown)


src/fits.cc
```diff
@@ -637,10 +667,25 @@ void Fits::forEachKey(HeaderIterationFunctor &functor) {
 
 namespace {
 
-bool isKeyIgnored(std::string const &key) {
-    return key == "SIMPLE" || key == "BITPIX" || key == "EXTEND" || key == "GCOUNT" || key == "PCOUNT" ||
-           key == "XTENSION" || key == "TFIELDS" || key == "BSCALE" || key == "BZERO" ||
-           key.compare(0, 5, "NAXIS") == 0;
+bool isKeyIgnored(std::string const &key, bool write=false) {
+    for (auto const & kk : ignoreKeys) {
```
ktlim: I'd feel a little happier if this was a O(1) hash lookup instead of an O(n) list iteration because this is going to be called for every key in every HDU.  I understand that would be harder for the `Starts` items, although you could hash the first 4 or 5 letters.

ktlim:
The new code is nice — except `std::set` and `std::map` are (at least) O(log n) as opposed to `std::unordered_set` and `std::unordered_map`.

PR #320, opened 2018-02-14
DM-13539: astshim fails to preserve SIP terms for some TAN SIP when writing FITS metadata

r-owen:
The test of SkyWcs.getTanWcs was only passing because TAN-SIP
was not being written out by AST, a problem fixed on this ticket.
For now remove the test for SkyWcs.getTanWcs();
tickets/DM-10765 has a much better test, but it relies on
a function that that is not available on master.


parejkoj:
This looks fine, especially with your comment in the commit about the better test existing in 10765. If you update that commit message to say what the better test is, you're good to merge.

PR #321, opened 2018-02-26
DM-12447: Remove TransformMap.transform, Camera.transform and Detector.transform

r-owen:
and CameraPoint


PaulPrice:
Can you state the reason for the removal in the commit log?

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -68,118 +73,73 @@ def getPupilFactory(self, visitInfo, pupilSize, npix, **kwargs):
...
-            # use camera transform map
-            return self._transformMap.transform(positionList, fromSys, toSys)
-        raise RuntimeError(
-            "Could not find mapping from %s to %s"%(fromSys, toSys))
-
-    def _transformSingleSys(self, cameraPoint, toSys):
-        """!Transform a CameraPoint with a CameraSys to another CameraSys.
+            return det.getTransformMap().getTransform(fromSys, toSys)
+        else:
+            return self.getTransformMap().getTransform(fromSys, toSys)
```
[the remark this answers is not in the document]
r-owen: Your suggested new code does not do the same thing because the `else` clause handles the camera's own transform map, not a detector's transform map.

I suppose we could retrieve the transform map. It will be a bit awkward since we then have to assume the code will never leave the transform map equal to None (which will be true initially, but perhaps not if the code evolves) or check for None even though it should be impossible. That is why I chose to use the current technique.

examples/cameraGeomExample.py
```diff
@@ -15,32 +15,16 @@
...
-# * To have the Camera find a detector, specify the target coordinate system as a CameraSysPrefix
-#   (e.g. cameraGeom.PIXELS). Camera will search for a detector that overlaps the point.
-#   If it finds exactly one detector then it will use that detector, and you can figure
-#   out which detector it used from the detector name in the returned CameraPoint.
-# If it finds no detectors, or more than one detector, then it will raise
-# an exception.
-detPixelsSys = det.makeCameraSys(cameraGeom.PIXELS)
-detPointOnSpecifiedDetector = camera.transform(fieldPoint, detPixelsSys)
-detPointOnFoundDetector = camera.transform(fieldPoint, cameraGeom.PIXELS)
-assert detPointOnFoundDetector.getCameraSys() == detPixelsSys  # same detector
```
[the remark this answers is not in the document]
r-owen: Having `Camera.transform` search for a detector is gone. I don't think it's practical to implement it with `CameraPoint` gone. In my opinion it was never a good idea, as it could be quite slow, which could surprise users. I would much rather have users explicitly search for a detector if necessary.

Note that I found no instances where this feature was used, other than this example.

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -68,118 +73,73 @@ def getPupilFactory(self, visitInfo, pupilSize, npix, **kwargs):
...
         """
-        # first convert to focalPlane since the point may be in another
-        # overlapping detector
-        nativePoint = self._transformSingleSys(
-            cameraPoint, self._nativeCameraSys)
+        # convert `point` to the native coordinate system
+        transform = self._getTransformFromOneTransformMap(cameraSys, self._nativeCameraSys)
+        nativePoint = transform.applyForward(point)
 
         detectorList = []
```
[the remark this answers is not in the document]
r-owen: Because a point *can* be on multiple detectors. I don't believe any instruments presently use that capability, but an imaging instrument with a beamsplitter certainly would.

RobertLuptonTheGood:
What is the functionality that you've removed (the one that you didn't find in the live)?

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -196,31 +156,19 @@ def getTransformMap(self):
         """
         return self._transformMap
 
-    def transform(self, cameraPoint, toSys):
```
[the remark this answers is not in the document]
r-owen: I added that in a subsequent commit.

r-owen:
@RobertLuptonTheGood Having Camera.transform search for a detector (for a camera system prefix).

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -106,12 +106,12 @@ def plotFocalPlane(camera, fieldSizeDeg_x=0, fieldSizeDeg_y=None, dx=0.1, dy=0.1
...
-        # pos is a pair of lsst.afw.geom.Angle
-        # but the units of FIELD_ANGLE are radians
-        posFieldAngle = afwGeom.Point2D(math.radians(pos[0]), math.radians(pos[1]))
-        fieldAngleToFocalPlane = camera.getTransform(FIELD_ANGLE, FOCAL_PLANE)
-        posFocalPlane = fieldAngleToFocalPlane.applyForward(posFieldAngle)
+
+    # compute focal plane positions corresponding to field angles field_gridx, field_gridy
+    posFieldAngleRadiansList = [afwGeom.Point2D(x.asRadians(), y.asRadians())
+                                for x, y in zip(field_gridx, field_gridy)]
+    posFocalPlaneList = camera.transform(posFieldAngleRadiansList, FIELD_ANGLE, FOCAL_PLANE)
```
PaulPrice: Why isn't `std::pair<Angle, Angle>` not a supported endpoint type?

src/cameraGeom/Detector.cc
```diff
@@ -101,6 +101,18 @@ std::shared_ptr<geom::TransformPoint2ToPoint2> Detector::getTransform(FromSysT c
...
+geom::Point2D Detector::transform(geom::Point2D const &point, FromSysT const &fromSys,
+                                  ToSysT const &toSys) const {
+    return _transformMap.transform(point, makeCameraSys(fromSys), makeCameraSys(toSys));
+}
+
+template <typename FromSysT, typename ToSysT>
+std::vector<geom::Point2D> Detector::transform(std::vector<geom::Point2D> const &points,
+                                               FromSysT const &fromSys, ToSysT const &toSys) const {
+    return _transformMap.transform(points, makeCameraSys(fromSys), makeCameraSys(toSys));
+}
```
PaulPrice: I'm surprised these aren't defined in the header, since they're so simple.

src/cameraGeom/Detector.cc
```diff
@@ -141,9 +153,12 @@ void Detector::_init() {
...
 // Explicit instantiations
 //
-#define INSTANTIATE(FROMSYS, TOSYS)                                                                 \
-    template std::shared_ptr<geom::TransformPoint2ToPoint2> Detector::getTransform(FROMSYS const &, \
-                                                                                   TOSYS const &) const;
+#define INSTANTIATE(FROMSYS, TOSYS)                                                                          \
+    template std::shared_ptr<geom::TransformPoint2ToPoint2> Detector::getTransform(FROMSYS const &,          \
+                                                                                   TOSYS const &) const;     \
+    template geom::Point2D Detector::transform(geom::Point2D const &, FROMSYS const &, TOSYS const &) const; \
+    template std::vector<geom::Point2D> Detector::transform(std::vector<geom::Point2D> const &,              \
```
PaulPrice: For the same reason we don't line up equal signs in a list of equations, I suggest not spacing out the backslashes.

tests/test_detector.py
```diff
@@ -105,6 +105,32 @@ def addBadCameraSys(dw):
...
+        """Test the transform method
+        """
+        dw = DetectorWrapper()
+        pixOffset = dw.orientation.getReferencePoint()
+        for xyMM in ((25.6, -31.07), (0, 0), (-1.234e5, 3.123e4)):
+            fpPoint = afwGeom.Point2D(*xyMM)
+            pixPoint = dw.detector.transform(fpPoint, cameraGeom.FOCAL_PLANE, cameraGeom.PIXELS)
+            for i in range(2):
+                self.assertAlmostEqual(
+                    fpPoint[i]/dw.pixelSize[i] + pixOffset[i], pixPoint[i])
```
PaulPrice: Instead of iteration, can't you do this with `Point` and `Extent` arithmetic?

python/lsst/afw/cameraGeom/camera.py
```diff
@@ -173,3 +165,25 @@ def getTransform(self, fromSys, toSys):
         fromSysToNative = self._getTransformFromOneTransformMap(fromSys, self._nativeCameraSys)
         nativeToToSys = self._getTransformFromOneTransformMap(self._nativeCameraSys, toSys)
         return fromSysToNative.then(nativeToToSys)
+
+    def getTransformMap(self):
```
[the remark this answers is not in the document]
r-owen: That is existing functionality (I think github is a bit confused about what changed).

It's a good idea, but I'd rather that be a separate RFC and ticket.

python/lsst/afw/cameraGeom/utils.py
r-owen: I think that adds too much complexity, especially because it also requires adding support for `std::vector<std::pair<Angle, Angle>>` and adding a new Endpoint type. In any case it is clearly out of scope for this ticket.

src/cameraGeom/Detector.cc
r-owen: I prefer to define functions in the .cc file unless there's a good reason to do otherwse (e.g. inlining a method. It reduces clutter in the header saves on compilation time if something changes.

src/cameraGeom/Detector.cc
r-owen: clang-format

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -106,12 +106,12 @@ def plotFocalPlane(camera, fieldSizeDeg_x=0, fieldSizeDeg_y=None, dx=0.1, dy=0.1
...
     pcolors = []
-    for pos in zip(field_gridx, field_gridy):
-        # pos is a pair of lsst.afw.geom.Angle
-        # but the units of FIELD_ANGLE are radians
-        posFieldAngle = afwGeom.Point2D(math.radians(pos[0]), math.radians(pos[1]))
-        fieldAngleToFocalPlane = camera.getTransform(FIELD_ANGLE, FOCAL_PLANE)
-        posFocalPlane = fieldAngleToFocalPlane.applyForward(posFieldAngle)
+
+    # compute focal plane positions corresponding to field angles field_gridx, field_gridy
+    posFieldAngleRadiansList = [afwGeom.Point2D(x.asRadians(), y.asRadians())
```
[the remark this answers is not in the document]
r-owen: I *am* Hungarian.

tests/test_detector.py
r-owen: Interesting idea. I tried `fpPoint/dw.pixelSize + pixOffset` but it was rejected with `TypeError: unsupported operand type(s) for /: 'lsst.afw.geom.coordinates.coordinates.Point2D' and 'lsst.afw.geom.coordinates.coordinates.Extent2D'`

I'll stick to this for now -- it is existing code and known to work.

r-owen:
I rebased my changes

PR #325, opened 2018-03-03
DM-13534: avoid usage of old ndarray PyConverter APIs

TallJimbo:
pybind11's own numpy APIs are now preferred, as that's what the ndarray pybind11 converters use.

### DIFF
--- a/include/lsst/afw/table/python/catalog.h
+++ b/include/lsst/afw/table/python/catalog.h
@@ -52,6 +52,22 @@ ndarray::Array<typename Field<T>::Value const, 1, 0> _getArrayFromCatalog(
     return out;
 }
 
+// Specialization of the above for Angle: have to return a double array (in
+// radians), since NumPy arrays can't hold Angles.
+template <typename Record>
+ndarray::Array<double const, 1, 0> _getArrayFromCatalog(
+    CatalogT<Record> const& catalog,  ///< Catalog
+    Key<Angle> const& key  ///< Key to column to extract
+) {
+    ndarray::Array<double, 1, 0> out = ndarray::allocate(catalog.size());
+    auto outIter = out.begin();
+    auto inIter = catalog.begin();
+    for (; inIter != catalog.end(); ++inIter, ++outIter) {
+        *outIter = inIter->get(key).asRadians();
+    }
+    return out;
+}
+
 
 /**
 Declare field-type-specific overloaded catalog member functions for one field type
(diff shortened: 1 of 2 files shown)


include/lsst/afw/table/python/catalog.h
```diff
@@ -39,11 +39,11 @@ using PyCatalog = pybind11::class_<CatalogT<Record>, std::shared_ptr<CatalogT<Re
 
 /// Extract a column from a potentially non-contiguous Catalog
 template <typename T, typename Record>
-ndarray::Array<typename Field<T>::Value const, 1, 0> _getArrayFromCatalog(
+ndarray::Array<typename Field<T>::Value const, 1, 1> _getArrayFromCatalog(
     CatalogT<Record> const& catalog,  ///< Catalog
     Key<T> const& key  ///< Key to column to extract
 ) {
-    ndarray::Array<typename Field<T>::Value, 1, 0> out = ndarray::allocate(catalog.size());
+    ndarray::Array<typename Field<T>::Value, 1, 1> out = ndarray::allocate(catalog.size());
```
pschella: How many more of these are out there? And would `auto` work nicely for them to prevent such issues in the future?

python/lsst/afw/table/baseColumnView/baseColumnView.cc
```diff
@@ -94,12 +81,26 @@ static void declareBaseColumnView(py::module &mod) {
...
     // do not have the same memory model as ours.
-    declareBaseColumnViewOverloads<lsst::afw::geom::Angle>(cls);
     declareBaseColumnViewArrayOverloads<std::uint8_t>(cls);
     declareBaseColumnViewArrayOverloads<std::uint16_t>(cls);
     declareBaseColumnViewArrayOverloads<int>(cls);
     declareBaseColumnViewArrayOverloads<float>(cls);
     declareBaseColumnViewArrayOverloads<double>(cls);
+    // Angle requires custom wrappers, because ndarray doesn't
+    // recognize it natively; we just return a double view
+    // (e.g. radians).
```
pschella: This is breaking the encapsulation of `Angle` a bit right? Or does the API for `getData` specify that the values are in radians?

include/lsst/afw/table/python/catalog.h
TallJimbo: These are not a big deal; it basically boils down to whether a for loop has `n += s` (where `s == 1`) instead of just `++n`.  I just happened to notice it here.

`auto` won't help at all and in fact would hurt quite a bit, because `ndarray::allocate` doesn't return an `Array` and we're relying on implicit conversions in an expression-template kind of way.

python/lsst/afw/table/baseColumnView/baseColumnView.cc
TallJimbo: Yes, it's breaking encapsulation, but it was broken before; I'm just breaking it differently now.

PR #327, opened 2018-03-09
DM-13744: Run clang-tidy checks and cleanup on afw

pschella:
I ran a subset of the available checks. The full list of possible checks can be found at https://clang.llvm.org/extra/clang-tidy/checks/list.html. Each check was run independently and only those that resulted in changes are listed as a separate commit (that can be easily removed if deemed too intrusive).

### DIFF
--- a/src/detection/Footprint.cc
+++ b/src/detection/Footprint.cc
@@ -21,6 +21,8 @@
  * see <https://www.lsstcorp.org/LegalNotices/>.
  */
 
+#include <memory>
+
 #include "lsst/afw/detection/Footprint.h"
 #include "lsst/afw/table/io/CatalogVector.h"
 #include "lsst/afw/table/io/OutputArchive.h"
@@ -289,7 +291,7 @@ std::unique_ptr<Footprint> Footprint::readSpanSet(afw::table::BaseCatalog const&
         }
         loadedSpanSet = std::make_shared<geom::SpanSet>(std::move(tempVec));
     }
-    auto loadedFootprint = std::unique_ptr<Footprint>(new Footprint(loadedSpanSet));
+    auto loadedFootprint = std::make_unique<Footprint>(loadedSpanSet);
     return loadedFootprint;
 }
 
--- a/src/detection/Footprint.cc
+++ b/src/detection/Footprint.cc
@@ -200,8 +200,8 @@ public:
     }
 
     // No copying
-    LegacyFootprintPersistenceHelper(const LegacyFootprintPersistenceHelper&) = delete;
-    LegacyFootprintPersistenceHelper& operator=(const LegacyFootprintPersistenceHelper&) = delete;
+    LegacyFootprintPersistenceHelper(LegacyFootprintPersistenceHelper const &) = delete;
+    LegacyFootprintPersistenceHelper& operator=(LegacyFootprintPersistenceHelper const &) = delete;
 
     // No moving
     LegacyFootprintPersistenceHelper(LegacyFootprintPersistenceHelper&&) = delete;
--- a/src/detection/Footprint.cc
+++ b/src/detection/Footprint.cc
@@ -233,7 +233,7 @@ std::pair<afw::table::Schema&, table::Key<int>&> spanSetPersistenceHelper() {
 
 class FootprintFactory : public table::io::PersistableFactory {
 public:
-    virtual std::shared_ptr<afw::table::io::Persistable> read(
+    std::shared_ptr<afw::table::io::Persistable> read(
             afw::table::io::InputArchive const& archive,
             afw::table::io::CatalogVector const& catalogs) const override {
         // Verify there are two catalogs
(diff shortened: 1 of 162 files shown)


src/detection/FootprintSet.cc
```diff
@@ -872,7 +872,7 @@ enum DIRECTION {
 template <typename MaskPixelT>
 class Startspan {
 public:
-    Startspan(geom::Span const *span, image::Mask<MaskPixelT> *mask, DIRECTION const dir);
+    Startspan(geom::Span const *span, image::Mask<MaskPixelT> *mask, DIRECTION dir);
```
kfindeisen: Is this actually discouraged in C++? I didn't know that. :thinking:

src/detection/Footprint.cc
```diff
@@ -219,11 +219,11 @@ class LegacyFootprintPersistenceHelper {
...
 std::pair<afw::table::Schema&, table::Key<int>&> spanSetPersistenceHelper() {
     static afw::table::Schema spanSetIdSchema;
-    static int initialize = true;
+    static int initialize = 1;
     static table::Key<int> idKey;
-    if (initialize) {
+    if (initialize != 0) {
         idKey = spanSetIdSchema.addField<int>("id", "id of the SpanSet catalog");
-        initialize = false;
+        initialize = 0;
```
kfindeisen: Why isn't `initialize` a boolean? Did `clang-tidy` detect a different usage somewhere?

src/detection/Footprint.cc
```diff
@@ -296,16 +296,16 @@ std::unique_ptr<Footprint> Footprint::readSpanSet(afw::table::BaseCatalog const&
...
         mapper.addMinimalSchema(PeakTable::makeMinimalSchema());
         afw::table::Key<float> oldX = peakCat.getSchema()["x"];
         afw::table::Key<float> oldY = peakCat.getSchema()["y"];
         afw::table::Key<float> oldPeakValue = peakCat.getSchema()["value"];
-        mapper.addMapping(oldX, "f.x");
-        mapper.addMapping(oldY, "f.y");
-        mapper.addMapping(oldPeakValue, "peakValue");
+        mapper.addMapping(oldX, true);
+        mapper.addMapping(oldY, true);
+        mapper.addMapping(oldPeakValue, true);
```
kfindeisen: This one really seems off...

src/fits.cc
```diff
@@ -52,7 +52,7 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
         auto name = (lastPeriod == std::string::npos) ? fullName : fullName.substr(lastPeriod + 1);
         std::type_info const &type = metadata.typeOf(name);
 
-        std::string out = "";
+        std::string out;
```
kfindeisen: I'd suggest keeping these in. While the default values for standard C++ types are fairly intuitive, I don't see any harm in being explicit.

src/detection/FootprintSet.cc
pschella: It is not, but isn't part of the signature so possibly confuses people. I don't know what we prefer though, so we may not apply this check.

src/detection/Footprint.cc
pschella: I don't know why it isn't a boolean. The check detected that it was an `int` and thus replaced the initialization. But we probably want to see why it isn't `bool` in the first place.

src/detection/Footprint.cc
pschella: Well, could be. Or `clang-tidy` detected that it would always convert to `bool` and call that overload? We should check.

src/fits.cc
pschella: Sounds good.

src/detection/Footprint.cc
kfindeisen: Ah, crap, because a C-style string is actually a pointer. :man_facepalming: I don't suppose `clang-tidy` has a detector for ambiguous or surprising overload resolution?

I suppose we could try manually converting these to C++14 string literals, and seeing if that makes `clang-tidy` happy.

src/detection/FootprintSet.cc
TallJimbo: I do occasionally find it useful to use a by-value const argument in order for the implementation to use the argument as a const variable.  And, of course, I dislike the fact that this is exposed in the signature.

So this feels like a situation where we should have a convention on whether this is allowed or not (I'm pretty ambivalent), and only do an automatic conversion if the convention discussion converges on prohibiting it.

src/detection/Footprint.cc
pschella: Yeah, I believe that is what is happening. We expect it to call `Key<T> addMapping(Key<T> const& inputKey, std::string const& outputName, bool doReplace = true);`, but instead it calls `Key<T> addMapping(Key<T> const& inputKey, bool doReplace = false);` because it prefers to cast to `bool` rather than create a `std::string` temporary.

timj:
Closing this pull request since there is no way the PR as is can be merged.

PR #328, opened 2018-03-09
DM-13750: Move BaseRecord.__str__ implementation to C++.

TallJimbo:
Old implementation was very inefficient, even in Python, but we can make it even faster by moving it to C++.

### DIFF
--- a/src/table/BaseRecord.cc
+++ b/src/table/BaseRecord.cc
@@ -1,6 +1,7 @@
 // -*- lsst-c++ -*-
 
 #include <cstring>
+#include <iostream>
 
 #include "lsst/pex/exceptions.h"
 #include "lsst/afw/table/BaseRecord.h"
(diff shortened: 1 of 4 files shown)


src/table/BaseRecord.cc
```diff
@@ -90,6 +91,38 @@ void BaseRecord::assign(BaseRecord const& other, SchemaMapper const& mapper) {
...
+private:
+    BaseRecord const * _record;
+    std::ostream * _os;
+};
+
+} // anonymous
+
+
+void BaseRecord::_stream(std::ostream & os) const {
+    PrintField functor(this, &os);
```
natelust: Personally I would find this much clearer and easier to read if you wrote this as a lambda with a capture on this and os. It would put the function right next to where it was used, and reduce many lines in this file. This seems exactly to be the kind of thing that lambdas were made for. It would also negate the need to get the address of os to pass into the functor, just to keep the size of the functor small, you could simply reference capture os, and have it as a reference all the way down.

Also, if a functor is used, please consider reverse the arguments of BaseRecord and std::ostream to put them in the same order as operator<<. I dont know if it is standard to pass this first, and if so fine, but it seems weirdly asymmetric.

natelust: Personally I would find this much clearer and easier to read if you wrote this as a lambda with a capture on this and os. It would put the function right next to where it was used, and reduce many lines in this file. This seems exactly to be the kind of thing that lambdas were made for. It would also negate the need to get the address of os to pass into the functor, just to keep the size of the functor small, you could simply reference capture os, and have it as a reference all the way down.

Also, if a functor is used, please consider reverse the arguments of BaseRecord and std::ostream to put them in the same order as operator<<. I dont know if it is standard to pass this first, and if so fine, but it seems weirdly asymmetric.

TallJimbo: Good idea!

I was thinking in C++11, which does not permit "universal lambdas" (lambdas with templated function-call operators) instead of C++14, which does.

PR #329, opened 2018-03-13
DM-13787: afw::table::SchemaMapper::addMapping calls wrong overload with string literal

pschella:
Prior to this fix calling addMapping with a string literal as the
second argument caused the wrong (bool) overload to be called.
Now a compile-time error is generated instead.


timj:
Is there a reason this code hasn't been merged?

pschella:
Yes, I forgot about it.

PR #332, opened 2018-03-16
DM-11162: Replace all use of Coord and subclasses with SpherePoint

r-owen:


python/lsst/afw/coord/__init__.py
```diff
@@ -20,6 +20,8 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
+import lsst.afw.geom
 
-from .coordLib import *
-from .utils import *
+from ._observatory import *
+from ._weather import *
+from ._weather import *
```
rearmstr: repeated line

src/geom/Endpoint.cc
```diff
@@ -197,26 +197,26 @@ ndarray::Array<double, 2, 2> IcrsCoordEndpoint::dataFromArray(Array const& arr)
     return data;
 }
 
-coord::IcrsCoord IcrsCoordEndpoint::pointFromData(std::vector<double> const& data) const {
+SpherePoint SpherePointEndpoint::pointFromData(std::vector<double> const& data) const {
     this->_assertNAxes(this->_getNAxes(data));
-    return coord::IcrsCoord(data[0] * radians, data[1] * radians);
+    return SpherePoint(data[0] * radians, data[1] * radians);
```
rearmstr: Any reason not to use the ```SpherePoint(data[0], data[1], radians)``` constructor here and below?

src/geom/detail/frameSetUtils.cc
```diff
@@ -31,10 +31,10 @@
 #include "astshim.h"
 
 #include "lsst/afw/formatters/Utils.h"
-#include "lsst/afw/coord/Coord.h"
 #include "lsst/afw/fits.h"
 #include "lsst/afw/geom/Angle.h"
 #include "lsst/afw/geom/Point.h"
+#include "lsst/afw/geom/SpherePoint.h"
```
rearmstr: Does this file need to be included here?

src/geom/Endpoint.cc
r-owen: Good idea. I'll do it.

src/geom/detail/frameSetUtils.cc
r-owen: No. Good catch.

PR #334, opened 2018-03-23
DM-13854: extend cache in Psf

PaulPrice:
Larger cache makes things go faster.

### DIFF
--- a/include/lsst/afw/detection/Psf.h
+++ b/include/lsst/afw/detection/Psf.h
@@ -29,6 +29,7 @@
 #include <memory>
 
 #include "lsst/daf/base.h"
+#include "lsst/utils/CacheFwd.h"
 #include "lsst/afw/geom/ellipses/Quadrupole.h"
 #include "lsst/afw/math/Kernel.h"
 #include "lsst/afw/image/Color.h"
@@ -37,6 +38,12 @@
 namespace lsst {
 namespace afw {
 namespace detection {
+namespace detail {
+
+/// Key for caching PSFs with lsst::utils::Cache
+struct PsfCacheKey;
+
+} // namespace detail
 
 class PsfFormatter;
 
@@ -86,12 +93,12 @@ public:
                       */
     };
 
-    Psf(Psf const&) = default;
-    Psf(Psf&&) = default;
+    Psf(Psf const&);
     Psf& operator=(Psf const&) = delete;
     Psf& operator=(Psf&&) = delete;
 
-    virtual ~Psf() = default;
+    Psf(Psf&&);
+    virtual ~Psf();
 
     /**
      *  Polymorphic deep-copy.
@@ -254,14 +261,27 @@ public:
                                                       std::string const& warpAlgorithm = "lanczos5",
                                                       unsigned int warpBuffer = 5);
 
+    /** Return the capacity of the caches
+     *
+     * Both the image and kernel image caches have the same capacity.
+     */
+    std::size_t getCacheCapacity() const;
+
+    /** Set the capacity of the caches
+     *
+     * Both the image and kernel image caches will be set to this capacity.
+     */
+    void setCacheCapacity(std::size_t capacity);
+
 protected:
     /**
      *  Main constructor for subclasses.
      *
      *  @param[in] isFixed  Should be true for Psf for which doComputeKernelImage always returns
      *                      the same image, regardless of color or position arguments.
+     *  @param[in] capacity  Capacity of the caches.
      */
-    explicit Psf(bool isFixed = false);
+    explicit Psf(bool isFixed = false, std::size_t capacity=100);
 
 private:
     //@{
@@ -284,12 +304,9 @@ private:
     //@}
 
     bool const _isFixed;
-    mutable std::shared_ptr<Image> _cachedImage;
-    mutable std::shared_ptr<Image> _cachedKernelImage;
-    mutable image::Color _cachedImageColor;
-    mutable image::Color _cachedKernelImageColor;
-    mutable geom::Point2D _cachedImagePosition;
-    mutable geom::Point2D _cachedKernelImagePosition;
+    using PsfCache = utils::Cache<detail::PsfCacheKey, std::shared_ptr<Image>>;
+    std::unique_ptr<PsfCache> _imageCache;
+    std::unique_ptr<PsfCache> _kernelImageCache;
 
     LSST_PERSIST_FORMATTER(PsfFormatter)
 };
--- a/src/detection/Psf.cc
+++ b/src/detection/Psf.cc
@@ -2,9 +2,53 @@
 #include <limits>
 #include <typeinfo>
 #include <cmath>
+#include <memory>
 
+#include "lsst/utils/Cache.h"
+#include "lsst/utils/hashCombine.h"
 #include "lsst/afw/detection/Psf.h"
 #include "lsst/afw/math/offsetImage.h"
+
+namespace lsst::afw::detection::detail {
+
+// Key for caching PSFs with lsst::utils::Cache
+//
+// We cache PSFs by their x,y position. Although there are placeholders
+// in the `Psf` class and here for `image::Color`, these are not used
+// in the cache because `image::Color` is not currently well-defined
+// or used.
+struct PsfCacheKey {
+    geom::Point2D const position;
+    image::Color const color;
+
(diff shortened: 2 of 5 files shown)


include/lsst/afw/detection/Psf.h
```diff
@@ -283,13 +298,29 @@ class Psf : public daf::base::Citizen,
...
+
     bool const _isFixed;
-    mutable std::shared_ptr<Image> _cachedImage;
-    mutable std::shared_ptr<Image> _cachedKernelImage;
-    mutable image::Color _cachedImageColor;
-    mutable image::Color _cachedKernelImageColor;
-    mutable geom::Point2D _cachedImagePosition;
-    mutable geom::Point2D _cachedKernelImagePosition;
+    mutable utils::Cache<CacheKey, std::shared_ptr<Image>> _imageCache;
+    mutable utils::Cache<CacheKey, std::shared_ptr<Image>> _kernelImageCache;
```
TallJimbo: Maybe `shared_ptr<Image const>`, because it can be?

include/lsst/afw/detection/Psf.h
```diff
@@ -29,6 +29,7 @@
 #include <memory>
 
 #include "lsst/daf/base.h"
+#include "lsst/utils/Cache.h"
```
TallJimbo: As discussed on the `utils` ticket, I'd prefer to forward-declare `Cache`, have `Psf` hold it by `unique_ptr`, and move this include to the .cc file.

include/lsst/afw/detection/Psf.h
```diff
@@ -254,14 +255,28 @@ class Psf : public daf::base::Citizen,
...
                                                       unsigned int warpBuffer = 5);
 
+    /// Return the capacity of the caches
+    ///
+    /// Both the image and kernel image caches have the same capacity.
+    std::size_t getCacheCapacity() const { return _kernelImageCache.capacity(); }
+
+    /// Set the capacity of the caches
+    ///
+    /// Both the image and kernel image caches will be set to this capacity.
```
TallJimbo: Use `/** */` for multi-line Doxygen comments.

include/lsst/afw/detection/Psf.h
```diff
@@ -283,13 +298,29 @@ class Psf : public daf::base::Citizen,
...
+
+        bool operator==(CacheKey const& other) const {
+            return position == other.position;  // Currently don't care about color
+        }
+
+        friend std::ostream & operator<<(std::ostream &os, CacheKey const &key) {
+            return os << key.position;
+        }
+    };
+
```
TallJimbo: Hopefully this can be moved to the .cc file, too.

python/lsst/afw/geom/coordinates/coordinates.cc
```diff
@@ -364,6 +366,8 @@ PyPoint<T, N> declarePoint(py::module &mod, std::string const &suffix) {
     cls.def("__ne__", [](Point<T, N> const &self, Point<T, N> const &other) { return self != other; },
             py::is_operator());
 
+    cls.def("__hash__", [](Point<T, N> const &self) { return boost::hash<Point<T, N>>()(self); });
+
```
TallJimbo: I'm not sure it's a good idea to add this to Python; Points are not immutable and hence putting them in dicts could definitely confuse the dict.

include/lsst/afw/detection/Psf.h
PaulPrice: Pulled it out of the class, into a preceding `namespace detail`. But it does mess up the top of the file a bit, especially since I have to get into `namespace std` to provide the `std::hash` support (it's more annoying to work with than `boost::hash`).

include/lsst/afw/detection/Psf.h
PaulPrice: Don't think I can, unless I cast away the same `const` in `Psf::computeImage`.

include/lsst/afw/detection/Psf.h
```diff
@@ -254,14 +299,30 @@ class Psf : public daf::base::Citizen,
...
+    std::size_t getCacheCapacity() const { return _kernelImageCache->capacity(); }
+
+    /** Set the capacity of the caches
+     *
+     * Both the image and kernel image caches will be set to this capacity.
+     */
+    void setCacheCapacity(std::size_t capacity) {
+        _imageCache->reserve(capacity);
+        _kernelImageCache->reserve(capacity);
+    }
```
TallJimbo: If you define `setCacheCapacity` and `getCacheCapacity` (and `~Psf()`, I think), you should no longer need to include `Cache.h` in other source files that use `Psf`.  I'm not totally sure that's enough, but the goal here is to make sure that anything that uses `Cache` as more than an opaque pointer is in `Psf.cc`, and that will make it unnecessary to include `Cache.h` anywhere but `Psf.cc`.

include/lsst/afw/detection/Psf.h
```diff
@@ -29,11 +29,56 @@
...
+      : position(position_), color(color_) {}
+
+    bool operator==(PsfCacheKey const& other) const {
+        return position == other.position;  // Currently don't care about color
+    }
+
+    friend std::ostream & operator<<(std::ostream &os, PsfCacheKey const &key) {
+        return os << key.position;
+    }
+};
```
TallJimbo: You'll need a forward declaration of this class in the header for the typedef later, but the rest can also go in `Psf.cc`.

include/lsst/afw/detection/Psf.h
```diff
@@ -283,13 +344,11 @@ class Psf : public daf::base::Citizen,
...
 
+
     bool const _isFixed;
-    mutable std::shared_ptr<Image> _cachedImage;
-    mutable std::shared_ptr<Image> _cachedKernelImage;
-    mutable image::Color _cachedImageColor;
-    mutable image::Color _cachedKernelImageColor;
-    mutable geom::Point2D _cachedImagePosition;
-    mutable geom::Point2D _cachedKernelImagePosition;
+    typedef utils::Cache<detail::PsfCacheKey, std::shared_ptr<Image>> PsfCache;
```
TallJimbo: `using PsfCache = utils::Cache<detail::PsfCacheKey, std::shared_ptr<Image>>` is preferred in C++11.

include/lsst/afw/detection/Psf.h
```diff
@@ -283,13 +344,11 @@ class Psf : public daf::base::Citizen,
...
+    mutable std::shared_ptr<PsfCache> _imageCache;
+    mutable std::shared_ptr<PsfCache> _kernelImageCache;
```
TallJimbo: I believe these no longer need to be `mutable`; the extra layer of indirection does that for you.  Since `Psf` is already noncopyable and immutable, `std::unique_ptr` should be used instead of `std::shared_ptr` as well.

src/detection/Psf.cc
```diff
@@ -12,19 +13,16 @@ namespace detection {
...
 bool isPointNull(geom::Point2D const &p) { return std::isnan(p.getX()) && std::isnan(p.getY()); }
 
 }  // anonymous
 
-Psf::Psf(bool isFixed) : daf::base::Citizen(typeid(this)), _isFixed(isFixed) {}
+Psf::Psf(bool isFixed, std::size_t capacity)
+  : daf::base::Citizen(typeid(this)),
+  _isFixed(isFixed),
+  _imageCache(new PsfCache(capacity)),
+  _kernelImageCache(new PsfCache(capacity))
```
TallJimbo: Best practice is to use `std::make_shared` to initialize smart pointers when you're constructing the pointee at the same time (or, if you switch to `std::unique_ptr`, `std::make_unique`).

include/lsst/afw/detection.h
```diff
@@ -26,6 +26,8 @@
 #ifndef LSST_AFW_DETECTION_H
 #define LSST_AFW_DETECTION_H
 
+#include "lsst/utils/Cache.h" // for afw::detection::Psf
+
```
TallJimbo: This should be removable now.

src/geom/Point.cc
```diff
@@ -21,6 +21,7 @@
  */
 
 #include <cmath>
+#include <boost/functional/hash.hpp>  // boost::hash
```
TallJimbo: Use `""`, not `<>`.

PR #337, opened 2018-03-29
Add basic sphinx/numpydoc support

parejkoj:


parejkoj:
I understand adding the relevant things to `manifest.yaml`, but do I have to make the separate directories and put an `index.rst` in each of them, each with its own `automodapi`? Or do I just use the one `lsst.afw/index.rst` and (eventually) put the relevant `automodapi` statements in it?

jonathansick:
Sorry I missed your question. Each public namespace has its own directory, and the `index.rst` in it has the same format as the `lsst.afw/index.rst` you had, but now the titles are customized to e.g. `lsst.afw.cameraGeom` and the `automodapi` directive only points to that module e.g.

```
.. automodapi:: lsst.afw.cameraGeom
```

And you'll delete the `/lsst.afw/index.rst` file you currently have.

The idea is that, from a Python API perspective, `lsst.afw` isn't a useful topic. But `lsst.afw.submodule` are, so that's why the module homepages are one namespace level deeper for `afw`.

jonathansick:
For background, the technical role of `manifest.yaml` is to tell the build process of the existence of the package and module documentation directories.

parejkoj:
Ok, I've updated to use the submodule directories, and updated the manifest and the `doc/index.rst` and `doc/afw/index.rst`, which seem to also need the submodules listed (I got sphinx-build errors otherwise).

jonathansick:
Thanks, looks great!

PR #342, opened 2018-04-05
tickets/DM-13834: Add minimum and maximum wavelength to FilterProperty.

isullivan:
This adds additional FilterProperties `lambdaMin` and `lambdaMax`, and should work in a backwards-compatible way. But, I am not confident working with C++, so I would appreciate an extra-careful look at my changes to make sure I haven't messed something up.

### DIFF
--- a/include/lsst/afw/image/Filter.h
+++ b/include/lsst/afw/image/Filter.h
@@ -93,11 +93,11 @@ public:
      */
     double getLambdaEff() const { return _lambdaEff; }
     /**
-     * Return the filter's minimum wavelength (nm)
+     * Return the filter's minimum wavelength (nm) where the transmission is above 1% of the maximum.
      */
     double getLambdaMin() const { return _lambdaMin; }
     /**
-     * Return the filter's maximum wavelength (nm)
+     * Return the filter's maximum wavelength (nm) where the transmission is above 1% of the maximum.
      */
     double getLambdaMax() const { return _lambdaMax; }
     /**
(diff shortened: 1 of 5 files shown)


tests/test_color.py
```diff
@@ -246,7 +246,7 @@ def setUp(self):
                             lambdaEff in wavelengths.items() if name == "g"][0]  # for tests
 
     def defineFilterProperty(self, name, lambdaEff, force=False):
-        return afwImage.FilterProperty(name, lambdaEff, force)
+        return afwImage.FilterProperty(name, lambdaEff, force=force)
```
ktlim: Would be nice to test the new functionality as well as the old.

include/lsst/afw/image/Filter.h
```diff
@@ -90,6 +92,14 @@ class FilterProperty {
      * Return the filter's effective wavelength (nm)
      */
     double getLambdaEff() const { return _lambdaEff; }
+    /**
+     * Return the filter's minimum wavelength (nm)
```
jmeyers314: Is it possible to be more precise here?

For example, are these the values beyond which < X% of the total throughput exists?  or where the fractional throughput first drops below X% of the peak?  After all, if we're being really pedantic, I think we expect some light leaks (at the ~10^-5 of peak throughput level) across pretty much the entire visible range from 350nm to 1100nm (but obviously it isn't useful to set lambdaMin=350 and lambdaMax=1100 for every filter).  If you do have a concrete procedure for calculating these, or even if the answer is "I eyeballed it" it would be good to document that.

PR #344, opened 2018-04-07
DM-13898: Remove the XYTransform classes

r-owen:


include/lsst/afw/geom/SpanSet.h
```diff
@@ -36,6 +36,7 @@
 #include "lsst/afw/image/Mask.h"
 #include "lsst/afw/table/io/Persistable.h"
 #include "lsst/afw/geom/ellipses/Ellipse.h"
+#include "lsst/afw/geom/ellipses/Quadrupole.h"
```
fred3m: I see the use of `ellipses::Quadrupole` below, but I don't see anything you did that would require adding this header. Was this broken before?

r-owen: This include should have been there all along. I'm not sure how `Quadrupole` was getting included before, but I assume it was via some indirect path involving `XYTransform`.

PR #348, opened 2018-04-25
DM-13905: Change SpherePoint.getVector to return a sphgeom UnitVector3d

r-owen:


include/lsst/afw/geom/SpherePoint.h
```diff
@@ -104,7 +106,28 @@ class SpherePoint final {
...
+    explicit SpherePoint(sphgeom::UnitVector3d const& unitVector);
+
+    /**
+     * Construct a SpherePoint from a sphgeom::LonLat.
+     *
+     * @param lonLat The lonLat
+     *
+     * @exceptsafe Provides strong exception guarantee.
+     */
+    SpherePoint(sphgeom::LonLat const &lonLat);
```
kfindeisen: I'm a bit concerned by the growing number of `SpherePoint` constructors -- we're up to six, not including the two copy-constructors. The sheer number of combinations is going to make the `SpherePoint` API hard to learn; why not keep a small number of "obviously different" constructors and let the user convert as necessary?

tests/test_spherePoint.cc
```diff
@@ -44,7 +45,7 @@ namespace geom {
  * is an identical but independent copy.
  */
 BOOST_AUTO_TEST_CASE(SpherePointCopyResult, *boost::unit_test::tolerance(1e-14)) {
-    SpherePoint original(Point3D(0.34, -1.2, 0.97));
+    SpherePoint original(sphgeom::Vector3d(0.34, -1.2, 0.97));
```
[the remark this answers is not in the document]
r-owen: I'd rather hold off adding anything to lsst::afw::geom until we at least move most of it into lsst::geom. Even then I'm a bit leery. I guess I can see how it makes it easier to construct SpherePoints from vectors.

include/lsst/afw/geom/SpherePoint.h
r-owen: I share your concern but I am not comfortable removing any.

The `LonLat` constructor makes `LonLat` implicitly convertible to `SpherePoint` in a way that avoids adding dependencies to `sphgeom`. This seems a useful feature as we integrate more of our code with `sphgeom`. That said, there is too much overlap, which is a wart.

The `sphgeom::UnitVector3d` constructor seems like clutter to me, but it avoids the need to normalize the argument so I reluctantly think it worth doing. It also seems a logical constructor to have.

The `sphgeom::Vector3d` constructor is one I would be loath to give up, as we make heavy use of it.

Which ones would you give up?

kfindeisen: While it is an option, I'm a bit wary of relying on factory methods as replacements for constructors because my experience (particularly with Gaia) is that most astronomers find them unintuitive. A more generally applicable objection is that documentation systems don't highlight factory methods in any way, so they're harder to find (basically, you've replaced the problem of deciding which constructor you want with the problem of deciding which methods you want).

As for what could be trimmed, `(Angle, Angle)`, `(double, double, AngleUnit)`, and `(LonLat)` convey basically the same information; I'd be inclined to remove one of the first two. I don't have a good sense for how valuable implicit conversions from `sphgeom`will be, but having a dedicated constructor for `LonLat` at least prevents users from mixing up longitude and latitude if that's a common conversion.

Likewise, `(Vector3d)` and `(UnitVector3d)` are redundant. We could keep just `(Vector3d)` on the grounds that it's more frequently used. I don't think the minor efficiency advantage of `(UnitVector3d)` is worth the clutter. If we later add implicit conversion of `UnitVector3d` to `Vector3d` (which seems a reasonable operation), then an overload that takes both `(Vector3d)` and `(UnitVector3d)` would be ambiguous anyway.

I don't like the zero-argument constructor, but I think that discussion is already closed.

r-owen: I will note one point in favor of `fromVector`: it would be a natural way to add a `SpherePoint` constructor that takes 3 floats: `(x, y, z)`. This would be one of the most heavily used constructors in Python, second only to `(lonAsFloat, latAsFloat, angleUnit)`, and it would save the bother of doing this: `SpherePoint(sphgeom.Vector3d(*values))`. Indeed in Python I suspect `SpherePoint(Vector3d)` would hardly ever be used if `(x, y, z)` was available.

`SpherePoint(float, float, angle)` is far too heavily used to consider removing. `SpherePoint(angle, angle)` is worth considering but I am against doing so because it feels like the most obvious constructor,and because it would greatly expand the scope of this ticket for what I see as very little gain.

I propose the following:
- Remove `SpherePoint(unitVector3d)`, as per @kfindeisen's suggestion
- Remove `SpherePoint(LonLat)` for now and make the implicit conversion one way (SpherePoint to LonLat). We can always add it if and when we find a need. We have no identified use case, though it may well become a need when we use `sphgeom` package more heavily.

TallJimbo: I think I'd cut that down to just removing `SpherePoint(unitVector3d)`; when a naive user in the future asks, "why is `SpherePoint` convertible to `LonLat` but not vice versa", I think "we felt `SpherePoint` had too many constructors` will feel like a poor excuse.

kfindeisen: After discussion with @r-owen, we've agreed that keeping `SpherePoint(LonLat)` is best. He pointed out that creating a `SpherePoint` from a `LonLat` using the vector- or angle-based constructors is actually pretty hard because `sphgeom` doesn't use `afw::geom` types anywhere.

PR #349, opened 2018-04-26
DM-14198: Modernize python in afw

r-owen:


python/lsst/gdb/afw/printers_oldgdb.py
```diff
@@ -2,10 +2,6 @@
 Code that works with gdb 7.1's python pretty printing.  When gdb >= 7.2 is widely available this
```
timj: Do we still need this file?

r-owen: No. I'll delete it.

r-owen:
Good suggestion to remove inheritance from object. I did that.

PR #350, opened 2018-05-02
DM-14275: The distortion in test_wcsUtils.py testDistortion is unreasonable

r-owen:


src/geom/transformFactory.cc
```diff
@@ -124,7 +124,7 @@ ast::PolyMap makeOneDDistortion(std::vector<double> const &coeffs) {
         polyCoeffs[i - 1][2] = i;
     }
 
-    return ast::PolyMap(polyCoeffs, 1, "IterInverse=1, TolInverse=1e-8, NIterInverse=20");
+    return ast::PolyMap(polyCoeffs, 1, "IterInverse=1, TolInverse=1e-8, NIterInverse=30");
```
parejkoj: I assume this runs 30 steps every time it computes the inverse? How much slower is it?

tests/test_wcsUtils.py
```diff
@@ -99,25 +99,27 @@ def testNoDistortion(self):
...
 
     def testDistortion(self):
         """Test makeDistortedTanWcs using a non-affine transform for pixelToFocalPlane
         """
-        # Compute a distorted wcs that matches self.tanWcs at the center
-        # of the field
-        focalPlaneToFieldAngle = afwGeom.makeRadialTransform([0.0, self.radPerMm, 0.0, self.radPerMm])
+        # Compute a distorted wcs that matches self.tanWcs at the center of the field;
+        # the amount of distortion is 10s of pixels over the detector
+        fieldAngleToFocalPlane = afwGeom.makeRadialTransform([0.0, 1/self.radPerMm, 0.0, 1000/self.radPerMm])
```
parejkoj: Ah the joys of picking a "reasonable" fake model. This seems like a better choice.

r-owen: It gives much more reasonable values, and is very similar to what LSST uses (the factor of 1000 is not in the LSST model, but the LSST model only shows significant distortion at much larger focal plane radius; I tuned it to get the distortion I wanted over the range of values this test uses).

src/geom/transformFactory.cc
r-owen: `NIterInverse` is an upper limit. For most transforms the solution will reach the tolerance specified by `TolInverse` long before it hits the limit on iterations and stop. However, our unit tests have found several examples where 20 is not enough and so the results are silently lousy. In each such case 30 has been more than enough. It's not a systematic approach, but 30 is clearly better than 20 for accuracy.

r-owen:
The transform was backwards before because I forgot to check transformed values to see if they looked reasonable. The test passed, so I didn't worry about it. I suppose I could have left it alone, but it sure makes debugging easier if the code does what it says it does.

PR #351, opened 2018-05-08
DM-14353: remove or fix bad C++ usage in FootprintSet.cc

TallJimbo:
There are a lot of stylistic things that I have not touched; I'm just trying to fix code that is incorrect, likely to emit warnings, or bad in a way that could affect performance.


RobertLuptonTheGood:
Please don't squash that "remove disabled ..." changeset.  The algorithm (a flood-fill) was used in peak culling and may need to be brought back to life.

Removing it now is fine.

src/detection/FootprintSet.cc
```diff
@@ -714,15 +714,15 @@ static void findFootprints(
         }
 
         if (in_span) {
-            auto sp = std::make_shared<IdSpan>(in_span, y, x0, width - 1, good);
+            IdSpan sp(in_span, y, x0, width - 1, good);
             spans.push_back(sp);
```
PaulPrice: Suggest using `emplace_back`, here and above.

PR #354, opened 2018-05-22
DM-14548: Make dimensionless quantities in `refraction` `floats`

isullivan:
Also fix return value docstring.

### DIFF
--- a/python/lsst/afw/coord/refraction.py
+++ b/python/lsst/afw/coord/refraction.py
@@ -69,7 +69,7 @@ def refraction(wavelength, elevation, observatory, weather=None):
         weather = defaultWeather(altitude*units.meter)
     reducedN = deltaN(wavelength, weather)/deltaRefractScale
     temperature = extractTemperature(weather, useKelvin=True)
-    atmosScaleheightRatio = float(4.5908E-6*temperature/units.Kelvin)
+    atmosScaleheightRatio = 4.5908E-6*temperature.to_value(units.Kelvin)
 
     # Account for oblate Earth
     # This replicates equation 10 of Stone 1996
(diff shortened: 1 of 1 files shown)


python/lsst/afw/coord/refraction.py
```diff
@@ -164,9 +164,9 @@ def densityFactorDry(weather):
     waterVaporPressure = humidityToPressure(weather)
     airPressure = weather.getAirPressure()*units.pascal
     dryPressure = airPressure - waterVaporPressure
-    eqn = (dryPressure/cds.mbar)*(57.90E-8 - 9.3250E-4*units.Kelvin/temperature +
-                                  0.25844*units.Kelvin**2/temperature**2.)
-    densityFactor = (1. + eqn)*(dryPressure/cds.mbar)/(temperature/units.Kelvin)
+    eqn = float((dryPressure/cds.mbar)*(57.90E-8 - 9.3250E-4*units.Kelvin/temperature +
+                                        0.25844*units.Kelvin**2/temperature**2.))
+    densityFactor = (1. + eqn)*float(dryPressure/cds.mbar)/float(temperature/units.Kelvin)
```
r-owen: From reading the astropy docs I expected to see `value.to_value(units)`, instead of using `float(value/units)`. For example:

    dryPressure.to_value(cds.mbar)

and this seems a clear way to express what you want done.

If what you've done is a standard way to get a float in particular units, can you please point me to the appropriate docs? I realize it works, but based on reading what I found, it doesn't seem obvious to me that it should.

python/lsst/afw/coord/refraction.py
```diff
@@ -212,8 +212,8 @@ def humidityToPressure(weather):
 
     Returns
     -------
-    `float`
-        The water vapor pressure in millibar
+    `astropy.units.quantity.Quantity`
```
r-owen: I suggest `astropy.units.Quantity` since that is the documented version. The existence of a `quantity` sub-module appears to be an internal detail of astropy.

python/lsst/afw/coord/refraction.py
isullivan: Yes, that looks more clear. I'll change to `to_value(units)` throughout.

PR #356, opened 2018-05-23
DM-14429: Implement RFC-460: move afw.geom content to new geom package

r-owen:


include/lsst/afw/geom/AffineTransform.h
```diff
@@ -37,249 +35,12 @@ namespace lsst {
...
-    TransformDerivativeMatrix dTransform(Extent2D const &input) const;
-
-private:
-    LinearTransform _linear;
-    Extent2D _translation;
-};
+// for backwards compatibility make lsst::geom headers and symbols available in lsst::afw::geom
 
-std::ostream &operator<<(std::ostream &os, lsst::afw::geom::AffineTransform const &transform);
+using lsst::geom::AffineTransform;
```
kfindeisen: Can you use `[[deprecated]]` here? The [documentation](https://en.cppreference.com/w/cpp/language/attributes#Standard_attributes) is kind of unclear whether it applies to using-declarations.

TallJimbo: I gave this a try with a small test program (I'll attach it to the JIRA issue), and at least on gcc 7.3.0, it looks like
```
[[deprecated("use lsst::geom::AffineTransform instead")]]
typedef lsst::geom::AffineTransform AffineTransform
```
does exactly what we want, but I no variant of `using` will compile if it is preceded by a `[[deprecated]]` tag.

I also think that if we add `[[deprecated]]` here, we might want to just squash those warnings via pragmas or compiler options in existing packages we haven't tried to convert, and that sort of defeats the purpose of not trying to convert them.  I'd love to somehow get a warning for new code only, but the goal here was explicitly to avoid breaking old code whenever possible, and adding lots of warnings is sort of a form of breakage.

kfindeisen: What? Not breaking code is the whole point of warnings (deprecation or otherwise). It will be nearly impossible to phase out the old API otherwise.

I couldn't parse the first half of your comment. Are you saying that a typedef can be marked deprecated but a using-declaration cannot?

TallJimbo: > What? Not breaking code is the whole point of warnings (deprecation or otherwise).
That's fair, but...

> It will be nearly impossible to phase out the old API otherwise.
... I simply don't think it'll be worthwhile to try to phase out the old API in old code in anything like the near future, and I don't want these warnings to obscure more important ones until we decide that's worthwhile.

> I couldn't parse the first half of your comment. Are you saying that a typedef can be marked deprecated but a using-declaration cannot?

Yes, at least for gcc.

kfindeisen: Then that makes it impractical, since templates can't be typedefed. :disappointed:

TallJimbo: Ah, good point.  We do tend to use most of the particular templates we've moved on this ticket almost exclusively via typedefs to specific template instantiations, but it certainly limits the usefulness of `[[deprecated]]` for projects like this.

r-owen: @kfindeisen @TallJimbo I'm inclined to skip trying to use [[deprecated]]. We can convert packages as they come up and we can find out what is broken by removing the deprecated aliases and fixing what shows up.

PR #357, opened 2018-06-04
DM-14072: Add getCutout method to Exposure

kfindeisen:
This PR adds a method, `Exposure::getCutout`, that returns a fixed-sized image centered on particular sky coordinates. To ensure the image always has the desired size, empty pixels (for coordinates near the edge of an exposure) are allowed.

Note that this implementation does not use the `Box2D` -> `Box2I` conversion suggested by @TallJimbo. That conversion prioritizes keeping particular pixels over keeping the desired size, and correctly clipping the `Box2I` back to the desired size would require complex conditions on the sky coordinates' location within a pixel.

### DIFF
--- a/include/lsst/afw/image/Exposure.h
+++ b/include/lsst/afw/image/Exposure.h
@@ -31,6 +31,8 @@
 #include "lsst/daf/base.h"
 #include "lsst/afw/image/MaskedImage.h"
 #include "lsst/afw/image/ExposureInfo.h"
+#include "lsst/geom/Extent.h"
+#include "lsst/geom/SpherePoint.h"
 
 namespace lsst {
 namespace afw {
@@ -390,6 +392,23 @@ public:
     static Exposure readFits(fits::MemFileManager& manager) {
         return Exposure<ImageT, MaskT, VarianceT>(manager);
     }
+
+    /**
+     * Return an Exposure that is a small cutout of the original.
+     *
+     * @param center desired center of cutout (in RA and Dec)
+     * @param size width and height (in that order) of cutout in pixels
+     *
+     * @return An Exposure of the requested size centered on `center` to within
+     *     half a pixel in either dimension. Pixels past the edge of the original
+     *     exposure will be zero.
+     *
+     * @throws lsst::pex::exceptions::LogicError Thrown if this Exposure does
+     *     not have a WCS.
+     * @throws lsst::pex::exceptions::InvalidParameterError Thrown if ``center``
+     *     falls outside this Exposure or if ``size`` is not a valid size.
+     */
+    Exposure getCutout(lsst::geom::SpherePoint const& center, lsst::geom::Extent2I const& size) const;
 
 private:
     LSST_PERSIST_FORMATTER(lsst::afw::formatters::ExposureFormatter<ImageT, MaskT, VarianceT>)
--- a/src/image/Exposure.cc
+++ b/src/image/Exposure.cc
@@ -24,6 +24,7 @@
 
 #include <memory>
 #include <stdexcept>
+#include <sstream>
 #include <cstdint>
 
 #include "boost/format.hpp"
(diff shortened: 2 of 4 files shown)


include/lsst/afw/image/Exposure.h
```diff
@@ -391,6 +393,23 @@ class Exposure : public lsst::daf::base::Persistable, public lsst::daf::base::Ci
...
 
+    /**
+     * Return an Exposure that is a small cutout of the original.
+     *
+     * @param center desired center of cutout (in RA and Dec)
+     * @param size width and height (in that order) of cutout in pixels
+     *
+     * @return An Exposure of the requested size centered on `center` to within
+     *     half a pixel in either dimension. Pixels past the edge of the original
+     *     exposure will be zero.
```
TallJimbo: I'd like to make sure I understand the desired behavior; I think you want to:
 - Ensure the given size is respected exactly.
 - Pad with zeros when the requested region is not wholly contained by the Exposure.
 - When a dimension of `size` is even (and hence the pixel-coordinate center of the returned subimage is a half-integer value in that dimension), the given `center` lands within 0.5 pixels of that half-integer value.
 - When a dimension of `size` is odd (and hence the pixel-coordinate center of the returned subimage is an integer value in that dimension), the given `center` lands within 0.5 pixels of that integer value.

src/image/Exposure.cc
```diff
@@ -196,6 +198,72 @@ void Exposure<ImageT, MaskT, VarianceT>::writeFits(fits::Fits &fitsfile,
...
+ * Create a box centered as closely as possible on a particular point.
+ *
+ * @param center The desired center of the box.
+ * @param size The desired width and height (in that order) of the box.
+ *
+ * @returns a box with size ``size`` and center within half a pixel of ``center`` in either dimension.
+ *
+ * @throws lsst::pex::exceptions::InvalidParameterError Thrown if ``size`` has invalid dimensions.
+ */
+lsst::geom::Box2I _makeCenteredBox(lsst::geom::Point2D const &center, lsst::geom::Extent2I const &size) {
```
TallJimbo: I think this would be a useful public addition to `Box2D`, either as a static member function or a free function in the same namespace.

Related methods that create `Image`, `Mask`, or `MaskedImage` cutouts given a non-integer center coordinate would be similarly useful, and with this function they'd be trivial.

src/image/Exposure.cc
```diff
@@ -196,6 +198,72 @@ void Exposure<ImageT, MaskT, VarianceT>::writeFits(fits::Fits &fitsfile,
...
+    return lsst::geom::Box2I(lsst::geom::Point2I(corner), size, false);
+}
+
+/**
+ * Copy all overlapping pixels from one Exposure to another.
+ *
+ * @param destination The Exposure to copy pixels to.
+ * @param source The Exposure whose pixels will be copied.
+ *
+ * @throws lsst::pex::exceptions::InvalidParameterError Thrown if the Exposures do not overlap.
```
TallJimbo: I think it might actually make sense for this to silently do nothing if there is no overlap.  I can certainly imagine use cases where it's undesirable for images that only slightly overlap to be treated qualitatively differently from those that do not overlap at all.

kfindeisen:
> I also have a comment on the comment for 8eee3ae: note that because this `Exposure` constructor makes a shallow copy of the `MaskedImage`'s pixels, it _can_ in fact modify the `MaskedImage`. But the change you've made is still a good one (a small step in the right direction within an inconsistent big picture).

I disagree. I was not aware that there was a good reason for the argument taking a non-const reference, and "risk of data corruption" is a pretty good reason. The only reason the non-const reference caused me trouble was because I wanted to construct an `Exposure` without having to assign the temporary `MaskedImage` to a variable, but that's by far the lesser evil.

src/image/Exposure.cc
kfindeisen: This goes back to the point you made by email about "conventions" (which apply to both even and odd dimensions, BTW: for an even size there are two or four equally good solutions if `center` lies on a pixel centerline, for an odd size there are multiple solutions if `center` lies on a pixel boundary). For the specific purpose of `getCutout`, it really doesn't matter which of four equally shifted bounding boxes is chosen (our application is source images for human consumption), and I was careful to neither promise nor test anything more specific.

A public method or function, on the other hand, may have use cases where the user does actually care which of the multiple solutions gets chosen (though, to be fair, I found in testing that floating-point error in the generation of `center` will almost always cause one of the four points to be slightly closer). It would need a more specific postcondition than what I've provided here, and the right answer might well depend on what the user wants to do with `center`.

src/image/Exposure.cc
kfindeisen: This behavior was requested by @mrawls on the grounds that an out-of-bounds bounding box for `getCutout` is almost never something you'd do on purpose.

TallJimbo:
> The only reason the non-const reference caused me trouble was because I wanted to construct an Exposure without having to assign the temporary MaskedImage to a variable, but that's by far the lesser evil.

But this does not provide any protection against data corruption either; the resulting `Exposure` can still modify the original `MaskedImage`, because the temporary `MaskedImage` is a view to the original, and the `Exposure` is a view to the temporary (and hence to the original).

src/image/Exposure.cc
mrawls: I think a use case where the user is confused about how to define bboxes correctly is significantly more likely than a use case where they wish to make a cutout of an exposure that has zero overlap with the original. If we do allow this, it should scream about it and not be silent.

src/image/Exposure.cc
TallJimbo: I'd be quite content having a public method that also documents that it does not make guarantees about which of these solutions are selected (and also does not guarantee that the behavior will not change in the future, if we want to start providing those guarantees).  I'd like to at least support the use cases that have the same (lenient) requirements as `getCutout`.

src/image/Exposure.cc
TallJimbo: I'll certainly concede that the case where the user is confused is also pretty likely, but I'm still a bit bothered that we're not also throwing if the overlap is, say, just a single column.  Maybe the criteria should be whether the center point lies within the image's bounding box?  (I suppose that'd have to be checked at a higher level.)

kfindeisen:
In this case the temporary `MaskedImage` is not (should not be?) a view to anything. I thought the [`MaskedImage(Box2I, MaskPlaneDict)` constructor](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/classlsst_1_1afw_1_1image_1_1_masked_image.html#af10ad1512869737fced97744de0b213d) didn't have a way to access the original `Exposure`, and it was always our intent that the cutout be an independent object.

src/image/Exposure.cc
kfindeisen: I'm a little worried by the wording "does not guarantee that the behavior will not change in the future" -- it sounds like you'd want to make arbitrary changes, rather than merely making the output more specific. :+1: if I misunderstood.

TallJimbo:
Yes, that constructor would be safe.  I was thinking you were planning to use [`MaskedImage(MaskedImage, Box2I, ImageOrigin)`](https://github.com/lsst/afw/blob/master/include/lsst/afw/image/MaskedImage.h#L731) (which strangely does not appear in Doxygen) or the regular copy constructor to make the temporary.

In any case, I suppose it's not crazy for us to disagree on whether it's better to try to preserve what little const protection the image classes provide or remove it all to eliminate the illusion that the const protections actually provide any guarantees.  I don't have a strong opinion about what to do in this small corner of the API in the absence of a plan to deal with the larger problem.

PR #361, opened 2018-06-14
DM-14690: Add ability to construct centered boxes

kfindeisen:
This PR modifies the implementation of `Exposure::getCutout` to use the `Box2I::makeCenteredBox` factory introduced by lsst/geom#4. However, as a side effect the `getCutout` method became considerably more complex (17 LLOC for input validation plus three logical processing steps); any suggestions on how to break it up would be appreciated.

### DIFF
--- a/src/image/Exposure.cc
+++ b/src/image/Exposure.cc
@@ -199,30 +199,6 @@ void Exposure<ImageT, MaskT, VarianceT>::writeFits(fits::Fits &fitsfile,
 
 namespace {
 /**
- * Create a box centered as closely as possible on a particular point.
- *
- * @param center The desired center of the box.
- * @param size The desired width and height (in that order) of the box.
- *
- * @returns a box with size ``size`` and center within half a pixel of ``center`` in either dimension.
- *
- * @throws lsst::pex::exceptions::InvalidParameterError Thrown if ``size`` has invalid dimensions.
- */
-lsst::geom::Box2I _makeCenteredBox(lsst::geom::Point2D const &center, lsst::geom::Extent2I const &size) {
-    if (size[0] <= 0 || size[1] <= 0) {
-        std::stringstream buffer;
-        buffer << "Cannot create bounding box with dimensions " << size;
-        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError, buffer.str());
-    }
-
-    lsst::geom::Point2D corner(center);
-    corner.shift(-0.5 * lsst::geom::Extent2D(size));
-    // compensate for Box2I's coordinate conventions (where max = min + size - 1)
-    corner.shift(lsst::geom::Extent2D(0.5, 0.5));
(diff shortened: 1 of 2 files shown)


src/image/Exposure.cc
```diff
@@ -249,15 +225,20 @@ Exposure<ImageT, MaskT, VarianceT> Exposure<ImageT, MaskT, VarianceT>::getCutout
...
                << getBBox();
         throw LSST_EXCEPT(pex::exceptions::InvalidParameterError, buffer.str());
     }
-    lsst::geom::Box2I bbox = _makeCenteredBox(pixelCenter, size);
+    if (size[0] <= 0 || size[1] <= 0) {
+        std::stringstream buffer;
+        buffer << "Cannot create bounding box with dimensions " << size;
+        throw LSST_EXCEPT(pex::exceptions::InvalidParameterError, buffer.str());
+    }
+    lsst::geom::Box2I bbox = lsst::geom::Box2I::makeCenteredBox(pixelCenter, size, false);
```
TallJimbo: I don't see a good way to simplify this, though I'm also not too bothered by that - the overall code complexity is the same; you've just moved this check from one place to another.

But it may be worth noting that given the check for both the center coordinate being in within the Exposure bbox and the size being positive, I believe it's guaranteed that `overlapBox.isEmpty()` is always `false` up in `_copyCommonPixels`, and could perhaps be turned into an `assert`, *if* this is the only caller.

kfindeisen: I'd prefer to leave it in, in case somebody later wants to promote `_copyCommonPixels` to a public method or function. I had quite a bit of trouble generalizing `makeCenteredBox` to work correctly with negative boxes, and `assign`'s behavior is pretty surprising.

TallJimbo: > `assign`'s behavior is pretty surprising

I had no idea that's how it worked until I saw your comment, and now that I do know, it makes me sad, too.

RobertLuptonTheGood:
What happens if the centre's on the Exposure but the box extends off the edge?  It's convenient for the user to return something sane under these circumstances (presumably with NaNs and mask bits set)

						R

kfindeisen:
As stated in the `getCutout` method documentation, pixels off the edge of the image will be zero. This behavior is consistent with the behavior of the "blank" `Exposure` constructors. Note that this PR does not change the behavior of `getCutout` in any way.

NaNs are not a terribly practical choice for the filler because we would need to adopt a different convention anyway for integer `Exposure`s (of which we have three mandatory instantiations ATM). It could be done, of course, but it would make the behavior more surprising to users.

Setting mask bits is not possible because `Exposure`/`MaskedImage` knows nothing about the meaning of specific bits -- those are determined by client code.

TallJimbo:
I should have thought of this this on the previous ticket, but I think the `NO_DATA` mask plane's meaning is sufficiently global and appropriate that it probably would make sense to set that for any off-the-exposure pixels.

PR #365, opened 2018-06-21
DM-14844: Look up data files relative to known locations

timj:
Without this you can't run `pytest $AFW_DIR` and expect it to work.

### DIFF
--- a/tests/test_footprint.py
+++ b/tests/test_footprint.py
@@ -19,6 +19,7 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 
+import os
 import unittest
 import tempfile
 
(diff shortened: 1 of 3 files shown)


tests/test_footprint.py
```diff
@@ -247,7 +248,8 @@ def testPersistence(self):
         del footprintFromFile
 
     def testLegacyFootprints(self):
-        fileName = 'tests/data/preSpanSetsFootprint.fits'
+        testPath = os.path.abspath(os.path.dirname(__file__))
+        fileName = os.path.join(testPath, 'data/preSpanSetsFootprint.fits')
```
r-owen: I suggest `fileName = os.path.join(testPath, 'data', 'preSpanSetsFootprint.fits')` instead

PR #366, opened 2018-06-29
DM-9937: Add noexcept specifiers to applicable methods in afw

kfindeisen:
This PR adds `noexcept` specifiers to several key types in `lsst::afw`. I have omitted these specifiers in a couple of key cases:
- any method that directly or indirectly uses dynamic memory allocation, or where it might be reasonable to want that in the future.  This includes methods that return `vector` or `string` by value.
- any method that could in principle take an invalid input (e.g., anything that takes an index) has been left potentially throwing
- some classes have implementations that I wasn't able to analyze (e.g., `ImageBase` and classes that compose it or its subclasses); to be on the safe side, I left these classes untouched even when a method "obviously shouldn't" throw

I was a bit surprised to conclude that `afw::table::FieldBase<T>` and `KeyBase<T>` are unconditionally nothrow-constructible and nothrow-copyable; please confirm that neither they nor any reasonable subclass can ever have a constructor failure for any `T`.

### DIFF
--- a/include/lsst/afw/geom/SpanPixelIterator.h
+++ b/include/lsst/afw/geom/SpanPixelIterator.h
@@ -42,13 +42,13 @@ namespace geom {
 class SpanPixelIterator : public boost::iterator_facade<SpanPixelIterator, lsst::geom::Point2I const,
                                                         boost::random_access_traversal_tag> {
 public:
-    explicit SpanPixelIterator(lsst::geom::Point2I const& p = lsst::geom::Point2I()) : _p(p) {}
+    explicit SpanPixelIterator(lsst::geom::Point2I const& p = lsst::geom::Point2I()) noexcept : _p(p) {}
 
-    SpanPixelIterator(SpanPixelIterator const&) = default;
-    SpanPixelIterator(SpanPixelIterator&&) = default;
-    SpanPixelIterator& operator=(SpanPixelIterator const&) = default;
-    SpanPixelIterator& operator=(SpanPixelIterator&&) = default;
-    ~SpanPixelIterator() = default;
+    SpanPixelIterator(SpanPixelIterator const&) noexcept = default;
+    SpanPixelIterator(SpanPixelIterator&&) noexcept = default;
+    SpanPixelIterator& operator=(SpanPixelIterator const&) noexcept = default;
+    SpanPixelIterator& operator=(SpanPixelIterator&&) noexcept = default;
+    ~SpanPixelIterator() noexcept = default;
 
 private:
     friend class boost::iterator_core_access;
--- a/include/lsst/afw/math/FunctionLibrary.h
+++ b/include/lsst/afw/math/FunctionLibrary.h
@@ -266,7 +266,7 @@ public:
         return os.str();
     }
 
-    virtual bool isPersistable() const { return true; }
+    virtual bool isPersistable() const noexcept override { return true; }
 
 protected:
     virtual std::string getPersistenceName() const;
@@ -379,7 +379,7 @@ public:
         return os.str();
     }
 
-    virtual bool isPersistable() const { return true; }
+    virtual bool isPersistable() const noexcept override { return true; }
 
 protected:
     virtual std::string getPersistenceName() const;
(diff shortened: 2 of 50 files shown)


include/lsst/afw/geom/SpanPixelIterator.h
```diff
@@ -61,7 +61,7 @@ class SpanPixelIterator : public boost::iterator_facade<SpanPixelIterator, lsst:
 
     void advance(int n) { _p.getX() += n; }
 
-    bool equal(SpanPixelIterator const& other) const {
+    bool equal(SpanPixelIterator const& other) const noexcept {
```
TallJimbo: If the `getX` and `getY` calls here are `noexcept`, they should be in `increment`, `decrement`, and `advance`, too, even if they're const overloads.

But also note [DM-14992](https://jira.lsstcorp.org/browse/DM-14992), which I just created after noticing the incorrect iterator tag for this class - I'm hope to do that soon, as it's a potentially serious problem, and that will probably involve this class being rewritten without `boost::iterator_facade` anyway.

include/lsst/afw/math/FunctionLibrary.h
```diff
@@ -836,7 +838,7 @@ class Chebyshev1Function2 : public BasePolynomialFunction2<ReturnT> {
     /**
      * Get x,y range
      */
-    lsst::geom::Box2D getXYRange() const {
+    lsst::geom::Box2D getXYRange() const noexcept {
```
TallJimbo: Looks like the `Box2D` constructor used here is not `noexcept` (though I believe it could be).

In addition, the `Point2D` constructors used hare are only conditionally `noexcept`, so this should be too, if you fix the `Box` issue - but I think it'd be better to `static_assert` against anything that could make the `Point2D` constructors throw instead, as per my comments on the `geom` PR.

include/lsst/afw/geom/SpanPixelIterator.h
kfindeisen: I thought that forward iterators must always be comparable (the [ForwardIterator spec](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) does not mention an equality domain, while InputIterator does), while iterator moves have cases where they're allowed to throw (e.g., the spec does not define what happens if you increment a past-the-end iterator).

However, the [proposed ForwardIterator concept](https://en.cppreference.com/w/cpp/experimental/ranges/iterator/ForwardIterator) explicitly says that only comparison of iterators over the same sequence is defined. I guess that means we should remove `noexcept` here, whether or not DM-14992 is implemented.

TallJimbo: ForwardIterators definitely need to be equality comparable, and I don't think any kind of iterator can be comparable unless a domain is specified (see previous conversation on pointer comparisons).

In any case, my comment here was just saying that *if* `equal` can be `noexcept` *given its implementation*, then the other methods could, too, given their implementations.  I wasn't thinking about whether any possible implementation would be `noexcept`, and on those grounds what you have here is fine with me.  (Making them all `noexcept` or making none of them `noexcept` is also fine with me, as I actually don't think there is any reasonable implementation of this class that could throw when the iterator is moved, and I don't think iterator concepts require `noexcept` anywhere).

include/lsst/afw/math/FunctionLibrary.h
kfindeisen: `Point2D` is `noexcept` (`2 != Eigen::Dynamic`, and `double` is nothrow-copyable). In fact, it's impossible to speak of a class as being conditionally `noexcept`, because there's nothing to condition on.

As for `Box2D`, I left that constructor `noexcept` because it could potentially throw when `max > min` (which is prevented in this case by `Box2D`'s class invariant). Though I guess you could argue that such a change in behavior would break backward-compatibility in ways other than unexpectedly throwing an exception...

kfindeisen: Yeah, on second thought I was being too pedantic. I'll update the `Box` constructors to guarantee that `max > min` never throws; I think that will leave `Box2I` overflow as the only failure mode.

PR #368, opened 2018-07-02
DM-14998: Fix documentation on Schema field name conventions.

TallJimbo:


doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
- *  layer, so we cannot distinguish between the two when we read tables from FITS).
- *  Schema provides extra functionality for names with period-separated elements; these elements can
- *  be accessed separately individually with the bracket operators.  More information on schema namespaces
- *  can be found in the Schema and SubSchema class documentation, and the testSchema.py unit test may
- *  also be a useful example.
- *
- *  Other field strings (documentation and units) are essentially arbitrary, but should not contain
- *  single quotes, as these may also confuse FITS parsers (even when escaped).
+ *
+ *  By convention, field names are CamelCase and have '_'-separated elements.
```
laurenam: Would it be more accurate to say (c)CamelCase (or some other way of emphasizing that the initial letter need not always be capitalized)?

Might it help to provide a list of examples that fall into the category of "element" in this case (no worries if you think the meaning of the word is clear/unambiguous in the context of code)?

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *  Only letters, numbers and underscores should be used.
```
laurenam: Comma after "numbers" (to be consistent with use below at line 171)?

include/lsst/afw/table/Schema.h
```diff
@@ -323,7 +325,7 @@ class Schema {
 /**
  *  A proxy type for name lookups in a Schema.
  *
- *  Elements of schema names are assumed to be separated by periods ("a.b.c.d");
+ *  Elements of schema names are assumed to be separated by underscores ("a.b.c");
```
laurenam: a.b.c -> a_b_c ?!

include/lsst/afw/table/Schema.h
```diff
@@ -132,7 +134,7 @@ class Schema {
      *  Return a set of field names in the schema.
      *
      *  If topOnly==true, return a unique list of only the part
-     *  of the names before the first period.  For example,
+     *  of the names before the first underscore.  For example,
```
laurenam: Examples below ['a.b.c', 'a.d', 'e.f'] -> ['a_b_c', 'a_d', 'e_f']

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *      particular field represents the x-x second-moment of the source.
+ *
+ *    - `deblend_nChild` is produced by `SourceDeblendTask`, where `deblend`
+ *      is an abbreviation of the Task, and `nChild` is a value produced by
+ *      it.  There are no hard rules for how to abbreviate the name of a Task
+ *      when generating its field names; we trust developer judgement in
+ *      selecting a prefix that is both concise and unambiguous.
+ *
+ *  Note that some underscore-separated elements are themselves multiple words,
+ *  such as `SdssShape` or `nChild`, and we use CamelCase, not more underscores,
```
laurenam: (c)CamelCase?

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *  to separate words.
+ *  A good rule of thumb is that if two words are not individually meaningful
+ *  (or mean something different when separated), join them with CamelCase.
```
laurenam: (c)CamelCase?

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *      should start with a lowercase letter.
+ *
+ *    - The last namespace element (the name of a particular value) should
+ *      begin with a lowercase letter.
+ *
+ *    - Names that correspond to a Python or C++ class that produce the field
+ *      should start with uppercase (to match the name of the class).
+ *
+ *  Schema provides extra functionality for names with underscore-separated
+ *  elements; these elements can be accessed separately individually with the
```
laurenam: "separately individually" -> redundant?

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *  bracket operators.  More information on this behavior be found in the
```
laurenam: "...behavior *can* be found..."

include/lsst/afw/table/Schema.h
```diff
@@ -32,11 +32,13 @@ class BaseRecord;
  *  Because offsets for fields are assigned when the field is added to the Schema,
  *  Schemas do not support removing fields, though they do allow renaming.
  *
- *  Field names in Schemas are expected to be dot-separated names (e.g. 'a.b.c').  The SubSchema
+ *  Field names in Schemas are expected to be underscore-separated names (e.g. 'a_b_c',
+ *  but see @ref afwTableFieldNames for the full conventions, including when to use
+ *  underscores vs. CamelCase).  The SubSchema
```
laurenam: (c)CamelCase?

doc/table.dox
TallJimbo: I'll just write "camel case" here and not imply anything with how its capitalized.  I think later uses are okay because they're right above the section saying that starting with lowercase is okay.

I think I prefer having the list of element examples in the next paragraph, just to keep this paragraph simple.

doc/table.dox
TallJimbo: I think I'll leave this one and the next as is; I don't want the punctuation to be confusing, either.

doc/table.dox
```diff
@@ -158,17 +158,58 @@ namespace lsst { namespace afw { namespace table {
...
+ *      the meas_base package (we drop "meas_" since so many algorithms are in
+ *      meas_* packages; see `lsst.meas.base.generateAlgorithmName`), and this
+ *      particular field represents the x-x second-moment of the source.
+ *
+ *    - `deblend_nChild` is produced by `SourceDeblendTask`, where `deblend`
+ *      is an abbreviation of the Task, and `nChild` is a value produced by
+ *      it.  There are no hard rules for how to abbreviate the name of a Task
+ *      when generating its field names; we trust developer judgement in
+ *      selecting a prefix that is both concise and unambiguous.
+ *
```
laurenam: Actually, could I ask that you include an example from one of the cases that pointed this issue out (just to be overly complete!)?

TallJimbo: Which cases do you mean?  The ones on DM-14997?  I'm vaguely aware that different Tasks do different things, which is why I've avoiding trying to declare a rule that I know we don't follow, but I haven't actually thought about which ones use good names and why.

laurenam: Yeah, like the calib_photometryUsed -> calib_photometry_used, for example.  A call has to be made there to homogenize (or do you think that requires an RFC.  I liked your <module-or-task>_<algorithm>_<state> argument...)

TallJimbo: I think there will need to be an RFC regardless, because changing the schemas produced by the pipeline breaks any code that reads those outputs; it's like an API change.

I actually think the reason to prefer `calib_photometry_used` over `calib_photometryUsed` (now that I think about it) is that you want underscores when you want the prefix to be considered a conceptual group.  I'll go ahead and add that rule of thumb as well.

laurenam: Ok, shall I write an RFC for which DM-14997 will be the implementation ticket?

TallJimbo: > Ok, shall I write an RFC for which DM-14997 will be the implementation ticket?

Yup, that sounds perfect.

PR #370, opened 2018-07-10
DM-14932: Add utility functions for creating SkyWcss from boresight/rotator + cameraGeom

r-owen:


include/lsst/afw/geom/SkyWcs.h
```diff
@@ -521,6 +521,26 @@ std::shared_ptr<SkyWcs> makeSkyWcs(daf::base::PropertySet &metadata, bool strip
...
+ * @param[in] crval  Center of projection on the sky.
+ * @param[in] projection  The name of the FITS WCS projection, e.g. "TAN" or "STG".
+ *
+ * @note Unlike makeCdMatrix, `orientation` is with respect to the focal plane axes, not the CCD axes.
+ *      This is because field angle is defined with respect to focal plane axes.
+ */
+std::shared_ptr<SkyWcs> makeSkyWcs(TransformPoint2ToPoint2 const &pixelsToFieldAngle,
+                                   lsst::geom::Angle const &orientation, bool flipX,
+                                   lsst::geom::SpherePoint const &crval,
+                                   std::string const &projection = "TAN");
```
TallJimbo: I'm a bit concerned that the name of this function and the types of its arguments really give no information about what it's for, or more importantly, how it differs from other `makeSkyWcs` overloads - and while the documentation of course does that, I think the method name should be a bit more descriptive.  I realize this is going against precedent (for this class, at least), but I'd prefer a name like `SkyWcs::fromCameraGeom`.  If you'd rather a free function to go along with what exists, perhaps `makeSkyWcsFromCameraGeom`?

That said, really excellent clarity in the documentation for what this does.  Thanks for taking the time to do that carefully.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -521,6 +521,26 @@ std::shared_ptr<SkyWcs> makeSkyWcs(daf::base::PropertySet &metadata, bool strip
...
 
+/**
+ * Construct a FITS SkyWcs from camera geometry
+ *
+ * @param[in] pixelsToFieldAngle  Transformation from \ref afwCameraGeomPIXELS "pixels"
+ *              to \ref afwCameraGeomFIELD_ANGLE "field angle" (in radians).
+ * @param[in] orientation  Position angle of focal plane +Y, measured from N through E at crval.
+ *                         At 0 degrees, +Y is along N and +X is along W/E if flipX false/true.
+ *                         At 90 degrees, +Y is along E and +X is along N/S if flipX false/true.
+ * @param[in] flipX  Fip x axis? See orientation for details.
```
TallJimbo: Typo: `fip` -> `flip`.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -521,6 +521,26 @@ std::shared_ptr<SkyWcs> makeSkyWcs(daf::base::PropertySet &metadata, bool strip
 std::shared_ptr<SkyWcs> makeSkyWcs(lsst::geom::Point2D const &crpix, lsst::geom::SpherePoint const &crval,
                                    Eigen::Matrix2d const &cdMatrix, std::string const &projection = "TAN");
 
+/**
+ * Construct a FITS SkyWcs from camera geometry
+ *
+ * @param[in] pixelsToFieldAngle  Transformation from \ref afwCameraGeomPIXELS "pixels"
+ *              to \ref afwCameraGeomFIELD_ANGLE "field angle" (in radians).
```
TallJimbo: `\ref` -> `@ref`

include/lsst/afw/geom/SkyWcs.h
r-owen: I plan to rename the `crval` argument to `boresight` since it is the sky position at field angle (0, 0). The current name seems more appropriate for a WCS whose reference sky position is on the detector.

Regarding the function name...I agree it's a bit vague but I have not thought of anything I like better. `makeSkyWcsFromCameraGeom` seems too long, but is tolerable as long as we violate our standards and abbreviate "geometry".

On your final review comments you suggested we should have another signature once `Detector` has a reference to camera geometry. Perhaps that one can be called `makeSkyWcsFromDetector` and this one can stay as it is. I think that one will have to go in `cameraGeom` because `cameraGeom` depends on `geom`.

PR #372, opened 2018-07-15
DM-14864: Make afw pybind11 wrappers compatible with pybind11 2.2

r-owen:


python/lsst/afw/math/stack.cc
```diff
@@ -32,6 +35,31 @@ using namespace py::literals;
...
+        std::vector<std::vector<PixelT>>& vectors,  ///< Vectors to process
+        Property flags,                                              ///< statistics requested
+        StatisticsControl const& sctrl = StatisticsControl(),        ///< control structure
+        std::vector<lsst::afw::image::VariancePixel> const& wvector =
+                std::vector<lsst::afw::image::VariancePixel>(0)  ///< vector containing weights
+) {
+        std::vector<std::shared_ptr<std::vector<PixelT>>> vcopy;
+        vcopy.reserve(vectors.size());
+        for (auto const & v : vectors) {
+            auto vshared = std::make_shared<std::vector<PixelT>>(std::move(v));
```
kfindeisen: I think this is unsafe, as it will empty the elements of `vectors`. I realize that the [original statisticsStack](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/namespacelsst_1_1afw_1_1math.html#aa7f9ddedbd14f76864aee051549b5cea) takes the vectors by modifiable reference, but surely that's not the change they're supposed to make.

python/lsst/afw/math/stack.cc
```diff
@@ -32,6 +35,31 @@ using namespace py::literals;
...
+            vcopy.emplace_back(vshared);
+        }
+        auto result = statisticsStack(vcopy, flags, sctrl, wvector);
+        return std::vector<PixelT>(result->begin(), result->end());
```
kfindeisen: This seems unnecessarily convoluted. Why not just `vector<PixelT>(*result)`? If anything, a constructor that knows it's dealing with a vector may be more optimal than one that works on a generic iterator.

python/lsst/afw/math/stack.cc
r-owen: The `vectors` argument is a copy that pybind11 makes, so I figured it was safe to alter it.

What do you suggest instead?

kfindeisen: If it really is a temporary copy than that's fine, but that's not at all obvious to me. I think at the very least a comment would be necessary.

TallJimbo: I think the `move` is safe, as long as this function is only called by pybind11.  I'm not sure you're actually moving, though, given that you're calling `std::move` on a const reference; that will just cast the object to `const T&&`, which will invoke the copy constructor.  I think changing the loop variable to `auto & v` should fix that.

All that said, it'd probably be best (and perhaps not even any more disruptive) to just change the main C++ API to take a vector-of-vectors.  Using `shared_ptr` was almost certainly a premature attempt at optimization, and quite possibly a counterproductive one.

r-owen: @TallJimbo Sounds good to me. Any objection to having that function return a std::vector instead of a shared_ptr<std::vector> as well? That would make it directly compatible with pybind11

TallJimbo: Oh, definitely do that as well.  I didn't realize that was also an issue.  I'd be surprised if we have anywhere in the stack where using a `shared_ptr<vector>` is actually a good idea (which is not to say that fixing this would always be easy, but I think it would be here, because there's probably very little C++ code that uses statisticsStack).

r-owen: OK. I have made that change (including returning a std::vector. It all seems to work and it's certainly an improvement. A global search strongly suggests it won't break anything but I'll start a Jenkins run to be sure.

python/lsst/afw/math/stack.cc
pschella: Or `return std::vector<PixelT>(*statisticsStack(vcopy, flags, sctrl, wvector))` to make sure the compiler knows it is a temporary enabling a move? It may or may not do that now too though.

PR #375, opened 2018-07-31
Fixed minor showCamera problems

RobertLuptonTheGood:


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -459,39 +459,27 @@ def getCcdImage(self, ccd, imageFactory=afwImage.ImageF, binSize=1):
...
-                if im:
-                    im = im.getMaskedImage().getImage()
-                else:
-                    raise e
-
-            if e:
                 if self.verbose:
-                    log.info("Reading %s: %s" % (ccd.getId(), e))
+                    print("Reading %s: %s" % (ccd.getId(), err))
+                    log.warn("Reading %s: %s" % (ccd.getId(), err))
```
jdswinbank: Whether the log message gets displayed or not should be up to the logger configuration, not up to the `self.verbose` flag (obviously, that's not the case for the `print`).

python/lsst/afw/cameraGeom/utils.py
```diff
@@ -459,39 +459,27 @@ def getCcdImage(self, ccd, imageFactory=afwImage.ImageF, binSize=1):
 
         im = None
         if self.butler is not None:
-            e = None
-            if self.type == "calexp":    # reading the exposure can die if the PSF's unknown
-                try:                     # need to switch to cid as below.  RHL has no calexp to test with
-                    fileName = self.butler.get(self.type + "_filename", ccd=ccd.getId(),
-                                               **self.kwargs)[0]
-                    im = imageFactory(fileName)
+            im = None
```
jdswinbank: It's already `None` (per line 460).

python/lsst/afw/cameraGeom/utils.py
timj: Also, it's preferable to defer the string formatting until the logger decides whether it is going to write the message or not. Remove the `%` and replace with standard arguments:
```python
log.warn("Reading %s: %s", ccd.getId(), err)
```

PR #381, opened 2018-08-16
Implement afwDisplay colourmap control

RobertLuptonTheGood:


python/lsst/afw/display/interface.py
```diff
@@ -299,6 +301,27 @@ def setDefaultMaskPlaneColor(name=None, color=None):
         #
         Display._defaultMaskPlaneColor[name] = color
 
+    @staticmethod
+    def setDefaultImageColormap(cmap):
+        """!Set the default colourmap for images
+        @param cmap Name of colourmap, as interpreted by the backend
```
stargaser: At some point we should re-do docstrings to make these work with the new documentation scheme (Numpy-style docstrings) but I think this is better left to another ticket.

PR #383, opened 2018-08-21
DM-14814: Change invalid pixel handling by Exposure::getCutout

kfindeisen:
This PR changes the behavior of `Exposure::getCutout` when creating cutouts that extend off the edge of the original exposure. Previously, the off-edge pixels were left blank; the new behavior is to set them to `math::edgePixel`. Note that this means `getCutout` now throws an exception if the `NO_DATA` flag has been deleted from the global mask plane dictionary.

### DIFF
--- a/tests/test_exposure.py
+++ b/tests/test_exposure.py
@@ -713,7 +713,8 @@ class ExposureTestCase(lsst.utils.tests.TestCase):
         wcs = self.smallExposure.getWcs()
 
         dimensions = [lsst.geom.Extent2I(100, 50), lsst.geom.Extent2I(15, 15), lsst.geom.Extent2I(0, 10),
-                      lsst.geom.Extent2I(25, 30), lsst.geom.Extent2I(15, -5)]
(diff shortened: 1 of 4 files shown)


tests/test_exposure.py
```diff
@@ -773,6 +800,24 @@ def _getExposureCenter(self, exposure):
...
+            The bounding box of the original image.
+        cutoutBox : `lsst.geom.Box2I`
+            The bounding box of the cutout.
+
+        Returns
+        -------
+        corners : iterable of `lsst.geom.Point2I`
+            The corners that are drawn from the original image.
+        """
+        # TODO: prefer set, but need to fix DM-9938 first
```
TallJimbo: I think you'd also need to make `Box` immutable in order to be able to use it (safely) in a `set`).

kfindeisen: I'll remove the comment for now (it was referring to the wrong issue anyway).

kfindeisen:
For the record, the behavior is *not* constrained by the implementation. It's just, as you say, less confusing for users if it matches what `edgePixel` itself would do.

PR #387, opened 2018-08-28
DM-15151: Set symbol visibility to hidden in pybind11 wrappers

r-owen:


include/lsst/afw/table/io/Persistable.h
```diff
@@ -203,8 +203,19 @@ class PersistableFacade {
...
+    /**
+     * Dynamically cast a shared_ptr
+     *
+     * You must provide an explicit template instantiation in the .cc file
+     * for each class that inherits from PersistableFacade.
+     * Designed to work around RTTI issues on macOS with hidden symbols;
+     *
+     * @throw lsst::pex::exceptions::LogicError if the cast fails
+     */
+    static std::shared_ptr<T> dynamicCast(std::shared_ptr<Persistable> const &ptr);
```
kfindeisen: Why is this method part of `PersistableFacade`? Can you provide a link for the "issues"?

kfindeisen: Why throw `LogicError` instead of `std::bad_cast` or, if you want an LSST exception, `pex::exceptions::TypeError`?

src/detection/HeavyFootprint.cc
```diff
@@ -333,6 +334,14 @@ typename HeavyFootprint<ImagePixelT, MaskPixelT, VariancePixelT>::Factory
...
 //
 #define INSTANTIATE(TYPE)                                                                                  \
+    } /* namespace detection */                                                                            \
+    namespace table {                                                                                      \
+    namespace io {                                                                                         \
+    template std::shared_ptr<detection::HeavyFootprint<TYPE>>                                              \
+    PersistableFacade<detection::HeavyFootprint<TYPE>>::dynamicCast(std::shared_ptr<Persistable> const&);  \
+    } /* namespace io */                                                                                   \
+    } /* namespace table */                                                                                \
+    namespace detection {                                                                                  \
```
kfindeisen: This seems more than a little bug prone, especially the initial closing brace. Can you make `INSTANTIATE` context-independent (e.g., declare `::lsst::table::io::PersistableFacade`)? Or allow argument-dependent lookup?

src/geom/Transform.cc
```diff
@@ -252,16 +253,25 @@ void Transform<FromEndpoint, ToEndpoint>::write(OutputArchiveHandle &handle) con
...
-    template std::ostream &operator<<<FromEndpoint, ToEndpoint>(                       \
-            std::ostream &os, Transform<FromEndpoint, ToEndpoint> const &transform);   \
-    namespace {                                                                        \
-    TransformFactory<FromEndpoint, ToEndpoint> registration##FromEndpoint##ToEndpoint( \
-            Transform<FromEndpoint, ToEndpoint>::getShortClassName());                 \
-    }                                                                                  \
-    INSTANTIATE_OVERLOADS(FromEndpoint, ToEndpoint, GenericEndpoint)                   \
-    INSTANTIATE_OVERLOADS(FromEndpoint, ToEndpoint, Point2Endpoint)                    \
+#define INSTANTIATE_TRANSFORM(FromEndpoint, ToEndpoint)                                    \
+    } /* namespace geom */                                                                 \
```
kfindeisen: Same question as above.

src/math/Function.cc
```diff
@@ -61,7 +62,19 @@ std::vector<double> PolynomialFunction2<ReturnT>::getDFuncDParameters(double x,
 }
 
 /// @cond
-#define INSTANTIATE(TYPE) \
+#define INSTANTIATE(TYPE)                                                                                  \
+    } /* namespace math */                                                                                 \
```
kfindeisen: Another namespace crossing outside macro boundaries.

tests/test_kernelIo1.py
```diff
@@ -74,22 +71,9 @@ def testFixedKernel(self):
...
-        persistence.persist(k, storageList, additionalData)
-
-        storageList2 = dafPersist.StorageList()
-        storage2 = persistence.getRetrieveStorage("XmlStorage", loc)
-        storageList2.append(storage2)
-        k2 = persistence.unsafeRetrieve(
-            "FixedKernel", storageList2, additionalData)
+        with lsst.utils.tests.getTempFilePath(".fits") as filename:
+            k.writeFits(filename)
+            k2 = afwMath.FixedKernel.readFits(filename)
```
kfindeisen: I know we're trying to get rid of Boost persistence in general, but why is this part of this ticket?

include/lsst/afw/table/io/Persistable.h
r-owen: This can only occur if the code has an internal bug so I thought `LogicError` seemed most appropriate to me.

We never even bothered to test it before, since it always worked. But symbol visibility issues broke it and the error was hard to diagnose down due to the lack of a test.

src/detection/HeavyFootprint.cc
r-owen: Based on further comments below you are upset about having namespace changes inside of macro boundaries. I am also uncomfortable with this. Unfortunately I could not figure out how to avoid it -- the naive things I tried (e.g. providing full namespaces) didn't work. It appears that the symbols have to be defined in the namespace they are to inhabit.

However, it might make things a bit less ugly if I managed the namespaces fully within the macro -- start from nothing and end with nothing.

src/geom/Transform.cc
r-owen: This is the one instantiation that I could not figure out how to really clean up. I need to provide the `FromEndpoint` and `ToEndpoint` arguments without a namespace in order for `registration##FromEndpoint##ToEndpoint` to work correctly, but I also need to get out of the `geom` namespace in order to instantiate `PersistableFacade`. 

I did get rid of the "table" and "io" namespaces here and everywhere else that I instantiate a `PersistableFacade` and I think that makes the code a bit cleaner.

tests/test_kernelIo1.py
r-owen: Because these changes broke boost persistence for kernels and I don't know how to fix it (nor do I want to spend any real time on it, since we are not using it and I am about to rip it out).

It might be better to rip it out on this ticket, but I was hoping to keep those tickets separate.

PR #389, opened 2018-08-30
DM-15599: Stop using boost persistence in afw

r-owen:


tests/test_imageIo1.py
```diff
@@ -191,6 +191,48 @@ def testBBoxFromMetadata(self):
                                                        lsst.geom.Extent2I(145, 200)),
                                        hdu=hdu)
 
+    @unittest.skipIf(dataDir is None, "afwdata not setup")
+    def testReadFitsWithOptions(self):
+        xy0Offset = lsst.geom.Extent2I(7, 5)
+        bbox = lsst.geom.Box2I(lsst.geom.Point2I(10, 11), lsst.geom.Extent2I(31, 22))
+
+        with tempfile.NamedTemporaryFile() as f:
```
TallJimbo: Why not `TestCase.getTempFilePath`?

tests/test_fitsCompression.py
```diff
@@ -56,7 +56,7 @@ def checkAstropy(image, filename, hduNum=0):
     hduNum : `int`
         HDU number of interest.
     """
-    print("Astropy currently doesn't read our compressed images perfectly.")
+#    print("Astropy currently doesn't read our compressed images perfectly.")
```
TallJimbo: This line may not be good as a `print`, but commenting it out seems worse.

python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -0,0 +1,182 @@
...
+from lsst.log import Log
+from lsst.afw.fits import ImageWriteOptions
+from . import image
+
+
+# This must be added to a class as a *classmethod*, for example:
+#
+# @continueclass
+# class MaskX:
+#     readFitsWithOptions = classmethod(imageReadFitsWithOptions)
```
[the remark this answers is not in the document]
r-owen: I am reluctant to do this because it will show up when a user does `help(image.readFitsWithOptions)`. In other words it's not part of the API.

python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -0,0 +1,182 @@
...
+def imageReadFitsWithOptions(cls, source, options):
+    """Read an Image, Mask, MaskedImage or Exposure from FITS, with options
+
+    Parameters
+    ----------
+    source : `string`, `lsst.afw.fits.File` or `lsst.afw.fits.MemFileManager`
```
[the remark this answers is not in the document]
r-owen: Support does come for free. I will remove all but `str` from the doc strings.

For the record, the `Fits` class is wrapped and the wrapper includes this comment:
```
// Not every feature is wrapped, only those that we guess might be useful.
// In particular, the header keyword read/write and table read/write are not wrapped.
```

tests/test_imageIo1.py
r-owen: OK. I have a mild preference for standard library functions when they do the job, but I admit it's a bit skanky to open the file twice (though it is documented to work on unix).

tests/test_fitsCompression.py
r-owen: Agreed. That was accidental (the voluminous output was annoying and slowed down a test I was running frequently)

r-owen:
Thank you for a very helpful review. I made all your suggested changes except adding the "how to use" bit to the doc strings (and will still do that if kf votes in favor in time).

PR #397, opened 2018-09-27
DM-15857 Use correct aliases for old schemas

parejkoj:


src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -99,6 +102,7 @@ FitsSchemaInputMapper::FitsSchemaInputMapper(daf::base::PropertyList &metadata,
         _impl->version = lsst::afw::table::Schema::VERSION;
     }
     _impl->version = metadata.get("AFW_TABLE_VERSION", _impl->version);
+    _impl->type = metadata.get("AFW_TYPE", _impl->type);
```
parejkoj: Although it's no longer used, we might as well keep it around.

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -708,9 +712,21 @@ bool endswith(std::string const &s, std::string const &suffix) {
...
+bool hasInstFluxUnits(FitsSchemaItem const & item) {
+    // helper lambda to make reading the real logic easier
+    auto includes = [](std::string const & s, char const * target) {
+        return s.find(target) != std::string::npos;
+    };
+    if (!includes(item.ttype, "flux")) return false;
+    // transform units to lowercase.
+    std::string units(item.tunit);
+    std::transform(units.begin(), units.end(), units.begin(), [](char c) { return std::tolower(c); } );
+    return includes(units, "count") || includes(units, "dn") || includes (units, "adu");
```
parejkoj: Oh joy: we use all of these units? Have we standardized this yet, or do we need an RFC for it?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -708,9 +712,21 @@ bool endswith(std::string const &s, std::string const &suffix) {
     return s.size() >= suffix.size() && s.compare(s.size() - suffix.size(), suffix.size(), suffix) == 0;
 }
 
-// Replace the last n characters of a string with a new suffix string, returning the result as a new string
-std::string replaceSuffix(std::string const &s, std::size_t n, std::string const &suffix) {
-    return s.substr(0, s.size() - n) + suffix;
+bool hasInstFluxUnits(FitsSchemaItem const & item) {
```
parejkoj: Maybe name this `isFluxField`, since it's doing more than just checking the units?

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: We certainly don't anymore.  I'm pretty sure that at one point we used `dn`, but that may predate `afw::table`, and I'm not certain we ever used `adu`.  I was just trying to be thorough.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: How about `isInstFluxField`?

TallJimbo:
> Did you revert the ps1-refcat-v1.fits file to AFW_TYPE=SOURCE?

I did not (yet).  Will do.

> I recommend deleting the existing sourceTable-v2.fits file and using the sourceCatalog-hsc file for all the v2 tests, since it's much less "fake".

Sorry, I forgot to mention that I tried this, and ran into trouble in that `sourceCatalog-hsc` uses `HsmSourceMoments` for the shape slot, and that algorithm happens not to provide uncertainties.  So I kept the old test data around instead of decreasing the test coverage.

> We should get a similar file for the v1 tests, since some of these problems could have been avoided if we had a "real" v1 file in here for testing. On the other hand, we probably won't care a year from now, so.

Yeah, for that reason I don't think this is worth the effort.

src/table/io/FitsSchemaInputMapper.cc
parejkoj: That works.

parejkoj:
You can just delete it, I don't think it's worth saving.

PR #400, opened 2018-10-05
DM-16023: Include alias maps in output schema when denormalizing matches

laurenam:


python/lsst/afw/table/catalogMatches.py
```diff
@@ -170,3 +174,37 @@ def matchesFromCatalog(catalog, sourceSlotConfig=None):
...
+    Parameters
+    ----------
+    inSchema : `lsst.afw.table.Schema`
+       The input schema whose `lsst.afw.table.AliasMap` is to be
+       copied to ``outSchema``.
+    outSchema : `lsst.afw.table.Schema`
+       The output schema into which the `lsst.afw.table.AliasMap`
+       from ``inSchema`` is to be copied (modified in place).
+    prefix : `str`, optional
+       An optional prefix to add to both sides of the alias names
```
isullivan: Here and in the description above, "both sides of the alias" is unclear. Maybe something like "... optionally appending a prefix to the name for both the input and output field entries"?

laurenam: I see your point, but I don't think referring to input/output really works (as that could be confused with the input/output schemas), so I reworded to:
```
An optional prefix to add to both the "from" and "to" names
of the alias (default is an empty string).
```

PR #401, opened 2018-10-08
DM-14363: make cameraGeom.Detector persistable with afw::table::io

TallJimbo:
Now rebased on and diffed against master.

### DIFF
--- a/include/lsst/afw/cameraGeom/TransformMap.h
+++ b/include/lsst/afw/cameraGeom/TransformMap.h
@@ -30,6 +30,7 @@
 #include "boost/iterator/transform_iterator.hpp"
 #include "astshim/FrameSet.h"
 
+#include "lsst/afw/table/io/Persistable.h"
 #include "lsst/afw/cameraGeom/CameraSys.h"
 #include "lsst/geom/Point.h"
 #include "lsst/afw/geom/Transform.h"
@@ -60,7 +61,7 @@ namespace cameraGeom {
  * @exceptsafe Unless otherwise specified, all methods guarantee only basic
  *             exception safety.
  */
-class TransformMap final {
+class TransformMap final : public table::io::PersistableFacade<TransformMap>, public table::io::Persistable {
 private:
 
     // Functor for boost::transform_iterator: given an entry in a std::map or unordered_map, return the key
@@ -161,9 +162,20 @@ public:
      */
     size_t size() const noexcept;
 
+
+    /**
+     * TransformMaps should always be Persistable.
+     */
+    bool isPersistable() const noexcept override { return true; }
+
 private:
 
-    // Private ctor, only called by Builder::build().
+    // Helper class used in persistence.
+    class Factory;
+
+    static Factory const registration;
+
+    // Private ctor, only called by Builder::build() and Factory.
     TransformMap(std::unique_ptr<ast::FrameSet> && transforms,
                  CameraSysFrameIdMap && frameIds,
                  std::vector<std::pair<int, int>> && canonicalConnections);
@@ -192,6 +204,13 @@ private:
      */
     std::shared_ptr<ast::Mapping const> _getMapping(CameraSys const &fromSys, CameraSys const &toSys) const;
 
+    std::string getPersistenceName() const override;
+
+    std::string getPythonModule() const override;
+
+    void write(OutputArchiveHandle& handle) const override;
+
+
     /// Allows conversions between LSST and AST data formats
     static lsst::afw::geom::Point2Endpoint _pointConverter;
 
@@ -210,7 +229,7 @@ private:
      * the TransformMap, so we can extract and save exactly those mappings
      * when persisting the TransformMap.
      */
-    std::vector<std::pair<int, int>> _canonicalConnections;
+    std::vector<std::pair<int, int>> const _canonicalConnections;
 };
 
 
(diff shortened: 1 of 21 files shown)


src/cameraGeom/TransformMap.cc
```diff
@@ -84,12 +88,160 @@ std::shared_ptr<ast::Mapping const> TransformMap::_getMapping(CameraSys const &f
...
+    handle.saveCatalog(sysCat);
+
+    auto connectionCat = handle.makeCatalog(keys.connectionSchema);
+    for (auto const & connectionPair : _canonicalConnections) {
+        auto connectionRecord = connectionCat.addNew();
+        connectionRecord->set(keys.from, connectionPair.first);
+        connectionRecord->set(keys.to, connectionPair.second);
+        auto transform = geom::TransformPoint2ToPoint2(
+            *_transforms->getMapping(connectionPair.first, connectionPair.second)
+        );
```
yalsayyad: weird formatting imo

src/cameraGeom/TransformMap.cc
```diff
@@ -259,10 +414,20 @@ std::shared_ptr<TransformMap const> TransformMap::Builder::build() const {
...
 
-
 }  // namespace cameraGeom
+
+namespace table {
+namespace io {
+
+template class PersistableFacade<cameraGeom::TransformMap>;
+
+} // namespace io
```
yalsayyad: Can't find the rule, but usee everywhere else in afw uses 2 spaces before `// namespace`

src/cameraGeom/TransformMap.cc
```diff
@@ -84,12 +88,160 @@ std::shared_ptr<ast::Mapping const> TransformMap::_getMapping(CameraSys const &f
...
+std::string makeFrameName(CameraSys const & sys) {
+    std::string r = "Ident=" + sys.getSysName();
+    if (sys.hasDetectorName()) {
+        r += "_";
+        r += sys.getDetectorName();
+    }
+    return r;
+}
+
+} // anonymous
```
yalsayyad: space space // namespace?

include/lsst/afw/cameraGeom/TransformMap.h
```diff
@@ -202,6 +223,13 @@ class TransformMap final {
      * Must have exactly one mapping for each Frame in `transforms`.
      */
     CameraSysFrameIdMap const _frameIds;
+
+    /**
+     * Sequence of from -> to frame IDs used to construct the TransformMap,
```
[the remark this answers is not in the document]
TallJimbo: Sorry; I should be overloading common prepositions with special meaning without at least quoting them.  How about "Sequence of `(fromFrameId, toFrameId)` that records the edges used to construct the `TransformMap`"?

src/cameraGeom/Detector.cc
```diff
@@ -143,6 +148,156 @@ std::vector<lsst::geom::Point2D> Detector::transform(std::vector<lsst::geom::Poi
...
+        table::AmpInfoCatalog amps(catalogs.back().getSchema());
+        amps.reserve(catalogs.back().size());
+        // we can't use amps.assign or amps.insert here because those
+        // require the input record to be a subclass of the output record
+        // to permit shallow assignment while this case is the opposte
+        for (auto const & amp : catalogs.back()) {
+            amps.addNew()->assign(amp);
+        }
+
+        auto flat = record.get(keys.crosstalk);
```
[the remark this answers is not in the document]
TallJimbo: Changed to `flattenedMatrix`.

src/cameraGeom/TransformMap.cc
TallJimbo: Huh.  I actually really like this form; it's the only way to get a really long single-argument in that I find readable (and it seems the style guide is silent on layout for *calling* functions).

I'm guessing you'd prefer just putting the closing `);` on the previous line, or were you expecting something else entirely?  (I think I must just have a strong aversion to not being able to find closing braces/brackets/parens).

src/cameraGeom/TransformMap.cc
TallJimbo: Done.

As a side note, I am a bit worried about the degree to which usage of clang-format's particular style seems to be becoming an implied MUST rather than the [MAY](https://developer.lsst.io/cpp/style.html#layout-may-be-automated-with-clang-format-using-the-lsst-configuration) that was actually RFC'd.  But this is not the hill I want my clang-format-free-workflow to die on, and I certainly can't pretend I have a good reason to differ from it on this particular case.

src/cameraGeom/Detector.cc
```diff
@@ -143,6 +148,156 @@ std::vector<lsst::geom::Point2D> Detector::transform(std::vector<lsst::geom::Poi
...
+    record->set(keys.bbox, getBBox());
+    record->set(keys.pixelSize, lsst::geom::Point2D(getPixelSize()));
+    auto orientation = getOrientation();
+    record->set(keys.fpPosition, orientation.getFpPosition());
+    record->set(keys.refPoint, orientation.getReferencePoint());
+    record->set(keys.yaw, orientation.getYaw());
+    record->set(keys.pitch, orientation.getPitch());
+    record->set(keys.roll, orientation.getRoll());
+    record->set(keys.transformMap, handle.put(getTransformMap()));
+    ndarray::Array<float, 2, 2> m = ndarray::copy(getCrosstalk());
```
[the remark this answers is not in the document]
TallJimbo: Yes, this is me trying and failing to get away with "short names in short scopes" without actually doing the work to make the surrounding code clear enough that the short name is sufficiently clear in context.

To that end, I've wrote a little one-use helper lambda to encapsulate the "flatten a matrix" operation and added some comments, in addition to using slightly longer names in it.

PR #404, opened 2018-10-11
DM-15767: remove usage of pex_policy

TallJimbo:


include/lsst/afw/formatters/Utils.h
```diff
@@ -75,50 +75,6 @@ bool extractOptionalFlag(std::shared_ptr<lsst::daf::base::PropertySet const> con
...
- * `"${itemName}.numSlices"` (a positive integer of type `int`) output tables exist and
- * are to be read in.
- *
- * @returns a list of table names
- * @see getTableName()
- */
-std::vector<std::string> getAllSliceTableNames(
-        std::shared_ptr<lsst::pex::policy::Policy const> const& policy,
-        std::shared_ptr<lsst::daf::base::PropertySet const> const& properties);
-
```
laurenam: Should the lines L55-57 also be removed?
```
namespace policy {
 class Policy;
```

include/lsst/afw/math/Random.h
```diff
@@ -126,23 +125,6 @@ class Random final {
...
-     * @param[in] policy    policy which contains the algorithm and seed to
-     *                      to use for random number generation
-     * @returns              a newly created random number generator
-     *
-     * @throws lsst::pex::exceptions::InvalidParameterError
-     *      Thrown if the requested algorithm is not supported.
-     * @throws std::bad_alloc
-     *      Thrown if memory allocation for internal generator state fails.
-     */
-    explicit Random(std::shared_ptr<pex::policy::Policy> const policy);
```
laurenam: Is this at L54-55 no longer valid?
```
... (or the "rngAlgorithm" and "rngSeed" keys in a policy)
```

python/lsst/afw/image/utils.py
```diff
@@ -69,29 +68,6 @@ def defineFilter(name, lambdaEff, lambdaMin=np.nan, lambdaMax=np.nan, alias=[],
...
-
-    filterPolicy.mergeDefaults(defPolicy.getDictionary())
-
-    for p in filterPolicy.getArray("Filter"):
-        Filter.define(FilterProperty(p.get("name"), p))
-        if p.exists("alias"):
-            for a in p.getArray("alias"):
-                Filter.defineAlias(p.get("name"), a)
-
-
```
laurenam: There are three uses of this in the **lsst/ctrl_prepdc3b** pakcage (whatever that is!):
https://github.com/lsst/ctrl_prepdc3b/blob/master/python/splitCfhtForDc3b2.py#L48
https://github.com/lsst/ctrl_prepdc3b/blob/master/python/splitCfhtForDc3b3.py#L55
https://github.com/lsst/ctrl_prepdc3b/blob/master/python/splitCfhtForDc3b4.py#L31

src/formatters/Utils.cc
```diff
@@ -159,33 +157,6 @@ bool extractOptionalFlag(std::shared_ptr<PropertySet const> const& properties, s
...
-    std::vector<std::string> names;
-    names.reserve(numSlices);
-    std::shared_ptr<PropertySet> props = properties->deepCopy();
-    for (int i = 0; i < numSlices; ++i) {
-        props->set("sliceId", i);
-        names.push_back(LogicalLocation(pattern, props).locString());
-    }
-    return names;
-}
-
```
laurenam: Some uses in **lsst/mops_nightmops**:
https://github.com/lsst/mops_nightmops/blob/master/src/MovingObjectPredictionFormatter.cc#L244
https://github.com/lsst/mops_nightmops/blob/master/src/MovingObjectPredictionFormatter.cc#L196

src/image/Filter.cc
```diff
@@ -191,7 +178,7 @@ void Filter::_initRegistry() {
     _nameMap = new NameMap;
     _idMap = new IdMap;
 
-    define(FilterProperty(unknownFilter, lsst::pex::policy::Policy(), true));
+    define(FilterProperty(unknownFilter, daf::base::PropertySet(), true));
```
laurenam: Why doesn't this one get an **lsst::** lead-in?  (not saying it's wrong...I just don't know!)

ups/afw.cfg
```diff
@@ -3,8 +3,8 @@
 import lsst.sconsUtils
 
 dependencies = {
-    "required": ["utils", "daf_persistence", "daf_base", "pex_exceptions", "geom", "log", "pex_config",
-                 "pex_policy", "eigen", "fftw", "ndarray", "numpy", "minuit2", "gsl", "cfitsio",
+    "required": ["utils", "daf_persistence", "daf_base", "pex_exceptions","geom", "log", "pex_config",
+                 "eigen", "fftw", "ndarray", "numpy", "minuit2", "gsl", "cfitsio",
                  "boost_filesystem", "boost_regex", "boost_serialization", "astshim", "sphgeom"],
```
laurenam: Looks like you accidentally removed the space before "geom".

include/lsst/afw/formatters/Utils.h
TallJimbo: Yes, good catch!

include/lsst/afw/math/Random.h
TallJimbo: Also a good catch - I think that whole paragraph has probably been invalid for a while, because those factory functions do not seem to exist.

python/lsst/afw/image/utils.py
TallJimbo: That package is definitely dead. I'm surprised we haven't moved it to `lsst-dm/legacy_` (I'm following up on this on slack).

src/image/Filter.cc
TallJimbo: In C++ when you're inside a namespace you don't usually need to use, so the `lsst::` was never necessary even on the old one.

PR #405, opened 2018-10-11
DM-16131: Add a unittest to address the issue of DM-16068

laurenam:


TallJimbo:
This was merged into tickets/DM-16068.

PR #410, opened 2018-10-29
DM-9938: Make some afw types hashable

kfindeisen:
This PR adds hash support to many value types in `afw` (defined as copyable types supporting `operator==`).

I've excluded types that define `operator==` in a non-leaf class, as the `std::hash` template would give inconsistent results unless we explicitly defined a specialization for every concrete subclass and possibly _every_ subclass. Hash support for these class hierarchies can still be added if somebody finds it useful later.

This PR requires lsst/utils#65, lsst/daf_base#45, and lsst/geom#17.

### DIFF
--- a/include/lsst/afw/cameraGeom/CameraSys.h
+++ b/include/lsst/afw/cameraGeom/CameraSys.h
@@ -71,7 +71,7 @@ public:
      *
      * @note Workhorse for std::hash<CameraSysPrefix>.
      */
-    size_t hash() const noexcept;
+    std::size_t hash_value() const noexcept;
 
 private:
     std::string _sysName;  ///< coordinate system name
@@ -145,7 +145,7 @@ public:
      *
      * @note Workhorse for std::hash<CameraSys>.
      */
-    size_t hash() const noexcept;
+    size_t hash_value() const noexcept;
 
 private:
     std::string _sysName;       ///< coordinate system name
@@ -211,12 +211,18 @@ std::ostream &operator<<(std::ostream &os, CameraSys const &cameraSys);
 namespace std {
 template <>
 struct hash<lsst::afw::cameraGeom::CameraSysPrefix> {
-    size_t operator()(lsst::afw::cameraGeom::CameraSysPrefix const &obj) const noexcept { return obj.hash(); }
+    using argument_type = lsst::afw::cameraGeom::CameraSysPrefix;
+    using result_type = size_t;
+    size_t operator()(lsst::afw::cameraGeom::CameraSysPrefix const &obj) const noexcept {
+        return obj.hash_value();
+    }
 };
 
 template <>
 struct hash<lsst::afw::cameraGeom::CameraSys> {
-    size_t operator()(lsst::afw::cameraGeom::CameraSys const &obj) const noexcept { return obj.hash(); }
+    using argument_type = lsst::afw::cameraGeom::CameraSys;
+    using result_type = size_t;
+    size_t operator()(lsst::afw::cameraGeom::CameraSys const &obj) const noexcept { return obj.hash_value(); }
 };
 }  // namespace std
 
--- a/include/lsst/afw/image/Pixel.h
+++ b/include/lsst/afw/image/Pixel.h
@@ -28,6 +28,8 @@
 #include <iostream>
 #include <type_traits>
 
+#include "lsst/utils/hashCombine.h"
+
 namespace lsst {
 namespace afw {
 namespace image {
@@ -231,6 +233,14 @@ public:
     template <typename T1>
     friend bool operator!=(Pixel const& lhs, T1 const& rhs) {
         return !(lhs == rhs);
+    }
+
+    /// Return a hash of this object.
+    std::size_t hash_value() const noexcept {
+        // Completely arbitrary seed
+        // Convert to double to avoid inconsistently hashing equal numbers of different types
+        return utils::hashCombine(17, static_cast<double>(image()), static_cast<double>(mask()),
+                                  static_cast<double>(variance()));
     }
 
     //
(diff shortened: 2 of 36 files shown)


include/lsst/afw/image/Pixel.h
```diff
@@ -233,6 +235,13 @@ class Pixel : public detail::MaskedImagePixel_tag {
         return !(lhs == rhs);
     }
 
+    /// Return a hash of this object.
+    std::size_t hash_value() const noexcept {
+        // Convert to double to avoid inconsistently hashing equal numbers of different types
```
TallJimbo: Are you trying to make image-like objects with different pixel types but the same values hash as equal?  I wouldn't have thought we had a use case for cross-type comparisons, but even if we do I'm not sure it's desirable (is that what our `operator==` does?!).

include/lsst/afw/table/arrays.h
```diff
@@ -125,6 +127,9 @@ class ArrayKey : public FunctorKey<ndarray::Array<T const, 1, 1> >,
     bool operator!=(ArrayKey<T> const& other) const noexcept { return !operator==(other); }
     //@}
 
+    /// Return a hash of this object.
+    std::size_t hash_value() const noexcept { return utils::hashCombine(17, _begin, _size); }
```
TallJimbo: The magic number `17` has now appeared often enough I'm wondering if it should be a named constant in `utils`, but I suppose that gets into a question about what `seed` ought to be that I've already asked there.

src/table/Schema.cc
```diff
@@ -705,6 +706,13 @@ int Schema::compare(Schema const &other, int flags) const {
     return result;
 }
 
+std::size_t Schema::hash_value() const noexcept {
+    std::size_t result = 17;
+    auto hasher = [&result](auto const &item) { result = utils::hashCombine(result, item.key); };
```
TallJimbo: I don't know how we survived before C++14 and universal lambdas :smile:.

include/lsst/afw/image/Pixel.h
```diff
@@ -652,4 +661,28 @@ std::ostream& operator<<(std::ostream& os, BinaryExpr<ExprT1, ExprT2, BinOp, Mas
...
+    using result_type = size_t;
+    size_t operator()(argument_type const& obj) const noexcept { return obj.hash_value(); }
+};
+
+template <typename ImagePixelT, typename MaskPixelT, typename VariancePixelT>
+struct hash<lsst::afw::image::pixel::SinglePixel<ImagePixelT, MaskPixelT, VariancePixelT>> {
+    using argument_type = lsst::afw::image::pixel::SinglePixel<ImagePixelT, MaskPixelT, VariancePixelT>;
+    using result_type = size_t;
+    size_t operator()(argument_type const& obj) const noexcept {
+        // Don't want to define SinglePixel::hash_value when SinglePixel::operator== doesn't exist
```
TallJimbo: In that case why are we defining `std::hash<SinglePixel<...>>`?  Is it used in hashing some other class somewhere I didn't spot?

include/lsst/afw/cameraGeom/CameraSys.h
```diff
@@ -71,7 +71,7 @@ class CameraSysPrefix final {
      *
      * @note Workhorse for std::hash<CameraSysPrefix>.
      */
-    size_t hash() const noexcept;
+    size_t hash_value() const noexcept;
```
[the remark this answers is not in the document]
kfindeisen: I prefer to not write `std::` when I don't have to... but looking more closely at this file, that shouldn't have compiled. Do we have a bad header or something that imports `std`? 😕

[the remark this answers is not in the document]
TallJimbo: Sadly, I don't think the C++ standard guarantees that the C standard library is not also available outside `std::`.

include/lsst/afw/image/Pixel.h
kfindeisen: Yup: https://github.com/lsst/afw/blob/master/include/lsst/afw/image/Pixel.h#L224-L228. This one actually kind of makes sense, especially if you're doing the comparison on the Python side -- programmers won't think of `Pixel<int>` and `Pixel<double>` as distinct types there.

In any case, one of the few hard rules for hash functions is that equal objects must have equal hashes, and apparently `4 == 4.0`. I've got a test case that covers this; we can remove it if we change the behavior of `operator==`.

src/table/Schema.cc
kfindeisen: Actually, I originally wrote this using an explicit functor similar to [`Stream`](https://github.com/lsst/afw/blob/master/src/table/Schema.cc#L727-L738). The only thing the universal lambda really does for you here is let you put the class template inside the function definition.

include/lsst/afw/image/Pixel.h
kfindeisen: Ugh, yes, the relationship between `Pixel` and `SinglePixel`... 😡

The problem (or, at least, the relevant one) is that `SinglePixel` *can* participate in equality tests -- but only on the right-hand side of `operator==(Pixel, T)`. And given all the hacking that went into making `Pixel` and `SinglePixel` interchangeable (e.g., [unsafe implicit conversions](https://github.com/lsst/afw/blob/tickets/DM-9938/include/lsst/afw/image/Pixel.h#L159-L165)), I wouldn't be surprised if there were some code *somewhere* that tries to mix them in ways where they might end up being hashed at the same time.

include/lsst/afw/table/arrays.h
kfindeisen: I'd actually like to avoid that, for fear that the number will get enshrined as something special. As noted in the `utils` thread, it does make sense to pick different seeds for classes that might get hashed together and would be collision-prone if you did so.

src/table/Schema.cc
TallJimbo: Yup, I just really like being able to do that :slightly_smiling_face:

PR #413, opened 2018-11-01
DM-16291: Use NumPy stringification to generate test results.

jdswinbank:
Previously, we were checking against hard-coded “guesses” of what NumPy might
return for a given array. That conflates two different tests:

- Are we correctly delegating formatting to NumPy?
- Is NumPy actually giving us what we want?

This modification means we are only testing the former, which, I suggest, is
appropriate. We could consider adding a test to verify that some future
version of NumPy doesn't change its output to something which is unhelpful to
us, but I suggest that that is out of scope for this work.

### DIFF
--- a/tests/test_maskedImage.py
+++ b/tests/test_maskedImage.py
@@ -817,7 +817,9 @@ class MaskedImageTestCase(lsst.utils.tests.TestCase):
         self.assertIn("image=", str(image))
         self.assertIn("mask=", str(image))
         self.assertIn("variance=", str(image))
-        self.assertIn("...", str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.image.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.mask.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.variance.dtype)), str(image))
         self.assertIn("bbox=%s"%str(image.getBBox()), str(image))
         self.assertIn("maskPlaneDict=%s"%str(image.mask.getMaskPlaneDict()), str(image))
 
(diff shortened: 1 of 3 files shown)


tests/test_maskedImage.py
```diff
@@ -817,6 +817,9 @@ def testString(self):
         self.assertIn("image=", str(image))
         self.assertIn("mask=", str(image))
         self.assertIn("variance=", str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.image.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.mask.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.variance.dtype)), str(image))
         self.assertIn("...", str(image))
```
kfindeisen: I think this line is obsolete now.

tests/test_maskedImage.py
```diff
@@ -817,6 +817,9 @@ def testString(self):
         self.assertIn("image=", str(image))
         self.assertIn("mask=", str(image))
         self.assertIn("variance=", str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.image.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.mask.dtype)), str(image))
+        self.assertIn(str(np.zeros((100, 100), dtype=image.variance.dtype)), str(image))
```
kfindeisen: I suspect that, as written, these three assertions are largely redundant (e.g., the image representation alone could cause all three to pass). Consider combining with the "image=" tests just above.

jdswinbank: I think that's not quite true, because the mask will be integral while the image and variance are floats. I do agree that the latter two are redundant, but I think it's clearer to include all thee.

kfindeisen: That depends on the type of `MaskedImage`; integer pixels *are* allowed. 🙂

PR #414, opened 2018-11-02
DM-16386: Update makeLimitedFitsHeader to force floats to correct type

timj:
In FITS a "0" is an integer so we must ensure that "0.0" becomes a floating point number when written as a FITS header card. This PR also includes some improvements to the tests and a fix to booleans.

### DIFF
--- a/include/lsst/afw/geom/detail/frameSetUtils.h
+++ b/include/lsst/afw/geom/detail/frameSetUtils.h
@@ -107,6 +107,20 @@ Copy values from an AST FitsChan into a PropertyList
 */
 std::shared_ptr<daf::base::PropertyList> getPropertyListFromFitsChan(ast::FitsChan& fitsChan);
 
+/**
+Construct AST FitsChan from PropertyList
+
+@param[in] metadata  Metadata to transfer; if this is a PropertyList then the order of items is preserved.
+@param[in] excludeNames  Names of entries to exclude from the returned header string.
+@param[in] options  Options string to pass to ast::FitsChan constructor.
+
+@return an ast::FitsChan representing this PropertyList
+
+*/
+ast::FitsChan getFitsChanFromPropertyList(daf::base::PropertySet& metadata,
+                                          std::set<std::string> const& excludeNames = {},
+                                          std::string options = "");
+
 }  // namespace detail
 }  // namespace geom
 }  // namespace afw
--- a/tests/test_makeLimitedFitsHeader.py
+++ b/tests/test_makeLimitedFitsHeader.py
@@ -21,6 +21,7 @@
 #
 
 import unittest
+import numpy as np
 
 import lsst.utils.tests
 from lsst.daf.base import PropertyList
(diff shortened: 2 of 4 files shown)


tests/test_makeLimitedFitsHeader.py
```diff
@@ -29,6 +29,30 @@
...
 
+    def assertHeadersEqual(self, header, expectedHeader, rtol=None):
+        """Compare 80 characters at a time"""
+        self.assertEqual(len(header), len(expectedHeader),
+                         msg="Compare header lengths")
+        start = 0
+        while start < len(header):
+            end = start + 80
+            # Strip trailing whitespace to make the diff clearer
+            with self.subTest("Testing individual headers"):
```
r-owen: I strongly suggest you provide the `from` and `to` strings as values for `subTest` instead of "Testing individual headers", e,g.

    this = header[start:end].rstrip()
    expected = expectedHeader[start:end].rstrip()
    with self.subTest(this=this, expected=expected):
        ...

That way you can just use assertEqual without a message (which is one big advantage to `subTest`).

I am also tempted to suggest you use `astropy.io.fits` for header comparison. The advantage is that the code would be more general (it would compare whether two headers were logically identical, without worrying about details that don't matter to FITS). That said, this code is simple enough and avoids using a 3rd party package, and since as you expect the headers to be truly identical then this is certainly acceptable and perhaps preferable.

tests/test_makeLimitedFitsHeader.py
```diff
@@ -63,17 +91,43 @@ def testBasics(self):
...
+
+        expectedLines = [  # without padding to 80 chars
+            "SINGLE  =              3.14159",
+            "SINGLEI =                  5.0",
+            "SINGLEE =             -5.9E+20",
+            "EXP     =               -5E+10",
+        ]
+        expectedHeader = "".join("%-80s" % val for val in expectedLines)
+
+        self.assertHeadersEqual(header, expectedHeader, rtol=1e-7)
```
r-owen: consider an rtol based on floating point information e.g. `rtol=np.finfo(np.float32).resolution` here and for double precision -- just to make it a bit clearer why you chose these values of rtol. A brief code comment would probably suffice, but finfo makes it truly obvious.

include/lsst/afw/geom/detail/frameSetUtils.h
```diff
@@ -107,6 +107,20 @@ Copy values from an AST FitsChan into a PropertyList
...
+@param[in] excludeNames  Names of entries to exclude from the returned header string.
+@param[in] options  Options string to pass to ast::FitsChan constructor.
+
+@return an ast::FitsChan representing this PropertyList
+
+*/
+std::shared_ptr<ast::FitsChan> getFitsChanFromPropertyList(daf::base::PropertySet& metadata,
+                                                           std::set<std::string> const& excludeNames = {},
+                                                           std::string options = "");
+
```
r-owen: I suggest you return a shared ptr instead of a FitsChan. Return value optimization should prevent copying and there is only one kind of FitsChan so polymorphism is not an issue. Or am I missing something?

timj: I was doing what was done with `getPropertyListFromFitsChan`.

r-owen: PropertyList is polymorphic so pointers are important. Not so for FitsChan.

r-owen: I withdraw the suggestion. Channels cannot be copied so it's safest not to pretend that's what we are doing here (even though it would not really be copied). I'm not even sure it would compile. What you have is likely better than what I suggested.

TallJimbo: I'm pretty sure this is a move, not a copy (and `FitsChan` does seem to be movable).  I think the argument that this would have been an elided copy would have been true in C++98 but is not true in C++11.

r-owen: @TallJimbo so do you have a preference? I think the pointer is a bit less obvious, but both work.

TallJimbo: Slight preference for returning by value.

timj: that would be dropping all the shared pointer stuff and returning `ast::FitsChan` directly?

r-owen: Yes.

PR #416, opened 2018-11-23
tickets/DM-16558: removeMaskPlane function in multiband.py does not work

laurenam:


python/lsst/afw/image/image/multiband.py
```diff
@@ -484,10 +484,31 @@ def removeMaskPlane(name):
...
+        ----------
+        name : `str`
+           Name of the mask plane to remove
+        removeFromDefault : `bool`, optional
+           Whether to remove the mask plane from the default dictionary.
+           Default is `False`.
+        """
+        # Clear all masks in MultibandMask but leave in default dict for now
+        for single in self.singles:
+            single.removeAndClearMaskPlane(name, False)
```
isullivan: I think `single.removeAndClearMaskPlane(name, removeFromDefault=False)` would be more clear.

python/lsst/afw/image/image/multiband.py
```diff
@@ -484,10 +484,31 @@ def removeMaskPlane(name):
...
+        # Now remove from default dict according to removeFromDefault
+        self._refMask.removeAndClearMaskPlane(name, removeFromDefault)
```
isullivan: Would `self.Mask().removeAndClearMaskPlane(name, removeFromDefault)` work?

laurenam: I’m not entirely sure, but I think not as all the non-static methods this class do it this way (e.g. **clearAllMaskPlanes**, **getMaskPlaneDict**, **getNumPlanesUsed**, etc...), and **removeAndClearMaskPlane** is a non-static method: 
https://github.com/lsst/afw/blob/master/include/lsst/afw/image/Mask.h#L477

laurenam: To be sure, I ran this by @TallJimbo and he concurred.

PR #420, opened 2018-12-14
DM-16650: Change PhotoCalib defintion to nanojansky

parejkoj:


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -167,84 +175,81 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
 
-    /// @overload Measurement instFluxToMaggies(double, double, lsst::geom::Point<double, 2> const &) const
-    Measurement instFluxToMaggies(double instFlux, double instFluxErr) const;
+    /// @overload Measurement instFluxToNanojansky(double, double, lsst::geom::Point<double, 2> const &) const
+    Measurement instFluxToNanojansky(double instFlux, double instFluxErr) const;
 
     /**
      * Convert `sourceRecord[instFluxField_instFlux]` (ADU) at location
-     * `(sourceRecord.get("x"), sourceRecord.get("y"))` (pixels) to maggies and maggie error.
+     * `(sourceRecord.get("x"), sourceRecord.get("y"))` (pixels) to nJy and magnitude error.
```
erykoff: This is nJy error, not magnitude error, right?

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -167,84 +175,81 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
-     * @returns    The flux in maggies.
+     * @returns    The flux in nJy.
      */
-    double instFluxToMaggies(double instFlux, lsst::geom::Point<double, 2> const &point) const;
+    double instFluxToNanojansky(double instFlux, lsst::geom::Point<double, 2> const &point) const;
 
-    /// @overload instFluxToMaggies(double, lsst::geom::Point<double, 2> const &) const;
-    double instFluxToMaggies(double instFlux) const;
+    /// @overload instFluxToNanojansky(double, lsst::geom::Point<double, 2> const &) const;
+    double instFluxToNanojansky(double instFlux) const;
```
erykoff: I know this was discussed on the RFC, but now seeing "nJy" as the units and "instFluxToNanojansky()" as the camelCase method, I don't like it.  Is "ToNanoJansky()" really not an option?

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -74,30 +75,37 @@ inline void assertNonNegative(double value, std::string const &name) {
...
+ * The spatially varying flux calibration has units of nJy/ADU, and is defined such that,
  * at a position (x,y) in the domain of the boundedField calibration and for a given measured source instFlux:
  * @f[
- *     instFlux*calibration(x,y) = flux [maggies]
+ *     instFlux*calibration(x,y) = flux [nJy]
  * @f]
  * while the errors (constant on the domain) are defined as:
  * @f[
- *     sqrt((instFluxErr/instFlux)^2 + (calibrationErr/calibration)^2)*flux = fluxErr [maggies]
+ *     sqrt((instFluxErr/instFlux)^2 + (calibrationErr/calibration)^2)*flux = fluxErr [nJy]
```
erykoff: I know this is just updated from before, but blindly adding the calibration error to the instFlux errors (in quadrature) is not often the right thing to do, because of local correlations.  Repeated observations that have been cross-calibrated are not independent samples from the calibrationErr.  Anyway, this is definitely beyond the scope of this ticket.

include/lsst/afw/image/PhotoCalib.h
parejkoj: If you're asking for `instFluxToNanoJansky` here, that's truly not a valid name: the unit name is `jansky` and the abbreviation is `Jy`: https://en.wikipedia.org/wiki/Jansky

If you meant `toNanojansky`, with no prefix, I don't like that because it doesn't tell you what the input to the method is (a significant flaw in the existing `Calib` object).

erykoff: Welp, I guess it just will have to look odd and asymmetric.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -386,25 +387,25 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
     double getCalibrationErr() const { return _calibrationErr; }
 
     /**
-     * Get the magnitude zero point (equal to 1 / mean(calibration)).
+     * Get the magnitude zero point (the instrument flux corresponding to 0 magnitude).
```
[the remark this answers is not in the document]
parejkoj: Hmmm... I was going from what I thought the definition of `Calib.getFluxMag0()` was, which is the "instFlux at magnitude 0". I'm not sure I fully understand the meaning you give in your first sentence.

src/image/PhotoCalib.cc
```diff
@@ -85,55 +89,57 @@ double toMagnitudeErr(double instFlux, double instFluxErr, double scale, double
...
+Measurement PhotoCalib::instFluxToNanojansky(double instFlux, double instFluxErr,
+                                             lsst::geom::Point<double, 2> const &point) const {
+    double calibration, err, nanojansky;
     calibration = evaluate(point);
-    maggies = toMaggies(instFlux, calibration);
-    err = toMaggiesErr(instFlux, instFluxErr, calibration, _calibrationErr, maggies);
-    return Measurement(maggies, err);
+    nanojansky = toNanojansky(instFlux, calibration);
+    err = toNanojanskyErr(instFlux, instFluxErr, calibration, _calibrationErr, nanojansky);
+    return Measurement(nanojansky, err);
```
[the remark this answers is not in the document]
parejkoj: This brings up a very good question: what *should* we call the error part of the `value, err[or]` pair in the `Measurement` object. I changed it to `error` as part of this work, because `value, error` seems more sensible to me than `value, err`, but we call the error component `blahErr` everywhere else, so I'm really not sure about it.

I think you're right though, that if we go with `error`, this one should be `error` here.

include/lsst/afw/image/PhotoCalib.h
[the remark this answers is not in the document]
erykoff: I was just confused.

src/image/PhotoCalib.cc
```diff
@@ -45,38 +46,41 @@ namespace image {
...
-double toMaggies(double instFlux, double scale) { return instFlux * scale; }
+double toNanojansky(double instFlux, double scale) { return instFlux * scale; }
 
-double toMagnitude(double instFlux, double scale) { return -2.5 * log10(instFlux * scale); }
+double toMagnitude(double instFlux, double scale) { return utils::nanojanskyToABMagnitude(instFlux * scale); }
 
-double fromMagnitude(double magnitude, double scale) { return pow(10, magnitude / -2.5) / scale; }
+double fromMagnitudeToInstFlux(double magnitude, double scale) {
+    return utils::ABMagnitudeToNanojansky(magnitude) / scale;
+}
```
erykoff: A quick look at the code, I see `ABMagnitudeToNanojansky` but this is converting to `instFlux` not `nJy`!  Then I notice the scale.  I think that a code comment here would be helpful for future readers.  (As above with the `toMagnitude`).  But why is that not `toMagnitudeFromInstFlux` for symmetry?

tests/test_photoCalib.py
```diff
@@ -33,23 +35,29 @@
 import lsst.pex.exceptions
 
 
-def computeMaggiesErr(instFluxErr, instFlux, calibrationErr, calibration, flux):
-    """Return the error on the flux (Maggies)."""
+def nJyToMagnitude(flux):
+    """Return an AB magnitude given a flux in nanojansky."""
+    return (flux*units.nJy).to(units.ABmag).value
+
```
erykoff: This routine is the only obvious missing piece.  I know it's defined in the header in utils, and it's here in the tests ... but it's not really transparent to the user on how to do this conversion in the stack.  That is, even a pure python pair of convenience routines that simply calls astropy (`nJyToMagnitude(flux)` and `abMagnitudeToNanojansky(mag)` or something, it's hard to do with the camel casing...) that are accessible from `lsst.afw.image` would be very useful.  These would replace what is currently there, `afwImage.abMagFromFlux()` and `afwImage.fluxFromABMag()` via `calib.cc`, which are unitless and confusing, but convenient.

tests/test_photoCalib.py
```diff
@@ -65,11 +73,18 @@ def setUp(self):
...
         self.calibration = 1e-3
         self.calibrationErr = 1e-4
         self.instFlux = 1000.
         self.instFluxErr = 10.
+        self.flux = 1.0
+        self.mag = nJyToMagnitude(self.flux)
+
+        # useful reference points: 575.44 nJy ~= 24.5 mag, 3630.78 * 10^9 nJy ~= 0 mag
+        self.flux2 = 575.439937337159
+        self.instFlux2 = self.instFlux*self.flux2
```
erykoff: I don't see any reference to `instFlux1` in the code, and if `instFlux1` was a calibrated flux of 1, then `instFlux2` should be a calibrated flux of 2...?

tests/test_photoCalib.py
```diff
@@ -119,87 +138,106 @@ def _testPhotoCalibCenter(self, photoCalib, calibrationErr):
...
+        self.assertFloatsAlmostEqual(self.mag, result.value)
+        self.assertFloatsAlmostEqual(errMag, result.error)
+
+        # test that we get a correct nJy err for flux2
+        errFlux2 = computeNanojanskyErr(self.instFluxErr,
+                                        self.instFlux2,
+                                        calibrationErr,
+                                        self.calibration,
+                                        self.flux2)
+        result = photoCalib.instFluxToNanojansky(self.instFlux2, self.instFluxErr)
```
erykoff: I worry this test is a bit circular ... it's comparing the version of the algorithm in the test file to that in the main code I guess?  Is there any verification that they are correct, or just match? (similar below)

tests/test_photoCalib.py
parejkoj: I was actually thinking to remove this method entirely and just use astropy directly in this test. The method only exists here because I originally wrote the math in it, before realizing I could just delegate it to astropy directly.

parejkoj:
> ...convenience functions for converting from AB magnitude to and from nJy which is a common operation...

Is it that common, though? I count a total of 10 calls to `abMagFromFlux`/`fluxFromABMag` together, outside of the existing `Calib` object. I'd suggest we just replace those (as part of the "replace Calib" work) with `astropy.units`, without making an astropy wrapper.

tests/test_photoCalib.py
erykoff: So I can't reply to the comment below for some reason.  Anyway, I think this would be useful, if only because the way to do it is non-obvious (sorry astropy).  But since dm is only releasing fluxes, the common user will want to do this, yes.  But will they use the stack to do it is the relevant question.

tests/test_photoCalib.py
parejkoj: I'm not sure of another approach to this test: error calculations like this aren't in astropy, so I don't know of an external way to verify it.

tests/test_photoCalib.py
parejkoj: You have no idea how long I dithered on this one. I can change `instFlux` to `instFlux1` if you think that's better: I just wanted two different values to test, one being "trivial" and the other not so trivial. Should I call the second one `flux575`?

erykoff: I think "1" and "2" (or even "0" and "1") are fine, but the comment should match the name of the variable.

src/image/PhotoCalib.cc
parejkoj: I've renamed it and add a hopefully clarifying comment.

parejkoj:
@erykoff : the very last commit I pushed (after a bunch of cleanups and rebase-squash force pushes) has the version updated. Would you please take a quick look at it?

I'm going to write a Community post noting the compatibility break. I think this is the last thing, other than the final Jenkins run which is now going.

erykoff:
I'm not sure about this one actually.  Does this mean the code will raise a RuntimeError if we try to read an old `photoCalib` file after this change?  If we know it's the old version in maggies, can't we do the conversion on read?

parejkoj:
Yes: this will raise if you try to ingest an old photoCalib with a version of the stack after this is merged. @TallJimbo and @jdswinbank both reluctantly agreed with this approach.

Given that we don’t really have a versioning system for anything, I personally think we're better off making this a breaking change here, and making this version the version to maintain compatibility with in the future. The number of existing uses of PhotoCalib is still small, but will soon not be.

We can convert the mean calibration on read, but not the bounded field, which is where the real trouble lies.

src/image/PhotoCalib.cc
```diff
@@ -297,8 +297,9 @@ class PhotoCalibFactory : public table::io::PersistableFactory {
         table::BaseRecord const &record = catalogs.front().front();
         PhotoCalibSchema const &keys = PhotoCalibSchema::get();
         int version = getVersion(record);
-        if (version != 0) {
-            throw(pex::exceptions::TypeError("Unsupported version: " + std::to_string(version)));
+        if (version < 1) {
+            throw(pex::exceptions::RuntimeError("Unsupported version (version 0 was defined in maggies): " +
+                                                std::to_string(version)));
```
TallJimbo: Would there be any value to instead doing
```
auto mean = record.get(keys.calibrationMean);
auto err = record.get(keys.calibrationErr);
auto field = archive.get<afw::math::BoundedField>(record.get(keys.field));
if (version < 1) {
    field = (*field) * MAGGIES_TO_nJy;  // or whatever you've called it
    mean *= MAGGIES_TO_nJy;
    err *= MAGGIES_TO_nJy;
}
```

I think that question boils down to:
 - do `PhotoCalib`s in the wild actually implement `operator=`?
 - does this do the wrong thing anyway with version 0 `PhotoCalib`s in the wild that use meas_mosaic's `FluxFitBoundedField`?

Anyhow, if it *is* just as easy as the above, let's do the backwards compatibility.  But it's not worth lots of extra work if it isn't.

parejkoj: The new version of `FluxFitBoundedField` has the conversion factor baked into it, so it would have to be special cased.

jdswinbank: I don't know the code well enough (read: at all), so I'm don't know how messy that special casing would be in practice — but if it's really just an `if` block with Jim's code in it, then it sounds like something we should do.

parejkoj: `FluxFitBoundedField` is defined in meas_mosaic, which means we can't make a special case for it, now that I think about it (dependency goes the wrong way).

erykoff: My thought was that we could check for that type and only raise there, and convert what we can.

parejkoj: The only thing I can think of that really matters is Dominique's pre-DC2 jointcal runs. Those are easy enough to redo.

Remember: the meas_mosaic PhotoCalib object is not used for any further processing that I'm aware of (e.g., the ffp object is still used to make coadds, even after DM-15751).

erykoff: There are no FGCM-produced PhotoCalibs that are out in the wild, and it would only take a few minutes to regenerate them anyway (since they're produced at the last step of processing). So if we can give a heads up to Dominique (or maybe update to slack somewhere) that might be useful (this would be before merger and the post-merger community post).

PR #422, opened 2018-12-15
DM-16830: Add versioning to PhotoCalib persistence

parejkoj:
All existing files are version 0: setting up versioning now will let the nanojansky version in DM-16650 (#420 ) become version 1 explicitly.


src/image/PhotoCalib.cc
```diff
@@ -285,6 +289,18 @@ class PhotoCalibFactory : public table::io::PersistableFactory {
...
+            std::string versionName = "version";
+            auto versionKey = record.getSchema().find<int>(versionName);
+            version = record.get(versionKey.key);
+        } catch (const pex::exceptions::NotFoundError &) {
+            // un-versioned files are version 0
+            version = 0;
+        }
+        if (version != 0) {
+            throw(pex::exceptions::TypeError("Unsupported version: " + std::to_string(version)));
+        }
```
TallJimbo: This block of new code seems like a good candidate for a separate method; even if it's only used once, just having a call like `int version = getVersion(record);` in `read` will improve code clarity a bit.

PR #424, opened 2018-12-19
DM-16856: Convert afw.coord to numpydoc

kfindeisen:
This PR adds the Python code in `lsst.afw.coord` to the documentation build, and rearranges the documentation to conform to current guidelines.


python/lsst/afw/coord/refraction.py
```diff
@@ -156,9 +170,19 @@ def densityFactorDry(weather):
 
     Returns
     -------
-    `float`
+    factor : `float`
```
isullivan: densityFactor would be preferable, here and for the next function.

PR #429, opened 2019-01-11
DM-16872: Fix numpy warnings in afw

kfindeisen:
This PR eliminates the following three warnings from `afw` unit tests:

> PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices  or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.

> FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.

> FutureWarning: `rcond` parameter will change to the default of machine precision times `max(M, N)` where M and N are the input matrix dimensions.
To use the future default and silence this warning we advise to pass `rcond=None`, to keep using the old, explicitly pass `rcond=-1`.

The last gave two options for fixing the warning; I chose `rcond=-1` to ensure continuity of results.

Changes have been tested with Numpy 1.14 and Numpy 1.15 and runs are warning-free in both cases.


kfindeisen:
`rcond=None` works as well (not surprising, given that the warning implies it's a lower tolerance).

PR #431, opened 2019-01-25
DM-17451: raise when columnviews access undefined keys

jdswinbank:


src/table/BaseColumnView.cc
```diff
@@ -69,6 +69,11 @@ std::shared_ptr<BaseTable> BaseColumnView::getTable() const { return _impl->tabl
 
 template <typename T>
 typename ndarray::ArrayRef<T, 1> const BaseColumnView::operator[](Key<T> const &key) const {
+    if (!key.isValid()) {
+        throw LSST_EXCEPT(
+            pex::exceptions::LogicError,
+            "Key is not valid (if this is a SourceCatalog, make sure slot aliases have been set up).");
```
parejkoj: Why `SourceCatalog` here? Can slots not be set on `SimpleCatalog`s? What other types of catalogs can or cannot have slots?

tests/test_simpleTable.py
```diff
@@ -271,6 +271,14 @@ def testColumnView(self):
             self.assertEqual(catalog[i].get(kI), 4)
             self.assertEqual(catalog[i].get(kD), f3v[i])
 
+        # Accessing an invalid key should raise.
+        for keyType in ["Angle", "ArrayB", "ArrayD", "ArrayF", "ArrayI",
+                        "ArrayU", "B", "D", "F", "Flag", "I", "L", "U"]:
+            # Default-constructed key is invalid
+            invalidKey = getattr(lsst.afw.table, f"Key{keyType}")()
```
parejkoj: Neat use of format strings. I need to use those more.

tests/test_simpleTable.py
```diff
@@ -271,6 +271,14 @@ def testColumnView(self):
...
+            self.assertFalse(invalidKey.isValid())
+            self.assertRaises(lsst.pex.exceptions.LogicError, catalog.get, invalidKey)
```
parejkoj: I find the context manager version easier to understand the intent:

```
with self.assertRaises(LogicError):
    catalog.get(invalidKey)
```

src/table/BaseColumnView.cc
```diff
@@ -69,6 +69,11 @@ std::shared_ptr<BaseTable> BaseColumnView::getTable() const { return _impl->tabl
 
 template <typename T>
 typename ndarray::ArrayRef<T, 1> const BaseColumnView::operator[](Key<T> const &key) const {
+    if (!key.isValid()) {
+        throw LSST_EXCEPT(
+            pex::exceptions::LogicError,
```
parejkoj: Should this raise `LogicError` or `KeyError`?

tests/test_sourceTable.py
```diff
@@ -639,7 +643,8 @@ def _testFluxSlot(self, slotName):
                                  type=np.float64, doc="flux uncertainty")
         flagKey = schema.addField("%s_flag" % (baseName,),
                                   type="Flag", doc="flux flag")
-        table = lsst.afw.table.SourceTable.make(schema)
+        catalog = lsst.afw.table.SourceCatalog(schema)
+        table = catalog.table
```
parejkoj: I wish someone would explain to me the difference between an Table and a Catalog...

tests/test_sourceTable.py
```diff
@@ -258,6 +258,10 @@ def testColumnView(self):
         self.assertFloatsEqual(cols2["c_xx"], cols2.getIxx())
         self.assertFloatsEqual(cols2["c_yy"], cols2.getIyy())
         self.assertFloatsEqual(cols2["c_xy"], cols2.getIxy())
+        self.catalog.table.schema.getAliasMap().erase("slot_Centroid")
+        self.catalog.table.schema.getAliasMap().erase("slot_Shape")
+        for quantity in ["X", "Y", "Ixx", "Iyy", "Ixy"]:
+            self.assertRaises(lsst.pex.exceptions.LogicError, getattr(self.catalog, f"get{quantity}"))
```
parejkoj: Again with the `with`. Please add a comment above this explaining what this test is doing: it took me a bit to figure it out.

tests/test_sourceTable.py
```diff
@@ -662,13 +667,22 @@ def _testFluxSlot(self, slotName):
...
         # And we should be able to delete the slot, breaking the mapping.
         table.schema.getAliasMap().erase("slot_%s" % (slotName,))
         self.assertFalse(getattr(table, "get%sSlot" % (slotName,))().isValid())
         self.assertNotEqual(getattr(table, "get%sSlot" % (slotName,))().getMeasKey(), instFluxKey)
         self.assertNotEqual(getattr(table, "get%sSlot" % (slotName,))().getErrKey(), errKey)
         self.assertNotEqual(getattr(table, "get%sSlot" % (slotName,))().getFlagKey(), flagKey)
 
+        with self.assertRaises(lsst.pex.exceptions.LogicError):
+            getattr(catalog, f"get{instFluxName}")()
+        with self.assertRaises(lsst.pex.exceptions.LogicError):
```
parejkoj: Why should these two raise? I would have thought they would work?

tests/test_sourceTable.py
jdswinbank: Roughy speaking (and I'm sure Jim or somebody else who knows better will be along to correct me in a moment):

- Tables are (confusingly named) record factories.
- Catalogs are containers of records.

tests/test_sourceTable.py
jdswinbank: These are referring to keys which are defined as part of the slot. When the slot is erased (at line 675, above), all the keys defined as part of it are invalidated. Accessing via those invalid keys raises.

I'll add a comment to that effect.

tests/test_simpleTable.py
timj: The testing guide prefers the context manager. Currently about 90% of the LSST code bases complies with that statement.

https://developer.lsst.io/python/testing.html?highlight=assertraises

PR #433, opened 2019-02-05
Add helper method for IdFactory.

TallJimbo:
This encapsulates the number of (usable) bits in a record ID in order to avoid having "64" or "63" magic numbers strewn about downstream code.

### DIFF
--- a/include/lsst/afw/table/IdFactory.h
+++ b/include/lsst/afw/table/IdFactory.h
@@ -48,6 +48,17 @@ public:
      */
     static std::shared_ptr<IdFactory> makeSource(RecordId expId, int reserved);
 
+    /**
+     * Return the number to pass as the 'reserved' argument to makeSource for
+     * an exposure ID with the given maximum number of bits.
+     *
+     * @param[in]  maxBits  The maximum number of bits an exposure ID can have.
+     */
+    static int computeReservedFromMaxBits(int maxBits) {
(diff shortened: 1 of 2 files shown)


include/lsst/afw/table/IdFactory.h
```diff
@@ -48,6 +48,17 @@ class IdFactory {
      */
     static std::shared_ptr<IdFactory> makeSource(RecordId expId, int reserved);
 
+    /**
+     * Return the number to pass as the 'reserved' argument to makeSource for
+     * an exposure ID with the given maximum number of bits.
+     *
+     * @param[in]  maxBits  The maximum number of bits an exposure ID can have.
```
timj: "exposure ID" does confuse me at times since I sometimes see this used ambiguously.  Is this "detector exposure ID" or "exposure ID"? The camera mappers define nbits for detector exposure ID and not exposure ID.  Does the system complain if mapper nbits exceeds this maxBits?

TallJimbo: This is sometimes detector_exposure_id and sometimes it has nothing at all to do with exposures (except maybe `afw.image.Exposure` of a coadd).

Agree that it's confusing, but it's a much bigger issue than this ticket.

TallJimbo: Oh, and yes, you will get an exception if the numbers of bits don't add up.

PR #435, opened 2019-02-08
DM-10346: Allow depersist of old Calib objects as PhotoCalib

parejkoj:


tests/data/makeTestExposure.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+nx = 10
+ny = 10
+exp = lsst.afw.image.ExposureF(nx, ny)
+
+# Fill the maskedImage
+exp.maskedImage.image.array = np.arange(nx*ny, dtype='f').reshape(nx, ny)
+exp.maskedImage.variance.array = np.ones((nx, ny), dtype='f')
+exp.maskedImage.mask.array[5, 5] = 5  # make one pixel non-zero
```
TallJimbo: What you have here is not correct, but e.g. `exp.image.array` works, too.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -488,6 +489,19 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
+/**
+ * Construct a PhotoCalib from FITS FLUXMAG0/FLUXMAG0ERR keywords.
+ *
+ * This provides backwards compatibility with the obsoleted Calib object that PhotoCalib replaced.
+ * It should not be used outside of reading old Exposures written before PhotoCalib existed.
+ *
+ * @param metadata FITS header metadata containing FLUXMAG0 and FLUXMAG0ERR keys.
+ * @param strip Strip FLUXMAG0 and FLUXMAG0ERR from `metadata`?
+ *
+ * @throws lsst::pex::exceptions::TypeError if the metadata does not contain FLUXMAG0.
```
TallJimbo: I think just returning `nullptr` might be a more natural way to express this condition.

In fact, it looks like the old `Calib` code handled this case by returning a valid `Calib` initialized with zeros.  I'm by no means convinced that was a good idea, but if we do change that behavior, we should be prepared for breakage in places we didn't expect (though it may be that you're already dealt with that in a part of this ticket I haven't gotten to yet).

include/lsst/afw/image/ExposureInfo.h
```diff
@@ -197,6 +197,12 @@ class ExposureInfo final {
     /// Set the exposure's transmission curve.
     void setTransmissionCurve(std::shared_ptr<TransmissionCurve const> tc) { _transmissionCurve = tc; }
 
+    /// Get the version of FITS serialization that this ExposureInfo understands.
+    static const int getFitsSerializationVersion() { return _fitsSerializationVersion; }
+
+    /// Get the version of FITS serialization version info name
+    static const std::string getFitsSerializationVersionName() { return _fitsSerializationVersionName; }
```
TallJimbo: This should return `std::string const &` (or just `std::string`) and the one above should just return `int`; there's no reason to add `const` to a by-value return type.

I would also recommend reimplementing these to use function-local static variables, instead of a static class data member.  In other words, you'd declare but not define these functions here, and then in `ExposureInfo.cc`, you'd have e.g.:
```
std::string const & ExposureInfo::getFitsSerializationVersionName() {
    static std::string const versionName("EXPINFO_V");
    return versionName;
}
```

What you've done here is not incorrect, and in fact it's very unlikely to cause problems.  But it leaves the door slightly open for one of the trickier and more obscure ways to get C++ to segfault: the [static initialization order problem](https://isocpp.org/wiki/faq/ctors#static-init-order), while what I've recommended instead closes that door.  (You may also note that link advocates a slightly different solution, which closes the door even more securely but makes `valgrind` generate spurious warnings; I'm okay with either but personally prefer to keep valgrind output clean in this case.)

src/image/ExposureFitsReader.cc
```diff
@@ -44,6 +44,17 @@ class ExposureFitsReader::MetadataReader {
...
+        int version;
+        if (primaryMetadata->exists(ExposureInfo::getFitsSerializationVersionName())) {
+            version = primaryMetadata->getAsInt(ExposureInfo::getFitsSerializationVersionName());
+            primaryMetadata->remove(ExposureInfo::getFitsSerializationVersionName());
+        } else {
+            version = 0;  // unversioned files are implicitly version 0
+        }
+        if (version > ExposureInfo::getFitsSerializationVersion()) {
+            throw LSST_EXCEPT(pex::exceptions::TypeError, "Cannot read Exposure FITS version >= MAX_VERSION");
+        }
```
TallJimbo: I'm a bit torn about whether we should throw this exception here; doing so declares that there can never be a forward-compatible version increment, while it seems that being forward-compatible is something we could do if it happens to be easy.  The alternative would be to defer the exception until the version compatibility caused a problem, and perhaps even then set an unrecognized component to empty and warn (when possible), if doing so permits the rest of the exposure to be read in.  That's what we already do when (e.g.) the PSF component cannot be read because the package that defines it wasn't `setup`, and it's what usually happens now (without the warning) if you currently use old code to read a newer `Exposure` that has an unrecognized component.

That may not be possible when we're talking about replacing a component rather than adding one, but maybe we should just warn here until we run into a case where we have to throw?

tests/data/makeTestExposure.py
parejkoj: "is not correct": did you mean "is not incorrect"?

include/lsst/afw/image/PhotoCalib.h
parejkoj: I took the idea of raising from SkyWcs:

``` * @throws lsst::pex::exceptions::TypeError if the metadata does not describe a celestial WCS.
```

Can we try to do the Calib->PhotoCalib switch with the raise in place and then see if it does break anything? I would guess that the only place it would cause a problem are very, very old files that may be broken/invalid in other ways.

Returning nullptr feels more dangerous to me, since calling code then has to check on what it got.

src/image/ExposureFitsReader.cc
parejkoj: If I keep my promises, the next version of ExposureInfo will replace Calib with PhotoCalib and will thus make forward compatibility impossible from 0->1. If a particular forward-compatibility is possible, we could generalize this code; say, make it maxVersion+1 for that particular change. I feel like we are causing ourselves problems by always trying to "work but warn" and then having things fail much further down then line, when they should have failed early (i.e. DM-14310).

I wonder if we should bring in @leannep or @jdswinbank for management-level commentary?

tests/data/makeTestExposure.py
TallJimbo: Yikes, sorry, yes, that's what I meant; fixed.

src/image/ExposureFitsReader.cc
TallJimbo: They're of course welcome to comment, but I'm happy to go with your preference on this; I certainly agree that the next anticipated version bump will not permit this kind of forward compatibility, and that's fine.

include/lsst/afw/image/PhotoCalib.h
TallJimbo: Fine with me.

My reasoning was that we'd actually (like WCS, I think?) run this on all files, and then overwrite it if we found a `Calib` or `PhotoCalib` in the afw::table::io binary tables, and I didn't want to have exceptions thrown in the usual case.  But that's an assumption about what we'll do on a later ticket.

In any case, note, that what matters in most cases is whether `Exposure.getCalib()` returns a `nullptr` when you couldn't read one, and I think that will be the case regardless of whether you throw or return `nullptr` here.

parejkoj: > My reasoning was that we'd actually (like WCS, I think?) run this on all files, and then overwrite it if we found a `Calib` or `PhotoCalib` in the afw::table::io binary tables...

I'm not sure I understand: I thought the purpose of this was to read old `Calibs`, but not modify the files.

> whether you throw or return `nullptr` here.

Throwing means the calling code is immediately informed of a problem. But yes, we can keep `Exposure.getCalib()/Exposure.getPhotoCalib()` returning nullptr/None if it couldn't be read, which I think is better than initializing to 0.

TallJimbo: I meant "overwrite" in the in-memory `Exposure`, but I just realized I was talking about logic that I changed a couple of months ago - so "overwrite" is definitely the wrong word to use (sorry about the confusion!), but I think my point (which is a very minor one!) remains.

To be more explicit, the current logic in `ExposureFitsReader` [first tries to extract a `SkyWcs` from the metadata, inside a try block](https://github.com/lsst/afw/blob/master/src/image/ExposureFitsReader.cc#L50).  If that method throws, the data member that holds the `wcs` remains null.  If `makeSkyWcs` had returned null directly, then we'd have avoided using exceptions for non-exceptional-case control flow, which is a good thing (and maybe be better able to use exceptions to signify the genuinely exceptional case where there appears to be a WCS in the header but it's corrupted).  And I think return null there would match how we normally treat exposure components in other contexts - we don't expect `Exposure.getWcs()` to throw if there is no `Wcs`; we expect it to return null/`None`.

Anyhow, regardless of whether or not that works, `ExposureFitsReader` will [later try to read a `SkyWcs` from the afw::table::io archive binary tables](https://github.com/lsst/afw/blob/master/src/image/ExposureFitsReader.cc#L214).  If that fails, it will fall back the one it already loaded from the metadata, which may still be null if the on-disk file didn't actually have a WCS.

But this is indeed a very minor point, and if we were worried about `PhotoCalib`-from-metadata being called *outside* `ExposureFitsReader` (I'm pretty sure its `SkyWcs` counterpart is), then preferring an exception to a null return would be a perfectly good reason to just ignore my comment here, and maintaining similarity with `SkyWcs` is probably a perfectly good reason as well.

To summarize, I suppose I'm saying that:
 - I think I'd prefer `makeSkyWcs` to return null on normal-failure, for similarity with `exposure.getWcs()` and to avoid using exceptions for normal control flow.
 - Given that `makeSkyWcs` does throw (and changing it here is definitely out of scope), it's entirely reasonable for the `PhotoCalib` equivalent to either throw to be more like it, or to return null, to be more like `exposure.getCalib()` and to avoid using exceptions for normal control flow.

parejkoj: `PhotoCalib` will not be persisted in the metadata, so there won't be a "fall back to metadata" option: the metadata reader is only for backwards compatibility with persisted `Calib`s.

Given that, I think you're right that it is better to return `nullptr` here instead of raising: existing Exposures that don't have that metadata won't behave any more oddly than they already do.

That said, throwing here doesn't feel like "normal flow control" to me: not having the values you expected when reading a file should be an "exceptional" case, but the old Exposure persistence wasn't careful enough for us to guarantee that. `PhotoCalibFactory::read`` will implicitly raise some kind of key error if the required fields are missing. Having fields missing from your persisted file is different from `get()`ing an (initialized or not) attribute on a constructed object.

PR #438, opened 2019-02-14
DM-17914: afw PhotoCalibTestCase failure using bleed environment

parejkoj:


gcomoretto:
All afw tests now pass.

PR #439, opened 2019-02-26
DM-18067: Add fluxMag0 PhotoCalib factory function

parejkoj:


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -502,6 +502,21 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
+ * It should not be used outside of tests that compare with old persisted Calib objects.
+ *
+ * @param fluxMag0 The instrumental flux at zero magnitude. If 0, the resulting `PhotoCalib` will have
+ *                 infinite calibrationMean and non-finite (inf or NaN) calibrationErr.
+ * @param fluxMag0Err The instrumental flux at zero magnitude error. If 0, the resulting `PhotoCalib` will
+ *                    have 0 calibrationErr.
+ *
+ * @returns Pointer to the constructed PhotoCalib.
+ */
+std::shared_ptr<PhotoCalib> makePhotoCalib(double fluxMag0, double fluxMag0Err);
```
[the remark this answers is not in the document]
kfindeisen: Just noticed the distinction between "flux" and "calibration". :dizzy_face: Now I think even "ZeroPoint" might not be explicit enough...

src/image/PhotoCalib.cc
```diff
@@ -397,6 +397,8 @@ void PhotoCalib::instFluxToMagnitudeArray(afw::table::SourceCatalog const &sourc
 std::shared_ptr<PhotoCalib> makePhotoCalib(daf::base::PropertySet &metadata, bool strip) {
     double calibration = 0.0, calibrationErr = 0.0;
 
+    // NOTE: we don't use makePhotoCalib(fluxMag0, fluxMag0Err) here because 0 were the "make a default"
+    // Calib values, and we don't want that to result in non-finite PhotoCalibs.
```
kfindeisen: I don't understand this comment. What defaults? I don't see any special handling of 0 appearing in the metadata.

tests/test_photoCalib.py
```diff
@@ -626,6 +626,30 @@ def testCreateFromMetadataRaisesNoKey(self):
         result = lsst.afw.image.makePhotoCalib(metadata)
         self.assertIsNone(result)
 
+    def testCreateFromFluxMag0(self):
+        """Test creating from the Calib-style fluxMag0/fluxMag0Err values."""
+        fluxMag0 = 12345
+        fluxMag0Err = 67890
+
+        referenceFlux = (0*u.ABmag).to_value(u.nJy)
```
kfindeisen: :laughing:

tests/test_photoCalib.py
```diff
@@ -626,6 +626,30 @@ def testCreateFromMetadataRaisesNoKey(self):
...
+
+        # create with non-zero fluxMag0, but zero err
+        photoCalib = lsst.afw.image.makePhotoCalib(fluxMag0, 0)
+        self.assertEqual(photoCalib.getInstFluxAtZeroMagnitude(), fluxMag0)
+        self.assertEqual(photoCalib.getCalibrationErr(), 0.0)
+
+        # create with non-zero fluxMag0 and err
+        photoCalib = lsst.afw.image.makePhotoCalib(fluxMag0, fluxMag0Err)
+        self.assertEqual(photoCalib.getInstFluxAtZeroMagnitude(), fluxMag0)
+        self.assertEqual(photoCalib.getCalibrationErr(), calibrationErr)
```
kfindeisen: I don't really like the fact that this test copies the formula you used in the implementation, it seems prone to human error. Could you write something like the `getInstFluxAtZeroMagnitude` tests, where you check that the resulting `PhotoCalib` behaves as implied by the definition of `fluxMag0Err`?

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -502,6 +502,21 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
  */
 std::shared_ptr<PhotoCalib> makePhotoCalib(daf::base::PropertySet &metadata, bool strip = false);
 
+/**
+ * Construct a PhotoCalib from the deprecated `Calib`-style fluxMag0/fluxMag0Err values.
+ *
+ * This provides backwards compatibility with the obsoleted Calib object that PhotoCalib replaced.
+ * It should not be used outside of tests that compare with old persisted Calib objects.
```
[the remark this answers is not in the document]
parejkoj: I put it in C++ because there wasn't a `photoCalibContinued.py` yet. Maybe this is a good reason to make one, though?

No C++ tests of PhotoCalib.

include/lsst/afw/image/PhotoCalib.h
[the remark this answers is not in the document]
parejkoj: What about `makePhotoCalibFromDeprecatedCalibZeroPoint()`? It's long, and pretty explicit. I don't want this interface to be too user-friendly, anyway, since it should only be used for backwards compatibility.

src/image/PhotoCalib.cc
parejkoj: `Calib` only had a default constructor, which set those to (0,0). This meant that there are plenty of default-constructed `Calib`s (including in the metadata). The difference between the metadata version and this one is that I assume that if the metadata has 0, it means "construct the default version of the thing", whereas for this version I wanted to make it behave exactly as if you'd constructed a (0,0) `Calib` (which is a nonsensical object, but I'm "giving the user what they ask for").

parejkoj: I'm not sure how to reword that though.

tests/test_photoCalib.py
parejkoj: If `getInstFluxAtZeroMagnitude` is correct, then the PhotoCalib will be correct (assuming that its tests itself pass). I think your suggestion would require copying the code from Calib over to here to compare it?

include/lsst/afw/image/PhotoCalib.h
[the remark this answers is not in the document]
kfindeisen: I was probably thinking of pybind11, but on second thought that would be a huge discoverability problem. Never mind.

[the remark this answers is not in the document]
kfindeisen: Probably not, the file shuffling needed to switch between extensions and no extensions is a real pain.

tests/test_photoCalib.py
kfindeisen: No, I meant something like trying to use the `PhotoCalib` to convert a specific flux (error?) where you know the right answer. Duplicating/sharing code between the system and its tests is what I'd like to avoid.

src/image/PhotoCalib.cc
kfindeisen: I still don't understand. If `metadata.exists("FLUXMAG0")`, then the code you have is *exactly* equivalent to calling `makePhotoCalib(fluxMag0, fluxMag0Err)`. If not, then it returns `nullptr` (though this is not at all obvious; until I wrote this comment I thought it returns `PhotoCalib(0.0, 0.0)`).

It seems to me that if you called `makePhotoCalib`, you'd not only simplify the code inside the `if` statement, but you could make the execution path easier to understand as well.

parejkoj: Ha! Ok, it looks like what I'd intended to write for the metadata version (what my comment above is trying to explain) was not what I actually implemented. So, yes, you're correct, and I'll just use this new function in the metadata one.

The nullptr-return for the metadata version is documented in its docstring. Relatedly, the name and docstring for `testCreateFromMetadataRaisesNoKey` need to be changed: I'd originally planned on raising for that case, but was convinced to return nullptr instead (for backwards compatibility with old files that don't have Calib info in them).

tests/test_photoCalib.py
parejkoj: One problem is that the old Calib was incorrect at the ~1e-5 level (the reference flux was wrong). I don't know that doing the (correct) calculation and putting that number in gains us anything vs. having the math here.

src/image/PhotoCalib.cc
```diff
@@ -394,25 +394,29 @@ void PhotoCalib::instFluxToMagnitudeArray(afw::table::SourceCatalog const &sourc
...
-            double fluxMag0Err = metadata.getAsDouble(key);
-            calibrationErr = utils::referenceFlux * fluxMag0Err / std::pow(fluxMag0, 2);
+            instFluxMag0Err = metadata.getAsDouble(key);
             if (strip) metadata.remove(key);
         }
     } else {
         return nullptr;
     }
 
+    return makePhotoCalibFromCalibZeroPoint(instFluxMag0, instFluxMag0Err);
```
kfindeisen: I was kind of hoping you'd move this inside the `if` statement... as it is, you've got code that's executed only if `"FLUXMAG0"` exists, followed by code that's executed only if it doesn't exist, followed by code that's executed only if it does exist. :spaghetti: As it is, it's not at all obvious that the default value of `instFluxMag0` never gets used.

parejkoj: Good point. How is it now?

PR #442, opened 2019-03-19
DM-10156: Replace all uses of Calib with PhotoCalib

parejkoj:


python/lsst/afw/image/utils.py
```diff
@@ -68,23 +67,6 @@ def defineFilter(name, lambdaEff, lambdaMin=np.nan, lambdaMax=np.nan, alias=[],
             Filter.defineAlias(name, a)
 
 
-class CalibNoThrow:
-    """A class intended to be used with python's with statement, to return NaNs for negative fluxes
-    instead of raising exceptions (exceptions may be raised for other purposes).
-
```
czwa: This is the default with `PhotoCalib`, correct?
(Just to make sure I'm understanding the differences correctly.)

python/lsst/afw/table/exposure/exposureContinued.py
```diff
@@ -22,8 +22,21 @@
...
+# class ExposureRecord:
+#     @deprecated(reason="Replaced with getPhotoCalib (will be removed in 18.0)", category=FutureWarning)
+#     def getCalib(self, *args, **kwargs):
+#         return self._getCalib(*args, **kwargs)
+
+#     @deprecated(reason="Replaced with setPhotoCalib (will be removed in 18.0)", category=FutureWarning)
+#     def setCalib(self, *args, **kwargs):
+#         return self._setCalib(*args, **kwargs)
+
+
```
czwa: As before, for reference can you note the ticket for this implementation?

src/image/ExposureInfo.cc
```diff
@@ -235,8 +237,8 @@ ExposureInfo::FitsWriteData ExposureInfo::_startWriteFits(lsst::geom::Point2I co
     /*
      * We need to define these keywords properly! XXX
      */
-    data.metadata->set("FLUXMAG0", getCalib()->getFluxMag0().first);
-    data.metadata->set("FLUXMAG0ERR", getCalib()->getFluxMag0().second);
+    // data.metadata->set("FLUXMAG0", getPhotoCalib()->getFluxMag0().first);
+    // data.metadata->set("FLUXMAG0ERR", getPhotoCalib()->getFluxMag0().second);
```
czwa: Will these be defined properly?  I assume nothing currently depends on these values?

python/lsst/afw/image/utils.py
parejkoj: Yes: `PhotoCalib` never throws due to negative fluxes.

src/image/ExposureInfo.cc
parejkoj: Oh yeah, those. They can and should be deleted. The comment there is very long standing, and now irrelevant.

PR #443, opened 2019-03-21
DM-16925 Provide backwards-compatibility with Calib API

parejkoj:


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -451,6 +452,80 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
+    }
+    /**
+     * @copydoc getInstFluxAtZeroMagnitude()
+     * Deprecated: For backwards compatibility with Calib.
+     */
+    [[deprecated(
+            "For backwards compatibility with Calib: use `getCalibrationMean`, `getCalibrationErr`, or "
+            "`getInstFluxAtZeroMagnitude. To be removed after v18.")]] std::pair<double, double>
+    getFluxMag0() const {
+        return std::make_pair<double, double>(getInstFluxAtZeroMagnitude(),
```
r-owen: I like the name `getInstFluxAtZeroMagnitude`. It is very clear.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -451,6 +452,80 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
+                        "removed after v18.")]] std::pair<double, double>
+    getMagnitude(double instFlux, double instFluxErr) const {
+        auto result = instFluxToMagnitude(instFlux, instFluxErr);
+        return std::make_pair<const double &, const double &>(result.value, result.error);
+    };
+    [[deprecated(
+            "For backwards compatibility with Calib; use `instFluxToMagnitude` instead. To be "
+            "removed after v18.")]] std::pair<ndarray::Array<double, 1>, ndarray::Array<double, 1>>
+    getMagnitude(ndarray::Array<double const, 1> const &instFlux,
+                 ndarray::Array<double const, 1> const &instFluxErr) const;
```
r-owen: Did the original C++ have a default of 0 for instFluxErr (as the Python does)? If so, please add it.

tests/test_photoCalib.py
```diff
@@ -650,6 +653,56 @@ def testMakePhotoCalibFromCalibZeroPoint(self):
...
+        msg = "PhotoCalib is immutable: create a new `PhotoCalib` with the calibration factor and error"
+        ", or create it like an old Calib object with `makePhotoCalibFromCalibZeroPoint`."
+        with self.assertRaisesRegex(RuntimeError, msg):
+            photoCalib.setFluxMag0(1)
+        with self.assertRaisesRegex(RuntimeError, msg):
+            photoCalib.setFluxMag0(1, 2)
+
+        # test getMagnitude()
+        instFlux = 1234
+        instFluxErr = 567
```
r-owen: Minor request: if these values are arbitrary please note that in a code comment (similarly for mag = 25 below)

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -451,6 +452,80 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
 
     friend std::ostream &operator<<(std::ostream &os, PhotoCalib const &photoCalib);
 
+    /* Backwards compatibility with old Calib object */
+
+    /// No-op: for backwards compatibility with Calib.
+    [[deprecated("No-op: PhotoCalib never throws on negative instFlux. Will remove after v18.")]] static void
+    setThrowOnNegativeFlux(bool raiseException) noexcept {
```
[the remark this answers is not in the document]
TallJimbo: I don't think we can make this raise on `true`, at least not easily; the usual pattern is (sometimes wrapped in a context manager):
```
Calib.setThrowOnNegativeFlux(False)`
# do stuff
Calib.setThrowOnNegativeFlux(True)
```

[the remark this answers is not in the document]
parejkoj: I think we can just let the (not yet existing) python-level deprecation warnings take care of letting people know that this is deprecated.

tests/test_photoCalib.py
```diff
@@ -650,6 +653,56 @@ def testMakePhotoCalibFromCalibZeroPoint(self):
...
+        except AttributeError:
+            self.fail("Expected setThrowOnNegativeFlux to be available as a no-op.")
+        self.assertFalse(lsst.afw.image.PhotoCalib.getThrowOnNegativeFlux())
+
+        result = photoCalib.getFluxMag0()
+        self.assertEqual(result[0], photoCalib.getInstFluxAtZeroMagnitude())
+        self.assertTrue(np.isnan(result[1]))  # we just return NaN for the second element now
+
+        msg = "PhotoCalib is immutable: create a new `PhotoCalib` with the calibration factor and error"
+        ", or create it like an old Calib object with `makePhotoCalibFromCalibZeroPoint`."
```
r-owen: Is this supposed to concatenate the two lines? I would not expect it to do so without a trailing `\` on the `msg =` line.

parejkoj: Multiple adjacent strings are automatically concatenated: https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation

include/lsst/afw/image/PhotoCalib.h
parejkoj: I don't see where you see a default value: neither this new deprecated code, nor the original `Calib` object has any defaults for `instFluxErr`.

Were you instead thinking of `setFluxMag0()`? That one did have a 0 default for the error, which remains here at the python level, but seems silly at the C++ level since it the function just immediately throws.

r-owen: I guess I must have been thinking of setFluxMag0 and it looks like the C++ already has a default of 0, which is great.

PR #444, opened 2019-03-28
DM-18678: Support boost 1.69 removal of gil_all.hpp

timj:


timj:
@TallJimbo switching out boost meant I also had to explicitly load `<set>` in one of the include files. Since I was watching the build I also fixed the deprecation warning from pybind11.

timj:
Yes, I wanted to make sure it worked with the current boost.  This lets me merge it without waiting for @fritzm to sign off on the boost update for Qserv.

PR #446, opened 2019-04-02
DM-18864: Now support UNDEF in FITS headers

timj:


timj:
@TallJimbo as discussed on Slack I've added some code to this PR that means we can read FITS headers that have identical keys but some of which are undefined. I've added a test. I've included warnings and the code looks a bit repetitive so you might want to take a look.

timj:
FITS can have cards with identical keys and differing values and sometimes they can be undefined in one place and defined in another.  I added the warnings not because FITS can't support that, but because people may be surprised that we can't.

PR #448, opened 2019-04-10
DM-18544: Add python-level deprecation warnings for Calib interface

parejkoj:


python/lsst/afw/image/photoCalibContinued.py
```diff
@@ -0,0 +1,62 @@
...
+__all__ = []  # import this module only for its side effects
+
+from lsst.utils import continueClass
+
+from deprecated.sphinx import deprecated
+
+from .photoCalib import PhotoCalib
+
+
+@continueClass  # noqa F811
```
timj: I think this should formally be `noqa: F811` (the colon is important to flake8 since it tells it to only mask F811 rather than ignoring all warnings).

python/lsst/afw/image/exposure/exposureContinued.py
```diff
@@ -99,14 +99,13 @@ def setVariance(self, variance):
...
+    @deprecated(reason="Replaced with getPhotoCalib (will be removed after v18)", category=FutureWarning)
+    def getCalib(self, *args, **kwargs):
+        return self._getCalib(*args, **kwargs)
 
-    # @deprecated(reason="Replaced with setPhotoCalib (will be removed in 18.0)", category=FutureWarning)
-    # def setCalib(self, *args, **kwargs):
-    #     return self._setCalib(*args, **kwargs)
+    @deprecated(reason="Replaced with setPhotoCalib (will be removed after v18)", category=FutureWarning)
+    def setCalib(self, *args, **kwargs):
+        self._setCalib(*args, **kwargs)
```
kfindeisen: Why force the wrapped methods to be all underscores instead of just following the style guide

    setCalib = deprecated(reason="Replaced with setPhotoCalib (will be removed after v18)", category=FutureWarning)(setCalib)

You had to make a lot of changes in the C++ file...

parejkoj: Because that doesn't work for methods on classes due to a docstring overwriting problem. @ktlim has what he thinks is a fix for upstream `deprecate`, but getting that into the stack is a problem because of upgrading the conda env being complicated.

python/lsst/afw/image/photoCalibContinued.py
parejkoj: It looks like we get this wrong all over in afw: I find 33 uses of `noqa ` (space) vs. 8 uses of `noqa: ` (colon space).

PR #451, opened 2019-04-11
DM-17431: Respect priority order when merging a footprint

yalsayyad:
that overlaps previously disconnected higher priority footprints.
Include unit test that simulates this case.

### DIFF
--- a/src/detection/FootprintMerge.cc
+++ b/src/detection/FootprintMerge.cc
@@ -270,14 +270,17 @@ void FootprintMergeList::addCatalog(std::shared_ptr<afw::table::SourceTable> sou
                 if (box.overlaps(foot->getBBox()) && (**iter).overlaps(*foot)) {
                     if (!first) {
                         first = *iter;
-                        // Add Footprint to existing merge and set flag for this band
+                        // Add new footprint to existing FootprintMerge without peaks, in order to connect
+                        // other now-overlapping FootprintMerges. If two FootprintMerges both overlap with
+                        // the new footprint, they are now guaranteed to overlap with each other.
+                        // Hold off adding foot's lower-priority peaks until the higher-priority existing
+                        // peaks merged into this first FootprintMerge. 
                         if (doMerge) {
-                            first->add(foot, _peakSchemaMapper, keyIter->second, minNewPeakDist,
-                                       maxSamePeakDist);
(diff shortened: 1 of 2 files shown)


src/detection/FootprintMerge.cc
```diff
@@ -270,25 +270,32 @@ void FootprintMergeList::addCatalog(std::shared_ptr<afw::table::SourceTable> sou
...
-        }
-
-        if (!first) {
+            } // while mergeList
+        } //     if checkForMatches
+
+        if (doMerge && first) {
+            // Now merge footprint including peaks into the newly-connected, higher-priority FootprintMerge
+            first->add(foot, _peakSchemaMapper, keyIter->second, minNewPeakDist, maxSamePeakDist);
+        } else {
```
TallJimbo: I assume `doMerge == bool(first)` is guaranteed? If so, is it worth asserting on that?

yalsayyad: *As written* no, that's not guaranteed. See new line 272. `bool(first)` says there was at least one overlapping FootprintMerge. `doMerge` is the input parameter that says "yes, merge it." Both have to be true for us to merge it.  This logic I didn't change, but I also found it confusing, and why I was proposing some reordering (to move the `doMerge` up and out of the loop). Because what's the point of looping over existing FootprintMerges if you don't plan on merging them!  Your comment demonstrated that doing that would be worthwhile. Going to put it on a different commit.

yalsayyad:
@TallJimbo I added the two additional commits I wanted to make. Will you take a look?  Checked there's no behavior change and I'm rerunning Jenkins now.

src/detection/FootprintMerge.cc
```diff
@@ -247,8 +247,8 @@ void FootprintMergeList::addCatalog(std::shared_ptr<afw::table::SourceTable> sou
                           (boost::format("Filter %s not in original list") % filter).str());
     }
 
-    // If list is empty don't check for any matches, just add all the objects
-    bool checkForMatches = (_mergeList.size() > 0);
+    // If list is empty or merging not requested, don't check for any matches, just add all the objects
+    bool checkForMatches = (_mergeList.size() > 0) && doMerge;
```
TallJimbo: Prefer `!_mergeList.empty()`; this comparison will generate a signed vs. unsigned warning on some compilers.

PR #453, opened 2019-04-12
DM-18393: Add some needed features for testing at SLAC

RobertLuptonTheGood:


include/lsst/afw/cameraGeom/Detector.h
```diff
@@ -89,21 +90,23 @@ class Detector final : public table::io::PersistableFacade<Detector>, public tab
              lsst::geom::Box2I const &bbox, lsst::afw::table::AmpInfoCatalog const &ampInfoCatalog,
              Orientation const &orientation, lsst::geom::Extent2D const &pixelSize,
              TransformMap::Transforms const &transforms,
-             CrosstalkMatrix const &crosstalk = CrosstalkMatrix());
+             CrosstalkMatrix const &crosstalk = CrosstalkMatrix(),
+             std::string const &physicalType="unknown");
```
TallJimbo: What do you think about just using an empty string to mean "unknown"?  It's an easier sentinel value to remember.

src/cameraGeom/Detector.cc
```diff
@@ -212,7 +243,7 @@ class DetectorFactory : public table::io::PersistableFactory {
 
     std::shared_ptr<table::io::Persistable> read(InputArchive const& archive,
                                                  CatalogVector const& catalogs) const override {
-        auto const & keys = PersistenceHelper::get();
+        auto const & keys = PersistenceHelper(catalogs.front().getSchema());
```
TallJimbo: While C++ lets you get away with binding a temporary to a const reference, for a local variable like this I think it's better to just assign it to a value (i.e. `auto const` instead of `auto const &`).

tests/test_detector.py
```diff
@@ -84,7 +86,7 @@ def duplicateAmpName(dw):
             DetectorWrapper(modFunc=duplicateAmpName)
 
         # These break in the pybind layer
-        for crosstalk in ([1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],  # 1D and not numpy
+        for crosstalk in ([11.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0],  # 1D and not numpy
```
TallJimbo: Was this change intentional?

RobertLuptonTheGood: No; used in setting a breakpoint when trying to find the numpy problem.  I'll revert it.

src/cameraGeom/Detector.cc
RobertLuptonTheGood: Cargo-culted from the previous lone, which worked because it was a singleton.  I'll change it

include/lsst/afw/cameraGeom/Detector.h
RobertLuptonTheGood: Should be OK in C++; yamlCamera can still assign "unknown" which I prefer in the end

PR #457, opened 2019-04-20
DM-10155: Implement in-place SourceCatalog operators in PhotoCalib

parejkoj:
Also vectorize catalog array operations for a big speedup on AST-backed BoundedFields.

### DIFF
--- a/include/lsst/afw/image/PhotoCalib.h
+++ b/include/lsst/afw/image/PhotoCalib.h
@@ -339,7 +339,7 @@ public:
     /**
      * Return a flux calibrated image, with pixel values in nJy.
      *
-     * Mask pixels are propogated directly from the input image.
+     * Mask pixels are propagated directly from the input image.
      *
      * @param maskedImage The masked image to calibrate.
      * @param includeScaleUncertainty Include the uncertainty on the calibration in the resulting variance?
--- a/include/lsst/afw/image/PhotoCalib.h
+++ b/include/lsst/afw/image/PhotoCalib.h
@@ -554,7 +554,12 @@ private:
     /**
      * Return the calibration evaluated at a sequence of points.
      */
-    ndarray::Array<double, 1> evaluateArray(ndarray::Array<double, 1> xx, ndarray::Array<double, 1> yy) const;
+    ndarray::Array<double, 1> evaluateArray(ndarray::Array<double, 1> const &xx,
+                                            ndarray::Array<double, 1> const &yy) const;
+    /**
+     * Return the calibration evaluated at the centroids of a SourceCatalog.
+     */
+    ndarray::Array<double, 1> evaluateCatalog(afw::table::SourceCatalog const &sourceCatalog) const;
 
     /// Returns the spatially-constant calibration (for setting _calibrationMean)
     double computeCalibrationMean(std::shared_ptr<afw::math::BoundedField> calibration) const;
@@ -589,8 +594,8 @@ std::shared_ptr<PhotoCalib> makePhotoCalibFromMetadata(daf::base::PropertySet &m
  *
  * @param instFluxMag0 The instrumental flux at zero magnitude. If 0, the resulting `PhotoCalib` will have
  *                     infinite calibrationMean and non-finite (inf or NaN) calibrationErr.
- * @param instFluxMag0Err The instrumental flux at zero magnitude error. If 0, the resulting `PhotoCalib` will
- *                        have 0 calibrationErr.
+ * @param instFluxMag0Err The instrumental flux at zero magnitude error. If 0, the resulting `PhotoCalib`
+ * will have 0 calibrationErr.
  *
  * @returns Pointer to the constructed PhotoCalib.
  */
--- a/include/lsst/afw/image/PhotoCalib.h
+++ b/include/lsst/afw/image/PhotoCalib.h
@@ -551,6 +551,10 @@ private:
      * Helper function to manage constant vs. non-constant PhotoCalibs
      */
     double evaluate(lsst::geom::Point<double, 2> const &point) const;
+    /**
+     * Return the calibration evaluated at a sequence of points.
+     */
+    ndarray::Array<double, 1> evaluateArray(ndarray::Array<double, 1> xx, ndarray::Array<double, 1> yy) const;
 
     /// Returns the spatially-constant calibration (for setting _calibrationMean)
     double computeCalibrationMean(std::shared_ptr<afw::math::BoundedField> calibration) const;
--- a/include/lsst/afw/image/PhotoCalib.h
+++ b/include/lsst/afw/image/PhotoCalib.h
@@ -250,8 +250,6 @@ public:
      *                            "PsfFlux_instFluxErr"
      * @param[in]  outField       The field to write the nJy and magnitude errors to.
      *                            Keys of the form "*_instFlux" and "*_instFluxErr" must exist in the schema.
-     *
-     * @warning Not implemented yet: See DM-10155.
      */
     void instFluxToNanojansky(afw::table::SourceCatalog &sourceCatalog, std::string const &instFluxField,
                               std::string const &outField) const;
@@ -330,8 +328,6 @@ public:
      * @param[in]  outField       The field to write the magnitudes and magnitude errors to.
      *                            Keys of the form "*_instFlux", "*_instFluxErr", *_mag", and "*_magErr"
      *                            must exist in the schema.
-     *
-     * @warning Not implemented yet: See DM-10155.
      */
     void instFluxToMagnitude(afw::table::SourceCatalog &sourceCatalog, std::string const &instFluxField,
                              std::string const &outField) const;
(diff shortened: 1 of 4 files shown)


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -349,6 +345,28 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
+     *
+     * @param catalog The catalog to compute calibrated fluxes of. This catalog is exactly reproduced in the
+     * output catalog.
+     * @param instFluxFields The fields to calibrate. If empty, every field ending with `_instFlux` will have
+     * its corresponding calibrated fields produced in the output catalog. If `_instFluxErr` also exists, it
+     * will be used to compute the `_fluxErr` and `_magErr` fields too.
+     *
+     * @return The calibrated catalog, with new flux and magnitude (and their errors) fields.
+     *
+     * @throws lsst::pex::exceptions::NotFoundError if any item in `instfluxField` does not have a
```
TallJimbo: `instfluxField` -> `instFluxFields`.

include/lsst/afw/image/PhotoCalib.h
```diff
@@ -555,6 +573,14 @@ class PhotoCalib : public table::io::PersistableFacade<PhotoCalib>, public table
...
      */
     double evaluate(lsst::geom::Point<double, 2> const &point) const;
+    /**
+     * Return the calibration evaluated at a sequence of points.
+     */
+    ndarray::Array<double, 1> evaluateArray(ndarray::Array<double, 1> xx, ndarray::Array<double, 1> yy) const;
+    /**
+     * Return the calibration evaluated at the centroids of a SourceCatalog.
+     */
+    ndarray::Array<double, 1> evaluateCatalog(afw::table::SourceCatalog const &sourceCatalog) const;
```
TallJimbo: Please add blank lines between methods.

src/image/PhotoCalib.cc
```diff
@@ -293,6 +293,85 @@ MaskedImage<float> PhotoCalib::calibrateImage(MaskedImage<float> const &maskedIm
...
+    std::vector<table::Key<double>> instFluxErrKeys;
+    instFluxErrKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> fluxKeys;
+    fluxKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> fluxErrKeys;
+    fluxErrKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> magKeys;
+    magKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> magErrKeys;
+    magErrKeys.reserve(instFluxFields.size());
```
TallJimbo: This isn't bad by any means, but it strikes me as a place where a little local `struct` to hold those keys, might be a bit better - i.e. you'd just have a single `vector` of that struct, and then the inner `for` loop could just be a range based `for`, and there'd be a lot more simple attribute assignments instead of paren-filled `push_back()` and `back()` calls.

src/image/PhotoCalib.cc
```diff
@@ -293,6 +293,85 @@ MaskedImage<float> PhotoCalib::calibrateImage(MaskedImage<float> const &maskedIm
...
+    for (auto const &field : instFluxFields) {
+        instFluxKeys.push_back(inSchema.find<double>(field + "_instFlux").key);
+        try {
+            instFluxErrKeys.push_back(inSchema.find<double>(field + "_instFluxErr").key);
+        } catch (pex::exceptions::NotFoundError) {
```
TallJimbo: Even though you're not doing anything with this exception, I think you still want to catch it by reference, i.e. `catch (pex::exceptions::NotFoundError&)`.  I'm not positive there would be any ill effects from this as is, and I don't even remember in detail what the ill effects of catching by value are, but it's safest to just always catch by reference.

src/image/PhotoCalib.cc
```diff
@@ -293,6 +293,85 @@ MaskedImage<float> PhotoCalib::calibrateImage(MaskedImage<float> const &maskedIm
...
+    std::vector<table::Key<double>> fluxKeys;
+    fluxKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> fluxErrKeys;
+    fluxErrKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> magKeys;
+    magKeys.reserve(instFluxFields.size());
+    std::vector<table::Key<double>> magErrKeys;
+    magErrKeys.reserve(instFluxFields.size());
+    for (auto const &field : instFluxFields) {
+        instFluxKeys.push_back(inSchema.find<double>(field + "_instFlux").key);
```
TallJimbo: Once this is switched to struct field assigment (per above comment), you could actually use an even nicer syntax for this:
```
s.instFlux = inSchema[field + "_instFlux"];  // assuming s is the struct, and instFlux is a field of type Key<double>.
```
That is, with the assignment form, it can infer the fact that it's a `double` from what you're assigning it to (which is why this kind of syntax isn't usable in Python as often).

I'd also say it'd be slightly better to use `inSchema.join(field, "instFlux")` instead of `field + "_instFlux"` to avoid referring directly to the delimiter.  We don't do that across the board, but I'm trying to encourage it in new code.

src/image/PhotoCalib.cc
```diff
@@ -293,6 +293,85 @@ MaskedImage<float> PhotoCalib::calibrateImage(MaskedImage<float> const &maskedIm
...
+
+    return output;
+}
+
+afw::table::SourceCatalog PhotoCalib::calibrateCatalog(afw::table::SourceCatalog const &catalog) const {
+    std::vector<std::string> instFluxFields;
+    for (auto const &name : catalog.getSchema().getNames()) {
+        // Pick every field ending in "_instFlux", grabbing everything before that prefix.
+        if (name.size() > 10 && name.compare(name.size() - 9, 9, "_instFlux") == 0) {
+            instFluxFields.emplace_back(name.substr(0, name.size() - 9));
```
TallJimbo: You could do something like
```
constexpr std::string SUFFIX = "_instFlux";
```
above the for loop, and then refer to `SUFFIX.size()` instead of using `9` and `10` explicitly.

src/image/PhotoCalib.cc
```diff
@@ -465,38 +544,68 @@ double PhotoCalib::evaluate(lsst::geom::Point<double, 2> const &point) const {
         return _calibration->evaluate(point);
 }
 
+ndarray::Array<double, 1> PhotoCalib::evaluateArray(ndarray::Array<double, 1> xx,
+                                                    ndarray::Array<double, 1> yy) const {
```
TallJimbo: Convention is to pass arrays by const reference: `ndarray::Array<double, 1> const & xx`.

It's not a big deal either (neither is our opposite convention of passing `shared_ptr` by value), but best to be consistent.

src/image/PhotoCalib.cc
```diff
@@ -465,38 +544,68 @@ double PhotoCalib::evaluate(lsst::geom::Point<double, 2> const &point) const {
...
+        ++i;
+    }
+    return evaluateArray(xx, yy);
+}
+
 void PhotoCalib::instFluxToNanojanskyArray(afw::table::SourceCatalog const &sourceCatalog,
                                            std::string const &instFluxField,
                                            ndarray::Array<double, 2, 2> result) const {
-    double instFlux, instFluxErr, nanojansky, calibration;
+    double instFlux, instFluxErr, nanojansky;  //, calibration;
```
TallJimbo: It's fine (i.e. just as efficient) and probably slightly better form to declare these on first use, even though that puts them inside the loop - it's always good to avoid uninitialized variables, and for stack variables of trivial types like `double` you don't need to worry about the cost of construction being different from the cost of assignment.

src/image/PhotoCalib.cc
```diff
@@ -465,38 +544,68 @@ double PhotoCalib::evaluate(lsst::geom::Point<double, 2> const &point) const {
...
+        ++iter;
+        ++i;
     }
 }
 
 void PhotoCalib::instFluxToMagnitudeArray(afw::table::SourceCatalog const &sourceCatalog,
                                           std::string const &instFluxField,
                                           ndarray::Array<double, 2, 2> result) const {
-    double instFlux, instFluxErr, calibration;
+    double instFlux, instFluxErr;  //, calibration;
```
TallJimbo: same comment as above about declaring these on first use inside the loop.

src/image/PhotoCalib.cc
parejkoj: Thanks for that suggestion. I think that does improve clarity.

src/image/PhotoCalib.cc
parejkoj: `constexpr std::string` is not allowed. Is it better to make it a non-const string, or use a `const char[]`? The latter makes for less nice syntax for computing the size.

TallJimbo: I think I'd use `static std::string const SUFFIX = "_instFlux";`, then.

PR #462, opened 2019-04-29
DM-19461: performance optimizations for afw::table FITS I/O

TallJimbo:


src/table/BaseRecord.cc
```diff
@@ -13,6 +13,55 @@ namespace table {
 
 namespace {
 
+// A Schema Functor used to set floating point-fields to NaN and initialize variable-length arrays
+// using placement new.  All other fields are left alone, as they should already be zero.
+struct RecordInitializer {
+    template <typename T>
+    static void fill(T *element, int size) {}  // this matches all non-floating-point-element fields.
+
+    static void fill(float *element, int size) {
```
PaulPrice: For my enlightenment: this doesn't need `template <>` because they're in a `struct` rather than free functions?

include/lsst/afw/table/io/FitsSchemaInputMapper.h
```diff
@@ -6,26 +6,54 @@
...
     FitsColumnReader(FitsColumnReader const &) = delete;
     FitsColumnReader(FitsColumnReader &&) = delete;
     FitsColumnReader &operator=(FitsColumnReader const &) = delete;
     FitsColumnReader &operator=(FitsColumnReader &&) = delete;
 
+    /**
+     *  Optionally read ahead and cache values from multiple rows.
+     *
+     *  Subclasses are not required to implement this method; if they do, they
+     *  it should indicate to readCell that cached values should be used
```
PaulPrice: "they it"

include/lsst/afw/table/io/FitsSchemaInputMapper.h
```diff
@@ -64,6 +92,19 @@ class FitsSchemaInputMapper {
...
+    /**
+     *  When processing each column, ask CFITSIO to read this many rows of
+     *  values from that column in a single call.
+     *
+     *  Both FITS binary tables and afw.table are stored row-major, so setting
+     *  this to something other than 1 leads to nonsequential reads.  But
+     *  given the way the I/O code is structured, we tend to get nonsequential
+     *  reads anyway, and it seems the per-call overload to CFITSIO is
+     *  sufficiently high that it's best to have *significantly*
+     *  non-sequential reads by making this >> 1.
```
PaulPrice: Add reasons not to set it to infinity?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -405,16 +428,32 @@ class AngleReader : public FitsColumnReader {
         }
     }
 
+    void prepRead(std::size_t firstRow, std::size_t nRows, fits::Fits & fits) override {
+        assert(_key.getElementCount() == 1u);
```
PaulPrice: We can't have arrays of `Angle`? Pairs of `Angle` could be used for sky coordinates?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -833,8 +872,16 @@ void FitsSchemaInputMapper::readRecord(BaseRecord &record, afw::fits::Fits &fits
             record.set(_impl->flagKeys[bit], _impl->flagWorkspace[bit]);
         }
     }
-    for (auto iter = _impl->readers.begin(); iter != _impl->readers.end(); ++iter) {
-        (**iter).readCell(record, row, fits, _impl->archive);
+    if (N_ROWS_TO_PREP != 1 && row % N_ROWS_TO_PREP == 0) {
+        // Give readers a chance to read and cache up to N_ROWS_TO_PREP rows-
+        // worth of values.
```
PaulPrice: Do we have tests that read both more and less than `N_ROWS_TO_PREP`, to ensure this is working>

include/lsst/afw/detection/Peak.h
```diff
@@ -46,6 +46,16 @@ class PeakRecord : public afw::table::BaseRecord {
     typedef afw::table::CatalogT<PeakRecord> Catalog;
     typedef afw::table::CatalogT<PeakRecord const> ConstCatalog;
 
+    /**
+     *  Constructor used by PeakTable.
+     *
+     *  While formally public, this constructor is conceptually and effectively
+     *  protected, due to the (protected) ConstructionToken argument.
+     */
```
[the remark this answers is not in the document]
TallJimbo: It's formally public only for `make_shared,` actually.  But I will add that comment.

include/lsst/afw/table/BaseRecord.h
```diff
@@ -29,7 +29,33 @@ namespace table {
...
+    // protected constructor would.  In particular, this allows make_shared to
+    // be used.
+    class ConstructionToken {
+    private:
+        friend class BaseTable;
+
+        // A private ctor for this class (and friendship) is necessary to
+        // prevent e.g. BaseRecord({}, ...) calls from code without access
+        // to ConstructionToken.
+        explicit ConstructionToken(int) noexcept {}
```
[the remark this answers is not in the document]
TallJimbo: Apparently (and this was something I learned on Stack Overflow while working on this), if there are no parameters, then you can pass `{}` when calling a function to construct a `ConstructionToken` on the spot, even if you don't have access to the default constructor.  I found that surprising, and it may have been a compiler bug, but the upshot is that this is safer.

src/table/BaseRecord.cc
TallJimbo: Because it's a function and not a class or struct, actually - because you can have function overloads that have nothing to do with templates, a fully-specialized "function template" is just a regular function overload, and the `template <>` syntax only applies to classes and structs.

include/lsst/afw/table/io/FitsSchemaInputMapper.h
TallJimbo: @erykoff's tests suggest that making it too large makes it go slower, especially when the width of the record is also large.  I'd guess that has something to do with disk page or CFITSIO buffer sizes.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: We now handle struct-like combinations of fields as a layer that sits on top of completely standalone fields instead, the the `FunctorKeys` classes in `lsst/afw/table/aggregates.h`.  That makes working with column arrays easier (i.e. because you can get `ra` and `dec` separately rather than as a 2xN matrix).

There might be a use case for more general arrays of angles, too, but I don't think we've encountered one yet, and until C++17 we're limited to 20 total field types by a hard-to-change upper limit in `boost::variant`.

src/table/io/FitsSchemaInputMapper.cc
TallJimbo: We definitely have tests that read less.  I'll wait until @erykoff's tests for the right value of `N_ROWS_TO_PREP` to see if I should look into adding one that reads more.

erykoff: I don't think I have any more tests planned than what I posted on the ticket: 32 is better for a typical `calexp_src` file (wide/short), and 256 is better for a giant table (narrow/tall).

PR #468, opened 2019-06-04
DM-20019: Fix pickling of String Fields

parejkoj:


python/lsst/afw/table/schema/schema.cc
```diff
@@ -82,6 +82,48 @@ void declareFieldBaseSpecializations(PyFieldBase<std::string> &cls) {
...
+                return py::make_tuple(self.getName(), self.getDoc(), self.getUnits());
+            },
+            [](py::tuple t) {
+                if (t.size() != 3) throw std::runtime_error("Invalid number of parameters when unpickling!");
+                return Field<T>(t[0].cast<std::string>(), t[1].cast<std::string>(), t[2].cast<std::string>());
+            }));
+}
+
+template <typename U>
+void declareFieldSpecializations(PyField<Array<U>> &cls) {
```
kfindeisen: `U` -> `T`

kfindeisen: As with the `makeKey` methods, you could make a template that covers both `Array` and `string`, and without the need for an explicit template argument.

python/lsst/afw/table/schema/schema.cc
```diff
@@ -82,6 +82,48 @@ void declareFieldBaseSpecializations(PyFieldBase<std::string> &cls) {
...
+    cls.def(py::pickle(
+            [](Field<Array<U>> const &self) {
```
kfindeisen: Consider `using Class = Field<Array<U>>` to reduce code duplication; see the [style guide](https://developer.lsst.io/pybind11/style.html#names-of-generic-class-types-should-be-called-class).

python/lsst/afw/table/schema/schema.cc
```diff
@@ -195,6 +237,30 @@ void declareKeySpecializations(PyKey<Array<U>> &cls) {
...
+            [](py::tuple t) {
+                if (t.size() != 2) throw std::runtime_error("Invalid number of parameters when unpickling!");
+                return detail::Access::makeKeyArray<U>(t[0].cast<int>(), t[1].cast<int>());
+            }));
+}
+
+void declareKeySpecializations(PyKey<std::string> &cls) {
+    declareKeyAccessors(cls);
+    cls.def_property_readonly("subfields", [](py::object const &) { return py::none(); });
+    cls.def_property_readonly("subkeys", [](py::object const &) { return py::none(); });
```
kfindeisen: Do you actually need these properties at all?

python/lsst/afw/table/schema/schema.cc
```diff
@@ -82,6 +82,48 @@ void declareFieldBaseSpecializations(PyFieldBase<std::string> &cls) {
...
+
+template <typename T>
+void declareFieldSpecializations(PyField<T> &cls) {
+    cls.def(py::pickle(
+            [](Field<T> const &self) {
+                /* Return a tuple that fully encodes the state of the object */
+                return py::make_tuple(self.getName(), self.getDoc(), self.getUnits());
+            },
+            [](py::tuple t) {
+                if (t.size() != 3) throw std::runtime_error("Invalid number of parameters when unpickling!");
```
kfindeisen: Report the actual size?

tests/test_schema.py
```diff
@@ -410,6 +415,10 @@ def testPickle(self):
         schema.addField("i", type="I", doc="int")
         schema.addField("f", type="F", doc="float", units="m2")
         schema.addField("flag", type="Flag", doc="a flag")
+        schema.addField("string", type="String", doc="A string field", size=42)
+        schema.addField("variable_string", type="String", doc="A variable-length string field", size=0)
+        schema.addField("array", type="ArrayF", doc="An array field", size=10)
+        schema.addField("variable_array", type="ArrayF", doc="A variable-length array field", size=0)
```
kfindeisen: Lots of code duplication; consider factoring the two schema definitions into their own function (`makeTestSchema`? `addCommonTestFields`?).

python/lsst/afw/table/schema/schema.cc
parejkoj: I think there are few enough uses (3 total) that that definition doesn't help, and it gives another layer of indirection for someone looking at the code to have to follow.

python/lsst/afw/table/schema/schema.cc
parejkoj: I copied them from the `PyKey<T>` specialization: that was being used `string` before I added this one, so I would assume I should keep it consistent with the new specialization.

PR #469, opened 2019-06-05
DM-20024: Simplify BackgroundList.readFits to avoid file leaks.

TallJimbo:


python/lsst/afw/math/backgroundList.py
```diff
@@ -164,64 +164,47 @@ def readFits(fileName, hdu=0, flags=0):
...
-
-            interpStyle = afwMath.Interpolate.Style(md.getScalar("INTERPSTYLE"))
-            undersampleStyle = afwMath.UndersampleStyle(
-                md.getScalar("UNDERSAMPLESTYLE"))
+        f = Fits(fileName, 'r')
+        nHdus = f.countHdus()
+        f.closeFile()
+        if nHdus % 3 != 0:
+            raise RuntimeError(f"BackgroundList FITS file {fileName} has {nHdus} HDUs;"
+                               f"expected a multiple of 3 (compression is not supported).")
```
timj: Is the `f`-string needed on this line?

TallJimbo: No, I just like to use or not use `f`-strings consistently when I'm relying on multi-line string concatenation, so changing the line break can't break things.

PR #471, opened 2019-06-07
Convert afw.table to numpydoc.

czwa:
Docstrings have been updated to numpydoc style, and boilerplate
copyright information added to all python code.

### DIFF
--- a/python/lsst/afw/table/match/matchContinued.py
+++ b/python/lsst/afw/table/match/matchContinued.py
@@ -1,9 +1,10 @@
+# This file is part of afw.
 #
-# LSST Data Management System
-# Copyright 2017 LSST/AURA.
-#
-# This product includes software developed by the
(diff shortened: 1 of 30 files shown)


python/lsst/afw/table/match/matchContinued.py
```diff
@@ -113,13 +114,25 @@ def packMatches(matches):
...
+        `.distance->float`.
+
+    Returns
+    -------
+    result :
+        Thecatalog of matches.
+
+    Notes
+    -----
+    This pure python implementation exists because SWIG could not
```
timj: Does @TallJimbo want to supply some new text for this note now that we are in the pybind11 era?

TallJimbo: "The pure python implementation exists as a historical artifact related to SWIG limitations."

PR #472, opened 2019-06-11
DM-19556: Use TemplateMeta with GenericMap

kfindeisen:
This PR adds non-templated interfaces to `GenericMapS`, `MutableGenericMapS`, and `SimpleGenericMapS`. These interfaces should make `GenericMap` more `dict`-like for most Python users.

### DIFF
--- a/python/lsst/afw/typehandling/_GenericMap.py
+++ b/python/lsst/afw/typehandling/_GenericMap.py
@@ -19,17 +19,33 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-__all__ = []
+__all__ = ["GenericMap", "MutableGenericMap"]
 
 from collections.abc import Mapping, MutableMapping
 
-from lsst.utils import continueClass
+from lsst.utils import TemplateMeta
 from ._typehandling import GenericMapS, MutableGenericMapS
 
-Mapping.register(GenericMapS)
 
-@continueClass  # noqa F811
-class GenericMapS:
+class GenericMap(metaclass=TemplateMeta):
+    """An abstract `~collections.abc.Mapping` for use when sharing a
+    map between C++ and Python.
+
+    For compatibility with C++, ``GenericMap`` has the following
+    restrictions:
+
+        - all keys must be of the same type
+        - values must be built-in types or subclasses of
+          `lsst.afw.typehandling.Storable`. Almost any user-defined class in
+          C++ or Python can have `~lsst.afw.typehandling.Storable` as a mixin.
+
+    As a safety precaution, `~lsst.afw.typehandling.Storable` objects that are
+    added from C++ may be copied when you retrieve them from Python, making it
+    impossible to modify them in-place. This issue does not affect objects that
+    are added from Python, or objects that are always passed by
+    :cpp:class:`shared_ptr` in C++.
+    """
+
     def __repr__(self):
         className = type(self).__name__
         return className + "({" + ", ".join("%r: %r" % (key, value) for key, value in self.items()) + "})"
@@ -54,10 +70,34 @@ class GenericMapS:
     items = Mapping.items
 
 
-MutableMapping.register(MutableGenericMapS)
+GenericMap.register(str, GenericMapS)
+Mapping.register(GenericMapS)
 
-@continueClass  # noqa F811
-class MutableGenericMapS:
+
+class MutableGenericMap(GenericMap):
+    """An abstract `~collections.abc.MutableMapping` for use when sharing a
+    map between C++ and Python.
+
+    For compatibility with C++, ``MutableGenericMap`` has the following
+    restrictions:
+
+        - all keys must be of the same type
+        - values must be built-in types or subclasses of
+          `lsst.afw.typehandling.Storable`. Almost any user-defined class in
+          C++ or Python can have `~lsst.afw.typehandling.Storable` as a mixin.
+
+    As a safety precaution, `~lsst.afw.typehandling.Storable` objects that are
+    added from C++ may be copied when you retrieve them from Python, making it
+    impossible to modify them in-place. This issue does not affect objects that
+    are added from Python, or objects that are always passed by
+    :cpp:class:`shared_ptr` in C++.
+
+    Notes
+    -----
+    Key-type specializations of ``MutableGenericMap`` are available as, e.g.,
+    ``MutableGenericMap[str]``.
+    """
+
     # Easier than making MutableGenericMap actually inherit from MutableMapping
     setdefault = MutableMapping.setdefault
     update = MutableMapping.update
@@ -73,3 +113,7 @@ class MutableGenericMapS:
                 return default
             else:
                 raise
+
+
+MutableGenericMap.register(str, MutableGenericMapS)
+MutableMapping.register(MutableGenericMapS)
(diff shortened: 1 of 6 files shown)


python/lsst/afw/typehandling/_GenericMap.py
```diff
@@ -73,3 +113,49 @@ def pop(self, key, default=None):
...
+    def __call__(cls, *args, **kwargs):  # noqa N805, non-self first param
+        if len(cls.TEMPLATE_PARAMS) != 1:
+            raise ValueError("AutoKeyMeta requires exactly one template parameter")
+        dtypeKey = cls.TEMPLATE_PARAMS[0]
+        dtype = kwargs.get(dtypeKey, None)
+
+        # Try to infer dtype if not provided
+        # Assume homogeneous keys; constructor will raise TypeError if they're not
+        if dtype is None and len(args) >= 1:
+            inputData = args[0]
```
parejkoj: The block of logic inside this if statement is complicated enough that it might warrant its own method for clarity. The nestedness of these make me want a bunch of unittests with different types of `inputData`.

python/lsst/afw/typehandling/_GenericMap.py
```diff
@@ -73,3 +113,49 @@ def pop(self, key, default=None):
...
+            if inputData:
+                if isinstance(inputData, Mapping):
+                    # mapping to copy
+                    firstKey = iter(inputData.keys()).__next__()
+                    dtype = type(firstKey)
+                elif not isinstance(inputData, str):
```
parejkoj: `if`, `elif`, but where's the `else`? What happens if `inputData` is a string?

python/lsst/afw/typehandling/_SimpleGenericMap.py
```diff
@@ -19,10 +19,48 @@
...
+        - all keys must be of the same type
+        - values must be built-in types or subclasses of
+          `lsst.afw.typehandling.Storable`. Almost any user-defined class in
+          C++ or Python can have `~lsst.afw.typehandling.Storable` as a mixin.
+
+    As a safety precaution, `~lsst.afw.typehandling.Storable` objects that are
+    added from C++ may be copied when you retrieve them from Python, making it
+    impossible to modify them in-place. This issue does not affect objects that
+    are added from Python, or objects that are always passed by
+    :cpp:class:`shared_ptr` in C++.
```
parejkoj: It's too bad that the above docstring has to be copied into each of the different `*GenericMap`s. I don't think there's a way around it though.

python/lsst/afw/typehandling/testUtils.py
```diff
@@ -102,6 +103,13 @@ def getTestData(cls, keyClass):
         """
         return {keyClass(key): value for key, value in cls._testData.items()}
 
+    @staticmethod
+    def getValidKeys(mapClass):
+        try:
+            return {mapClass.dtype}
+        except AttributeError:
+            return {cls.dtype for cls in mapClass.__subclasses__()}
```
parejkoj: Why is this a `set`? I would have thought there was exactly one valid key type.

tests/test_simpleGenericMap.py
```diff
@@ -42,58 +42,84 @@ def makeMap(mapType, values):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.targets = {(str, SimpleGenericMapS)}
+        cls.implicitKeyTargets = {SimpleGenericMap}  # Must be mapping-constructible
+        cls.targets = {SimpleGenericMap[str]}  # Must be default-constructible
```
parejkoj: Would you ever have more than one target in this `set`?

tests/test_simpleGenericMap.py
```diff
@@ -42,58 +42,84 @@ def makeMap(mapType, values):
...
+    def testUnknownKeys(self):
+        for target in self.implicitKeyTargets:
+            with self.assertRaises(TypeError):
+                target()
+            # Should not raise
+            target(dtype=str)
+
+    def testMixedKeys(self):
+        badData = {"What is the product of six and eight?": "Ultimate Question",
+                   42: "Ultimate Answer",
```
parejkoj: `badData`? Don't you mean `bestData`?

tests/test_simpleGenericMap.py
```diff
@@ -42,58 +42,84 @@ def makeMap(mapType, values):
...
-        for (keyType, target) in self.targets:
-            keys = self.getTestData(keyType).keys()
-            for value in self.getTestData(keyType).values():
-                self.checkFromKeys(keyType, target, keys, value,
-                                   msg=" class=%s, value=%r" % (target, value))
-            self.checkFromKeysDefault(keyType, target, keys, msg=" class=%s, no value" % (target))
+        for target in self.targets:
+            for keyType in self.getValidKeys(target):
+                keys = self.getTestData(keyType).keys()
+                for value in self.getTestData(keyType).values():
```
parejkoj: Triply nested loops in tests make me nervous. I don't know if there's a better approach, but this tends to make debugging failed tests more difficult.

python/lsst/afw/typehandling/_GenericMap.py
kfindeisen: I already cover both inputs in testing. 👍 for factoring the code.

python/lsst/afw/typehandling/_GenericMap.py
kfindeisen: The map constructor will fail, then, same as if you tried to make a `dict` from a string.

python/lsst/afw/typehandling/_SimpleGenericMap.py
kfindeisen: Well, in pybind11 I could factor the common text into its own function. Would it make sense to do the same here (assigning to `__doc__`)?

python/lsst/afw/typehandling/testUtils.py
kfindeisen: There can be multiple types if `mapClass` is a `TemplateMeta` class. Should I test for that directly? I thought this way was more Pythonic.

tests/test_simpleGenericMap.py
kfindeisen: In `targets`, certainly. `implicitKeyTargets` I'll admit is less likely. I guess I could make that one a scalar.

python/lsst/afw/typehandling/_GenericMap.py
parejkoj: Maybe it will be clear in refactoring, but a comment about the missing `else` may be worthwhile.

python/lsst/afw/typehandling/_SimpleGenericMap.py
parejkoj: All of the various alternate approaches I've seen feel not great and/or hacky to me, so I think we're stuck with it. Unless @jonathansick has any better ideas?

python/lsst/afw/typehandling/testUtils.py
parejkoj: Ah, that's fine. Maybe a comment about that difference here (because in my head this class is always a `GenericMap`, but that's clearly not true).

kfindeisen: Oh hey, this method doesn't have a docstring. Shame on me!

tests/test_simpleGenericMap.py
kfindeisen: I provide the `msg` argument to the assertions for exactly that reason. I think it's fine as-is.

PR #475, opened 2019-07-10
DM-20499: Add basic stringification to SkyWcs

parejkoj:


python/lsst/afw/geom/skyWcs/skyWcs.cc
```diff
@@ -149,6 +150,10 @@ PYBIND11_MODULE(skyWcs, mod) {
     // Do not wrap readStream or writeStream because C++ streams are not easy to wrap.
     cls.def_static("readString", &SkyWcs::readString);
     cls.def("writeString", &SkyWcs::writeString);
+
+    utils::python::addOutputOp(cls, "__str__");
+    // Alternative repr: call writeString for the full AST output (which is very long).
```
kfindeisen: The comment's a bit unclear, given that this seems to *not* be what you're doing. Maybe something like "We could implement repr using writeString instead of operator<<..."?

src/geom/SkyWcs.cc
```diff
@@ -334,7 +334,26 @@ std::shared_ptr<typehandling::Storable> SkyWcs::cloneStorable() const {
...
+    if (isFits()) {
+        os << "FITS standard SkyWcs:";
+    } else {
+        os << "Non-standard SkyWcs (Frames: ";
+        // Print the frames in index order (frames are numbered from 1).
+        std::string delimiter = "";
+        for (size_t i = 1; i <= getFrameDict()->getAllDomains().size(); ++i) {
+            os << delimiter << getFrameDict()->getFrame(i)->getDomain();
+            delimiter = ", ";
+        }
```
kfindeisen: This looks very complicated. Could you add a description or an example to the documentation comments for `toString` and `operator<<`? At the very least, it would be helpful for the user to know that the string has line breaks (so, e.g., you shouldn't print a `SkyWcs` in a sentence).

Handy convention from the language that shall not be named: "The exact details of the representation are unspecified and subject to change, but the following may be regarded as typical:"

src/geom/SkyWcs.cc
```diff
@@ -334,7 +334,26 @@ std::shared_ptr<typehandling::Storable> SkyWcs::cloneStorable() const {
...
+        os << "): ";
+    }
+    std::string delimiter = "\n";
+    os << delimiter << "Sky Origin: " << getSkyOrigin();
+    os << delimiter << "Pixel Origin: " << getPixelOrigin();
+    os << delimiter << "Pixel Scale: " << getPixelScale().asArcseconds() << " arcsec/pixel";
```
kfindeisen: Did you mean to have an `<< endl` after each statement?

tests/test_skyWcs.py
```diff
@@ -545,6 +545,15 @@ def testAgainstAstropyWcs(self):
...
             self.assertSkyWcsAstropyWcsAlmostEqual(skyWcs=skyWcs, astropyWcs=astropyWcs, bbox=bbox)
 
+    def testStr(self):
+        """Test that we can get something coherent when printing a SkyWcs.
+        """
+        cdMatrix = makeCdMatrix(scale=self.scale)
+        skyWcs = makeSkyWcs(crpix=self.crpix, crval=self.crvalList[0], cdMatrix=cdMatrix)
+        self.assertIn(f"Sky Origin: {self.crvalList[0]}", str(skyWcs))
+        self.assertIn(f"Pixel Origin: {self.crpix}", str(skyWcs))
+        self.assertIn(f"Pixel Scale: ", str(skyWcs))
```
kfindeisen: Derive the expected value from `cdMatrix`?

src/geom/SkyWcs.cc
parejkoj: No, I intentionally put the delimiter first (in part so that whatever is produced in the `if isFits()` block is independent of the last three lines).

tests/test_skyWcs.py
parejkoj: I don't understand this comment. You mean derive the pixel scale from it? It's potentially more complicated than just `cdMatrix` (e.g. optical distortions). I could put in a call to `getPixelScale()` here, but I decided I didn't want to be too specific in the unittest (e.g. units conversions).

src/geom/SkyWcs.cc
parejkoj: I put in the newlines because it seemed much more readable that way. For complicated objects, I'm never quite sure whether to try to keep it on one line, or expand it into a multi-line but more readable thing.

parejkoj: I've added an example of what the output could look like to the `toString()` docs.

src/geom/SkyWcs.cc
kfindeisen: Ok, I totally missed that. In the future, please use `endl` instead of `"\n"`. I realize the Stack is single-platform anyway, but it's the principle of the thing.

include/lsst/afw/geom/SkyWcs.h
```diff
@@ -619,9 +628,7 @@ std::shared_ptr<TransformPoint2ToPoint2> getPixelToIntermediateWorldCoords(SkyWc
                                                                            bool simplify = true);
 
 /**
- * Print a SkyWcs to an ostream
- *
- * For now it just prints "SkyWcs"; eventually it would be nice to have a summary
+ * Print a SkyWcs to an ostream (delegates to SkyWcs.toString()).
```
kfindeisen: This won't link correctly. You want `SkyWcs::toString` (with or without parentheses).

src/geom/SkyWcs.cc
parejkoj: `endl` and `\n` mean the same thing, except the former flushes the stream. Where you thinking of `\n` vs. `\r`?

include/lsst/afw/geom/SkyWcs.h
kfindeisen: Huh. Never mind, [it works](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/namespacelsst_1_1afw_1_1geom.html#a9386e6373e2f43d46d287e8a64867ad2). Undocumented feature in Doxygen?

src/geom/SkyWcs.cc
kfindeisen: Aha, apparently getting the right line ending is the *stream*'s responsibility: https://stackoverflow.com/a/8033980. I didn't know that, never mind.

PR #480, opened 2019-07-31
DM-17982: Reiplement ExposureInfo using GenericMap

kfindeisen:
This PR changes the implementation of `ExposureInfo` to store components in a `GenericMap` rather than in dedicated pointers, and provides a generic API to add and remove components. The `filter`, `metadata`, and `visitInfo` components have not been migrated, because `metadata` cannot be made compatible with `GenericMap` due to package dependencies, `filter` cannot be round-tripped through `table::io` persistence, and both `filter` and `visitInfo` have a custom persisted form that would be broken by migration.

I'm a bit concerned about the API introduced in 8dcab8b23ab28486d1938f6b570b50c825f981fb. Not only does it expose the `GenericMap` back-end (which may be inevitable, given the requirements for this epic), but it's also fairly complex:
- setComponent(Key, shared_ptr<? extends Storable>)
- ~setComponent(string, ?)~
- ~hasComponent(Key<shared_ptr<? extends Storable>>)~
- hasComponent(Key\<other>)
- getComponent(Key<shared_ptr<? extends Storable>>)
- removeComponent(Key)

Note that there will be at least two additional declarations needed once it's safe to store components by value.

I also haven't addressed the problem of how to coordinate names of components across packages. Are the methods used for table columns, measurement plugins, mask planes, etc. adequate?

### DIFF
--- a/include/lsst/afw/image/ExposureInfo.h
+++ b/include/lsst/afw/image/ExposureInfo.h
@@ -68,8 +68,6 @@ class TransmissionCurve;
  *  The constness semantics of the things held by ExposureInfo are admittedly a bit of a mess,
  *  but they're that way to preserve backwards compatibility for now.  Eventually I'd like to make
  *  a lot of these things immutable, but in the meantime, here's the summary:
- *   - Psf is held by non-const pointer, and you can get a non-const pointer via a
- *     non-const member function accessor and a const pointer via a const member function accessor.
  *   - PhotoCalib is held by const pointer and only returned by const pointer (it's immutable).
  *   - Detector is held by const pointer and only returned by const pointer (but if you're
  *     in Python, SWIG will have casted all that constness away).
@@ -91,6 +89,8 @@ class ExposureInfo final {
 public:
     /// Standard key for looking up the Wcs.
     static typehandling::Key<std::string, std::shared_ptr<geom::SkyWcs const>> const KEY_WCS;
+    /// Standard key for looking up the point-spread function.
+    static typehandling::Key<std::string, std::shared_ptr<detection::Psf const>> const KEY_PSF;
 
     /// Does this exposure have a Wcs?
     bool hasWcs() const;
@@ -149,17 +149,13 @@ public:
     void setMetadata(std::shared_ptr<daf::base::PropertySet> metadata) { _metadata = metadata; }
 
     /// Does this exposure have a Psf?
-    bool hasPsf() const { return static_cast<bool>(_psf); }
+    bool hasPsf() const;
 
     /// Return the exposure's point-spread function
-    std::shared_ptr<detection::Psf> getPsf() const { return _psf; }
+    std::shared_ptr<detection::Psf const> getPsf() const;
 
     /// Set the exposure's point-spread function
-    void setPsf(std::shared_ptr<detection::Psf const> psf) {
-        // Psfs are immutable, so this is always safe; it'd be better to always just pass around
-        // const or non-const pointers, instead of both, but this is more backwards-compatible.
-        _psf = std::const_pointer_cast<detection::Psf>(psf);
-    }
+    void setPsf(std::shared_ptr<detection::Psf const> psf);
 
     /// Does this exposure have a valid Polygon
     bool hasValidPolygon() const { return static_cast<bool>(_validPolygon); }
@@ -452,7 +448,6 @@ private:
         }
     }
 
-    std::shared_ptr<detection::Psf> _psf;
     std::shared_ptr<PhotoCalib const> _photoCalib;
     std::shared_ptr<cameraGeom::Detector const> _detector;
     std::shared_ptr<geom::polygon::Polygon const> _validPolygon;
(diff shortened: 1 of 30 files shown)


python/lsst/afw/table/_exposure.cc
```diff
@@ -177,7 +177,7 @@ PyExposureCatalog declareExposureCatalog(WrapperCollection &wrappers) {
 }  // anonymous namespace
 
 void wrapExposure(WrapperCollection &wrappers) {
-    wrappers.addSignatureDependency("lsst.afw.geom");
+    // wrappers.addSignatureDependency("lsst.afw.geom");
```
TallJimbo: I assume this is another "uncomment once afw.geom uses WrapperCollection"?

include/lsst/afw/image/ExposureInfo.h
```diff
@@ -344,6 +345,11 @@ class ExposureInfo final {
     std::shared_ptr<ApCorrMap> _apCorrMap;
     std::shared_ptr<image::VisitInfo const> _visitInfo;
     std::shared_ptr<TransmissionCurve const> _transmissionCurve;
+
+    // Class invariant: all values in _components are Storable or shared_ptr<Storable>
+    // Class invariant: if a key is absent from this object, the key may either
+    //     be absent from _components or may map to a null pointer.
```
TallJimbo: I don't necessarily object, but I'm curious why you decided to have two ways to represent an absent component.

include/lsst/afw/image/ExposureInfo.h
```diff
@@ -213,6 +213,125 @@ class ExposureInfo final {
...
+     *       `hasComponent([return value])` shall return `false` after this
+     *       method returns. This is for compatibility with old ExposureInfo
+     *       idioms, which often use assignment of null to indicate no data.
+     */
+    template <class T>
+    typehandling::Key<std::string, T> setComponent(std::string const& key, T const& object) {
+        auto typedKey = typehandling::makeKey<T>(key);
+        setComponent(typedKey, object);
+        return typedKey;
+    }
```
TallJimbo: What do you think about removing this overload, in order to encourage a convention (which we'd also have to document somewhere) that packages that add new components to `ExposureInfo` should defined a module/namespace-level `Key` constant that should be used to access it?  That would cut down on the use of string literals, add a bit of type safety, and simplify the `ExposureInfo` interface a bit.

python/lsst/afw/image/exposureInfo.cc
```diff
@@ -45,6 +46,44 @@ namespace {
 
 using PyExposureInfo = py::class_<ExposureInfo, std::shared_ptr<ExposureInfo>>;
 
+// Template methods where we can use pybind11's overload resolution (T is input)
+template <class T>
+void declareGenericMethods(PyExposureInfo &cls) {
+    using Class = PyExposureInfo::type;
+    // py::overload_cast can't deduce return type... because Key is not pybind11-wrapped?
```
TallJimbo: FWIW, I've (rarely) seen the same problem with `overload_cast` on other methods (can't remember where, unfortunately) and went with the same workaround.  I also couldn't figure out what was wrong.

kfindeisen:
> And while I too am a bit bothered by not having and kind of namespace conventions in place for new components, I think it's fair to say that not only do we not have a name collision problem in the other cases you mentioned in practice

I wasn't worried about name collisions, actually, but about the opposite problem. For example, somebody who tries to call `getComponent("wcs")` from Python might conclude an exposure doesn't have a WCS, when they should have called `getComponent("SKYWCS")`.

> One big-picture question I didn't ask inline: what make `VisitInfo`'s persistence incompatible with being a generic omponent? I thought it'd be like `SkyWcs`, with its metadata representation always written but not read if the archive representation is present.

No, `VisitInfo` is [only persisted to and from metadata](https://github.com/lsst/afw/blob/master/src/image/ExposureInfo.cc#L232-L235). The `table::io` code exists, but is not used.

include/lsst/afw/image/ExposureInfo.h
kfindeisen: I think I was too tired to properly separate the interface and implementation. I'll switch to making `setComponent(nullptr)` remove the member.

include/lsst/afw/image/ExposureInfo.h
kfindeisen: I'm surprised to hear you request this, since having such an adder for `MutableGenericMap` was your idea. But I'm fine with removing it (it's redundant in Python anyway).

TallJimbo: Yeah, I suppose it's just the slight distinction between the roles of a bare `GenericMap` and `ExposureInfo` that swings me in opposite directions.  It isn't a terribly strongly preference in either case.

TallJimbo:
> No, VisitInfo is only persisted to and from metadata. The table::io code exists, but is not used.

:slightly_frowning_face:  I guess that's fodder for another ticket, but maybe it should wait until ongoing `astro_metadata_translator.ObservationInfo` vs. `VisitInfo` questions have shaken out a bit.

PR #485, opened 2019-09-13
DM-21181: Add getLocalCalibration function to photoCalib.

morriscb:
Add localCalib unittest.

Fix flux names in local calib tests.

Add getLocalCalibration function to header.

Add pybindings.

Add test comment.

Remove superfluous test.

Debug getLocalCalibration

clang-format of .h file


morriscb:
I mean, I change the commit message? It just has all the individual commits squashed in but the "title" reflects what is actually going in there. How would you recommend fixing it?

jdswinbank:
Looks to me like the commit message here should simply be `Add getLocalCalibration function to photoCalib' — the rest is unnecessary.

morriscb:
Done.

PR #486, opened 2019-09-17
DM-19582: Add cross-language GenericMap unit tests

kfindeisen:
This PR adds unit tests to `GenericMap` that pass objects from Python to C++ and/or vice versa. This exposes behavior that was not adequately tested with previous tests.

Bugs fixed:
- Building `afw` with Scons no longer issues an "ignoring a non-existent file" warning
- `GenericMap`'s policy on type conversions is now explicitly described in the API docs
- Python builtin integers are no longer demoted to `int`, nor Python floats to `float`, when inserted into a `GenericMap` (but see DM-21268, below)
- `Storable` values (i.e., non-pointers) can now be retrieved from a `GenericMap` in Python
- Storage and retrieval of integers in C++ now work consistently across platforms

Bugs deferred:
- `GenericMap`'s weakly-typed `insert` method turns C-strings into `bool` [[DM-21216](https://jira.lsstcorp.org/browse/DM-21216)].
- `GenericMap` cannot consistently handle integers and floats from Python, particularly Numpy's explicit-length types [[DM-21268](https://jira.lsstcorp.org/browse/DM-21268)].
- A Python `Storable` may be prematurely garbage-collected [[DM-21314](https://jira.lsstcorp.org/browse/DM-21314)].

### DIFF
--- a/include/lsst/afw/typehandling/GenericMap.h
+++ b/include/lsst/afw/typehandling/GenericMap.h
@@ -26,7 +26,6 @@
 #define LSST_AFW_TYPEHANDLING_GENERICMAP_H
 
 #include <algorithm>
-#include <cstdint>
 #include <functional>
 #include <ostream>
 #include <memory>
@@ -409,8 +408,9 @@ public:
      * type (which may be `void`). Through any combination of overloading or
      * templates, the visitor must accept values of the following types:
      *      * either `bool` or `bool const&`
-     *      * either `std::int32_t` or `std::int32_t const&`
-     *      * either `std::int64_t` or `std::int64_t const&`
+     *      * either `int` or `int const&`
+     *      * either `long` or `long const&`
+     *      * either `long long` or `long long const&`
      *      * either `float` or `float const&`
      *      * either `double` or `double const&`
      *      * `std::string const&`
@@ -499,7 +499,9 @@ protected:
      *
      * Keys of any subclass of Storable are implemented using PolymorphicValue to preserve type.
      */
-    using StorableType = boost::variant<bool, std::int32_t, std::int64_t, float, double, std::string,
+    // Use int, long, long long instead of int32_t, int64_t because C++ doesn't
+    // consider signed integers of the same length but different names equivalent
+    using StorableType = boost::variant<bool, int, long, long long, float, double, std::string,
                                         PolymorphicValue, std::shared_ptr<Storable const>>;
 
     /**
--- a/include/lsst/afw/typehandling/GenericMap.h
+++ b/include/lsst/afw/typehandling/GenericMap.h
@@ -237,7 +237,11 @@ public:
     /**
      * Return a reference to the mapped value of the element with key equal to `key`.
      *
-     * @tparam T the type of the element mapped to `key`
+     * @tparam T the type of the element mapped to `key`. It may be the exact
+     *           type of the element, if known, or any type to which its
+     *           references or pointers can be implicitly converted (e.g.,
+     *           a superclass). For example, references to built-in types are
+     *           not convertible, so you can't retrieve an `int` with `T=long`.
      * @param key the key of the element to find
      *
      * @return a reference to the `T` mapped to `key`, if one exists
--- a/python/lsst/afw/typehandling/_GenericMap.cc
+++ b/python/lsst/afw/typehandling/_GenericMap.cc
@@ -176,7 +176,7 @@ void declareMutableGenericMap(utils::python::WrapperCollection& wrappers, std::s
     wrappers.wrapType(PyClass(wrappers.module, className.c_str(), docstring.c_str()),
                       [](auto& mod, auto& cls) {
                           // Don't rewrap members of GenericMap
-                          declareMutableGenericMapTypedMethods<std::shared_ptr<Storable>>(cls);
+                          declareMutableGenericMapTypedMethods<std::shared_ptr<Storable const>>(cls);
                           declareMutableGenericMapTypedMethods<bool>(cls);
                           // TODO: int32 and float are suppressed for now, see DM-21268
                           declareMutableGenericMapTypedMethods<std::int64_t>(cls);  // chosen for builtins.int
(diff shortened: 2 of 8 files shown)


tests/test_simpleGenericMap.py
```diff
@@ -216,6 +218,148 @@ def testClear(self):
...
+                        'question': PyStorable('Unknown'),
+                        }
+        testmap.update(storableData)
+
+        for key, value in storableData.items():
+            self.assertIn(key, testmap, msg=msg)
+            self.assertEqual(value, testmap[key], msg='key=' + key + ', ' + msg)
+            # Exercise C++ equality operator
+            cppLib.assertKeyValue(testmap, key, PyStorable(value.value))
+            # TODO: move to test_Storable.py
```
TallJimbo: Was this TODO note intended for this ticket or a future one?

include/lsst/afw/typehandling/GenericMap.h
```diff
@@ -409,8 +412,9 @@ class GenericMap {
      * type (which may be `void`). Through any combination of overloading or
      * templates, the visitor must accept values of the following types:
      *      * either `bool` or `bool const&`
-     *      * either `std::int32_t` or `std::int32_t const&`
-     *      * either `std::int64_t` or `std::int64_t const&`
+     *      * either `int` or `int const&`
+     *      * either `long` or `long const&`
+     *      * either `long long` or `long long const&`
```
TallJimbo: It's a pity this problem with type equivalency bleeds into the public interface; I'd have preferred to keep the interface in terms of `std::int32_t` and `std::int64_t`.  But I also don't see a good way around it; it's probably not worth fighting the language's native treatment of integer types to that extent, especially in the short term with DM-21327 on the horizon.  But if you do see a good way to do that, it might be worth ticketing.

python/lsst/afw/typehandling/_GenericMap.cc
```diff
@@ -74,7 +74,16 @@ template <typename K>
...
+    auto variant = (self.*callable)(key);
+    try {
+        // py::cast can't handle this implicit conversion; do it by hand
+        Storable const& value = boost::get<PolymorphicValue const&>(variant);
+        // Prevent segfaults when assigning a Key<Storable> to Python variable, then deleting from map
+        // No existing code depends on being able to modify an item stored by value
+        return py::cast(value.cloneStorable());
+    } catch (boost::bad_get const&) {
+        return py::cast(variant);
+    }
```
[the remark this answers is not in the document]
kfindeisen: There is. As noted in the main `GenericMap` code, I did not use it in the belief that there's no `std::variant` equivalent, but it turns out [it exists under another name](https://en.cppreference.com/w/cpp/utility/variant/get_if).

tests/test_simpleGenericMap.py
kfindeisen: This one. I could swear I removed it in 51b1cdaae25e7e80fa500bbf465c01a72345416d... :confused:

include/lsst/afw/typehandling/GenericMap.h
kfindeisen: If I did, I would have done it. Honestly, even the existing level of overloading (e.g., `float` vs. `double`) is problematic.

Note that outside of a testing context, you could write these visitors with templates, and the distinction gets swept under the rug.

python/lsst/afw/typehandling/_GenericMap.cc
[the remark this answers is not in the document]
kfindeisen: Hmm... there's a problem. While I should be able to switch the the pointer-based version of `get` here and in the C++ accessor for `Storable`, if I do it in the `shared_ptr<Storable>` accessor as well, I can't easily distinguish between a variant that contains something other than a pointer and a variant that contains `nullptr` -- `get` will return null in both cases.

I could keep the throwing `get` there, or I could use [`which`](https://www.boost.org/doc/libs/1_71_0/doc/html/boost/variant.html#id-1_3_46_5_4_1_1_16_1-bb) (which requires an integer-based index, and is therefore brittle), or I could use [`type`](https://www.boost.org/doc/libs/1_71_0/doc/html/boost/variant.html#id-1_3_46_5_4_1_1_16_3-bb) (which involves more RTTI). I assume that will also be too much of a performance penalty for you?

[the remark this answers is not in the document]
TallJimbo: Would a direct `py::cast` (as in the `catch` block here) also do what you want in the case that the variant contains a `nullptr`?  After all, you can't clone that either.

If not, just leave as-is with the throwing `get`.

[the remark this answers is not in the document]
kfindeisen: I ended up only changing the pybind11 code. There's less incentive to use a pointer `get` in the C++ accessors (because the response on a mismatch is to throw an exception anyway), and the inability to implicitly convert `PolymorphicValue*` to `Storable*` makes the pointer-based code much more awkward. Somehow that proxy class isn't working out as well as one would expect... :disappointed:

include/lsst/afw/typehandling/GenericMap.h
kfindeisen: Hmm... you know, I think we might be able to decouple the API from the `boost::variant` by providing an internal adapter that has explicit methods for each `variant` type but then delegates to the user-provided callable in a way that allows implicit conversions (I'm not completely sure this is possible, as the adapter would have to be templated on the callable, though not on the value type). That might be a worthwhile ticket.

TallJimbo: I'm pretty sure it's doable in C++; making it work with Python callable would at least be harder.  Worth a ticket, though I don't think I'd make it a priority right now.

kfindeisen: Written up as [DM-21382](https://jira.lsstcorp.org/browse/DM-21382).

PR #488, opened 2019-09-30
DM-21355: Try to prevent NaN values turning up in FITS headers

timj:
The FITS standard doesn't allow NaN as a value. NaNs turning up in a PropertyList can confuse things so this change traps this and either drops the entire item or replaces with FITS undefined value.

### DIFF
--- a/src/fits.cc
+++ b/src/fits.cc
@@ -66,10 +66,28 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
         } else if (type == typeid(int)) {
             out += (boost::format("%20d") % metadata.get<int>(name)).str();
         } else if (type == typeid(double)) {
-            // use G because FITS wants uppercase E for exponents
-            out += (boost::format("%#20.17G") % metadata.get<double>(name)).str();
+            double value = metadata.get<double>(name);
+            if (!std::isnan(value)) {
+                // use G because FITS wants uppercase E for exponents
+                out += (boost::format("%#20.17G") % value).str();
+            } else {
+                LOGLS_WARN("afw.fits",
+                           boost::format("In %s, found NaN in metadata item '%s'") %
+                                         BOOST_CURRENT_FUNCTION % name);
+                // Convert it to FITS undefined
+                out += " ";
+            }
         } else if (type == typeid(float)) {
-            out += (boost::format("%#20.15G") % metadata.get<float>(name)).str();
+            float value = metadata.get<float>(name);
+            if (!std::isnan(value)) {
+                out += (boost::format("%#20.15G") % value).str();
+            } else {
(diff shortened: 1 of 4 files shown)


src/fits.cc
```diff
@@ -66,10 +66,28 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
...
+            double value = metadata.get<double>(name);
+            if (!std::isnan(value)) {
+                // use G because FITS wants uppercase E for exponents
+                out += (boost::format("%#20.17G") % value).str();
+            } else {
+                LOGLS_WARN("afw.fits",
+                           boost::format("In %s, found NaN in metadata item '%s'") %
+                                         BOOST_CURRENT_FUNCTION % name);
+                // Convert it to FITS undefined
+                out += " ";
```
erykoff: Is this FITS undefined, or is it `"BLANK   "`? https://heasarc.gsfc.nasa.gov/docs/fcg/standard_dict.html
It could be that there's another translation layer somewhere else in the bowels of afw, or that I'm misreading something.

src/fits.cc
```diff
@@ -66,10 +66,28 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
...
+                out += (boost::format("%#20.15G") % value).str();
+            } else {
+                LOGLS_WARN("afw.fits",
+                           boost::format("In %s, found NaN in metadata item '%s'") %
+                                         BOOST_CURRENT_FUNCTION % name);
+                // Convert it to FITS undefined
+                out += " ";
+            }
         } else if (type == typeid(std::nullptr_t)) {
             out += " ";
```
erykoff: Hmmm, I guess this is what has always been used in `afw`, so I clearly don't have the full picture.

tests/test_frameSetUtils.py
```diff
@@ -73,6 +73,18 @@ def getCrpix(self, metadata):
...
+        """
+        metadata = self.makeMetadata()
+        metadata["ISNAN"] = float("NaN")
+
+        # This will issue a warning from C++ but that warning can not be
+        # redirected to python logging machinery.
+        # This code causes a SEGV without DM-21355 implemented
+        frameSet1 = readFitsWcs(metadata, strip=False)
+        self.assertEqual(type(frameSet1), ast.FrameSet)
+
```
erykoff: It makes sense that this test `SIGABRT`s on master, this was an untested edge case that could lead to an `SIGABRT` and it should now be tested (and fixed).

src/fits.cc
timj: FITS undefined (there is a keyword and an `=` sign)

src/fits.cc
timj: I added support for FITS undefined a few months back.

PR #496, opened 2019-11-05
DM-21314: Investigate GC problems with Storable

kfindeisen:
This PR fixes a bug that prevented `Storable` objects from being usable unless they were continuously referenced by Python code. The fix depends on lsst/utils#75.

It also implements and tests cloning of Python `Storable`s, which was left unsupported due to an oversight.

### DIFF
--- a/include/lsst/afw/typehandling/python.h
+++ b/include/lsst/afw/typehandling/python.h
@@ -61,7 +61,7 @@ public:
         PYBIND11_OVERLOAD_NAME(std::size_t, Base, "__hash__", hash_value, );
     }
 
-    bool equals(Base const& other) const noexcept override {
+    bool equals(Storable const& other) const noexcept override {
         PYBIND11_OVERLOAD_NAME(bool, Base, "__eq__", equals, other);
     }
 };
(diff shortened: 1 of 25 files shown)


include/lsst/afw/typehandling/python.h
```diff
@@ -51,7 +51,12 @@ class StorableHelper : public Base {
 public:
     using Base::Base;
 
-    // Can't wrap clone(); PYBIND11_OVERLOAD chokes on unique_ptr return type
+    std::shared_ptr<Storable> cloneStorable() const override {
+        /* __deepcopy__ takes an optional dict, but PYBIND11_OVERLOAD_* doesn't
+         * allow translation between different argument lists in C++ and Python
+         */
```
TallJimbo: Are you saying this will fail if `__deepcopy__` is called with that argument?  (Not unreasonable - using it would at least be significantly more complex, if it's even possible - just asking for clarification in the comment.)

kfindeisen: I'm not sure if this answers your question, but the macro will not compile if you give it arguments (like `py::none()` or `py::dict()`) that are not present in `cloneStorable`'s argument list.

PR #499, opened 2019-11-14
DM- 22070: Add PixelAreaBoundedField and deprecate PixelScaleBoundedField

TallJimbo:


include/lsst/afw/math/PixelScaleBoundedField.h
```diff
@@ -44,7 +44,8 @@ namespace math {
  *
  * Typically used to move an image or source flux between surface brightness and fluence space.
  */
-class PixelScaleBoundedField : public BoundedField {
+class [[deprecated("Replaced by PixelAreaBoundedField; will be removed on DM-22192.")]]
```
erykoff: I thought that deprecation messages should say what release they'll be deprecated before (e.g., prior to 20.0)

python/lsst/afw/math/mathLib.py
```diff
@@ -49,3 +50,10 @@
 from .background import *
 import lsst.afw.image.pixel  # for SinglePixel, needed by the warping functions
 from .warpExposure import *
+
+from lsst.utils import deprecate_pybind11 as _deprecate_pybind11  # don't pollute package namespace
+
+PixelScaleBoundedField = _deprecate_pybind11(
+    PixelScaleBoundedField,
+    reason="Replaced by PixelAreaBoundedField; will be removed on DM-22192."
```
erykoff: Same as above for the version mention.

PR #501, opened 2019-11-15
DM-22147: Add method to ChebyshevBoundedField to approximate another BoundedField

erykoff:
Add a python-only method to ChebyshevBoundedField as a convenience to approximate another BoundedField.  Will typically be used to approximate a PixelAreaBoundedField which is very slow to evaluate directly.


erykoff:
I was thinking that it would be good to specify _what_ is being approximated in the name, but I see your point that "BoundedField" is implied by the name of the first argument.  So I think that `approximate()` is most consistent with the rest of the class naming and should still be clear enough given the context.

PR #502, opened 2019-11-19
DM-22260: Add metadata to temporary catalog to be persisted.

erykoff:
Also add a persistence test to confirm metadata is round-tripped.

### DIFF
--- a/src/table/Exposure.cc
+++ b/src/table/Exposure.cc
@@ -228,6 +228,9 @@ void ExposureFitsWriter::_writeTable(std::shared_ptr<BaseTable const> const &t, 
     }
     _mapper = _helper.makeWriteMapper(inTable->getSchema());
     std::shared_ptr<BaseTable> outTable = BaseTable::make(_mapper.getOutputSchema());
+    // Put the metadata from inTable in outTable
+    std::shared_ptr<daf::base::PropertyList> metadata = inTable->getMetadata();
+    outTable->setMetadata(metadata);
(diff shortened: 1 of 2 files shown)


tests/test_exposureTable.py
```diff
@@ -189,6 +192,8 @@ def testPersistence(self):
             self.assertIsNone(cat1[1].getVisitInfo())
             self.assertIsNone(cat1[0].getDetector())
             self.assertDetectorsEqual(cat1[1].getDetector(), self.detector)
+            for key in self.plist.keys():
```
timj: Don't need the `keys()` here.  Also please add a comment that you are doing it this way rather than direct equality because the property list read from the file will include extra keys.

src/table/Exposure.cc
```diff
@@ -228,6 +228,9 @@ void ExposureFitsWriter::_writeTable(std::shared_ptr<BaseTable const> const &t,
     }
     _mapper = _helper.makeWriteMapper(inTable->getSchema());
     std::shared_ptr<BaseTable> outTable = BaseTable::make(_mapper.getOutputSchema());
+    // Put the metadata from inTable in outTable
+    std::shared_ptr<daf::base::PropertyList> metadata = inTable->getMetadata();
+    outTable->setMetadata(metadata);
```
ktlim: More compact to say simply `outTable->setMetadata(inTable->getMetadata());`, no?
And if you keep the assignment, use `auto`.

PR #503, opened 2019-12-06
DM-22519: Fix GroupView iterator

SimonKrughoff:


python/lsst/afw/table/multiMatch.py
```diff
@@ -259,7 +259,7 @@ def __len__(self):
         return len(self.ids)
 
     def __iter__(self):
-        return self.ids
+        yield from self.ids
```
PaulPrice: I would have expected `return iter(self.ids)`.

tests/test_tableMultiMatch.py
```diff
@@ -134,6 +134,8 @@ def testGroupViewBuild(self):
         """
         allMatches = afwTable.GroupView.build(self.matchedCatalog)
         self.assertTrue(len(allMatches) > 0)
+        test_ids = [el for el in allMatches]
+        self.assertEqual(test_ids, list(allMatches.ids))
```
PaulPrice: Suggest not using `under_scores` where the convention is `camelCase`.

python/lsst/afw/table/multiMatch.py
SimonKrughoff: Aren't they essentially equivalent?  I'm happy to do either.

SimonKrughoff: I've implemented your recommendation.

PaulPrice: They may well be equivalent, but given that the method is called `__iter__`, I'm wired to expect to see `return iter(...)`.

PR #510, opened 2020-01-23
DM-23173: change W504 to W503 and fix the resulting flake8 warnings

r-owen:


r-owen:
The afw Python unit tests all pass, so I think it is safe to review.

However, Jenkins is still running (and I won't merge until it passes): https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/31070/pipeline/

If you'd rather wait, that's fine.

python/lsst/afw/geom/transformConfig.py
```diff
@@ -142,8 +142,8 @@ def validate(self):
             return
         if len(self.coeffs) == 1 or self.coeffs[0] != 0 or self.coeffs[1] == 0:
             raise RuntimeError(
-                "invalid radial transform coeffs %s: " % (self.coeffs,) +
-                "need len(coeffs)=0 or len(coeffs)>1, coeffs[0]==0, "
+                "invalid radial transform coeffs %s: " % (self.coeffs,)
```
timj: If you changed these lines to f-strings you could remove the `+` completely.

tests/test_background.py
```diff
@@ -602,8 +602,8 @@ def testBadRows(self):
             self.assertEqual(
                 np.mean(bkgdImage[0:100, 0:100].array), initialValue)
             if debugMode:
-                afwDisplay.Display(frame=frame).mtv(bkgdImage, title=self._testMethodName + " bkgdImage: " +
-                                                    interpStyle.__str__())
+                afwDisplay.Display(frame=frame).mtv(bkgdImage, title=self._testMethodName + " bkgdImage: "
+                                                    + interpStyle.__str__())
```
timj: consider an f-string here to allow the both of the `+` to go away.

tests/test_footprint2.py
```diff
@@ -477,8 +477,10 @@ def checkPeaks(self, dwidth=0, dheight=0, frame=3):
             # is treated as a Peak
             #
             if (dwidth != 0 or dheight != 0):
-                if (foot.getBBox().getMinX() == 0 or foot.getBBox().getMaxX() == self.im.getWidth() - 1 or
-                        foot.getBBox().getMinY() == 0 or foot.getBBox().getMaxY() == self.im.getHeight() - 1):
+                if (foot.getBBox().getMinX() == 0
+                        or foot.getBBox().getMaxX() == self.im.getWidth() - 1
+                        or foot.getBBox().getMinY() == 0
+                        or foot.getBBox().getMaxY() == self.im.getHeight() - 1):
```
timj: This reorganization does look easier to read.

tests/test_kernel.py
```diff
@@ -870,12 +869,10 @@ def basicTestComputeImageRaise(self, kernel, doRaise, kernelDescr=""):
         try:
             kernel.computeImage(kImage, True)
             if doRaise:
-                self.fail(kernelDescr +
-                          ".computeImage should have raised an exception")
+                self.fail(kernelDescr + ".computeImage should have raised an exception")
```
timj: f-string?

tests/test_kernel.py
```diff
@@ -870,12 +869,10 @@ def basicTestComputeImageRaise(self, kernel, doRaise, kernelDescr=""):
...
         except pexExcept.Exception:
             if not doRaise:
-                self.fail(kernelDescr +
-                          ".computeImage should not have raised an exception")
+                self.fail(kernelDescr + ".computeImage should not have raised an exception")
```
timj: f-string?

tests/test_offsetImage.py
```diff
@@ -146,8 +146,8 @@ def testOffsetGaussian(self):
                     im -= refIm
 
                     if display:
-                        afwDisplay.Display(frame=1).mtv(im, title=algorithm +
-                                                        ": diff image (dx, dy) = (%f, %f)" % (dx, dy))
+                        afwDisplay.Display(frame=1).mtv(im, title=algorithm
+                                                        + ": diff image (dx, dy) = (%f, %f)" % (dx, dy))
```
timj: f-string?

r-owen:
I will add f-strings where you suggest. I initially did not because it makes the code different than surrounding code and there are over 400 instances of % formatted strings in afw and I didn't want to change them all.

tests/test_background.py
r-owen: Note that an f-string does make the line very long (in fact I have to move title=f"..." to the next line). But it squeaks by.

timj: You can split the f-string over two lines like any other explicit quoted string.  You don't need the `+` to do it though.

r-owen:
I would like to make the f string changes a separate commit in order to avoid mixing f strings and % formatting in the files.

PR #511, opened 2020-01-24
DM-22708: Fix decam gen3 ingest

parejkoj:
This pybind11 addition makes parsing multi-HDU DECam files much faster.

### DIFF
--- a/python/lsst/afw/fits/fits.cc
+++ b/python/lsst/afw/fits/fits.cc
@@ -177,6 +177,14 @@ void declareFits(py::module & mod) {
             "strip"_a=false);
     cls.def("createEmpty", &Fits::createEmpty);
 
+    cls.def("readImageI", [](Fits &self) {
+        ndarray::Vector<int, 2> const offset;  // initialized to zero by default
+        ndarray::Vector<ndarray::Size, 2> shape = self.getImageShape<2>();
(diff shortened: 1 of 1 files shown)


python/lsst/afw/fits/fits.cc
```diff
@@ -177,7 +177,15 @@ void declareFits(py::module & mod) {
...
-    cls.def("gotoFirstHdu", [](Fits & self) { self.setHdu(DEFAULT_HDU); });
+    cls.def("readImageI", [](Fits &self) {
+        ndarray::Vector<int, 2> const offset;  // initialized to zero by default
+        ndarray::Vector<ndarray::Size, 2> shape = self.getImageShape<2>();
+        ndarray::Array<int, 2, 2> result = ndarray::allocate(shape[0], shape[1]);
+        self.readImage(result, offset);
+        return result;
+    });
+
+    cls.def("gotoFirstHdu", [](Fits &self) { self.setHdu(DEFAULT_HDU); });
```
timj: Why did you change this line to remove the space? It makes it inconsistent now with other definitions earlier. If you want to change `Fits & self` to `Fits &self` can you please make it a separate commit and also change the other instances (eg readMetadata)?

parejkoj: That's just what clang-format did. Not everybody is running clang-format on save, and I don't always catch the changes it makes.

PR #512, opened 2020-01-24
DM-23178: Update code to use f strings

r-owen:


r-owen:
Thank you. I already did that check. I found only one active ticket branch with a minor change that would not be affected at all.

PR #513, opened 2020-01-28
DM-20841: Support for approximating WCS as SIP

PaulPrice:


python/lsst/afw/geom/calculateSipWcsHeader.py
```diff
@@ -0,0 +1,80 @@
...
+    header.set("CRVAL1", crval.getRa().asDegrees())
+    header.set("CRVAL2", crval.getDec().asDegrees())
+    header.set("CRPIX1", sip.getPixelOrigin().getX() + 1)  # +1 for LSST --> FITS
+    header.set("CRPIX2", sip.getPixelOrigin().getY() + 1)  # +1 for LSST --> FITS
+    for xx, yy in itertools.product(range(2), range(2)):
+        header.set(f"CD{xx + 1}_{yy + 1}", sip.getCdMatrix()[xx][yy])
+    coefficientsToHeader(sip.getOrder(), sip.getA(), header, "A")
+    coefficientsToHeader(sip.getOrder(), sip.getB(), header, "B")
+    coefficientsToHeader(sip.getOrder(), sip.getAP(), header, "AP")
+    coefficientsToHeader(sip.getOrder(), sip.getBP(), header, "BP")
```
TallJimbo: instead of modifying the header directly, did you consider using `makeTanSipWcs` (defined in `afw/geom/SkyWcs.h`) to make a TAN-SIP SkyWcs, and then using `getFitsMetadata()` on that?  I could imagine that being a fair bit slower, as it'll initialize a bunch of AST stuff, but it might be better at handling FITS edge cases.  There's nothing obviously wrong with your code here, but I'm a bit worried about non-obvious things.

tests/test_sipApproximation.py
```diff
@@ -425,6 +426,580 @@ def setUp(self):
...
+                   Begin UnitMap
+                      Nin = 2
+                      IsSimp = 1
+                   IsA Mapping
+                   End UnitMap
+             End CmpMap
+       End CmpMap
+ End FrameSet
+"""
+
```
TallJimbo: That's a lot of string data for file that contains code.  Mind dropping it into a separate text file?

tests/test_sipApproximation.py
```diff
@@ -516,6 +1091,77 @@ def run(md, **kwds2):
...
+            self.assertFloatsAlmostEqual(offsets[0].mean(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(offsets[1].mean(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(offsets.mean(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(offsets[0].std(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(offsets[1].std(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(offsets.std(), 0.0, atol=skyTol)
+            self.assertFloatsAlmostEqual(np.abs(offsets).max(), 0.0, atol=skyTol)
+
+            # Not comparing to absolute round-trip, but relative to the original WCS round-trip.
+            # This is an important distinction: we're recreating the original WCS, along with all its flaws.
```
TallJimbo: Apparently we're not checking this in Travis in afw, but I think it's best to break comments at 79 chars in new code.

python/lsst/afw/geom/calculateSipWcsHeader.py
PaulPrice: Oh. I didn't realise that was available...
But it doesn't work out-of-the-box. I'm investigating...

PaulPrice: After another fix to `SipApproximation`, it's working.

tests/test_sipApproximation.py
PaulPrice: That's a bit annoying (the comment block starts at 15 chars in, which means the effective width is only 64 chars), but OK.

src/geom/SipApproximation.cc
```diff
@@ -264,24 +264,24 @@ double SipApproximation::getA(int p, int q) const {
...
 }
 
 double SipApproximation::getAP(int p, int q) const {
-    return _solution->ap[_solution->a.getBasis().index(p, q)];
+    return _solution->ap[_solution->ap.getBasis().index(p, q)];
 }
 
 double SipApproximation::getBP(int p, int q) const {
-    return _solution->bp[_solution->a.getBasis().index(p, q)];
+    return _solution->bp[_solution->bp.getBasis().index(p, q)];
```
TallJimbo: Yikes!  Glad you caught these.  I wonder what I was doing in my tests that made it going unnoticed.  Maybe just unlucky float values that were all close to zero?

PR #514, opened 2020-02-05
DM-21697 Fix SpanSet Creation from large circles

natelust:
Fix an issue where data-types were not large enough to hold values
that may be plausibly used.

### DIFF
--- a/src/geom/SpanSet.cc
+++ b/src/geom/SpanSet.cc
@@ -692,9 +692,13 @@ std::shared_ptr<SpanSet> SpanSet::fromShape(int r, Stencil s, lsst::geom::Point2
     tempVec.reserve(2 * r + 1);
     switch (s) {
         case Stencil::CIRCLE:
-            for (auto dy = -r; dy <= r; ++dy) {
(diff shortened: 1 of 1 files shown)


src/geom/SpanSet.cc
```diff
@@ -692,9 +692,13 @@ std::shared_ptr<SpanSet> SpanSet::fromShape(int r, Stencil s, lsst::geom::Point2
     tempVec.reserve(2 * r + 1);
     switch (s) {
         case Stencil::CIRCLE:
-            for (auto dy = -r; dy <= r; ++dy) {
-                int dx = static_cast<int>(sqrt(r * r - dy * dy));
-                tempVec.push_back(Span(dy + offset.getY(), -dx + offset.getX(), dx + offset.getX()));
+            {
+                double dr = static_cast<double>(r);
```
gpdf: I think a comment is necessary to clarify what problem is being avoided by the parallel increment of the integer and the double.  And I might name the variable that holds "r-squared" something like `r_sq`.

Issue #516, opened 2020-02-20
Raise descriptive error for broken display install


bsipocz:
Consider the following code and the scenario of the `'matplotlib'` backend and a broken install where `lsst.display` is not available.
https://github.com/lsst/afw/blob/master/python/lsst/afw/display/interface.py#L85-L106

It raises the somewhat cryptic error message:

``RuntimeError: Unable to set backend to matplotlib: "module 'matplotlib' has no attribute 'DisplayImpl'"``

While the real underlying issue is that lsst.display is not available, matplotlib should never be expected to have a `DisplayImpl` attribute. 

My naive approach of shuffling the order of elements in the tuple probably makes no sense, adding another try/except might be enough to raise a more relevant error.

Let me know if you would like a quick PR. (The install has since been fixed, so I won't have a good way to test it).

jdswinbank:
Thank you! We're now tracking this internally as [DM-23592](https://jira.lsstcorp.org/browse/DM-23592). Fix coming up...

PR #517, opened 2020-02-20
DM-23592: Raise descriptive error for broken display install 

jdswinbank:
Addressing #516.


timj (approved):
Looks like a reasonable clean up.

PR #528, opened 2020-04-29
DM-24565: Un-deprecate old APIs related to afw.math.Background

kfindeisen:
The PR reverts the portions of #505 that relate to `Background`, then removes all mention of deprecation from the corresponding documentation.

### DIFF
--- a/include/lsst/afw/math/Background.h
+++ b/include/lsst/afw/math/Background.h
@@ -104,9 +104,8 @@ public:
                                   nySample));
         }
     }
-    // And now the two old APIs (preserved for backward compatibility)
     /**
-     * @deprecated New code should specify the interpolation style in getImage, not the BackgroundControl ctor
+     * Overload constructor to provide interp style.
      *
      * @param style Style of the interpolation
      * @param nxSample Num. grid samples in x
@@ -138,8 +137,6 @@ public:
 
     /**
      * Overload constructor to handle strings for both interp and undersample styles.
-     *
-     * @deprecated New code should specify the interpolation style in getImage, not the BackgroundControl ctor
      *
      * @param style Style of the interpolation
      * @param nxSample num. grid samples in x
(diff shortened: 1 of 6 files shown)


include/lsst/afw/math/Background.h
```diff
@@ -506,7 +492,7 @@ class BackgroundMI : public Background {
      * @param y y-pixel coordinate (row)
      * @returns an estimated background at x,y (double)
      *
-     * @deprecated Don't call this image (not even in test code).
+     * @warning Don't call this image (not even in test code).
```
timj, line 495: I do worry about this message. "not even in test code" is fairly extreme. Either people should be allowed to use it or we really should deprecate this one.

kfindeisen: Yes, I was wondering about it myself. But IIRC this is also a case where migration is tricky, because `getPixel(x, y)` and `getImage()(x, y)` aren't exactly equivalent. I don't remember why, sorry.

kfindeisen: Hmm... looking at 488744f7d9a89d799963403493304a4b7e211849, it seems my main concern was uncertainty about which coordinate system was being used. A straight substitution passed unit tests, at least.

@timj, given that it goes against what was agreed in RFC-687, is there anything extra I would need to do if I selectively reverted the revert of `getPixel`?

timj: I'm not entirely sure what to do about this. I think we all agree that the warning is nonsensical as it stands. Maybe @jdswinbank or @ktlim  can comment here.

jdswinbank: Assuming I've understood the issue here (that leaving this function as deprecated would formally violate the terms of RFC-687, but that the function is useless/dangerous and ought to be removed), then I suggest that pragmatism should win over process in this case: leave it deprecated. I think that's an entirely defensible position (and one which I will defend, if necessary).

kfindeisen: Ok, I've reinstated all changes related to `getPixel` only. I'm leaving them as separate commits just in case.

Assuming Jenkins passes, I expect to merge tomorrow morning.

PR #530, opened 2020-05-19
Drop WARN to DEBUG on uppercasing keywords on read

brianv0:
When reading a fits keyword, we log only debug messages due to
proliferation of old files.

### DIFF
--- a/src/fits.cc
+++ b/src/fits.cc
@@ -73,8 +73,8 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
                 out += (boost::format("%#20.17G") % value).str();
             } else {
                 LOGLS_WARN("afw.fits",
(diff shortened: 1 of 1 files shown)


src/fits.cc
```diff
@@ -795,7 +795,7 @@ void Fits::forEachKey(HeaderIterationFunctor &functor) {
         std::string upperKey(key);
         boost::to_upper(upperKey);
         if (upperKey.compare(key) != 0){
-            LOGLS_WARN("afw.fits",
+            LOGLS_DEBUG("afw.fits",
                        boost::format("In %s, standardizing key '%s' to uppercase '%s' on read.") %
                        BOOST_CURRENT_FUNCTION % key % upperKey);
```
erykoff, line 800: Is the indentation correct here?

brianv0: ah  - I probably need to fix that, I'll check (I'm not sure)

PR #531, opened 2020-05-23
DM-6981: Implement assignment to Flag columns in the pybind11 layer.

TallJimbo:
This only works when the key is a `str` name or a `Key` instance; because support for slice assignment for all other fields currently delegates so much to NumPy, it's a lot harder to make that work for Flags.

### DIFF
--- a/tests/test_simpleTable.py
+++ b/tests/test_simpleTable.py
@@ -799,6 +799,26 @@ class SimpleTableTestCase(lsst.utils.tests.TestCase):
         with self.assertRaises(IndexError):
             del catalog[50]
 
+    def testSetFlagColumn(self):
+        schema = lsst.afw.table.Schema()
+        key = schema.addField("a", type="Flag", doc="doc for 'a'")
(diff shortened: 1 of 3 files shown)


tests/test_simpleTable.py
```diff
@@ -799,6 +799,26 @@ def testDelete(self):
         with self.assertRaises(IndexError):
             del catalog[50]
 
+    def testSetFlagColumn(self):
+        schema = lsst.afw.table.Schema()
+        key = schema.addField("a", type="Flag", doc="doc for 'a'")
```
erykoff, line 804: I wonder if we should add another layer to this test, with a `b` flag as well as an `a` flag, to ensure that the bit-packing behind the scenes is doing what we think it should be doing.  (However, I did confirm that this does work with multiple flags).

TallJimbo: The bit-packing is implemented in a level below this that has its own tests, so I think one is sufficient.

PR #536, opened 2020-08-05
DM-26260: ExposureFitsReader should not read the mask/variance if they are missing

timj:
If a compressed single image FITS file is read an attempt is made to read the mask and variance because additional extensions can be present.   This will fail because they aren't really masks but on failing an assumption is made that the file
was registered with the reader. This is not set because nextHdu can return a null pointer.  Therefore add protections such that we do not attempt to read the mask/variance if the fits file attribute is null.


src/image/MaskedImageFitsReader.cc
```diff
@@ -202,32 +202,37 @@ MaskedImage<ImagePixelT, MaskPixelT, VariancePixelT> MaskedImageFitsReader::read
...
+          try {
+              mask = std::make_shared<Mask<MaskPixelT>>(
+                  _maskReader.read<MaskPixelT>(bbox, origin, conformMasks, allowUnsafe)
+              );
+          } catch (fits::FitsError& e) {
+              if (needAllHdus) {
+                  LSST_EXCEPT_ADD(e, "Reading Mask");
+                  throw e;
+              }
+              LOGL_WARN(_log, "Mask unreadable (%s); using default", e.what());
```
TallJimbo, line 216: Looking at this code in isolation, it seems like it'd be appropriate to emit this warning when `_maskReader._fitsFile` is null, too.  But in the higher-level context of the ticket, it seems we load `Exposure` or `MaskedImage` objects from single-plane FITS files all the the time, and don't want warnings when we do that, so maybe the right fix is to remove the warning from both code paths?

In any case, we should probably obey the `needAllHdus` logic when `_maskFitsReader._fitsFile` is null, and throw an exception.  I don't think that's used very often (if at all), but it'd certainly be confusing to pass `needAllHdus=True` and then silently only get the image plane loaded.

PR #537, opened 2020-08-19
DM-25253: Warn when weights are ignored in statisticsStack

MorganSchmitz:
Jenkins run [#32492](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/32492/pipeline).

[Jira ticket](https://jira.lsstcorp.org/browse/DM-25253).

Paircoded with @arunkannawadi!

### DIFF
--- a/src/math/Stack.cc
+++ b/src/math/Stack.cc
@@ -34,8 +34,13 @@
 #include "lsst/pex/exceptions.h"
 #include "lsst/afw/math/Stack.h"
 #include "lsst/afw/math/MaskedVector.h"
+#include "lsst/log/Log.h"
 
 namespace pexExcept = lsst::pex::exceptions;
+
+namespace {
+    LOG_LOGGER _log = LOG_GET("afw.math.Stack");
+}
 
 namespace lsst {
 namespace afw {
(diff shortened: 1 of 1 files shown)


src/math/Stack.cc
```diff
@@ -260,6 +265,11 @@ void statisticsStack(image::MaskedImage<PixelT> &out,
                                                                 wvector);  // use wvector
         }
     } else {
+        if (!wvector.empty()) {
+            LOGLS_WARN(_log,
+                       "Weights passed on to statisticsStack are ignored as sctrl.getWeighted() is False."
+                       "Set sctrl.setWeighted(True) for them to be used.");
+        }
```
TallJimbo, line 272: You can just use `LOG_WARN` here.  The `LOGL_` and `LOGLS_` are just different ways of doing string formatting on the message, but you're not doing any here.

arunkannawadi: The use of LOG_WARN appears pretty minimal, with no occurrence except in its own unit test. Furthermore, I'm not sure if we want this to go into a default logger if there is one since it appears multiple times. So, we use LOGL_WARN and log it into the appropriate log file instead.

TallJimbo: Sorry, I should have suggested `LOGL_WARN`, not `LOG_WARN`; I hadn't realized that the "L" suffix controlled whether a log object was accepted and only the "S" suffix controlled the string formatting.  But what you've merged is fine.

PR #541, opened 2020-10-08
DM-22114: Remove deprecated `makeDistortedTanWcs`

arunkannawadi:
This PR removes the function `makeDistortedTanWcs` and merges its unit tests with those of `computePixelToDistortedPixel`.

### DIFF
--- a/tests/test_wcsUtils.py
+++ b/tests/test_wcsUtils.py
@@ -33,7 +33,7 @@ import lsst.utils.tests
 from lsst.afw.geom.wcsUtils import createTrivialWcsMetadata, deleteBasicWcsMetadata, \
     getCdMatrixFromMetadata, getSipMatrixFromMetadata, getImageXY0FromMetadata, \
     hasSipMatrix, makeSipMatrixMetadata, makeTanSipMetadata, \
-    computePixelToDistortedPixel, makeDistortedTanWcs
+    computePixelToDistortedPixel
 
 
 def makeRotationMatrix(angle, scale):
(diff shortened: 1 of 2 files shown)


parejkoj (commented):
I'm confused: why didn't you keep the existing `ComputePixelToDistortedPixelTestCase` and remove the `MakeDistortedTanWcsTestCase` class instead? It's hard to tell with this diff whether the new class is exactly the same as the deleted one or not.

arunkannawadi:
The messy diff happens to be due to git trying to behave too smart, it looks like. I unstaged the previous changes and removed the `MakeDistortedTanWcsTestCase` completely and edited the `ComputePixelToDistortedPixelTestCase`, but because the line numbers coincide, git really thinks I edited former one and deleted the latter one. In any case, the git diffs are messy either ways the two test cases are merged.

arunkannawadi:
Okay, I've split the changes into two commits that make more sense logically. The PR diff is still messy as before, but the commit diffs might be helpful to see the differences.

tests/test_wcsUtils.py
```diff
@@ -91,8 +91,8 @@ class ComputePixelToDistortedPixelTestCase(BaseTestCase):
     def testNoDistortion(self):
         """Test computePixelToDistortedPixel without distortion
 
-        Use an affine transform for pixelToFocalPlane; the transform
-        returned by computePixelToDistortedPixel should be the identity transform
+        Use an affine transform for pixelToFocalPlane; the transform returned
```
parejkoj, line 94: What happened here?

parejkoj (approved):
One minor comment about a docstring, but otherwise this looks good. Thank you for the detailed comments in the new tests regarding how you generate and apply the transforms, they are very helpful.

tests/test_wcsUtils.py
arunkannawadi: I miscalculated the number of characters and thought that line 95 was exceeding 79 characters. I'll drop this change and merge it to master.

arunkannawadi: No, I was right the first time. Line 95 as it was exceeds 79 characters, and now it doesn't.

PR #544, opened 2020-10-31
DM-22814: Remove afw APIs deprecated in DM-17566

parejkoj:


include/lsst/afw/table/io/python.h
```diff
@@ -41,45 +41,10 @@ namespace table {
...
-                   (std::shared_ptr<T>(*)(fits::MemFileManager &, int)) & PersistableFacade<T>::readFits,
-                   "manager"_a, "hdu"_a = fits::DEFAULT_HDU);
-}
-
 /**
  * Add table::io::Persistable and PersistableFacade methods to the pybind11 wrapper for a class
  *
- * Use this instead of declarePersistableFacade to avoid circular import issues in Python;
- * it allows your class to be used without importing lsst.afw.table.
+ * This allows your class to be used without importing lsst.afw.table.
```
kfindeisen, line 47: Without the previous statement this sentence isn't very clear. Maybe append "in Python", to distinguish from imports needed inside pybind11 code?

tests/test_gaussianPsf.py
```diff
@@ -37,6 +37,7 @@
 import lsst.geom
 import lsst.afw.table
 import lsst.afw.fits
+import lsst.afw.image
```
kfindeisen, line 40: Could this be moved to its own commit?

include/lsst/afw/table/Source.h
```diff
@@ -474,10 +474,10 @@ class SourceColumnViewT : public ColumnViewT<RecordT> {
     }
 
     ndarray::Array<double, 1> const getX() const {
-        return this->operator[](this->getTable()->getCentroidKey().getX());
+        return this->operator[](this->getTable()->getCentroidSlot().getMeasKey().getX());
```
kfindeisen, line 477: I'm not sure I understand why these changes are on their own commit, given that you remove other self-uses on the "Remove Deprecated Functions" commit. Could you rearrange the commits so that these come first? That way the code works at each step.

kfindeisen (approved):
Looks good, minor comments only.

tests/test_gaussianPsf.py
parejkoj: This was necessary to make the test run: somehow that import was unnecessary before (likely circular afw-internal dependencies), but is necessary now.

include/lsst/afw/table/Source.h
parejkoj: Oh, sorry: I'd meant to squash these with the other commit, to make it all one block. I just forgot (I made incremental commits while I was fixing things).

src/math/offsetImage.cc
```diff
@@ -102,7 +102,10 @@ std::shared_ptr<ImageT> offsetImage(ImageT const& inImage, float dx, float dy,
 
     offsetKernel->setKernelParameters(std::make_pair(dKerX, dKerY));
 
-    convolve(*convImage, *buffImage, *offsetKernel, true, true);
+    ConvolutionControl convolutionControl;
+    convolutionControl.setDoNormalize(true);
+    convolutionControl.setDoCopyEdge(true);
+    convolve(*convImage, *buffImage, *offsetKernel, convolutionControl);
```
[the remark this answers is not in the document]
parejkoj: Dunno. `tests/sourceMatch1.cc` also had to be fixed, as did some inlined things in `Source.h`

PR #547, opened 2020-11-03
DM-27359 Make python-derived Storable classes persistable

jmeyers314:


tests/test_psf_trampoline.py
```diff
@@ -158,6 +197,17 @@ def testClone(self):
...
+    def testPersistence(self):
+        for pgp in self.pgps:
+            assert cppLib.isPersistable(pgp)
+            im = ExposureF(10, 10)
+            im.setPsf(pgp)
+            assert im.getPsf() == pgp
+            with lsst.utils.tests.getTempFilePath(".fits") as tmpFile:
+                im.writeFits(tmpFile)
+                newIm = ExposureF(tmpFile)
+                assert newIm.getPsf() == im.getPsf()
```
TallJimbo, line 210: We prefer to use the `unittest` module's assert methods (e.g. `self.assertEqual`) in our test code (at least for consistency; I also think it provides better printing of failures in at least some contexts).

include/lsst/afw/typehandling/python.h
```diff
@@ -84,10 +88,121 @@ class StorableHelper : public Base {
...
 std::string declareGenericMapRestrictions(std::string const& className, std::string const& keyName);
 
+/**
+ * StorableHelper persistence.
+ *
+ * We cannot directly override `StorableHelper::write` or `StorableHelperFactory::read` in python using
+ * PYBIND11_OVERLOAD* macros as the required argument c++ types (OutputArchiveHandle, InputArchive,
+ * CatalogVector) are not bound to python types.  Instead, we allow python subclasses of Storable to define
+ * `_write` and `_read` methods that return/accept a string serialization of the object.  The `_write` method
+ * should take no arguments (besides self) and return a string.  The `_read` method should be a static method
```
TallJimbo, line 122: I'd say "bytes" or "byte string" instead of "string" so nobody thinks we're talking about text-based persistence.

include/lsst/afw/typehandling/python.h
```diff
@@ -84,10 +88,121 @@ class StorableHelper : public Base {
     bool equals(Storable const& other) const noexcept override {
         PYBIND11_OVERLOAD_NAME(bool, Base, "__eq__", equals, other);
     }
+
```
TallJimbo, line 91: Preexisting, but up at line 71, I think we should mark that constructor as explicit.  Same for the similar constructor on `PsfTrampoline`.

include/lsst/afw/typehandling/python.h
```diff
@@ -84,10 +88,121 @@ class StorableHelper : public Base {
...
+        table::io::PersistableFactory(name),
+        _module(module),
+        _name(name)
+    {}
+
+    std::shared_ptr<table::io::Persistable> read(
+        InputArchive const &archive,
+        CatalogVector const &catalogs
+    ) const override {
+        // Do I need to acquire the GIL here?
```
TallJimbo, line 172: Yes, you do.  Glad you thought about this, because I wouldn't have if you hadn't brought it up.

TallJimbo (approved):
Just a few minor comments; looks good!

PR #549, opened 2020-11-10
DM-27168: Phase out use of FilterProperty

czwa:
Stop usage of FilterProperty, as stated in RFC-730.

Allow filter bounds to be defined from a tranmsission curve.

### DIFF
--- a/tests/test_exposure.py
+++ b/tests/test_exposure.py
@@ -33,7 +33,6 @@ import lsst.utils
 import lsst.utils.tests
 import lsst.geom
 import lsst.afw.image as afwImage
-from lsst.afw.image.utils import defineFilter
 from lsst.afw.coord import Weather
 import lsst.afw.geom as afwGeom
 import lsst.afw.table as afwTable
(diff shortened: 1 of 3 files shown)


tests/test_exposure.py
```diff
@@ -88,11 +87,6 @@ def setUp(self):
         # test with ExtentI(100, 100) too
         self.exposureCrOnly = afwImage.ExposureF(lsst.geom.ExtentI(100, 100))
 
-        afwImage.Filter.reset()
-        afwImage.FilterProperty.reset()
-
-        defineFilter("g", 470.0)
-
```
kfindeisen, line 95: I'm not sure I understand what's going on here. The test case has assertions that depend on the `g` filter (and I do need those assertions to check backwards-compatibility, so please don't remove them!). However, you've removed the definition of that filter. How does the test still work?

kfindeisen (changes requested):
I think you may have been a bit overzealous in counting the unit tests of `FilterProperty` as a use of the class. That's not a usage we would be removing during the switch to `FilterLabel`.

The use of `FilterProperty` in the exposure tests I'm less sure about. We certainly need to keep using `Filter` there, but I'm surprised the test works with the changes you've made. Did you have any trouble getting those changes to build?

czwa:
Withdrawing this PR, as the review has made the changes unnecessary.

PR #550, opened 2020-11-25
DM-16857: Enable Python API docs and change to numpydoc

kherner:


python/lsst/afw/detection/utils.py
```diff
@@ -23,7 +23,22 @@
 
 
 def writeFootprintAsDefects(fd, foot):
```
kfindeisen, line 25: This function is not included in the docs at the moment, because [`lsst.afw.detection` does not include `lsst.afw.detection.utils`](https://github.com/lsst/afw/blob/master/python/lsst/afw/detection/__init__.py). If you think it's worth it, you could add an `.. automodapi:: lsst.afw.detection.utils` entry for it; see `afw.table` for an example.

doc/detection.dox
```diff
@@ -0,0 +1,16 @@
...
+/**
+\section afwDetection Detection Overview
+
+Detected pixels are manipulated using the following classes:
+ - Footprint : A set of pixels above (or below) a Threshold
+ - FootprintFunctor : A functor to process the pixels in a Footprint
+ - FootprintSet : A set of Footprints associated with a MaskedImage
+ - Peak : A peak pixel in the %image
+ - Source : The properties of a source of electrons (e.g. a star or cosmic ray)
+ - Threshold : An object describing the threshold used to define a set of Footprint%s
```
kfindeisen, line 13: This is still Doxygen, not Sphinx. You could create a file similar to https://github.com/lsst/afw/blob/master/doc/lsst.afw.cameraGeom/cameraGeom.rst... though now that I read the text, I think this is all redundant with the listing that Sphinx produces automatically (class descriptions don't show up yet for C++ types, but that's being worked on). So I, at least, would be fine with deleting it entirely.

kfindeisen (changes requested):
Thanks for catching the formatting problems in the previously-converted docstrings. However, `afw.detection` still has package docs in Doxygen; I can't really approve the PR until they are either moved to Sphinx or deleted (apologies for not thinking of the latter possibility sooner). There are also some minor formatting issues with the new docstrings.

python/lsst/afw/detection/utils.py
```diff
@@ -23,7 +23,22 @@
...
+    Write foot as a set of Defects to fd
+
+    Inputs
+    ------
+    fd: file object
+    foot: a footprint
+
+    Returns
+    -------
+    formatted string written to the file object specified in the argument
```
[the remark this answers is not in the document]
kherner: Description modified accordingly

python/lsst/afw/detection/utils.py
```diff
@@ -23,7 +23,22 @@
...
+
+    See Also
+    --------
+    `lsst.afw.detection.footprintToBBoxList`
```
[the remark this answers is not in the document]
kherner: Will be in next commit

doc/detection.dox
kherner: Right, this confused me somewhat because I was taking lsst.afw.table as my model, and it still has a .dox file that looks more or less like the one in question (though it also has an .rst file). I'm also fine with deleting this; I can just add any bits from this file that aren't already in the existing .rst file to the latter.

python/lsst/afw/detection/_footprintMerge.py
```diff
@@ -29,7 +29,33 @@
...
+        Parameters
+        ----------
+        catalogs : `list`
+            list of catalogs
+        filters : `list`
+            list of filters
+        peakDist : `float` or list of floats
+            distance between source peaks
+        schema : :py:class:`lsst.afw.table.Schema`
+        idFactory : :py:class:`lsst.afw.table.IdFactory`
```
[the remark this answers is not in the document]
kherner: Will remove the :py:class: additions

doc/manifest.yaml
```diff
@@ -1,4 +1,5 @@
 # Documentation manifest.
+package: "afw"
```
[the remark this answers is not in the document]
kherner: I had something fail to build earlier without it, and then after adding it, it worked. I think the original failure was just due to not running `package-docs clean` before a build, and sure enough just now I took this line out and everything is still fine. I'll remove it.

doc/mainpage.dox
```diff
@@ -281,11 +281,9 @@ but \b not possible to read it back from a non-default HDU.
 */
 }}}
 
-//-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
-
 namespace lsst { namespace afw { namespace detection {
 /**
-\page afwSecDetection Detection
+\page afwDetection Detection
```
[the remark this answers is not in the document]
kherner: Really just to match most of the other existing sections, no other reason. If we're getting rid of the other doxygen stuff (see below) then I think we can just remove the line entirely, no?

python/lsst/afw/detection/utils.py
kherner: For this initial round of review I left it out because I had tried adding that before, but the build broke because it complains about a missing doctree file.

Full details:

> FileNotFoundError: [Errno 2] No such file or directory: '/project/kherner/afw/doc/_build/doctree/py-api/lsst.afw.detection.utils.footprintToBBoxList.doctree'

What's interesting is that lsst.afw.detection.footprintToBBoxList.doctree does get created, but not a version with the .utils added. The relevant line in the code (python/lsst/afw/detection/utils.py) seems to be 

`from . import footprintToBBoxList`

So I suspect I need to tweak that somehow to get this all to play nicely together.

kherner: Figured out a workaround: adding `:skip: footprintToBBoxList` allows everything to build and gets the missing function included.

kfindeisen: Ah, that's because the file doesn't have an `__all__` definition. I suggest adding it even if you have a workaround, since it's good practice (though it doesn't seem to be in the style guide 😕).

kherner: Good catch, thanks! Adding `__all__ ["writeFootprintAsDefects"]` fixed it and the workaround is no longer necessary.

python/lsst/afw/detection/_footprintMerge.py
```diff
@@ -29,7 +29,33 @@
...
+
+        Returns
+        -------
+        mergedList : :py:class:`lsst.afw.table.SourceCatalog`
+            merged list of sources
+
+        See also
+        --------
+        footprintMerge: the other class attributes and members are defined in the C code in this package.
+        afwTable : :py:class:`lsst.afw.table`
```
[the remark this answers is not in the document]
kherner: Thinking about it some more, I agree the footprintMerge line didn't really need to be there. I changed the second one to point to lsst.afw.table.SourceCatalog, which is the intermediate product in this case.

kherner:
I've incorporated and pushed all suggested changes now; please have another look. I'll kick off a CI check in the meantime, too.

doc/mainpage.dox
```diff
@@ -8,7 +8,6 @@ by an image processing system.  In particular:
    - \ref afwSecImage : Representation of Images (and Masks)
    - \ref afwSecPyImage : How to manipulate images from python
    - \ref afwSecDisplay : How to display images
-   - \ref afwSecDetection : Detection of sources in images
```
kfindeisen, line 11: I assume you built the Doxygen docs (`scons doc`) and made sure there were no other links to the deleted section?

python/lsst/afw/detection/utils.py
```diff
@@ -19,11 +19,30 @@
...
+    Does not return anything.
+
+    Parameters
+    ----------
+    fd : `typing.TextIO`
+    foot : `lsst.afw.detection.Footprint`
+
+    Returns
+    -------
+
```
kfindeisen, line 41: Normally when we don't have a return value, we leave out the section entirely (I suspect it doesn't format correctly in the built docs, though I can't check). There's also no need to explicitly say "does not return anything."

kfindeisen (approved):
Looks good, just minor comments. Thanks for your patience with this!

doc/mainpage.dox
kherner: I did indeed; it successfully built and grep turned up no other links.

python/lsst/afw/detection/utils.py
kherner: Will adjust accordingly.

PR #551, opened 2020-11-26
DM-27169: Use FilterLabel in Exposure/ExposureInfo

kfindeisen:
This PR replaces both the internal state and the persistence format of `Exposure` to use `FilterLabel` instead of `Filter`. Accessors for `Filter` remain for backwards-compatibility, but are re-implemented in terms of `FilterLabel`. I've also done some refactoring of the generic component code that should make it easier to apply in the future.


src/image/ExposureFitsReader.cc
```diff
@@ -48,8 +52,155 @@ bool _contains(std::array<T, N> const& array, T const& value) {
...
+ * @param name The name to test.
+ */
+bool _isBand(std::string const& name) {
+    static std::set<std::string> const BANDS = {"u", "g", "r", "i", "z", "y", "SH", "PH", "VR", "white"};
+    // Standard band
+    if (BANDS.count(name) > 0) {
+        return true;
+    }
+    // Looks like a narrow-band band
+    if (std::regex_match(name, std::regex("N\\d+"))) {
```
parejkoj, line 74: Subaru also has `I945`, which I'm guessing might be "intermediate-band" (which I'd normally call "medium band"). Not sure if that's worth a check here too?

src/image/ExposureFitsReader.cc
```diff
@@ -48,8 +52,155 @@ bool _contains(std::array<T, N> const& array, T const& value) {
...
+/**
+ * Convert an old-style Filter to a FilterLabel.
+ *
+ * @param filter Any Filter object.
+ * @returns The closest equivalent FilterLabel, given available information.
+ */
+// TODO: compatibility code to be removed in DM-27177
+std::shared_ptr<FilterLabel> makeFilterLabel(Filter const& filter) {
+    // Filter has no self-consistency guarantees whatsoever, and most methods
+    // are unsafe. Program extremely defensively.
```
parejkoj, line 109: Heh... that's a useful, though scary comment.

src/image/ExposureFitsReader.cc
```diff
@@ -48,8 +52,155 @@ bool _contains(std::array<T, N> const& array, T const& value) {
...
+ */
+std::shared_ptr<FilterLabel> makeFilterLabel(std::string const& name) {
+    if (_AFW_NAMES.count(name) > 0) {
+        return std::make_shared<FilterLabel>(_AFW_NAMES.at(name));
+    }
+    // else name is either a band, a physical filter, or a deprecated alias
+
+    try {
+        // To ease the transition to FilterLabel, use Filter to get all the names
+        // TODO: after DM-27177, leave only the catch block
```
parejkoj, line 160: Maybe worth noting here (or in `makeFilterLabel(Filter)` above?) that what makes this work is the pre-registration of filters by `FilterDefinition`, so that there is extra information available from the singleton. It took me a bit to sort out how you were getting that extra information, and it would be very non-obvious to someone who isn't familiar with `Filter`.

src/image/ExposureFitsReader.cc
```diff
@@ -92,8 +243,17 @@ class ExposureFitsReader::MetadataReader {
...
         }
 
-        filter = Filter(metadata, true);
-        detail::stripFilterKeywords(metadata);
+        // Earlier versions persisted Filter as header keyword, version 2 persists FilterLabel as a Storable
+        if (version < 2) {
+            std::string key = "FILTER";
+            if (metadata->exists(key)) {
+                // Original Filter code depended on Boost for string trimming.
+                // DIY to avoid making this module depend on Boost.
```
parejkoj, line 251: Thanks for the clarification (and also, there's not a string trim in the STL???).

src/image/ExposureFitsReader.cc
```diff
@@ -223,12 +413,9 @@ class ExposureFitsReader::ArchiveReader {
 
         // Not safe to call getAll if a component cannot be unpersisted
         // Instead, look for the archives registered in the metadata
```
parejkoj, line 415: Given how you reworked this with `readComponent`, is this comment still necessary?

parejkoj (commented):
A handful of questions/comments, but this all looks quite good.

The logic for building a `FilterLabel` from a `Filter` is even more tortured than I'd feared, but I'm glad you were able to get it sorted out.

src/image/ExposureFitsReader.cc
kfindeisen: Good catch, though I'll have to be careful to exclude "I2".

src/image/ExposureFitsReader.cc
kfindeisen: I think so? If I could call `getAll`, I wouldn't need to manually build up a Map of results.

PR #552, opened 2020-12-02
DM-27837: Add integer array indexing of Catalog

PaulPrice:


python/lsst/afw/table/_base.py
```diff
@@ -115,6 +115,13 @@ def __getitem__(self, key):
         elif isinstance(key, np.ndarray):
             if key.dtype == bool:
                 return self.subset(key)
+            if np.issubdtype(key.dtype, np.integer) and key.ndim == 1:
+                # Return a shallow copy containing the desired records
+                new = type(self)(self.schema)
```
TallJimbo, line 120: I think you at least want to use `self.table` rather than `self.schema` here, since this is a shallow copy.  That should transfer shared metadata as well, though I'm a little worried that there might be subclasses for which `type(self)(...)` is not a reliable way to transfer all desired state to the new class.  I'm not sure there is - it's just something the base class can't guarantee, and that's the reason the other branches in this function call a method that derived classes can override.

python/lsst/afw/table/_base.py
```diff
@@ -115,6 +115,13 @@ def __getitem__(self, key):
...
+                new.reserve(len(key))
```
TallJimbo, line 121: I don't think we want this for a shallow copy - that will both reserve space in the `vector`-of-`shared_ptr` (_maybe_ what we want, _if_ we're willing to assume it's unlikely that this will be followed by more appends, but the performance without it is guaranteed to scale fine anyway) _and_ allocate new backing-store space for new _deep_-copied records (not what we want at all).

TallJimbo (commented):
I suspect you'll get much better performance with an implementation that just translates the given array of integers into an array of bools using numpy-vectorized calls, and then calls `self.`subset` on that - that will keep you from having any `for` loops down in Python.

That will also sidestep the concerns of my inline comments.

timj:
Closing the pull request because the associated Jira ticket was marked as Won't Fix.

PR #553, opened 2020-12-09
Add two new functions for matching and reindexing catalogs

fred3m:
This PR adds the function `reIndexCatalog` adopted from code by @PaulPrice to apply a numpy array as an index to rows in an afw Catalog and the `matchCatalogsExact` function to use the unique combination of `(parentId, peakId, patch)` to exactly match two catalogs derived from the same `mergeDet` catalog. The main use case is to test changes in deblending and measurement algorithms on the same exact set of sources.

### DIFF
--- a/python/lsst/afw/table/catalogMatches.py
+++ b/python/lsst/afw/table/catalogMatches.py
@@ -20,7 +20,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 __all__ = ["makeMergedSchema", "copyIntoCatalog",
-           "matchesToCatalog", "matchesFromCatalog", "copyAliasMapWithPrefix"]
+           "matchesToCatalog", "matchesFromCatalog", "copyAliasMapWithPrefix",
+           "reindexCatalog"]
 
 import os.path
 
@@ -254,3 +255,26 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
         outSchema.getAliasMap().set(prefix + k, prefix + v)
 
     return outSchema
+
+
+def reindexCatalog(catalog, indices, deep=True):
+    """Apply a numpy index array to an afw Catalog
+
+    Parameters
+    ----------
+    catalog : `lsst.afw.table.SourceCatalog`
+        Catalog to reindex.
+    indices : `numpy.ndarray` of `int`
+        Index array.
+    deep : `bool`
+        Whether or not to make a deep copy of the original catalog.
+
+    Returns
+    -------
+    new : subclass of `lsst.afw.table.BaseCatalog`
+        Reindexed catalog. Records are shallow copies of those in ``catalog``.
+    """
+    new = SourceCatalog(catalog.table.clone() if deep else catalog.table)
+    records = [catalog[int(ii)] for ii in indices]
+    new.extend(records, deep=deep)
+    return new


python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+        In the gen3 butler this is done already, in gen2
+        it is recommended to use `patch2Int`, assuming that
+        the patches are the same structure as HSC, that range
+        from '0,0' to '9,9'.
+
+    Returns
+    -------
+    result: tuple of `lsst.afw.table.BaseCatalog`
+        Both catalogs reordered such that all rows are
+        matched to the same source.
```
TallJimbo, line 315: Looks like this actually returns a `list` of `SourceMatch` instances.

Given that, I think you should document `catalog1` and `catalog2` as being specifically `SourceCatalog`, unless you're sure this actually does work on other catalog types.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+    For a more robust solution, as long as the downstream catalog is
+    derived from the same mergeDet catalog, exact source matching
+    can be done via the unique ``(parent, deblend_peakID)``
+    combination. So this function performs this exact matching for
+    all sources both catalogs.
+
+    Parameters
+    ----------
+    catalog1, catalog2: subclass of `lsst.afw.table.BaseCatalog`
+        The two catalogs to merge
```
TallJimbo, line 302: I haven't seen this style in numpydoc parameters before.  Are you sure it's allowed?

Also, these aren't actually subclasses; they're instances (subclasses would be types).  But (see below), I think this should just be `SourceCatalog` anyway, because I'm skeptical the return type is compatible with any other catalog type.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+    indices : `numpy.ndarray` of `int`
+        Index array.
+
+    Returns
+    -------
+    new : subclass of `lsst.afw.table.BaseCatalog`
+        Reindexed catalog. Records are shallow copies of those in ``catalog``.
+    """
+    new = type(catalog)(catalog.schema)
+    new.reserve(len(indices))
```
TallJimbo, line 277: Probably don't want this `reserve` line, because you're making shallow copies and hence don't need to allocate new memory (there will be memory allocated for the pointers, but I don't think you need to worry about allocating that up-front; `std::vector` will be smart enough to do it plenty efficiently without any help).

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+        Catalog to reindex.
+    indices : `numpy.ndarray` of `int`
+        Index array.
+
+    Returns
+    -------
+    new : subclass of `lsst.afw.table.BaseCatalog`
+        Reindexed catalog. Records are shallow copies of those in ``catalog``.
+    """
+    new = type(catalog)(catalog.schema)
```
TallJimbo, line 276: I am not confident this is a reliable way to construct a new catalog that contains everything you might want from the previous one.

Or, rather, this one definitely doesn't, and it's not appropriate for shallow-copies - `type(catalog)(catalog.table)` would be closer, but I think the safe thing to do is to just say `SourceCatalog(catalog.table)`, which is definitely safe, and document this as something that doesn't work for other catalog types.

TallJimbo (changes requested):

python/lsst/afw/table/catalogMatches.py
natelust: I think the default is to make deep copies

fred3m: Yeah, I added the `deep` argument to `new.extend` because otherwise it returned non-contiguous catalog that I couldn't do anything with. But left it as a parameter in case it is useful to a user for speed.

natelust: Are you constructing the catalogs and then accessing them through a numpy interface?

TallJimbo: Ah, I didn't even notice that you were actually providing the option of doing deep copies below (I just saw "shallow" in the return docs above).

If you want to optionally do deep copies, do this:
```
new = SourceCatalog(catalog.table.clone() if deep else catalog.table)
records = [catalog[int(ii)] for ii in indices]
new.extend(records, deep=deep)
```

TallJimbo: (and then also document the `deep` parameter)

python/lsst/afw/table/catalogMatches.py
fred3m: That's a shame, but I'm fine with changing it to `SourceCatalog`. If someone does need to reindex a different type of catalog at least this will provide a starting point.

python/lsst/afw/table/catalogMatches.py
fred3m: @natelust yes. This was originally intended to be used by the `matchCatalogsExact` function until I learned that `pipe_analysis` requires a `Match` object. But I still think it's useful to be able to go `afwCatalog[indices]` and get back a table that is matched in some way.

python/lsst/afw/table/catalogMatches.py
fred3m: Good point on the subclasses, I'll update this.

What do you mean about the style for the numpydoc parameters? You mean listing two parameters on the same line? From https://numpydoc.readthedocs.io/en/latest/format.html \:

> When two or more input parameters have exactly the same type, shape and description, they can be combined:
```
x1, x2 : array_like
    Input arrays, description of `x1`, `x2`.
```

python/lsst/afw/table/catalogMatches.py
fred3m: Yup, I forgot to update the docs after I realized what pipe_analysis expected.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+
+    For a more robust solution, as long as the downstream catalog is
+    derived from the same mergeDet catalog, exact source matching
+    can be done via the unique ``(parent, deblend_peakID)``
+    combination. So this function performs this exact matching for
+    all sources both catalogs.
+
+    Parameters
+    ----------
+    catalog1, catalog2: subclass of `lsst.afw.table.BaseCatalog`
```
PaulPrice, line 301: I think you need a space either side of the `:`.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+        matched to the same source.
+    """
+    # Only match the individual sources, the parents will
+    # already be matched by the mergeDet catalog
+    sidx1 = catalog1["parent"] != 0
+    sidx2 = catalog2["parent"] != 0
+
+    # Create the keys used to merge the catalogs
+    parents1 = np.array(catalog1["parent"][sidx1])
+    peaks1 = np.array(catalog1["deblend_peakId"][sidx1])
```
PaulPrice, line 324: You're using column names specific to certain processing, which I think is counter to the concept of afw as a low-level library. Can you make the column names to be function parameters? If not, I think this needs to get hoisted up to a more suitable package.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+    index1 = np.arange(len(catalog1))[sidx1]
+    parents2 = np.array(catalog2["parent"][sidx2])
+    peaks2 = np.array(catalog2["deblend_peakId"][sidx2])
+    index2 = np.arange(len(catalog2))[sidx2]
+
+    if patch1 is not None:
+        if patch2 is None:
+            msg = "If the catalogs are from different patches then patch1 and patch2 must be specified"
+            msg += ", got {} and {}".format(patch1, patch2)
```
PaulPrice, line 333: I think people generally prefer `f`-strings to `format`. It reads much easier.

python/lsst/afw/table/catalogMatches.py
```diff
@@ -254,3 +256,112 @@ def copyAliasMapWithPrefix(inSchema, outSchema, prefix=""):
...
+            raise ValueError(msg)
+        patch1 = patch1[sidx1]
+        patch2 = patch2[sidx2]
+
+        key1 = np.rec.array((parents1, peaks1, patch1, index1),
+                            dtype=[('parent', np.int64), ('peakId', np.int32),
+                                   ("patch", patch1.dtype), ("index", np.int32)])
```
PaulPrice, line 340: Not at all important, but why do you mix single quotes and double quotes in the same code?

PaulPrice (changes requested):
I'll defer to @TallJimbo on `reindexCatalog`, but I think `matchCatalogsExact` needs to be made generic or moved to a different package.

python/lsst/afw/table/catalogMatches.py
fred3m: I think that this particular function is general in the sense that this is the _only_ way to exactly match two source catalogs built from the same `mergeDet` catalog. So it seems to fit in this module to me. I _could_ make it more general and just allow the user to specify the keys to match on, but I'm not sure that would be useful.

python/lsst/afw/table/catalogMatches.py
fred3m: It depends on the people. I find `format` statements much more readable and frequently use them in stack code.

python/lsst/afw/table/catalogMatches.py
fred3m: Laziness. I must have copied that code from somewhere because I rarely use single quotes, then replaced the first two column names and added more columns myself with double quotes.

python/lsst/afw/table/catalogMatches.py
natelust: Note that our docs are not exactly numpydoc format as I found out, so you need to consult our style guide

python/lsst/afw/table/catalogMatches.py
natelust: Jim, does that need to be a list comp or could a generator exp work fine?

python/lsst/afw/table/catalogMatches.py
PaulPrice: The point is that `mergeDet` catalogs are generated in pipe_tasks. afw shouldn't know about details from pipe_tasks or require a specific catalog schema: it is a general-purpose toolkit.

TallJimbo: Hmm.  The `mergeDet` concept is defined only far downstream in `pipe_tasks`, so I think I'm with Paul on this not really belonging in afw, but putting it in the same module or package as `MergeDetectionsTask` seems more appropriate than trying to invent something more general/hypothetical for this to claim to do.

python/lsst/afw/table/catalogMatches.py
TallJimbo: Probably _better_ for it to be a list or a tuple, because it'll be iterated over in pybind11, so considerations like being able to use [`PySequence_Fast`](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast) in the C API are probably more relevant than the usual pure-Python wisdom of preferring lazy iterators.

python/lsst/afw/table/catalogMatches.py
natelust: a separate comment though, dont use `+=` for multiline strings do:
```
msg = ("this is  a really long string "
            "and this is the rest of it")
```

fred3m:
I closed this PR in order to move these functions, with the requested changes, to pipe_tasks. I'll mention the reviewers from this channel once I create the PR and wait until late Monday before I merge in order to give them a chance to review the code if they so desire.

fred3m:
I lied. Sorry about the extra noise, but I realized that the `reindexCatalog` function really does below in afw. So I'll push an update with the changes requested by Jim.

fred3m:
Changes pushed here, https://github.com/lsst/pipe_tasks/pull/441/files, and https://github.com/lsst-dm/pipe_analysis/pull/66.

PR #554, opened 2020-12-12
DM-28004: ExposureInfo may persist dummy FilterLabels

kfindeisen:
This PR enforces the convention that `ExposureInfo::getFilter() == Filter()` corresponds to `ExposureInfo::getFilterLabel() == nullptr` at the translator level rather than the `ExposureInfo` constructor level. This prevents spurious "\_unknown\_" `FilterLabels` from getting persisted as part of new `Exposures`.

### DIFF
--- a/include/lsst/afw/image/Exposure.h
+++ b/include/lsst/afw/image/Exposure.h
@@ -296,7 +296,12 @@ public:
     void setDetector(std::shared_ptr<lsst::afw::cameraGeom::Detector const> detector) {
         _info->setDetector(detector);
     }
-    /// Set the Exposure's filter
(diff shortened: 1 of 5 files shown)


parejkoj (commented):
My only question is whether we could have a test here (or in obs_base?) that confirms/demonstrates this behavior. If that requires persisting a FITS file without a Filter defined, then don't bother, but if it's a trivial test?

Also, does this require changing the docstrings for ExposureInfo or the FitsReader? 
A quick skim of the docs (and a reminder that the guts are a `StorableMap` now) suggests not, but I'm not positive.

PR #555, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


python/lsst/afw/image/utils.py
```diff
@@ -50,12 +52,16 @@ def clipImage(im, minClip, maxClip):
             mi.getImage(), ds.getFootprints(), maxClip)
 
 
+@deprecated(reason=("Removed with no replacement (FilterLabels do not need to be reset)."
+                    " Will be removed after v23."), category=FutureWarning)
```
kfindeisen, line 56: Missing `version` keyword.

python/lsst/afw/image/utils.py
```diff
@@ -50,12 +52,16 @@ def clipImage(im, minClip, maxClip):
...
 def resetFilters():
     """Reset registry of filters and filter properties"""
     Filter.reset()
     FilterProperty.reset()
 
 
+@deprecated(reason=("Removed with no replacement (but see lsst::afw::image::TransmissionCurve)."
+                    " Will be removed after v23."), category=FutureWarning)
 def defineFilter(name, lambdaEff, lambdaMin=np.nan, lambdaMax=np.nan, alias=[], force=False):
     """Define a filter and its properties in the filter registry"""
```
kfindeisen, line 66: I'm not sure `TransmissionCurve` is an even approximate analogue for this functionality. This function is fundamentally about editing a central registry so that you can get the info from `Filter` later (AFAIK it's only being called by the obs packages and by unit tests). `TransmissionCurve` is only useful when you're trying to _retrieve_ transmission info, which is not something you do during obs package setup.

kfindeisen (approved):
Looks good, minor comments only.

include/lsst/afw/image/Exposure.h
```diff
@@ -236,8 +236,10 @@ class Exposure {
         return _info->getDetector();
     }
     /// Return the Exposure's filter
-    // TODO: deprecate in DM-27170, remove in DM-27177
-    Filter getFilter() const { return _info->getFilter(); }
+    // TODO: remove in DM-27177
+    [[deprecated("Replaced with getFilterLabel. Will be removed after v23.")]] Filter getFilter() const {
```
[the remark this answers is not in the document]
kfindeisen: Delegated to RFC-748.

include/lsst/afw/image/Color.h
```diff
@@ -58,8 +58,11 @@ class Color final {
 
     /** Return the effective wavelength for this object in the given filter
      */
-    double getLambdaEff(Filter const &  ///< The filter in question
-                        ) const {
+    [
+            [deprecated("Removed with no replacement (but see lsst::afw::image::TransmissionCurve). Will be "
+                        "removed after v23.")]] double
```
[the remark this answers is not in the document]
parejkoj: Yay clang_format? It's weird, but at least I didn't have to line-wrap them myself.

python/lsst/afw/image/utils.py
parejkoj: I guess I was thinking of the wavelengths that get passed in here. I've tweaked it.

parejkoj:
@kfindeisen : I changed how I deprecated FilterProperty in a new commit: it seems to have fixed the bug I was getting building afw in the new conda environment (seem #dm-build-problems). If you're happy with it, I'll squash it.

kfindeisen (approved):
Icky, but ok. I wonder if you should separately deprecate `FilterProperty.reset` (and other static methods), just to make sure those calls are spotted as well.

PR #556, opened 2021-01-05
DM-28208: Add python access to readComponent() and associated Storable tests.

erykoff:


jmeyers314 (approved):
LGTM!

First time I've seen a dataclass in the wild, so that was fun :)

One question though: how would one get a `Blob` from a butler?  I suppose it's too much to ask that `blob = butler.get("BLOB", dataId=dataId)` would work?

erykoff:
For gen2, it's just not possible, and won't be.  You have to do the read-an-exposure-with-a-single-pixel-bbox, and get it out of the `getInfo()`.  For gen3, I'm working on it, and the `readComponent()` here should make that possible.  I hope.

include/lsst/afw/image/ExposureFitsReader.h
```diff
@@ -147,6 +147,9 @@ class ExposureFitsReader {
     /// Read the Exposure's detector.
     std::shared_ptr<cameraGeom::Detector> readDetector();
 
+    /// Read an arbitrary non-standard component by name.
+    std::shared_ptr<typehandling::Storable> readComponent(std::string const &componentName);
```
kfindeisen, line 151: Should this be non-standard, or any component?

tests/test_Storable.py
```diff
@@ -148,6 +153,69 @@ def testNewValue(self):
         cppLib.assertCppValue(self.testbed, newstr)
 
 
+@dataclass
+class Blob(Storable):
```
kfindeisen, line 157: What does this class add that `DemoStorable`, above, does not?

kfindeisen (commented):
Thanks for taking care of this. Just a few questions.

include/lsst/afw/image/ExposureFitsReader.h
erykoff: Well, this is by string name, and the standard ones are by enumerated component, so I think this has to be non-standard.  I certainly couldn't figure out the appropriate names to input to get the "standard" ones, which is why I put that comment.

timj: It's a shame because if readComponent could read any component it would simplify the ExposureFormatter code a lot.

erykoff: That seems like a lot more work here, because of passing enumerations up the chain, I think.  But the `readComponent(component)` could also be exposed in python, I suppose.

kfindeisen: `ExposureInfo` [exposes keys](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/classlsst_1_1afw_1_1image_1_1_exposure_info.html#pub-static-attribs) for looking up standard components, though because of my short-sightedness `ArchiveReader::getComponent` takes a string instead of a full key. Assuming you don't change that, you'd need to call something like `reader.readComponent(ExposureInfo::KEY_WCS.getId())` in C++ or `reader.readComponent(ExposureInfo.KEY_WCS)` in Python, but it certainly can be done.

tests/test_Storable.py
erykoff: That isn't persistable, afaict.

include/lsst/afw/image/ExposureFitsReader.h
kfindeisen: Not that it matters for your case, but having `ExposureFitsReader::readComponent` take a key would make it more useful in C++; I'm happy to take that on as a separate ticket.

tests/test_Storable.py
kfindeisen: Could it be made persistable? Or does Python-side support for `Persistable` require something (e.g., `@dataclass`) that's incompatible with the existing class?

erykoff: I suppose it can, but I also thought it was useful to have an example here that was a straight dataclass with yaml persistence...it's a bit redundant, is that a problem?

include/lsst/afw/image/ExposureFitsReader.h
TallJimbo: :+1: for making this take enums and handle standard components in another ticket.  IIRC, at least some of the standard components are also _unusual_ components here, at least in terms of how the affect they metadata reading and hence interact with each other, and really the only safe way to handle them here would be to intercept them and delegate to the special method for that component.  And "intercept _all_ standard components" feels like a case where you really want to be using enums somehow to make sure the check is exhaustive (and remains exhaustive even if new standard components are added in other parts of the codebase).

TallJimbo (approved):

PR #558, opened 2021-01-15
DM-28368: Fix id check so that all components can use readComponent.

erykoff:
Note that the TransmissionCurve is special in some way and
does not work with the current implementation.

### DIFF
--- a/src/image/ExposureFitsReader.cc
+++ b/src/image/ExposureFitsReader.cc
@@ -401,6 +401,10 @@ public:
      *
      * @throws pex::exceptions::NotFoundError Thrown if the component is
      *     registered in the file metadata but could not be found.
+     *
+     * @note When accessing from python, components with derived subclasses, such
+     *       as ``TransmissionCurve``, are not properly type converted and thus
+     *       the specialized reader must be used instead of ``readComponent``.
      */
     // This method takes a string instead of a strongly typed Key because
     // readExtraComponents() gets its keys from the FITS metadata.
--- a/src/image/ExposureFitsReader.cc
+++ b/src/image/ExposureFitsReader.cc
@@ -411,7 +411,7 @@ public:
             return nullptr;
         }
 
-        if (_extraIds.count(c) > 0) {
+        if (_genericIds.count(c) > 0) {
             int archiveId = _genericIds.at(c);
             return _archive.get<T>(archiveId);
         } else {
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -32,7 +32,7 @@ from lsst.afw.table import ExposureTable
 from lsst.afw.image import (Image, Mask, Exposure, LOCAL, PARENT, MaskPixel, VariancePixel,
                             ImageFitsReader, MaskFitsReader, MaskedImageFitsReader, ExposureFitsReader,
                             Filter, FilterLabel, PhotoCalib, ApCorrMap, VisitInfo, TransmissionCurve,
-                            CoaddInputs)
+                            CoaddInputs, ExposureInfo)
 from lsst.afw.image.utils import defineFilter
 from lsst.afw.detection import GaussianPsf
 from lsst.afw.cameraGeom.testUtils import DetectorWrapper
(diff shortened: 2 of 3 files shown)


tests/test_readers.py
```diff
@@ -182,18 +182,29 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
         self.assertEqual(exposureIn.getMetadata().toDict(), reader.readMetadata().toDict())
         self.assertWcsAlmostEqualOverBBox(exposureIn.getWcs(), reader.readWcs(), self.bbox,
                                           maxDiffPix=0, maxDiffSky=0*degrees)
+        self.assertWcsAlmostEqualOverBBox(exposureIn.getWcs(), reader.readComponent('SKYWCS'),
```
kfindeisen, line 185: I recommend using `ExposureInfo.KEY_WCS`, if only to minimize cargo-culting of magic values.

tests/test_readers.py
```diff
@@ -182,18 +182,29 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
...
+        self.assertEqual(exposureIn.getInfo().getValidPolygon(), reader.readComponent('VALID_POLYGON'))
         self.assertCountEqual(exposureIn.getInfo().getApCorrMap(), reader.readApCorrMap())
+        self.assertCountEqual(exposureIn.getInfo().getApCorrMap(), reader.readComponent('AP_CORR_MAP'))
         self.assertEqual(exposureIn.getInfo().getVisitInfo().getExposureTime(),
                          reader.readVisitInfo().getExposureTime())
         point = Point2D(2.3, 3.1)
         wavelengths = np.linspace(4000, 5000, 5)
         self.assertFloatsEqual(exposureIn.getInfo().getTransmissionCurve().sampleAt(point, wavelengths),
                                reader.readTransmissionCurve().sampleAt(point, wavelengths))
+        # Note: readComponent('TRANSMISSION_CURVE') does not work.
```
kfindeisen, line 206: Nothing obviously jumps out at me; it should behave the same as every other component. Are you getting `None`, or some kind of error?

The only possibility I can think of is that the logic in https://github.com/lsst/afw/blob/master/src/image/ExposureFitsReader.cc#L360-L363 is failling. Can you manually inspect the file and see if it has an `ARCHIVE_ID_TRANSMISSION_CURVE` keyword, and that it matches the old-style `TRANSMISSION_CURVE_ID`?

kfindeisen (commented):
Sorry, I don't know why `TransmissionCurve` is behaving differently. Given the existing code I assume there should be one in the first place. 😕

tests/test_readers.py
erykoff: It returns a generic `Storable` that hasn't been casted to a `TransmissionCurve`.

kfindeisen: I think that means there's a bug in `TransmissionCurve`'s pybind11 wrapper. The formal C++ type is `Storable`, but that shouldn't matter in Python.

erykoff: I don't know how this works, and looking at the `transmissionCurve.cc` it seems okay to me compared to others like it.  Any hints greatly appreciated!

TallJimbo: The problem is that many of the `TransmissionCurve` subclasses are defined in the .cc file (https://github.com/lsst/afw/blob/master/src/image/TransmissionCurve.cc#L58) and are never declared to pybind11, because their interfaces don't extend beyond that of the base class, and users are only supposed to construct them via public factory methods.  The sad thing is that this is a very _good_ pattern for C++ alone (including, I think, in this case), but pybind11 just doesn't support it well : it will only automatically downcast from base (e.g. `Storable`) to most-derived (e.g. `ProductTransmissionCurve`) if the latter is known to it, not some intermediate base class it knows about (e.g. `TransmissionCurve`).  Also bums me out that this is something Boost.Python could do, and pybind11 just thought it was okay to drop it when they set out to replace Boost.Python.

Anyhow, I'm not sure where that leaves us.  I think I'd just continue to special-case everything we have so far in the formatter, and maybe add some warnings in the docs (`TransmissionCurve` may not be the only component affected - new subclasses of other components that aren't declared to pybind11 will suffer the same fate).  We could refactor `TransmissionCurve` itself to avoid the immediate problem, but that feels like a step backward for the C++ interface that doesn't really address the problem anyway.

kfindeisen: Ah, that would explain it. I was looking at the docs and didn't realize `TransmissionCurve` has subclasses.

Unfortunately, aside from including a special case (perhaps in the pybind11 wrapper for `ExposureFitsReader`) or registering the subclasses with pybind11 I don't see an obvious solution. My proposal to make `readComponent` take a `Key` won't solve this problem, because the pybind11 wrapper would still be working with strings and `Storable`.

erykoff: Ah, I see.  This is rather unfortunate.  Where does this go into the docs?

TallJimbo: How about a `Notes` section for this `readComponent` and the class docs for `Storable`?  If there's anywhere else @jmeyers314 wrote instructions for trampolining, I'd put it there, too.

TallJimbo: One can add strings to the pybind11 wrapper `.def` call itself as one of the arguments, but you're right that we don't do that regularly (it'd involve a lot of duplication or code generation), and I was thinking we'd just add it to the C++ Doxygen docs.  I'm _hoping_ the Doxygen/Sphinx integration that's in the works will at least associate those docstrings with the right Python classes/methods, though I have no idea how it could possibly handle non-trivial signature mappings.

erykoff: Is this where the sphinx docs are? https://github.com/lsst/afw/blob/master/src/image/ExposureFitsReader.cc#L393-L404

kfindeisen: If you mean the Doxygen docs, then yes.

kfindeisen: I'd also add an implementation note to `TransmissionCurve` that says most objects are of a subclass, and that this may affect pybind11 behavior. It may, as @TallJimbo says, be a good pattern in general, but it's very unusual in our code.

erykoff: I added https://github.com/lsst/afw/pull/558/commits/53cd70ac15c11308f8797bde5728ecdc097fcef5 for the `readComponent` note, but I need explicit direction on these other doc edits since it's further in the weeds of c++ that I don't have the vocabulary/knowledge to describe what is going on here.

src/image/ExposureFitsReader.cc
```diff
@@ -401,6 +401,10 @@ class ExposureFitsReader::ArchiveReader {
      *
      * @throws pex::exceptions::NotFoundError Thrown if the component is
      *     registered in the file metadata but could not be found.
+     *
+     * @note When accessing via %pybind11, components with derived subclasses, such
```
kfindeisen, line 405: I'd say "from Python"; I'm not sure all our Python developers think of themselves as pybind11 users.

EDIT: sorry for putting my original comment on the commit rather than the PR!

tests/test_readers.py
kfindeisen: I'm not sure what to say for `Storable` (it's not really an issue specific to that class), but I can push a blurb for `TransmissionCurve`.

erykoff: That would be awesome, thanks!  (And I agree, this isn't a `Storable` issue per-se, though I can see that the doc string needs to be modified to say that these pybind11 wrappers are non-trivial in some cases?)

erykoff:
@kfindeisen Thanks!  So is this good to go when Jenkins is green?

kfindeisen (approved):

PR #560, opened 2021-01-21
DM-27896: Add summary statistics component to Exposure

erykoff:


timj (approved):

python/lsst/afw/image/exposureSummary.py
```diff
@@ -0,0 +1,88 @@
...
+__all__ = ("ExposureSummary", )
+
+
+def _default_corners():
+    return [float('nan')]*4
+
+
+@dataclass
+class ExposureSummary(Storable):
+    _factory = StorableHelperFactory(__name__, "ExposureSummary")
```
kfindeisen, line 39: Since `StorableHelperFactory` has no documentation, I'm not sure I can say anything about its use. I gather that it's a generic `PersistableFactory` for anything that implements `_read`?

python/lsst/afw/image/exposureSummary.py
```diff
@@ -0,0 +1,88 @@
...
+    skyNoise: float = float('nan')
+    # Mean variance of the weight plane (ADU**2)
+    meanVar: float = float('nan')
+    # Right Ascension of bounding box corners (degrees)
+    raCorners: list[float] = field(default_factory=_default_corners)
+    # Declination of bounding box corners (degrees)
+    decCorners: list[float] = field(default_factory=_default_corners)
+
+    def __post_init__(self):
+        Storable.__init__(self)  # required for trampoline
```
kfindeisen, line 72: I disagree with the comment here. *Any* Python class that inherits from a C++ class and has its own `__init__` must explicitly call the base class constructor, trampoline or no.

python/lsst/afw/image/exposureSummary.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def isPersistable(self):
+        return True
+
+    def _getPersistenceName(self):
+        return "ExposureSummary"
```
kfindeisen, line 78: Does this name have to be the same as the argument to `StorableHelperFactory`? If so, I suggest factoring it out.

tests/test_exposure.py
```diff
@@ -843,6 +843,7 @@ def setUp(self):
         self.photoCalib = afwImage.PhotoCalib(1.5)
         self.psf = DummyPsf(2.0)
         self.detector = DetectorWrapper().detector
+        self.summary = afwImage.ExposureSummary()
```
kfindeisen, line 846: I think the tests will be a bit more meaningful if this is not the default value. It looks like `ExposureSummary`'s constructor can take keyword arguments?

python/lsst/afw/image/exposureInfoContinued.py
```diff
@@ -0,0 +1,40 @@
...
+
+from lsst.utils import continueClass
+from .exposureInfo import ExposureInfo
+
+__all__ = []  # import this module only for its side effects
+
+
+@continueClass  # noqa: F811
+class ExposureInfo:  # noqa: F811
+    KEY_SUMMARY = 'SUMMARY'
```
kfindeisen, line 31: Should this be `KEY_SUMMARY_STATS`? "Summary" seems generic enough that it could conflict with some other kind of summary in the future.

kfindeisen (commented):
I'm not happy about `ExposureInfo` needing to know about something specific to a particular algorithm, but I assume that's the best that can be done at present. Please document the new methods; remaining comments are mostly code style.

python/lsst/afw/image/exposureInfoContinued.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def getSummary(self):
+        return self.getComponent('SUMMARY')
+
+    def setSummary(self, summary):
+        self.setComponent('SUMMARY', summary)
+
+    def hasSummary(self):
+        return self.hasComponent('SUMMARY')
```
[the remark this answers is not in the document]
erykoff: I added the `getSummary()`/`setSummary()` because that was the only way I could figure out how to cleanly get the components in `obs_base`, yes.  The transmission curve thing is different and has to do with the `readComponent` (from the persisted file) rather than the `getComponent` (from the `exposureInfo`).

Also, the reason the class has to be defined here (rather than downstream) is to create a cleaner dependency tree.  I first had it in `pipe_tasks` but after thinking it through that was even more confusing/circular.  If you have a better idea where to put `ExposureSummary` I'm all ears.

python/lsst/afw/image/exposureSummary.py
erykoff: This is what @jmeyers314 put together, I believe.  He helped me with https://github.com/lsst/afw/blob/master/tests/test_Storable.py#L156-L216 which is what this is based on.

python/lsst/afw/image/exposureInfoContinued.py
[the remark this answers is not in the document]
kfindeisen: I agree that if you have an `ExposureInfo.getSummary` method, then it's better for `ExposureSummary` to be here than in `pipe_tasks`.

python/lsst/afw/image/exposureSummary.py
jmeyers314: There's a comment here https://github.com/lsst/afw/blob/master/include/lsst/afw/typehandling/python.h#L115 describing it.  Sorry this didn't end up in doxygen.  My doxygen-fu is weak.  Is it possible to have the comment (which describes both `StorableHelperFactory::read` and `StorableHelper::write` together) be captured as is?  Or would one need to break it apart into read/write components to be captured?

kfindeisen: It is technically possible, but most such comments in the existing Stack code don't actually make sense outside the context of a particular code file and its layout (e.g., they refer to method parameters that don't exist, and you'd need to do a search to figure out where they do exist). This is especially true in this case because `StorableHelperPersistenceHelper` (which is what I assumed that comment was for) is an implementation detail and should be hidden, while both `StorableHelper` and `StorableHelperFactory` need to be visible to class authors.

I would recommend having separate documentation for both `StorableHelper` and `StorableHelperFactory` (_not_ just their `write` and `read` methods, because things like the `_factory` line above are non-obvious), and take advantage of the fact that code documentation allows linking. I'd be happy to write something myself if you're willing to review it for accuracy.

erykoff: Presumably, this should go on another ticket?

python/lsst/afw/image/exposureInfoContinued.py
erykoff: A question: Do you think this should be renamed from `getSummary()` to `getSummaryStats()` everywhere?

kfindeisen: Yes, it should probably be consistent everywhere. (FWIW, I'm not particularly attached to "summary stats" as the name; I just gave it as an example.)

PR #571, opened 2021-02-23
DM-28658: modernize pybind wrappers for afw.display, cameraGeom and geom

mwittgen:


python/lsst/afw/cameraGeom/_camera.py
```diff
@@ -21,10 +21,10 @@
 
 # Camera object below is the same one exported by the pybind11 camera
 # module, so we don't need to re-export it here.
-__all__ = []
+__all__ = ['Camera']
```
TallJimbo, line 24: This change looks inconsistent with the comment above it.  If we need the change, we should remove or modify the comment.

python/lsst/afw/cameraGeom/_detector.cc
```diff
@@ -0,0 +1,209 @@
...
+void declareDetectorBuilder(PyDetector &parent) {
+    PyDetectorBuilder cls(parent, "Builder");
+    cls.def("setBBox", &Detector::Builder::setBBox);
+    cls.def("setType", &Detector::Builder::setType);
+    cls.def("setSerial", &Detector::Builder::setSerial);
+    cls.def("setPhysicalType", &Detector::Builder::setPhysicalType);
+    cls.def("setCrosstalk", &Detector::Builder::setCrosstalk);
+    cls.def("unsetCrosstalk", &Detector::Builder::unsetCrosstalk);
+    cls.def("getAmplifiers", &Detector::Builder::getAmplifiers);
+    // TODO: __iter__ defined in pure-Python extension
```
TallJimbo, line 154: This looks preexisting, but it looks like the "TODO" part can be removed; the `__iter__` is on `DetectorBase` in the pure-Python extension, so it should be inherited by this class.

TallJimbo (approved):

PR #572, opened 2021-02-23
DM-28727: Clean up pybind11 wrappers for afw.image

mwittgen:


python/lsst/afw/image/apCorrMap/_apCorrMap.cc
```diff
@@ -67,7 +67,7 @@ void wrapApCorrMap(lsst::utils::python::WrapperCollection &wrappers) {
     });
 }
 
-PYBIND11_MODULE(apCorrMap, mod) {
+PYBIND11_MODULE(_apCorrMap, mod) {
```
TallJimbo, line 71: I had forgotten that we had these separate subpackage builds in `afw.image` with their own `SConscript`s inside directories (`apCorrMap`, `exposure`, `image`, `maskedImage`, `pixel`).  With the exception of `pixel`, these are not intended to appear as public namespaces in Python (they aren't namespaces in C++), and I don't think there's any reason for them to be separate anymore.  Could you merge those packages (`apCorrMap`, `exposure`, `image`, `maskedImage`) into the `_imageLib` build and take them out of their subdirectories?

python/lsst/afw/image/image/_imageLib.cc
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "pybind11/pybind11.h"
+#include "lsst/utils/python.h"
+
+namespace lsst {
+namespace afw {
+namespace image {
+  //void wrapApCorrMap(lsst::utils::python::WrapperCollection &);
```
TallJimbo: Looks like a relic, but I think it'll just come back after you address my other comment.

TallJimbo (approved):
Looks good, but I'd like to go a step further and pull out some of the remaining subpackages, too (see in-line comment).

python/lsst/afw/image/apCorrMap/_apCorrMap.cc
mwittgen: I has some dependency problems when doing this. I would push this back and revisit when all other sub packages are using the pybind wrappers.

PR #573, opened 2021-03-03
DM-28717: Allow for Stamps formatter to handle bbox 

MorganSchmitz:
One of several PRs for [DM-28717](https://jira.lsstcorp.org/browse/DM-28717), the others being in daf_butler, meas_algorithms and obs_base.


python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -60,22 +64,24 @@ def imageReadFitsWithOptions(cls, source, options):
         If options contains "llcX" and is missing any of
         "llcY", "width", or "height".
     """
+    origin = ImageOrigin.PARENT
     bbox = lsst.geom.Box2I()
-    if options.exists("llcX"):
+    if "bbox" in options.keys():
+        bbox = options["bbox"]
+    elif isinstance(options, PropertySet) and options.exists("llcX"):
```
timj, line 71: This should support `dict` and not be restricted to PropertySet. You can use `in` so this will work just like the check for `bbox` above:

```python
elif "llcx" in options:
```

python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -52,6 +53,9 @@ def imageReadFitsWithOptions(cls, source, options):
         - imageOrigin: one of "LOCAL" or "PARENT" (has no effect unless
             a bbox is specified by llcX, etc.)
 
+        Alternatively, a bounding box () can be passed on with the
+        ``"bbox"'' key.
```
timj, line 57: You should document what type the value should be.

python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -60,22 +64,24 @@ def imageReadFitsWithOptions(cls, source, options):
...
-    origin = ImageOrigin.PARENT
-    if options.exists("imageOrigin"):
-        originStr = options.getString("imageOrigin")
-        if (originStr == "LOCAL"):
-            origin = ImageOrigin.LOCAL
-        elif (originStr == "PARENT"):
-            origin = ImageOrigin.PARENT
-        else:
-            raise RuntimeError("Unknown ImageOrigin type {}".format(originStr))
+        if options.exists("imageOrigin"):
```
timj, line 77: Change this to be `dict` compatible as `if "imageOrigin" in options`

python/lsst/afw/image/image/fitsIoWithOptions.py
```diff
@@ -25,6 +25,7 @@
 import lsst.geom
 from lsst.log import Log
 from lsst.afw.fits import ImageWriteOptions
+from lsst.daf.base import PropertySet
```
timj, line 28: This line is not needed if you use the dict compatibility interface of PropertySet

PR #576, opened 2021-03-30
DM-29501: replace unary_function, binary_function, deprecated in C++17

mwittgen:


examples/integrateCartesian.cc
```diff
@@ -67,18 +66,20 @@ class Parab1D : public std::unary_function<IntegrandT, IntegrandT> {
...
+  typedef IntegrandT second_argument_type;
+  typedef IntegrandT result_type;
     // declare coefficients at instantiation.
     Parab2D(IntegrandT const k, IntegrandT const kx, IntegrandT const ky) : _k(k), _kx(kx), _ky(ky) {}
 
     // operator() must be overloaded to return the evaluation of the function
     // ** This is the function to be integrated **
     // @note *need*  ... operator()() __const__ {
-    IntegrandT operator()(IntegrandT const x, IntegrandT const y) const {
+    IntegrandT operator()(IntegrandT const x, IntegrandT const y) const  {
```
timj, line 82: Why the extra space added?

mwittgen: Obsolete

erykoff (changes requested):
So the changes are good so far, but I fear that they are incomplete, in particular regarding `Integrate.h` and friends.  I can't comment on the non-diff code, so I have links here.
The basic problem is that there are two different interrelated code-paths here.  First, we have `integrate` which is a front-end for `int1d`; `integrate2d` which does a double loop over `integrate`.  As best as I can tell, this is the code path that is used in the stack and all the tests (except for the one example that has been commented out).
Second, we have `int2d` and `int3d` which (a) aren't used; (b) aren't tested; (c) aren't updated.  (In particular, https://github.com/lsst/afw/blob/e607ee5ed4d591f8b7f91192dae065ca16b8a65a/include/lsst/afw/math/Integrate.h#L646-L664 are not going to compile with c++17 but aren't used so don't fail tests).
Relatedly, there is code comment documentation https://github.com/lsst/afw/blob/e607ee5ed4d591f8b7f91192dae065ca16b8a65a/include/lsst/afw/math/Integrate.h#L144-L165 that is now out of date, and furthermore refers to the code path that isn't actually used so has been out of date for some untold years.

My thought would be to remove the incorrect documentation and the unused, untested, and probably broken anyway `int2d` and `int3d` code paths and friends (but of course leave `int1d` which is used by `integrate` and `integrate2d`).

tests/test_integrate.cc
```diff
@@ -0,0 +1,61 @@
+/*
```
erykoff, line 1: Should this test go into `tests/integrate.cc`?

mwittgen: Cleaned up unused code, added unit test for the int1d over Gaussian example in the comments to `tests/integrate.cc? `

erykoff: Yeah, I think that this test here should probably be in `tests/integrate.cc` and not its own file, unless you have a good reason otherwise.

mwittgen: Done. The ? was a typo. Copy/paste from your comment.

erykoff (approved):
Looks good, with all the dead code that I know of now removed.  Thanks!
All good assuming that this compiles and tests pass with both C++14 and C++17.

PR #580, opened 2021-04-14
DM-29737: Perform a brute-force search if binary search fails

arunkannawadi:
Instead of assuming that the catalog is always sorted, the updated `find` method first performs a binary search hoping that the catalog is sorted, and if unsuccessful, performs a brute-force linear search before declaring it unsuccessful. This fixes the bug of not finding record IDs for unsorted catalog, but fails slower for sorted catalogs.


TallJimbo (approved):
Looks good!

(Saw the PR and decided to approve without being asked before I run off to do chores; feel free to consider it reviewed on Jira, too, if you don't need further changes or they're trivial.)

PR #581, opened 2021-04-19
DM-29563: replace std::variant, fix clang compilation

mwittgen:


include/lsst/afw/typehandling/detail/refwrap_utils.h
```diff
@@ -0,0 +1,127 @@
...
+// contain std::reference_wrapper types.  Because reference_wrapper doesn't
+// define operator==, std::variant::operator== can't do this for us.
+struct refwrap_equals {
+
+    template <typename T>
+    bool operator()(std::reference_wrapper<T> a, std::reference_wrapper<T> b) {
+        return a.get() == b.get();
+    }
+
+    template <typename T, typename U>
```
kfindeisen, line 46: I realize this shouldn't be instantiated if `T == U`, but can you add an `enable_if` as an extra safeguard?

include/lsst/afw/typehandling/detail/refwrap_utils.h
```diff
@@ -0,0 +1,127 @@
...
+#include <functional>
+#include <variant>
+
+namespace lsst {
+namespace afw {
+namespace typehandling {
+namespace detail {
+
+// An equality comparison function object for std::variant visitors that
+// contain std::reference_wrapper types.  Because reference_wrapper doesn't
```
kfindeisen, line 37:
```suggestion
// contain only std::reference_wrapper types.  Because reference_wrapper doesn't
```

include/lsst/afw/typehandling/SimpleGenericMap.h
```diff
@@ -105,7 +104,10 @@ class SimpleGenericMap final : public MutableGenericMap<K> {
 protected:
     ConstValueReference unsafeLookup(K key) const override {
         try {
-            return _storage.at(key);
+            return std::visit(
+                [](auto const & v) { return ConstValueReference(std::cref(v)); },
```
kfindeisen, line 108: Again, it seems strange that this is necessary. Both `reference_wrapper` and `variant` have implicit constructors, so why can't you implicitly convert a `StorableType` to a `ConstValueReference`?

include/lsst/afw/typehandling/detail/refwrap_utils.h
```diff
@@ -0,0 +1,127 @@
...
+// a lot harder to follow.
+//
+// Construct with the make_refwrap_visitor function.
+template <typename F>
+class refwrap_visitor {
+private:
+    F _func;
+public:
+
+    explicit refwrap_visitor(F && func) : _func(std::move(func)) {}
```
kfindeisen, line 83: Should this be `std::forward`?

include/lsst/afw/typehandling/detail/refwrap_utils.h
```diff
@@ -0,0 +1,127 @@
...
+};
+
+// Specialization of refwrap_visitor for functions passed by lvalue reference.
+//
+// This is necessary because you can't perfectly forward an argument that you
+// need to capture in order to use it multiple times; if it's an rvalue
+// reference, it needs to be moved into a value (not just passed along into
+// another function as an rvalue reference).  Without a specialization, lvalue
+// references end up being coped, so modifications to the callable would not be
+// reflected to the caller.
```
kfindeisen, line 99: I'm not sure I understand. Is this to support stateful callables?

python/lsst/afw/typehandling/_GenericMap.cc
```diff
@@ -66,26 +51,48 @@ template <typename K>
...
+// A combination std::variant visitor and helper function (apply()) that
+// invokes it, to implement item lookup for Python.
+//
+// We copy/clone because there's no good way to prevent other ++ calls from
+// destroying the any reference we might give to Python.
+//
+// Pybind11 does have its own automatic caster for std::variant, but it doesn't
+// handle reference_wrapper types within it (let alone polymorphic Storables),
+// and since we need a visitor to deal with that, there's no point in going
+// back to std::variant instead of straight to Python.
```
kfindeisen, line 68: Is this actually any better than using the custom caster that was removed? It looks a lot more complicated.

python/lsst/afw/typehandling/_GenericMap.cc
```diff
@@ -66,26 +51,48 @@ template <typename K>
...
 template <typename K>
-py::object get(GenericMap<K>& self, K const& key) {
-    auto callable = static_cast<typename Publicist<K>::ConstValueReference (GenericMap<K>::*)(K) const>(
+class Getter {
+public:
+
+    template <typename T>
+    py::object operator()(std::reference_wrapper<T const> const & value) const {
+        // Generic case: unwrap the reference_wrapper, then copy-construct and
+        // cast to Python.
```
kfindeisen, line 76: Redundant comment.

include/lsst/afw/typehandling/detail/refwrap_utils.h
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// A visitor function wrapper for callables with two arguments, where the
+// second is the target of a std::reference_wrapper (as is the case for
+// GenericMap's visitation interface).
+//
+// With sufficient cleverness, this could probably be made to work with any
+// number of arguments of reference_wrapper and non-reference_wrapper types,
+// unwrapping only the former.  But that cleverness would probably make this
+// a lot harder to follow.
```
kfindeisen, line 74: Unproductive comment; I think it suffices to describe what is actually there.

kfindeisen (approved):
I'm surprised that `reference_wrapper`'s implicit conversions to and from references don't help here. Unfortunately, I don't have any good insights into how to deal with that.

include/lsst/afw/fits.h
```diff
@@ -669,8 +669,8 @@ class Fits {
  * - names in `second`, omitting "COMMENT" and "HISTORY" if valid versions appear in `first`
  */
 std::shared_ptr<daf::base::PropertyList> combineMetadata(
-        std::shared_ptr<const daf::base::PropertyList> first,
-        std::shared_ptr<const daf::base::PropertyList> second);
+        const std::shared_ptr<const daf::base::PropertyList>& first,
+        const std::shared_ptr<const daf::base::PropertyList>& second);
```
kfindeisen, line 673: DM style is `std::shared_ptr<const daf::base::PropertyList> const&`

include/lsst/afw/image/ImageBase.h
```diff
@@ -249,8 +249,9 @@ class ImageBase {
      * declare this function private
      */
     ImageBase& operator=(const ImageBase& rhs);
-    ImageBase& operator=(ImageBase&& rhs);
-    /// Set the %image's pixels to rhs
+    ImageBase& operator=(ImageBase&& rhs) noexcept;
+
+    virtual /// Set the %image's pixels to rhs
     ImageBase& operator=(const PixelT rhs);
```
kfindeisen, line 255: This looks corrupted. Shouldn't it be:
```suggestion
    /// Set the %image's pixels to rhs
    virtual ImageBase& operator=(const PixelT rhs);
```

src/image/Image.cc
```diff
@@ -119,7 +110,7 @@ ImageBase<PixelT>::ImageBase(ImageBase const& rhs, bool const deep
 }
 // Delegate to copy-constructor for backwards compatibility
 template <typename PixelT>
-ImageBase<PixelT>::ImageBase(ImageBase&& rhs) : ImageBase(rhs, false) {}
+ImageBase<PixelT>::ImageBase(ImageBase&& rhs) noexcept : ImageBase(rhs, false) {}
```
kfindeisen, line 113: If this is delegating to the copy-constructor, it probably can't guarantee it's not throwing.

src/image/Image.cc
```diff
@@ -743,7 +736,7 @@ bool imagesOverlap(ImageBase<T1> const& image1, ImageBase<T2> const& image2) {
     INSTANTIATE_OPERATOR(*=, T); \
     INSTANTIATE_OPERATOR(/=, T)
 
-#define INSTANTIATE2(T1, T2) template bool imagesOverlap<T1, T2>(ImageBase<T1> const&, ImageBase<T2> const&);
+#define INSTANTIATE2(T1, T2) template bool imagesOverlap<T1, T2>(ImageBase<T1> const&, ImageBase<T2> const&)
```
kfindeisen, line 739: Is this a good idea? IIRC omitting a semicolon for macro definitions confuses Doxygen.

include/lsst/afw/fits.h
```diff
@@ -312,7 +312,7 @@ class Fits {
     std::string getFileName() const;
 
     /// Return the current HDU (0-indexed; 0 is the Primary HDU).
-    int getHdu();
+    int getHdu() const;
```
TallJimbo, line 315: I don't think this is conceptually correct even if it compiles; the implementation calls a C library function via a pointer that can change state that we want to consider part of the object.

include/lsst/afw/image/Image.h
```diff
@@ -208,7 +208,7 @@ class Image : public ImageBase<PixelT> {
      * private
      */
     Image& operator=(const Image& rhs);
-    Image& operator=(Image&& rhs);
+    Image& operator=(Image&& rhs) noexcept;
```
TallJimbo, line 211: I get that we might want this to be `noexcept`, but have you checked that the implementation can actually guarantees it?  If we can't default it, then I'd guess not - though that may be something we could change if we migrate away from `boost::gil`.

include/lsst/afw/image/ImageBase.h
```diff
@@ -237,7 +237,7 @@ class ImageBase {
      *  will be deep-copied in Python, but the constructor will fail to compile in pure C++.
      */
     explicit ImageBase(Array const& array, bool deep = false,
-                       lsst::geom::Point2I const& xy0 = lsst::geom::Point2I());
+                       lsst::geom::Point2I  xy0 = lsst::geom::Point2I());
```
TallJimbo, line 240: This seems like an unnecessary ABI change.

include/lsst/afw/image/ImageBase.h
```diff
@@ -249,8 +249,9 @@ class ImageBase {
      * declare this function private
      */
     ImageBase& operator=(const ImageBase& rhs);
-    ImageBase& operator=(ImageBase&& rhs);
-    /// Set the %image's pixels to rhs
+    ImageBase& operator=(ImageBase&& rhs) noexcept;
+
+    virtual /// Set the %image's pixels to rhs
```
TallJimbo, line 254: This is not wanted here, and would be a weird place for it anyway.  Same for all of the ones below.

TallJimbo (changes requested):
Sorry it took me so long to get to this.

But I started looking at the clang-tidy changes, and almost every one I've seen is actually undesirable for one reason or another (or at least questionable) - usually it's a case where our code is doing something that isn't exactly best practice, but fixing it naively (as clang-tidy has done) is simply incorrect.

The only exceptions so far are the changes from iterator-based to range-based `for`; those do look okay.

I think it might be best to remove those clang-tidy commits from this ticket entirely, and perhaps run it much more carefully (or at least less aggressively) on a future ticket.  Were any of them needed to get the variant and filesystem changes compiling at all?

include/lsst/afw/image/Image.h
kfindeisen: This ties in to my comment above -- a lot of these move-constructors and move-assignments were added some time after we'd already migrated to C++11, as part of implementing the new [explicit default or delete rule](https://developer.lsst.io/cpp/style.html#b-classes-should-explicitly-declare-or-delete-compiler-generated-members). Out of perhaps an excessive preoccupation with backwards-compatibility, I had the new move constructors delegate to the copy-constructors, emulating the behavior when only custom copy-constructors were defined.

In the long run we probably do want to undo that and switch to a move-based implementation, but the current implementation is most definitely not `noexcept`.

include/lsst/afw/image/ImageBase.h
mwittgen: I'll take those out. It was a mistake to do blind automatic fixing in those cases.

mwittgen:
@TallJimbo removed all the clang-tidy stuff and also the cherry-pick for arm-osx compilation for now as is changes the compile flag -D_LIBCPP_DISABLE_AVAILABILITY=1 for all platforms and all the arm osx stuff should go into a proper ticket.

include/lsst/afw/typehandling/GenericMap.h
```diff
@@ -544,9 +537,10 @@ class GenericMap {
     // No return value, const GenericMap
     template <class Visitor, typename std::enable_if_t<std::is_void<_VisitorResult<Visitor>>::value, int> = 0>
     void _apply(Visitor&& visitor) const {
+        auto wrapped = detail::make_refwrap_visitor(std::forward<Visitor>(visitor));
```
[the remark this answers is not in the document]
kfindeisen: Yes, you're right; template instantiation ignores implicit conversions. 😖

include/lsst/afw/typehandling/GenericMap.h
```diff
@@ -409,8 +408,8 @@ class GenericMap {
                       "Due to implementation constraints, const keys are not supported.");
         try {
             auto foo = unsafeLookup(key.getId());
-            return boost::get<T const&>(foo);
-        } catch (boost::bad_get const&) {
+            return std::get<std::reference_wrapper<T const>>(foo);
+        } catch (std::bad_variant_access const&) {
```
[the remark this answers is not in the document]
mwittgen: Change to use `holds_alternative`.

include/lsst/afw/typehandling/detail/refwrap_utils.h
mwittgen: Change to `std::forward?`

python/lsst/afw/typehandling/_GenericMap.cc
mwittgen: Remove.

include/lsst/afw/table/types.h
```diff
@@ -47,11 +46,14 @@ namespace lsst {
 namespace afw {
 namespace table {
 
-/// An MPL vector of scalar field types.
-typedef boost::mpl::vector<AFW_TABLE_SCALAR_FIELD_TYPES> ScalarFieldTypes;
+template <typename ...E> struct TypeList {};
+
+/// A compile-time list of scalar field types.
+using ScalarFieldTypes = TypeList<AFW_TABLE_SCALAR_FIELD_TYPES>;
```
mwittgen, line 52: `ScalarFieldTypes` is never used.

include/lsst/afw/typehandling/GenericMap.h
```diff
@@ -406,10 +406,10 @@ class GenericMap {
     T const& _at(Key<K, T> const& key) const {
         static_assert(!std::is_const<T>::value,
                       "Due to implementation constraints, const keys are not supported.");
-        try {
-            auto foo = unsafeLookup(key.getId());
+        auto foo = unsafeLookup(key.getId());
+        if(std::holds_alternative<std::reference_wrapper<T const>>(foo)) {
```
TallJimbo, line 410: Our style has a space between `if` and `(`.

TallJimbo (approved):
Looks good!  Only remaining comment is a tiny style one; feel free to merge after that's addressed (assuming Jenkins is green, etc.).

PR #582, opened 2021-04-19
DM-29775: Fix style issues introduced in DM-29737 and log output for unsorted catalogs

arunkannawadi:
This PR a) fixes a few style errors introduced in a recent merge DM-29737 and b) create a log message if the catalog is found to be unsorted. The log message is to hint the user of plausible inefficiencies in finding a desired record as a result of having catalog unsorted by the Id.

### DIFF
--- a/include/lsst/afw/table/Catalog.h
+++ b/include/lsst/afw/table/Catalog.h
@@ -760,14 +760,15 @@ typename CatalogT<RecordT>::iterator CatalogT<RecordT>::find(typename Field<T>::
     detail::KeyExtractionFunctor<RecordT, T> f = {key};
     // Iterator adaptor that makes a CatalogT iterator work like an iterator over field values.
     typedef boost::transform_iterator<detail::KeyExtractionFunctor<RecordT, T>, iterator> SearchIter;
-    // Try binary search for log n search assuming the table is sorted.
-    // If the search is unsuccessful, try a brute-force search before quitting.
+    /* Try binary search for log n search assuming the table is sorted.
+     * If the search is unsuccessful, try a brute-force search before quitting.
+     */
     SearchIter i = std::lower_bound(SearchIter(begin(), f), SearchIter(end(), f), value);
-    if (i.base() == end() || *i != value)
-    {
+    if (i.base() == end() || *i != value) {
         i = std::find(SearchIter(begin(), f), SearchIter(end(), f), value);
-        if (i.base() == end() || *i != value)
+        if (i.base() == end() || *i != value) {
             return end();
+        }
     }
     return i.base();
 }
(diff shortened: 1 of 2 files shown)


include/lsst/afw/table/Catalog.h
```diff
@@ -15,6 +15,7 @@
 #include "lsst/afw/table/io/FitsWriter.h"
 #include "lsst/afw/table/io/FitsReader.h"
 #include "lsst/afw/table/SchemaMapper.h"
+#include "lsst/log/Log.h"
```
arunkannawadi, line 18: I couldn't find an example where logging is done from another header file. I guess it makes sense in general, but since the implementation of the `find` method is in this header file, and the info about whether or not the catalog was sorted is not propagated up, the logging has to be done from here (if it's done).

src/image/PhotoCalib.cc
```diff
@@ -95,7 +95,7 @@ void toNanojanskyVariance(ndarray::Array<float const, 2, 1> const &instFlux,
 }
 
 double toMagnitudeErr(double instFlux, double instFluxErr, double scale, double scaleErr) {
-    return 2.5 / log(10.0) * hypot(instFluxErr / instFlux, scaleErr / scale);
+    return 2.5 / std::log(10.0) * hypot(instFluxErr / instFlux, scaleErr / scale);
```
arunkannawadi, line 98: This was necessary to use the natural logarithm and not a logging function, and whether or not we decide to log, we should perhaps keep this change.

include/lsst/afw/table/Catalog.h
```diff
@@ -760,14 +761,16 @@ typename CatalogT<RecordT>::iterator CatalogT<RecordT>::find(typename Field<T>::
...
     SearchIter i = std::lower_bound(SearchIter(begin(), f), SearchIter(end(), f), value);
-    if (i.base() == end() || *i != value)
-    {
+    if (i.base() == end() || *i != value) {
         i = std::find(SearchIter(begin(), f), SearchIter(end(), f), value);
-        if (i.base() == end() || *i != value)
+        if (i.base() == end()) {
             return end();
+        }
+        LOGL_INFO("afw.table.Catalog", "Catalog is not sorted by the key. Finding a record may be slow.");
```
arunkannawadi, line 773: One of the motivations for logging this at some low level is that this is (probably?) the only direct indication we have to let us know that the catalog is unsorted and I think it'd be useful to keep that at TRACE level at least.

include/lsst/afw/table/Catalog.h
```diff
@@ -760,14 +760,15 @@ typename CatalogT<RecordT>::iterator CatalogT<RecordT>::find(typename Field<T>::
...
+     */
     SearchIter i = std::lower_bound(SearchIter(begin(), f), SearchIter(end(), f), value);
-    if (i.base() == end() || *i != value)
-    {
+    if (i.base() == end() || *i != value) {
         i = std::find(SearchIter(begin(), f), SearchIter(end(), f), value);
-        if (i.base() == end() || *i != value)
+        if (i.base() == end() || *i != value) {
             return end();
+        }
```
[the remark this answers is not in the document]
arunkannawadi: Totally understandable. I wasn't familiar with C++ style guidelines, so I made it a point to go back and check after merging, just in case.

PR #584, opened 2021-05-06
Replace assert with exception for invalid argument

arunkannawadi:
The methods `setNumIter` and `setNumSigmaClip` must throw an exception
if the input parameters are invalid instead of asserting that they are
valid.


MorganSchmitz (approved):
Good to go as soon as you got rid of the changes to `_statistics.cc`!

PR #585, opened 2021-05-07
DM-29936: getChildren without repeatedly checking if the SourceCatalog is sorted

arunkannawadi:


fred3m (approved):
I really like your implementation. It's a very clean approach using generators.

PR #588, opened 2021-05-11
DM-28716: Add pybind wrapper to afw archive that handles adding set of HDUs

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-28716)

[Jenkins run #34190](https://jira.lsstcorp.org/browse/DM-28716)

### DIFF
--- a/python/lsst/afw/table/io/_inputArchive.cc
+++ b/python/lsst/afw/table/io/_inputArchive.cc
@@ -1,6 +1,4 @@
 /*
- * This file is part of afw.
- *
  * Developed for the LSST Data Management System.
  * This product includes software developed by the LSST Project
  * (https://www.lsst.org).
@@ -25,28 +23,24 @@
 
 #include "lsst/utils/python.h"
 
+#include "lsst/afw/table/io/InputArchive.h"
+#include "lsst/afw/fits.h"
+
 namespace py = pybind11;
-using namespace pybind11::literals;
+using namespace py::literals;
 
 namespace lsst {
 namespace afw {
 namespace table {
 namespace io {
 
-using utils::python::WrapperCollection;
+using PyInputArchive = py::class_<InputArchive, std::shared_ptr<InputArchive>>;
 
-void wrapFits(WrapperCollection&);
-void wrapPersistable(WrapperCollection&);
-void wrapInputArchive(WrapperCollection&);
-void wrapOutputArchive(WrapperCollection&);
-
-PYBIND11_MODULE(_io, mod) {
-    WrapperCollection wrappers(mod, "lsst.afw.table.io");
-    wrapPersistable(wrappers);
-    wrapFits(wrappers);
-    wrapInputArchive(wrappers);
-    wrapOutputArchive(wrappers);
-    wrappers.finish();
+void wrapInputArchive(utils::python::WrapperCollection &wrappers) {
+    wrappers.wrapType(PyInputArchive(wrappers.module, "InputArchive"), [](auto &mod, auto &cls) {
+        cls.def("get", &InputArchive::get<Persistable>);
+        cls.def("readFits", &InputArchive::readFits);
+    });
 }
 
 }  // namespace io
--- a/src/image/MaskedImage.cc
+++ b/src/image/MaskedImage.cc
@@ -454,6 +454,7 @@ void processPlaneMetadata(std::shared_ptr<daf::base::PropertySet const> metadata
     }
     hdr->set("INHERIT", true);
     hdr->set("EXTTYPE", exttype);
+    hdr->set("EXTNAME", exttype);
 }
 
 }  // namespace
--- a/src/table/io/OutputArchive.cc
+++ b/src/table/io/OutputArchive.cc
@@ -99,6 +99,8 @@ public:
         } catch (pex::exceptions::NotFoundError &) {
             iter->getTable()->getMetadata()->add("AR_NAME", name, "Class name for objects stored here");
         }
+        // Also add an EXTNAME. The most recent AR_NAME given will be used.
+        iter->getTable()->getMetadata()->set("EXTNAME", name);
         indexRecord->set(indexKeys.row0, iter->size());
         indexRecord->set(indexKeys.catArchive, catArchive);
         iter->insert(iter->end(), catalog.begin(), catalog.end(), false);
@@ -142,6 +144,7 @@ public:
     Impl() : _nextId(1), _map(), _index(ArchiveIndexSchema::get().schema) {
         std::shared_ptr<daf::base::PropertyList> metadata(new daf::base::PropertyList());
         metadata->set("EXTTYPE", "ARCHIVE_INDEX");
+        metadata->set("EXTNAME", "ARCHIVE_INDEX");
         metadata->set("AR_CATN", 0, "# of this catalog relative to the start of this archive");
         _index.getTable()->setMetadata(metadata);
     }
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -179,6 +179,10 @@ class FitsReaderTestCase(lsst.utils.tests.TestCase):
         reader = ExposureFitsReader(fileName)
         self.assertIn('EXPINFO_V', reader.readMetadata().toDict(), "metadata is automatically versioned")
         reader.readMetadata().remove('EXPINFO_V')
+        self.assertIn('EXTNAME', reader.readMetadata().toDict(), "EXTNAME is added upon writing")
+        reader.readMetadata().remove('EXTNAME')
+        self.assertIn('EXTVER', reader.readMetadata().toDict(), "EXTVER is added upon writing")
+        reader.readMetadata().remove('EXTVER')
         self.assertGreaterEqual(reader.readSerializationVersion(), 0)
         self.assertEqual(exposureIn.getMetadata().toDict(), reader.readMetadata().toDict())
         self.assertWcsAlmostEqualOverBBox(exposureIn.getWcs(), reader.readWcs(), self.bbox,
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -23,6 +23,7 @@ import unittest
 
 import os
 import numpy as np
+import astropy.io.fits
 
 import lsst.utils.tests
 from lsst.daf.base import PropertyList
@@ -179,10 +180,17 @@ class FitsReaderTestCase(lsst.utils.tests.TestCase):
         reader = ExposureFitsReader(fileName)
         self.assertIn('EXPINFO_V', reader.readMetadata().toDict(), "metadata is automatically versioned")
         reader.readMetadata().remove('EXPINFO_V')
+        # ensure EXTNAMEs can be read and make sense
+        extnames = set(('PRIMARY', 'IMAGE', 'MASK', 'VARIANCE', 'ARCHIVE_INDEX',
+                        'Detector', 'TransformMap', 'TransformPoint2ToPoint2',
+                        'FilterLabel', 'SkyWcs', 'ApCorrMap', 'PhotoCalib',
+                        'ChebyshevBoundedField', 'CoaddInputs', 'GaussianPsf',
+                        'Polygon', 'VisitInfo'))
+        with astropy.io.fits.open(fileName) as astropyReadFile:
+            for hdu in astropyReadFile:
+                self.assertIn(hdu.name, extnames)
         self.assertIn('EXTNAME', reader.readMetadata().toDict(), "EXTNAME is added upon writing")
         reader.readMetadata().remove('EXTNAME')
-        self.assertIn('EXTVER', reader.readMetadata().toDict(), "EXTVER is added upon writing")
-        reader.readMetadata().remove('EXTVER')
         self.assertGreaterEqual(reader.readSerializationVersion(), 0)
         self.assertEqual(exposureIn.getMetadata().toDict(), reader.readMetadata().toDict())
         self.assertWcsAlmostEqualOverBBox(exposureIn.getWcs(), reader.readWcs(), self.bbox,
--- a/include/lsst/afw/table/io/python.h
+++ b/include/lsst/afw/table/io/python.h
@@ -47,8 +47,9 @@ namespace python {
  * This allows your class to be used without importing lsst.afw.table in Python.
  *
  * Use as follows:
- * - When declaring the pybind11 class that wraps your Class do *not* list
- *   table::io::PersistableFacade<Class> and table::io::Persistable as base classes.
+ * - When declaring the pybind11 class that wraps your Class in afw, do *not* list
+ *   table::io::PersistableFacade<Class> and table::io::Persistable as base classes
+ *   unless you have to, and in that case be prepared for possible trouble.
  * - Call this function to wrap the methods that make your object persistable.
  */
 template <typename Class, typename... Args>
--- a/python/lsst/afw/geom/_transform.cc
+++ b/python/lsst/afw/geom/_transform.cc
@@ -32,6 +32,7 @@
 #include "ndarray/pybind11.h"
 
 #include "lsst/afw/table/io/python.h"
+#include "lsst/afw/table/io/Persistable.h"
 #include "lsst/afw/geom/Endpoint.h"
 #include "lsst/afw/geom/Transform.h"
 
@@ -78,7 +79,7 @@ void declareTransform(lsst::utils::python::WrapperCollection &wrappers) {
 
     std::string const pyClassName = Class::getShortClassName();
     wrappers.wrapType(
-            py::class_<Class, std::shared_ptr<Class>>(wrappers.module, pyClassName.c_str()),
+            py::class_<Class, std::shared_ptr<Class>, table::io::Persistable>(wrappers.module, pyClassName.c_str()),
             [](auto &mod, auto &cls) {
                 std::string const pyClassName = Class::getShortClassName();
                 cls.def(py::init<ast::FrameSet const &, bool>(), "frameSet"_a, "simplify"_a = true);
--- a/python/lsst/afw/table/io/SConscript
+++ b/python/lsst/afw/table/io/SConscript
@@ -3,5 +3,5 @@ from lsst.sconsUtils import scripts
 scripts.BasicSConscript.pybind11(
     ['_io', ],
     # Need .cc here or Scons will look for the wrong file type
-    extraSrc={'_io': ['_fits.cc', '_persistable.cc'], },
+    extraSrc={'_io': ['_fits.cc', '_persistable.cc', '_inputArchive.cc', '_outputArchive.cc'], },
     addUnderscore=False)
--- a/python/lsst/afw/table/io/_outputArchive.cc
+++ b/python/lsst/afw/table/io/_outputArchive.cc
@@ -1,6 +1,4 @@
 /*
- * This file is part of afw.
- *
  * Developed for the LSST Data Management System.
  * This product includes software developed by the LSST Project
  * (https://www.lsst.org).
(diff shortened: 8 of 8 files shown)


python/lsst/afw/table/io/_inputArchive.cc
```diff
@@ -0,0 +1,52 @@
...
+
+namespace lsst {
+namespace afw {
+namespace table {
+namespace io {
+
+using PyInputArchive = py::class_<InputArchive, std::shared_ptr<InputArchive>>;
+
+void wrapInputArchive(utils::python::WrapperCollection &wrappers) {
+    // TODO: uncomment once afw.fits uses WrapperCollection
```
timj, line 40: Please refer to the Jira ticket that is going to be doing this work.

src/image/MaskedImage.cc
```diff
@@ -454,6 +454,8 @@ void processPlaneMetadata(std::shared_ptr<daf::base::PropertySet const> metadata
     }
     hdr->set("INHERIT", true);
     hdr->set("EXTTYPE", exttype);
+    hdr->set("EXTNAME", exttype);
+    hdr->set("EXTVER", 1);
```
timj, line 458: We are going to have to have a scheme that allows the EXTVER to be set by the caller somehow if they are writing a new set of images into an existing FITS file. I think we should probably not set EXTVER in the fits writer at all and leave it unset. That way the caller can manually set it to something and we can think later about how to explicitly control EXTVER when persisting at some other time. I'm pretty sure this all means we should not be forcing EXTVER anywhere on this PR (except maybe leaving something in the writeFits docs to say that the caller needs to set EXTVER if they are doing multiple writes to a single file).

src/table/io/OutputArchive.cc
```diff
@@ -99,6 +100,15 @@ class OutputArchive::Impl {
         } catch (pex::exceptions::NotFoundError &) {
             iter->getTable()->getMetadata()->add("AR_NAME", name, "Class name for objects stored here");
         }
+        // Also add it as the EXTNAME.
+        try {
```
timj, line 104: Are we doing this sort of "overwrite existing else add new one" logic in other places? I'm a bit confused because I thought the PropertyList `set()` method did this already. The `add()` method is designed for "add a new keyword here in addition to what's already there" but `set()` is "replace anything there already with this new one". So why isn't this code using `set()`? (I understand this is copying the implementation above).

tests/test_readers.py
```diff
@@ -179,6 +179,10 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
         reader = ExposureFitsReader(fileName)
         self.assertIn('EXPINFO_V', reader.readMetadata().toDict(), "metadata is automatically versioned")
         reader.readMetadata().remove('EXPINFO_V')
+        self.assertIn('EXTNAME', reader.readMetadata().toDict(), "EXTNAME is added upon writing")
```
timj, line 182: I would really like a test that checks that EXTNAME has the expected value.

timj (changes requested):
This looks good but I do have some things that need changing:

* Do not set EXTVER -- let the caller do it and add a note about that until we can sort out a real API. I worry that if people call persist multiple times into a single file we'll be opening ourselves up to confusion. Does the Stamps writing work fine on this PR if EXTVER is continually being forced to 1?
* Add some tests that we are getting a good EXTNAME. I think it might be worth doing a test with astropy.io.fits since if you do that you will be assured of wider compatibility because you can use the `hdu.name` property directly without reading the header explicitly for EXTNAME.
* `PropertyList.set()` vs `PropertyList.add()` confusion.

src/image/MaskedImage.cc
MorganSchmitz: I removed places where EXTVER was set. As for a mention in the docs, I assume it would be in the writeFits of Exposure...? So [there](https://github.com/lsst/afw/blob/4939cbb3d9e4416b6dd5c2f40f6cf6e070342a3c/include/lsst/afw/image/Exposure.h#L358-L409)? Should I add it for every overload?

timj: I don't really know the best place since adding it to every `writeFits` is going to be painful. I'm not familiar with the documentation in this part of the code. Maybe @TallJimbo can advise.

src/table/io/OutputArchive.cc
MorganSchmitz: Yeah, I was just stupidly copying the above. I changed it to  `set` and it works fine; do you want me to do the same change for AR_NAME?

timj: Yes, please change AR_NAME as well -- it will continue to confuse people I think. If it is like that for some reason we can't comprehend I'd hope someone would put a more explicit comment in there. `set()` is used elsewhere in the file.

timj: `git blame` indicates that @TallJimbo wrote the code so maybe there's some subtlety here we are not understanding.

python/lsst/afw/table/io/_inputArchive.cc
MorganSchmitz: I checked with Kryzstof; this is actually out of date as of DM-20703. Removed.

tests/test_readers.py
MorganSchmitz: I gave it a shot, but I'm pretty sure that's not what you had in mind. I essentially have a predetermined list of acceptable EXTNAMEs, and I doubt it is exhaustive (`PiffPsf` should probably be in there, though i'm not sure how we persist those and now that I think of it I doubt it's as part of Archives...?). If this is really kind of what we want, how can I make sure I've got every kind of Persistable we have?

It  works as is with all extensions from our regular processing calexps and coadds, fwiw.

I'm not sure how to be less stupid about this - since I can't know beforehand which HDU should correspond to which element.

MorganSchmitz:
> Does the Stamps writing work fine on this PR if EXTVER is continually being forced to 1?

Yep. Stamps only have the one Archive that contains all Transforms (or whatever else we want to store alongside the stamp images). Each set of IMAGE/MASK/VARIANCE then has its EXTVER incremented upon writing, so there's no conflict possible there. See my example on the Jira ticket:

```No. Name Ver Type Cards Dimensions Format
0 PRIMARY 1 PrimaryHDU 6 ()
1 ARCHIVE_INDEX 1 BinTableHDU 42 2R x 7C [1J, 1J, 1J, 1J, 1J, 64A, 64A]
2 TransformPoint2ToPoint2 1 BinTableHDU 18 2R x 1C [1QB(123)]
3 IMAGE 1 ImageHDU 19 (11, 11) float32
4 MASK 1 ImageHDU 26 (11, 11) int32
5 VARIANCE 1 ImageHDU 19 (11, 11) float32
6 IMAGE 2 ImageHDU 19 (11, 11) float32
7 MASK 2 ImageHDU 26 (11, 11) int32
8 VARIANCE 2 ImageHDU 19 (11, 11) float32```

File: tests/test_readers.py (hunk 4 above)
Comment (timj) [reply tests/test_readers.py:197]: You do know the set of EXTNAMEs you expect for your specific test dataset don't you?

At the very least reading the file with astropy.io.fits and iterating over each HDU to check that it is returning an actual name for `.name` property seems useful. It would show that this ticket is doing something and would also break if somehow we stopped filling in a name.

File: src/image/MaskedImage.cc (hunk 2 above)
Comment (TallJimbo) [reply src/image/MaskedImage.cc]: There's no good way to let the user control EXTVER that I can see; anything we'd do would be a big API change.  But I also don't see why they would need to; if they've got more than one archive in a single FITS file they're using the library incorrectly anyway.

File: src/table/io/OutputArchive.cc (hunk 3 above)
Comment (TallJimbo) [reply src/table/io/OutputArchive.cc]: No idea, I'm afraid.  I agree it looks like it should just be a `set`.

File: tests/test_readers.py [hunk 5]
```diff
@@ -179,6 +180,18 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
         reader = ExposureFitsReader(fileName)
         self.assertIn('EXPINFO_V', reader.readMetadata().toDict(), "metadata is automatically versioned")
         reader.readMetadata().remove('EXPINFO_V')
+        # ensure EXTNAMEs can be read and make sense
+        astropyReadFile = astropy.io.fits.open(fileName)
```
timj, line 184: Please use a context manager here so you don't need to do the explicit close (which would never run if the assert fails).

```python
extnames = {...}
with astropy.io.fits.open(fileName) as astropyReadFile:
    for hdu in astropyReadFile[1:]:
        assert
```

tests/test_readers.py
```diff
@@ -179,6 +180,18 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
...
+        extnames = ('IMAGE', 'MASK', 'VARIANCE', 'ARCHIVE_INDEX',
+                    'Detector', 'TransformMap', 'TransformPoint2ToPoint2',
+                    'FilterLabel', 'SkyWcs', 'ApCorrMap', 'PhotoCalib',
+                    'ChebyshevBoundedField', 'CoaddInputs', 'GaussianPsf',
+                    'Polygon', 'VisitInfo')
+        for hdu in astropyReadFile[1:]:
```
timj, line 190: Maybe add a comment here that you are skipping the primary HDU?

tests/test_readers.py
```diff
@@ -179,6 +180,18 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
         reader = ExposureFitsReader(fileName)
         self.assertIn('EXPINFO_V', reader.readMetadata().toDict(), "metadata is automatically versioned")
         reader.readMetadata().remove('EXPINFO_V')
+        # ensure EXTNAMEs can be read and make sense
+        astropyReadFile = astropy.io.fits.open(fileName)
+        extnames = ('IMAGE', 'MASK', 'VARIANCE', 'ARCHIVE_INDEX',
```
timj, line 185: Please use a `set` rather than `tuple` -- it's slightly more efficient for the `in` operator.

timj (approved):

src/table/io/OutputArchive.cc
MorganSchmitz: Using `set` (for AR_NAME) causes [this test](https://github.com/lsst/afw/blob/4939cbb3d9e4416b6dd5c2f40f6cf6e070342a3c/tests/test_tableArchives.cc#L714-L716) to fail:

```_____________________________________________________________ UtilsBinaryTester.test_exe__project_mschmitz_repos_afw_tests_test_tableArchives _____________________________________________________________

args = (<test_executables.UtilsBinaryTester testMethod=test_exe__project_mschmitz_repos_afw_tests_test_tableArchives>,)
self = <test_executables.UtilsBinaryTester testMethod=test_exe__project_mschmitz_repos_afw_tests_test_tableArchives>

    def test_executable_runs(*args):
        self = args[0]
>       self.assertExecutable(executable)

/software/lsstsw/stack_20210415/stack/miniconda3-py38_4.9.2-0.5.0/Linux64/utils/21.0.0-6-g0bf7090+18535a8d22/python/lsst/utils/tests.py:246: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/software/lsstsw/stack_20210415/stack/miniconda3-py38_4.9.2-0.5.0/Linux64/utils/21.0.0-6-g0bf7090+18535a8d22/python/lsst/utils/tests.py:218: in assertExecutable
    self.fail(msg)
E   AssertionError: Bad exit status from '/project/mschmitz/repos/afw/tests/test_tableArchives': 201
------------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------------
Running executable '/project/mschmitz/repos/afw/tests/test_tableArchives' with no arguments...
Running 15 test cases...
afw.image.fits.ExposureFitsReader WARN: Could not read PSF; setting to null: PersistableFactory with name 'DummyPsf' not found, and import of module 'testTableArchivesLib' failed (possibly because Python calls were not available from C++). {0}; loading object with id=1, name='DummyPsf' {1}
tests/test_tableArchives.cc(714): error: in "ArchiveMetadata": check names.size() == 2u has failed [1 != 2]
tests/test_tableArchives.cc(715): error: in "ArchiveMetadata": check names[0] == "Footprint" has failed [HeavyFootprintF != Footprint]
unknown location(0): fatal error: in "ArchiveMetadata": memory access violation at address: 0x00000030: no mapping at fault address
tests/test_tableArchives.cc(716): last checkpoint
```

timj: Memory access violation is really interesting. If you run the test outside pytest do you get a stack trace? The code assumes that getMetadata returns something so it's blowing my mind a bit that this is somehow dying in `set`.

MorganSchmitz: Um... How do I run a  `.cc` test on its own?

timj: I mean run "python tests/test_file_that_fails.py"

MorganSchmitz: Hmmmm! No error:

```.Running executable '/project/mschmitz/repos/afw/tests/test_tableArchives' with no arguments...
--------------------------------------------------------------------------------------
The following warning is expected, and is an indication the test has passed.
--------------------------------------------------------------------------------------

*** No errors detected
Running 15 test cases...
afw.image.fits.ExposureFitsReader WARN: Could not read PSF; setting to null: PersistableFactory with name 'DummyPsf' not found, and import of module 'testTableArchivesLib' failed (possibly because Python calls were not available from C++). {0}; loading object with id=1, name='DummyPsf' {1}
```

timj: are you running it from a built afw? That library should be loadable.

MorganSchmitz: Yes, locally rebuilt afw on top of the shared stack on lsst-devl

timj: I'll see if I can do a local build of this branch.

tests/test_readers.py
```diff
@@ -179,6 +180,18 @@ def checkExposureFitsReader(self, exposureIn, fileName, dtypesOut):
...
+        # ensure EXTNAMEs can be read and make sense
+        extnames = set(('PRIMARY', 'IMAGE', 'MASK', 'VARIANCE', 'ARCHIVE_INDEX',
+                        'Detector', 'TransformMap', 'TransformPoint2ToPoint2',
+                        'FilterLabel', 'SkyWcs', 'ApCorrMap', 'PhotoCalib',
+                        'ChebyshevBoundedField', 'CoaddInputs', 'GaussianPsf',
+                        'Polygon', 'VisitInfo'))
+        with astropy.io.fits.open(fileName) as astropyReadFile:
+            for hdu in astropyReadFile:
+                self.assertIn(hdu.name, extnames)
+        astropyReadFile.close()
```
timj, line 192: You should remove this close since it's not needed with a context manager.

src/table/io/OutputArchive.cc
MorganSchmitz: Thanks! I repushed with the change (commit 17610c8)

timj: Okay. Maybe I just built the wrong version then
```
$ rebuild -r tickets/DM-28716 afw
Fetching Products...
Fetching afw...
                                            afw ok [tickets/DM-28716] (2.2 sec).
Fetching daf_base...
Fetching daf_persistence...
Fetching pex_config...
Fetching geom...
Fetching log...
Fetching sphgeom...
Fetching utils...
Fetching astshim...
                                               pex_config ok [master] (1.2 sec).
Fetching afwdata...
                                                      log ok [master] (1.3 sec).
Fetching base...
                                                  sphgeom ok [master] (1.3 sec).
Fetching sconsUtils...
                                                  astshim ok [master] (1.3 sec).
                                                 daf_base ok [master] (1.5 sec).
Fetching pex_exceptions...
                                                     geom ok [master] (1.5 sec).
                                                    utils ok [master] (1.5 sec).
                                          daf_persistence ok [master] (1.6 sec).
                                           pex_exceptions ok [master] (0.9 sec).
                                                     base ok [master] (1.1 sec).
                                               sconsUtils ok [master] (1.1 sec).
                                                  afwdata ok [master] (1.2 sec).
Performing git-lfs pulls...
Pulling afwdata LFS data...
                                                           afwdata ok (0.1 sec).
# BUILD ID: b5356
Saving environment information in /Volumes/MediaHD/Users/timj/work/lsstsw/build/builds/b5356.env
             afwdata: 21.0.0-2-g294ccec (already installed).
          sconsUtils: 21.0.0-11-ge0be345 (already installed).
             astshim: 21.0.0-2-g45278ab+64c1bc5aa5 (already installed).
                base: 21.0.0-7-gdf92d54+64c1bc5aa5 (already installed).
             sphgeom: 21.0.0-1-g8760c09+64c1bc5aa5 (already installed).
      pex_exceptions: 21.0.0-2-gde069b7+4f46bdaea8 (already installed).
               utils: 21.0.0-6-g0bf7090+18535a8d22 (already installed).
            daf_base: 21.0.0-3-gd7ab7aa+fdc5edd43f (already installed).
                geom: 21.0.0-5-g8c1d971+7b9a448d34 (already installed).
                 log: 21.0.0-5-gcc89fd6+fdc5edd43f (already installed).
     daf_persistence: 21.0.0-8-g5674e7b+8087abed30 (already installed).
          pex_config: 21.0.0-1-ga51b5d4+5491f2a448 (already installed).
                 afw: tickets.DM-28716-g1a1c2439e9+fb0e51fe34 .................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................ok (1425.4 sec).
# BUILD b5356 completed.
```

timj: I've finally read the error message and got it to fail. I thought it was a segv but it's actually a simple C++ test failure.

```
./test_tableArchives 
Running 15 test cases...
--------------------------------------------------------------------------------------
The following warning is expected, and is an indication the test has passed.
--------------------------------------------------------------------------------------
afw.image.fits.ExposureFitsReader WARN: Could not read PSF; setting to null: PersistableFactory with name 'DummyPsf' not found, and import of module 'testTableArchivesLib' failed (possibly because Python calls were not available from C++). {0}; loading object with id=1, name='DummyPsf' {1}
tests/test_tableArchives.cc:714: error: in "ArchiveMetadata": check names.size() == 2u has failed [1 != 2]
tests/test_tableArchives.cc:715: error: in "ArchiveMetadata": check names[0] == "Footprint" has failed [HeavyFootprintF != Footprint]
tests/test_tableArchives.cc:716: error: in "ArchiveMetadata": check names[1] == "HeavyFootprintF" has failed [ != HeavyFootprintF]

*** 3 failures are detected in the test module "table - archives"
```
and these tests are reading AR_NAME header I guess? @TallJimbo is the DummyPsf warning a red herring? I think I need to read the test because it naively looks like it thought it was getting two headers for the name but is actually only getting one and it says HeavyFootprintF.

timj:
```c++
    auto names = metadata.getArray<std::string>("AR_NAME");
    std::sort(names.begin(), names.end());
    BOOST_CHECK_EQUAL(names.size(), 2u);
    BOOST_CHECK_EQUAL(names[0], "Footprint");
    BOOST_CHECK_EQUAL(names[1], "HeavyFootprintF");
```
so I don't understand how this test is consistent with the code we removed that explicitly stated that it wanted to ensure only one AR_NAME.

timj: I think we misunderstood the original motivation for the code we replaced for AR_NAME. I think we thought that it was trying to ensure only one AR_NAME is used. What it's actually doing is let you set multiple distinct AR_NAME values but ensures that it doesn't end up with setting a value that is already set. I guess we need to put the AR_NAME code back but leave the EXTNAME code using set and state that the most recent AR_NAME given is the value we want to use.

TallJimbo: Ah, sorry, I should have remembered this.  In retrospect it's probably too clever, but the system can save objects of different types to the same HDU if they have the same on-disk representation, and that does happen once in a while.

PR #591, opened 2021-05-26
DM-30339: replace boost::any with std::any

mwittgen:


python/lsst/afw/display/simpleFits.cc
```diff
@@ -39,9 +39,8 @@ using namespace posix;
 #include <cstring>
 #include <cassert>
 #include <cctype>
-
+#include <any>
 #include "lsst/pex/exceptions.h"
```
kfindeisen, line 43: I suggest leaving a blank line between the standard library and LSST-specific includes.

kfindeisen (approved):
This migration looks unnervingly straightforward...

PR #592, opened 2021-05-26
DM-30004: replace boost::regex with std::regex

mwittgen:


src/math/warpExposure.cc
```diff
@@ -160,11 +160,11 @@ std::string NearestWarpingKernel::NearestFunction1::toString(std::string const &
...
 std::shared_ptr<SeparableKernel> makeWarpingKernel(std::string name) {
     typedef std::shared_ptr<SeparableKernel> KernelPtr;
-    boost::cmatch matches;
-    static const boost::regex LanczosRE("lanczos(\\d+)");
+    std::cmatch matches;
+    static const std::regex LanczosRE("lanczos(\\d+)");
     if (name == "bilinear") {
         return KernelPtr(new BilinearWarpingKernel());
-    } else if (boost::regex_match(name.c_str(), matches, LanczosRE)) {
+    } else if (std::regex_match(name.c_str(), matches, LanczosRE)) {
```
ktlim, line 167: Can the `c_str()` here and the later `istringstream` be dropped since the pattern is so simple?

src/table/io/FitsSchemaInputMapper.cc
```diff
@@ -585,9 +585,9 @@ template <typename T, int N>
 class CovarianceConversionReader : public FitsColumnReader {
 public:
     static std::string guessUnits(std::string const &oldUnits) {
-        static boost::regex const regex("(.*)(\\^(\\d+))?", boost::regex::perl);
-        boost::smatch m;
-        if (!boost::regex_match(oldUnits, m, regex)) {
+        static std::regex const regex("(.*)(\\\\^(\\d+))?");
```
ktlim, line 588: Is this right?  I thought the `^` caret was being escaped.  If it is, a comment would help future readers.

ktlim (approved):

src/table/io/FitsSchemaInputMapper.cc
mwittgen: This was wrong. Should indeed be `\\^`

PR #593, opened 2021-06-03
DM-30426: New Psf bbox methods

jmeyers314:
Adds `computeImageBBox`, which returns `Psf->computeImage()->getBBox()`, but allows Psf subclasses to implement this without requiring the intermediate call to `computeImage()`.

Also adds `computeKernelBBox` for symmetry, though this is equal to `computeBBox`.


PaulPrice (approved):
I have some concerns, but the general approach makes sense.

PR #597, opened 2021-06-23
DM-30829: Add persistence for WarpingControl

kfindeisen:
This PR makes `WarpingControl` a `Persistable` and implements persistence. Since a `WarpingControl` may contain arbitrary kernels (not just the designated "warping" kernels), representing the state required making the kernels typically used with `WarpingControl` persistable as well. These were given only skeletal support, however, on the assumption that they would only be persisted as subobjects.


src/math/warpExposure.cc
```diff
@@ -158,6 +162,64 @@ std::string NearestWarpingKernel::NearestFunction1::toString(std::string const &
...
+        static LanczosKernelPersistenceHelper const instance;
+        return instance;
+    }
+
+private:
+    LanczosKernelPersistenceHelper()
+            : schema(), order(schema.addField<int>("order", "order of Lanczos function")) {}
+};
+
+class : public table::io::PersistableFactory {
```
jmeyers314, line 181: I had no idea you could make an anonymous class like this.  Neat.

src/math/warpExposure.cc
```diff
@@ -158,6 +162,64 @@ std::string NearestWarpingKernel::NearestFunction1::toString(std::string const &
     return os.str();
 }
 
+namespace {
+
+struct LanczosKernelPersistenceHelper {
```
jmeyers314, line 167: When I implemented `StorableHelperPersistenceHelper`, I deleted the move and copy operations (which I'm sure I just copied from `PersistenceFactory`).  I don't know if that was strictly necessary then, but I thought I'd at least point this out so you can think about it.

jmeyers314 (approved):
Looks good to me.  Just a couple small inline questions for my own c++ education.

While you're at this, in Persistable.h:67, could you change `PersistenceFactory` to `PersistableFactory`?  This has bothered me for a while but I've been too lazy to do anything about it.  Now seems like a good time.

include/lsst/afw/math/warpExposure.h
```diff
@@ -83,8 +84,16 @@ class LanczosWarpingKernel : public SeparableKernel {
      */
     int getOrder() const;
 
+    bool isPersistable() const noexcept override { return true; }
+
 protected:
     void setKernelParameter(unsigned int ind, double value) const override;
+
+    std::string getPersistenceName() const override { return "LanczosWarpingKernel"; }
```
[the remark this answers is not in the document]
kfindeisen: Yes, exactly.

src/math/warpExposure.cc
kfindeisen: Yes, a singleton must be non-copyable. However, I don't really like the `*Helper` formalism (just too much boilerplate), and since this class is only reachable in this one file, I feel the readability benefits of skipping those bits are worth it.

Actually, by that argument perhaps these classes shouldn't attempt to be singleton-like at all.

jmeyers314: Great, thanks for the answer.  Minimizing boilerplate sounds good to me, including removing the singleton-ness if you like (as long as everything still tests fine of course).

PR #599, opened 2021-07-13
DM-11419: Allow larger than 2GB offsets in tables - use std::size_t for offsets…

mwittgen:
Use std::size_t for offsets and sizes

### DIFF
--- a/tests/test_ticket11419.py
+++ b/tests/test_ticket11419.py
@@ -1,0 +1,54 @@
+# This file is part of afw.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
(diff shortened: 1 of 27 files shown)


tests/test_ticket11419.py
```diff
@@ -0,0 +1,54 @@
...
+import lsst.afw.table as afwTable
+import lsst.utils.tests
+
+
+class SchemaOffsetTest(unittest.TestCase):
+
+    def setUp(self):
+        self.schema = afwTable.Schema()
+        self.schema.addField('test1', type='ArrayF', size=370010025)
+        self.schema.addField('test2', type='ArrayF', size=370010025)
```
andy-slac, line 33: I guess this covers case when size of two fields is larger than 2GB. Maybe add a case when single field exceeds 2GB too?

andy-slac (approved):
Looks good but please check my comments, I think there may be couple more leftover `int`s.

PR #601, opened 2021-07-18
DM-22486: Ensure pointer to a table in a catalog is always initialized

mwittgen:
Ensure pointer to a table in a catalog is always initialized
and table contains a minimal schema


erykoff (approved):
This is a very nice convenience addition, as well as fixing an annoying bug.

PR #602, opened 2021-07-28
Use the intended error estimate for clipped variance in Statistics::getResult

jjbuchanan:


parejkoj:
`tests/test_statistics.py:testVarianceClip()` should test for this, but it doesn't check the errors. Would you be able to provide an added python test for this calculation? I'm not sure if the errors on the clipped variance are really just `sqrt(second)`: see the `todo` comment in `Statistics.h::getResult()`.

The test would be modified something like this:

```
varianceClip = afwMath.makeStatistics(image, afwMath.VARIANCECLIP | afwMath.ERRORS).getResult()
self.assertAlmostEqual(varianceClip[0], self.clippedVariance3*std**2, delta=2*delta)
expectedVarianceError = ???
self.assertAlmostEqual(varianceClip[1], expectedVarianceError)
```

We should also have a similar test for the STDEVCLIP errors above this section in the same method.

parejkoj:
Note that I'm tracking this on Jira here: https://jira.lsstcorp.org/browse/DM-31358

PR #603, opened 2021-08-17
DM-31417: Add makeHpxWcs() function to make a valid HPX wcs from a single pixel.

erykoff:
This code is adapted from AladinSrc.jar version 11.024.


python/lsst/afw/geom/_hpxUtils.py
```diff
@@ -0,0 +1,400 @@
...
+    # Finalize
+    return (np.rad2deg((np.pi/4.)*(l_d0h + l_in_d0h/float(nside_tile))),
+            np.rad2deg((np.pi/4.)*(h_d0h + h_in_d0h/float(nside_tile))))
+
+
+class _ZOrderCurve2DInt(object):
+    """
+    Z-Order 2D curve for 32-bit integer values.
+
+    Code is ported from AladinSrc.jar version 11.024,
```
timj, line 163: What is the code license for the aladin code?

erykoff: GPL3, which is compatible with us, right?  Where do I put that?

timj: Yes. We might want to note that explicitly in the comment because it effectively blocks us from changing the afw license in the future. Maybe even an explicit note at the top of the file so we won't miss it.

erykoff: I have added links to http://aladin.u-strasbg.fr/COPYING .  But the file is already GPLv3 as noted at the top, so I think that there's nothing else to add?

timj: That's great. The point is that we continue to ponder license changes to BSD 3 clause in parts of the code and this would put an explicit brake on that for afw.

erykoff: FYI, if that's possibly a problem for the future, I can put this routine elsewhere.

gpdf (commented):
I am speculating that taking a look at WCSLIB, and perhaps the HPXcvt program it provides, might shed additional light on the opaque code in Aladin.

Ultimately what matters is that common community programs such as DS9, Aladin, and Firefly (whose HPX implementation is in progress) understand our outputs, of course.

python/lsst/afw/geom/_hpxUtils.py
```diff
@@ -0,0 +1,407 @@
...
+    # a non-trivial computation, and typically is outside of the
+    # tile pixel itself.
+    cent_ra, cent_dec = _hpx_center(order, pixel, tilepix)
+
+    md = PropertySet()
+    md['CD1_1'] = -cos45_scale
+    md['CD1_2'] = -cos45_scale
+    md['CD2_1'] = cos45_scale
+    md['CD2_2'] = -cos45_scale
+    md['CTYPE1'] = 'RA---HPX'
```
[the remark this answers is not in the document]
erykoff: I can add those references.  But the Calabretta paper is only dealing with whole-sky mapping, not the single-pixel mapping.  Same with `HPXcvt` which says what the `CD` matrix should be, but has nothing about partial sky mapping.
My unit tests confirm that the outputs are compliant at least via AST.  I can add additional tests with the astropy wcs code to definitely confirm everything is up-to-snuff.
I have also generated images using these WCSs and confirmed that ds9 does everything correctly, but that's harder to do in unit tests.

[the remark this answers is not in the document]
timj: DS9 uses AST for WCS so I'd hope it agreed.

womullan (commented):
I think this is  nside - the number of pixel on the side of a healpix of a given order. It would be good to stick to the vernacular ..   12 * nside ** 2 = npix .. but you do calcuate nside later  correctly as 2**order. So this may be some other value then it shoudl have a better comment and say its not part of the healpix world.

womullan (commented):
is this  pix2ang  (its in the healpy and java libraries) .. Aladin had a some functions implemented outside the library  - pix2and returns the angel at the center of the pixel in radians..   I would have to dig out the paper to see if the calculation is the same which I can do later unless you are certain this is a different projection.

erykoff:
@womullan Oh I wish this was `pix2ang`.  It is definitely not, the `cent_ra`, `cent_dec` values are typically not inside the pixel, they can be many degrees away.
As for your question on "I think this is nside" which line are you referring to?

womullan (approved):
I think this is pretty clear as updated thanks

python/lsst/afw/geom/_hpxUtils.py
```diff
@@ -0,0 +1,426 @@
...
+    # define full-sky healpix projections.  For single pixels we
+    # use the code from AladinSrc.jar Tile2HPX.java to compute
+    # CRPIX1, CRPIX2.
+
+    # The nside of the sub-pixels is the product of the tile nside
+    # and the number of sub-pixels on a side.
+    nside_pix = nside*tilepix
+    # All tiles are rotated 45 degrees.
+    cos45 = np.sqrt(2.0)/2.0
+    # This defines the pixel scale.
```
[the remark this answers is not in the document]
erykoff: I think the best thing to do is to add an explicit reference to Table 5 in https://www.ivoa.net/documents/HiPS/20170519/REC-HIPS-1.0-20170519.pdf which gives the areas.  The relevant columns are the "Tile angular size" (how big is the cell that will be in the file) and "Tile pixel angular size" (the pixel scale in the HPX projection, assuming a default `shift_order=9`/`hips_tilepix=512`).

python/lsst/afw/geom/_hpxUtils.py
```diff
@@ -0,0 +1,426 @@
...
+    if pixel < 0 or pixel >= 12*nside_tile*nside_tile:
+        raise ValueError(f"pixel value {pixel} out of range.")
+
+    twice_depth = np.left_shift(order, 1)
+    xy_mask = np.left_shift(1, twice_depth) - 1
+    fc = _ZOrderCurve2DInt()
+
+    d0h = np.int32(np.right_shift(pixel, twice_depth))
+    _hash = fc.hash2ij(pixel & xy_mask)
+    i_in_d0h = fc.ij2i(_hash)
```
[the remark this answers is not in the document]
erykoff: Yes, definitely.  In my tests I tried to get close to the pole and things seem to agree, so it's not obviously wrong, but we should be on the lookout!

PR #604, opened 2021-08-27
DM-31542: Allow writeFitsWithOptions for image/mask to select the options section

timj:
Image and Mask previously always used an "image" section of the
compression options. This made it impossible to reuse the compression
settings from an Exposure without copying the relevant section to
image. This change allows the section to be explicitly specified
so that a user can declare that they want the variance settings
to be applied. Mask is changed to always prefer "mask" section
before falling back to "image".

### DIFF
--- a/python/lsst/afw/image/image/_mask.py
+++ b/python/lsst/afw/image/image/_mask.py
@@ -48,7 +48,27 @@ class Mask(metaclass=TemplateMeta):
 
     readFitsWithOptions = classmethod(imageReadFitsWithOptions)
 
-    writeFitsWithOptions = imageWriteFitsWithOptions
+    def writeFitsWithOptions(self, dest, options, item=None):
+        """Write an Mask to FITS, with options
+
+        Parameters
+        ----------
+        dest : `str`
+            Fits file path to which to write the mask.
+        options : `lsst.daf.base.PropertySet`
+            Write options. The item ``item`` is read.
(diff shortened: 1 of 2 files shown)


python/lsst/afw/image/image/_mask.py
```diff
@@ -48,7 +48,27 @@ def __repr__(self):
...
 
-    writeFitsWithOptions = imageWriteFitsWithOptions
+    def writeFitsWithOptions(self, dest, options, item=None):
+        """Write an Mask to FITS, with options
+
+        Parameters
+        ----------
+        dest : `str`
+            Fits file path to which to write the mask.
+        options : `lsst.daf.base.PropertySet
```
PaulPrice, line 58: Missing backtick.

python/lsst/afw/image/image/_mask.py
```diff
@@ -48,7 +48,27 @@ def __repr__(self):
...
+            Write options. The item ``item`` is read.
+            It must contain an `lsst.daf.base.PropertySet` with data for
+            ``lsst.afw.fits.ImageWriteOptions``.
+        item : `str`, optional
+            Item to read from the ``options`` `lsst.daf.base.PropertySet`.
```
PaulPrice, line 63: You have mentioned the type of `options` multiple times...

PaulPrice (approved):

PR #605, opened 2021-09-02
DM-31535: Deprecate default-position-argument interface to Psf methods.

jmeyers314:


python/lsst/afw/detection/_detectionContinued.py
```diff
@@ -0,0 +1,47 @@
...
+
+for method in [
+    'computeImage', 'computeKernelImage', 'computePeak', 'computeApertureFlux',
+    'computeShape', 'computeBBox', 'computeImageBBox', 'computeKernelBBox',
+    'getLocalKernel'
+]:
+    setattr(
+        Psf,
+        method,
+        deprecate_pybind11(
```
ktlim, line 37: If this is going to show up too frequently, I think it is possible to wrap these differently and explicitly issue a warning only on the bad overload.  It's a bit complex, though, which is why the Dev Guide doesn't mention it, and it may not be worth it.

ktlim (approved):

python/lsst/afw/detection/_detectionContinued.py
jmeyers314: How would you suggest this work?  My thought was to use a lambda when binding to manually emit a warning before passing through to the wrapped method.  Maybe there's a way to do this just within python though?

ktlim: Here's an example of how this was done in the past using a Python-only patch: https://github.com/lsst/afw/commit/488744f7d9a89d799963403493304a4b7e211849#diff-3428060cb78fe305ebd4f720718710a2421cc76d4b355393edb26ab6a74ce0d0

jmeyers314: Sure, I considered inspecting the args like this but decided against it since it doesn't accomplish the sphinx decoration that `deprecate_pybind11` does.

jmeyers314: But if we're okay with ignoring the sphinx docs mangling, I think I'd prefer doing the same thing without the argument inspection; i.e., in _psf.cc something like:

```c++
                // cls.def("computeImage", py::overload_cast<>(&Psf::computeImage, py::const_));
                cls.def("computeImage",
                        [](const Psf& psf) {
                            auto warnings = py::module::import("warnings");
                            auto builtins = py::handle(PyEval_GetBuiltins());
                            auto FutureWarning = builtins["FutureWarning"];
                            warnings.attr("warn")(
                                "This is the warning",
                                "category"_a=FutureWarning,
                                "stacklevel"_a=2
                            );
                            return psf.computeImage();
                        }
                );
```

I assume there's a way to make this into a template too, though that might be a bit beyond my skillset.

ktlim: Any of the three options is OK with me: `deprecate_pybind11`, warn in Python, warn in the wrapper.

PR #606, opened 2021-09-05
DM-29511: apply clang tidy cleanups

mwittgen:


src/geom/polygon/Polygon.cc
```diff
@@ -193,7 +193,7 @@ struct Polygon::Impl {
             std::vector<BoostPolygon> const& boostPolygons);
 
     template <class PolyT>
-    bool overlaps(PolyT const& other) const {
+    [[nodiscard]] [[nodiscard]] bool overlaps(PolyT const& other) const {
```
TallJimbo, line 196: Is this double-attribute correct?

TallJimbo (approved):
Looks good.  I'm happy with all of these fixes except (possibly):

- I'd like to try to make the range-based for conversion use `auto const &` a little more often (see line comment).

- I'm not sure of the value of `[[nodiscard]]` for us; I think it'll be visual noise for a lot of readers, and since we don't use return values for error propagation really at all, I don't think we're getting much.  I can think of one big exception, but it's not relevant for `afw`: in `geom` and `sphgeom`, we have a lot of classes with both in-place and returning versions of many operations (e.g. `flip` operates in place, `flipped` returns a new geometry), and `[[nodiscard]]` on the returning versions would be very nice.  I don't feel very strongly about this, though, so if it's a pain to remove just that commit I don't mind if it gets merged as well.

PR #610, opened 2021-10-11
DM-31801: Use lsst prefix for loggers

timj:
Still uses `lsst.log`

### DIFF
--- a/src/fits.cc
+++ b/src/fits.cc
@@ -72,7 +72,7 @@ std::string makeLimitedFitsHeaderImpl(std::vector<std::string> const &paramNames
                 // use G because FITS wants uppercase E for exponents
                 out += (boost::format("%#20.17G") % value).str();
             } else {
-                LOGLS_WARN("afw.fits",
+                LOGLS_WARN("lsst.afw.fits",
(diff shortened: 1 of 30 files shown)


src/fits.cc
```diff
@@ -1072,7 +1072,7 @@ void writeKeyFromProperty(Fits &fits, daf::base::PropertySet const &metadata, st
         }
     } else {
         // FIXME: inherited this error handling from fitsIo.cc; need a better option.
-        LOGLS_WARN("afw.writeKeyFromProperty",
+        LOGLS_WARN("lsst.afw.writeKeyFromProperty",
```
andy-slac, line 1075: This logger name is different from all others, I presume this is intentional?

Also maybe define a constant for "lsst.afw.fits" logger name instead of repeating it 10 times?

andy-slac (approved):
Looks great.

src/fits.cc
timj: I wasn't sure how much I wanted to change the logger names but I'll take a second look at this one. It probably should have `fits` in it.

PR #614, opened 2021-11-02
DM-13943: Deprecate VisitInfo.getExposureId()

kfindeisen:
This PR adds C++ and Python deprecation warnings to `VisitInfo.getExposureId`, removes internal use of that method aside from tests of `VisitInfo` itself, and fixes some bugs in `VisitInfo` (in particular, a broken implementation of `operator==`; as in lsst/geom#35, simply banning NaN values is not an option) that were interfering with lsst/obs_base#396.

### DIFF
--- a/include/lsst/afw/coord/Observatory.h
+++ b/include/lsst/afw/coord/Observatory.h
@@ -85,10 +85,17 @@ public:
     std::string toString() const;
 
     bool operator==(Observatory const& rhs) const noexcept {
-        auto deltaLongitude = (_latitude - rhs.getLatitude()).wrapCtr();
-        auto deltaLatitude = (_longitude - rhs.getLongitude()).wrapCtr();
+        // Observatory may be initialized to NaN values as a placeholder, or to indicate "unknown".
+        auto deltaLongitude = std::isnan(_latitude.asRadians()) && std::isnan(rhs._latitude.asRadians())
+                                      ? 0.0 * lsst::geom::degrees
+                                      : (_latitude - rhs._latitude).wrapCtr();
+        auto deltaLatitude = std::isnan(_longitude.asRadians()) && std::isnan(rhs._longitude.asRadians())
+                                     ? 0.0 * lsst::geom::degrees
+                                     : (_longitude - rhs._longitude).wrapCtr();
         return (deltaLongitude == 0.0 * lsst::geom::degrees) &&
-               (deltaLatitude == 0.0 * lsst::geom::degrees) && ((_elevation - rhs._elevation) == 0.0);
+               (deltaLatitude == 0.0 * lsst::geom::degrees) &&
+               ((std::isnan(_elevation) && std::isnan(rhs._elevation)) ||
+                (_elevation - rhs._elevation) == 0.0);
     }
     bool operator!=(Observatory const& rhs) const noexcept { return !(*this == rhs); }
 
--- a/src/coord/Weather.cc
+++ b/src/coord/Weather.cc
@@ -22,6 +22,7 @@
  * see <http://www.lsstcorp.org/LegalNotices/>.
  */
 
+#include <cmath>
 #include <sstream>
 
 #include "lsst/utils/hashCombine.h"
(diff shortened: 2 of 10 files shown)


include/lsst/afw/coord/Observatory.h
```diff
@@ -85,10 +85,18 @@ class Observatory final {
...
-        auto deltaLongitude = (_latitude - rhs.getLatitude()).wrapCtr();
-        auto deltaLatitude = (_longitude - rhs.getLongitude()).wrapCtr();
+        auto deltaLongitude =
+                std::isfinite(_latitude.asRadians()) && std::isfinite(rhs.getLatitude().asRadians())
+                        ? (_latitude - rhs.getLatitude()).wrapCtr()
+                        : 0.0 * lsst::geom::degrees;
+        auto deltaLatitude =
+                std::isfinite(_longitude.asRadians()) && std::isfinite(rhs.getLongitude().asRadians())
+                        ? (_longitude - rhs.getLongitude()).wrapCtr()
+                        : 0.0 * lsst::geom::degrees;
```
spenczar, line 95: This logic is very surprising to me. When comparing two Observatories, one with `+Inf` latitude and longitude is equal to _all_ other Observatories? I would have expected Inf to only equal Inf (maybe ignoring sign, I guess).
```

File: src/coord/Weather.cc [hunk 2]
```diff
@@ -38,8 +39,13 @@ Weather::Weather(double airTemperature, double airPressure, double humidity)
...
 bool Weather::operator==(Weather const& other) const noexcept {
-    return (_airTemperature == other.getAirTemperature() && _airPressure == other.getAirPressure() &&
-            _humidity == other.getHumidity());
+    bool tempMatch = (std::isnan(_airTemperature) && std::isnan(other.getAirTemperature())) ||
+                     _airTemperature == other.getAirTemperature();
+    bool presMatch = (std::isnan(_airPressure) && std::isnan(other.getAirPressure())) ||
+                     _airPressure == other.getAirPressure();
+    bool humiMatch =
+            (std::isnan(_humidity) && std::isnan(other.getHumidity())) || _humidity == other.getHumidity();
+    return (tempMatch && presMatch && humiMatch);
```
spenczar, line 48: I have the same concerns about NaN equality here as in my comment at https://github.com/lsst/geom/pull/35#pullrequestreview-795905964.

src/image/VisitInfo.cc
```diff
@@ -472,11 +472,31 @@ VisitInfo::VisitInfo(daf::base::PropertySet const& metadata)
...
+ * Test whether two SpherePoints are exactly equal or invalid.
+ *
+ * This function is needed because SpherePoint::operation== is specifically
+ * designed to be ill-behaved when NaN values are involved.
+ */
+bool _eqOrNan(lsst::geom::SpherePoint const& lhs, lsst::geom::SpherePoint const& rhs) noexcept {
+    return (!lhs.isFinite() && !rhs.isFinite()) || lhs == rhs;
+}
+
 bool VisitInfo::operator==(VisitInfo const& other) const {
```
spenczar, line 493: Again, same question about whether NaN equality is desirable, yadda yadda yadda.

src/image/VisitInfo.cc
```diff
@@ -472,11 +472,31 @@ VisitInfo::VisitInfo(daf::base::PropertySet const& metadata)
...
+
+/**
+ * Test whether two SpherePoints are exactly equal or invalid.
+ *
+ * This function is needed because SpherePoint::operation== is specifically
+ * designed to be ill-behaved when NaN values are involved.
+ */
+bool _eqOrNan(lsst::geom::SpherePoint const& lhs, lsst::geom::SpherePoint const& rhs) noexcept {
+    return (!lhs.isFinite() && !rhs.isFinite()) || lhs == rhs;
+}
```
spenczar, line 491: The name says `_eqOrNan`, but the code seems to be `_eqOrNotFinite`, or something like that. To be fair, I don't really know what an equality test between +Inf and -Inf would mean for SpherePoints, so maybe this is fine.

src/image/VisitInfo.cc
```diff
@@ -472,11 +472,31 @@ VisitInfo::VisitInfo(daf::base::PropertySet const& metadata)
...
+
 bool VisitInfo::operator==(VisitInfo const& other) const {
-    return _exposureId == other.getExposureId() && _exposureTime == other.getExposureTime() &&
-           _darkTime == other.getDarkTime() && _date == other.getDate() && _ut1 == other.getUt1() &&
-           _era == other.getEra() && _boresightRaDec == other.getBoresightRaDec() &&
-           _boresightAzAlt == other.getBoresightAzAlt() && _boresightAirmass == other.getBoresightAirmass() &&
+    return _exposureId == other.getExposureId() && _eqOrNan(_exposureTime, other.getExposureTime()) &&
```
spenczar, line 494: This mostly makes me wonder why `_id == other.getId()` is not sufficient :) Not worth digging into that here.

src/coord/Weather.cc
kfindeisen: Here I think there's a much stronger case, because `Weather`, `Observatory`, and `VisitInfo` cannot even be thought of as mathematical objects or number-like in any way.

In these objects, NaNs are used unofficially to mean "unknown" or "not applicable". This is especially true for `VisitInfo`, where the pybind11 wrapper allows for any field to be undefined by default. This whole rabbit-hole started because two `VisitInfo` objects did not compare equal because a _single field_ was set to NaN in both.

spenczar (approved):

PR #618, opened 2021-11-22
DM-32662: rework catalog column array access to improve type dispatch performance

TallJimbo:


include/lsst/afw/table/types.h
```diff
@@ -46,12 +46,47 @@ namespace lsst {
...
+/// A compile-time list of types.
+template <typename ...E> struct TypeList;
+
+/// Partial specialization of TypeList used for recursive implementation of members.
+template <typename T, typename ...E> struct TypeList<T, E...> {
+
+    /// A constexpr variable that evaluates to true if U is in the list (no
+    /// checking for const, reference, mutable, etc.)
+    template <typename U>
+    static constexpr bool contains = std::is_same_v<T, U> || TypeList<E...>::template contains<U>;
```
natelust, line 58: I know I am missing something, and it probably is not important to the review, but I cant see where contains is actually used anywhere. Don't take too much time, but if you could explain it or point me to something I would appreciate satisfying my curiosity.

natelust (approved):
A few suggestions and questions but looks good.

PR #620, opened 2021-12-06
DM-32823: Change ds9 calls to use non-zero frame number

mfisherlevine:


tests/test_ds9.py
```diff
@@ -36,9 +36,11 @@
 except Exception:
     ds9 = None
 
+FRAME_NUM = 731  # largish, random number, unlikely to be in use by a user
```
erykoff, line 39: This is an arbitrary number, not a random number.

erykoff (approved):
Minor pedantry.

tests/test_ds9.py
mfisherlevine: Lol, I'll fix that once Jenkins passes and then merge without rerunning, I think that's fine, and not worth restarting Jenkins for.

PR #625, opened 2022-01-25
DM-33193: Exposure.getReadoutCorner method returns incorrect values for some LSST CCD amplifiers.

czwa:
Update the amplifier readout corner if data was flipped.


python/lsst/afw/cameraGeom/_amplifier.py
```diff
@@ -181,6 +181,20 @@ def transform(self, *, outOffset=None, outFlipX=False, outFlipY=False):
         bbox.shift(self.getRawXYOffset() + shift - outOffset)
 
         getattr(self, f"setRaw{bboxName}BBox")(bbox)
+
+    # Update the Readout Corner if we've flipped anything.
+    outReadoutCorner = self.getReadoutCorner()
+    if self.getRawFlipX() != outFlipX:
+        xFlipMapping = {ReadoutCorner.LL: ReadoutCorner.LR, ReadoutCorner.LR: ReadoutCorner.LL,
```
arunkannawadi, line 188: Would it make more sense to do this operation in L166, so it happens right after `bbox` is flipped?

python/lsst/afw/cameraGeom/_amplifier.py
```diff
@@ -181,6 +181,20 @@ def transform(self, *, outOffset=None, outFlipX=False, outFlipY=False):
...
+                        ReadoutCorner.UR: ReadoutCorner.UL, ReadoutCorner.UL: ReadoutCorner.UR}
+        outReadoutCorner = xFlipMapping[outReadoutCorner]
+    if self.getRawFlipY() != outFlipY:
+        yFlipMapping = {ReadoutCorner.LL: ReadoutCorner.UL, ReadoutCorner.LR: ReadoutCorner.UR,
```
arunkannawadi, line 192: Similarly, this could happen after current L180.

arunkannawadi (approved):
Looks good, overall. Just some minor comments about where the changes should go, assuming that the code will still be correct. Is it fairly straightforward to add a few lines in `test_amplifier.py` that checks that the readout corners are coming out correctly?

czwa:
I agree on the unit test comment, and have added two lines to the existing test that check that the readout corners match expectations.

python/lsst/afw/cameraGeom/_amplifier.py
arunkannawadi: I meant this could go before the loop. The `bbox` loop begins only at L171, right?

PR #626, opened 2022-01-28
DM-33443: Correct longitude stringification to deg East.

ktlim:


isullivan (approved):
This fix looks fine. It's probably beyond the scope of this ticket, but I'll note that it's confusing that in the declaration on lines 39-41 above these changes we list longitude followed by latitude, but then _latitude followed by _longitude

ktlim:
At one point I thought we had a standard that said to order the instance variable constructors in the order of their definition in the class (which this is), but I don't see that anymore.  I'll leave that alone at this time.

PR #628, opened 2022-01-31
DM-33470: Remove unused daf_persistence dependency

timj:


src/formatters/Utils.cc
```diff
@@ -32,132 +32,14 @@
...
-bool extractOptionalFlag(std::shared_ptr<PropertySet const> const& properties, std::string const& name) {
-    if (properties && properties->exists(name)) {
-        return properties->getAsBool(name);
-    }
-    return false;
-}
-
-int countFitsHeaderCards(lsst::daf::base::PropertySet const& prop) { return prop.paramNames(false).size(); }
-
 ndarray::Array<std::uint8_t, 1, 1> stringToBytes(std::string const& str) {
```
timj, line 43: Obvious question is whether to move what's left of this file into another location because it's not really anything to do with `formatters` any more.

ktlim: Seems like the place to move it would be somewhere under afw.table.  It's not really as connected with typehandling.

TallJimbo (approved):
I've got no insider knowledge on where these might or might not be used, so if this gets through Jenkins it's fine with me.

src/formatters/Utils.cc
timj: I see that these two functions are only used for tables so I suppose I could try to move them somewhere. Unless you have a good idea where I'm inclined to merge and get the easy win of daf_persistence avoidance.

TallJimbo: I don't understand.  If you've removed them, but they are still "used for tables", why isn't something breaking?

timj: We are talking about the two functions that are left in the file. They seem to be used by WCS for example to convert to bytes to stuff into a table and then convert back to strings to read out of a table.

TallJimbo: Ah, ok.  I'd lean towards moving them into the source files that use them, if it's just a couple of source files (not headers), but if it's much work at all, then leaving them where they are is probably best, since reducing churn is good, too.

timj: No. There are four places:
```
include/lsst/afw/typehandling/python.h
178:        std::string stringRep = formatters::bytesToString(record.get(keys.bytes));
202:    record->set(keys.bytes, formatters::stringToBytes(o));

src/math/TransformBoundedField.cc
112:        auto frameSetStr = formatters::bytesToString(record.get(keys.frameSet));
137:    record->set(keys.frameSet, formatters::stringToBytes(getTransform().writeString()));

src/geom/SkyWcs.cc
98:        std::string stringRep = formatters::bytesToString(record.get(keys.wcs));
385:    record->set(keys.wcs, formatters::stringToBytes(writeString()));

src/geom/Transform.cc
232:        std::string stringRep = formatters::bytesToString(record.get(keys.bytes));
244:    record->set(keys.bytes, formatters::stringToBytes(writeString()));
```

timj: So I'll merge.

ktlim: I think the right thing to do might have been to look at the common usage in the above four locations.  It's always getting or setting in a BaseRecord, so that might be the place to move it.

PR #629, opened 2022-02-08
DM-33591: Add getConvexPolygon() convenience method to exposure.

erykoff:


python/lsst/afw/image/exposure/_exposureContinued.py
```diff
@@ -94,6 +96,25 @@ def setVariance(self, variance):
 
     variance = property(getVariance, setVariance)
 
+    def getConvexPolygon(self):
+        """Get the convex polygon associated with the bounding box corners.
+
+        Returns
+        -------
+        convexPolygon : `lsst.sphgeom.ConvexPolygon`
+            Returns `None` if exposure does not have a valid WCS.
```
parejkoj, line 105: Should this be a `raise` instead?

python/lsst/afw/image/exposure/_exposureContinued.py
```diff
@@ -94,6 +96,25 @@ def setVariance(self, variance):
 
     variance = property(getVariance, setVariance)
 
+    def getConvexPolygon(self):
+        """Get the convex polygon associated with the bounding box corners.
```
parejkoj, line 100: The docs for this should probably note that this polygon is not a perfect sky-match to the detector boundaries. If people use this to check whether a sky point is within a detector, they need to also do a direct `skyToPixel` check afterwards, if they care about ~few pixel accuracy.

On that note: what's the rough value of the inaccuracy of this polygon? I'd expect a handful of pixels at most, and I hope that it's always strictly larger than the projected detector, but I don't have a good head for sky polygons.

tests/test_exposure.py
```diff
@@ -777,6 +777,39 @@ def testGetCutout(self):
...
+        x, y = np.meshgrid(np.arange(bbox.getBeginX(), bbox.getEndX()),
+                           np.arange(bbox.getBeginY(), bbox.getEndY()))
+        wcs = self.exposureMiWcs.wcs
+        ra, dec = wcs.pixelToSkyArray(x.ravel().astype(np.float64),
+                                      y.ravel().astype(np.float64))
+
+        poly = self.exposureMiWcs.convex_polygon
+        contains = poly.contains(ra, dec)
+
+        self.assertTrue(np.all(contains))
```
parejkoj, line 796: I wonder if this might provide more useful fail output if it was `assertEqual(contains, np.ones(len(contains), dtype=bool)`? And similarly for the negative version just below.

python/lsst/afw/image/exposure/_exposureContinued.py
erykoff: I first wrote this with a raise, but then I thought about it some more and decided that if it's okay that `wcs` is `None` then it would make sense that `convex_polygon` would also be `None`.  Though I could change it, I see both ways, but as a property I think it's better if it doesn't raise.

python/lsst/afw/image/exposure/_exposureContinued.py
erykoff: The inaccuracy depends on the size of the exposure and the latitude, I think.  If you need a cushion you can always dilate it, there are routines in `sphgeom`.

python/lsst/afw/image/exposure/_exposureContinued.py
```diff
@@ -94,6 +96,25 @@ def setVariance(self, variance):
...
+            Returns `None` if exposure does not have a valid WCS.
+        """
+        wcs = self.wcs
+        if wcs is None:
+            return None
+
+        bbox = lsst.geom.Box2D(self.getBBox())
+        corners = [wcs.pixelToSky(corner).getVector()
+                   for corner in bbox.getCorners()]
+        return lsst.sphgeom.ConvexPolygon(corners)
```
natelust, line 114: Since you are not padding this there is a chance that because distortions and such that the spherical 'rectangle" created by this will actually not contain pixels from the edges of the bbox. Is this something we should worry about, or just put a message? In the past we have dilated such that you can be sure that all points in the image are in the region, but the converse may not be true. This requires 2 checks, but you can be sure no sph points that are rejected are not actually in the image.

tests/test_exposure.py
```diff
@@ -777,6 +777,39 @@ def testGetCutout(self):
...
+
+        # Check that points outside of the bounding box are not in the polygon
+        x = np.array([bbox.getBeginX() - 1,
```
ktlim, line 799: Isn't the distortion potentially biggest in the middle of the sides rather than the corners?

python/lsst/afw/image/exposure/_exposureContinued.py
erykoff: I did some tests on HSC detectors in the center and the edge, and found that growing the bounding box by 7 pixels was sufficient to ensure that the polygon contained the entire detector.  Therefore, I set a default padding of 10 pixels.

One thing that I'm now unsure of is that this will be a larger/different polygon than is retrieved by the skymap `patch_info.outer_sky_polygon`, which doesn't need a buffer.

parejkoj: I suppose there's no way for this to know whether it's operating on a patch or a single exposure? (I hear Jim saying "Yes we should have had different classes for those things!" in the background)

erykoff: What's hard is making this work idiomatically, but one could say "if this is a SIP-type WCS we need padding", but what could the parameter be called in this case?

natelust (approved):

python/lsst/afw/image/exposure/_exposureContinued.py
parejkoj: Please update the docs for this, per the above discussion.

erykoff: @parejkoj are you asking just to update the docstring with something like "The returned polygon has additional padding to ensure that the bounding box is entirely contained within it." or something else?

parejkoj: Yes, that and something about the user needing to do a second check on the original pixel bbox, "is this sky point on the detector?" is the goal.

erykoff:
```
        Get the convex polygon associated with the bounding box corners.

        The returned polygon has additional padding to ensure that the
        bounding box is entirely contained within it. To ensure a set of
        ra/dec points are entirely contained within a detector, a second
        check of points contained within the convex polygon must be
        projected to x/y and checked against the exposure bounding box.
```

tests/test_exposure.py
parejkoj: poking this one...

erykoff: Already done, note the outdated badge.  I guess I didn't click an emoji...

python/lsst/afw/image/exposure/_exposureContinued.py
parejkoj: Second sentence, maybe be explicit about it?
```
...entirely contained within this exposure, run ``exposure.wcs.skyToPixel()`` to get the points in pixel coordinates, and check that those points are in the exposure's pixel box with ``exposure.getBBox().contains()``.
```

erykoff:
I've done some refactoring and I think this needs a re-review.  Thanks @natelust and @parejkoj !

python/lsst/afw/image/exposure/exposureUtils.py
```diff
@@ -0,0 +1,112 @@
...
+    degrees : `bool`, optional
+            Input ra, dec arrays are degrees if True.
+    padding : `int`
+       Pixel padding to ensure that bounding box is entirely contained
+       within the resulting polygon.
+
+    Returns
+    -------
+    contained : `np.ndarray`
+       Boolean indicating which points are contained in the bounding box.
```
parejkoj, line 85: "Boolean array ..."

python/lsst/afw/image/exposure/exposureUtils.py
```diff
@@ -0,0 +1,112 @@
...
+    ------
+    ValueError if exposure does not have a valid wcs.
+    """
+    poly = bbox_to_convex_polygon(bbox, wcs, padding=padding)
+
+    if poly is None:
+        raise ValueError("Exposure does not have a valid wcs.")
+
+    if degrees:
+        _ra = np.deg2rad(np.atleast_1d(ra)).astype(np.float64)
```
parejkoj, line 97: Why the astype here? If you're worried that everything should be double arithmetic, wouldn't you want to do that on `ra` itself?

python/lsst/afw/image/exposure/exposureUtils.py
```diff
@@ -0,0 +1,112 @@
...
+        _dec = np.deg2rad(np.atleast_1d(dec)).astype(np.float64)
+    else:
+        _ra = np.atleast_1d(ra).astype(np.float64)
+        _dec = np.atleast_1d(dec).astype(np.float64)
+
+    radec_contained = poly.contains(_ra, _dec)
+
+    x_in, y_in = wcs.skyToPixelArray(_ra, _dec, degrees=False)
+
+    xy_contained = ((x_in >= bbox.minX)
```
parejkoj, line 107: Why not `box.contains(x_in, y_in)`? Yes DM-33726, but we can fix that and shouldn't use propagate this more complicated and error-prone version.

tests/test_exposure.py
```diff
@@ -777,6 +777,88 @@ def testGetCutout(self):
                     with self.assertRaises(pexExcept.InvalidParameterError, msg=msg):
                         self.smallExposure.getCutout(cutoutCenter, cutoutSize)
 
+    def testGetConvexPolygon(self):
+        """Test the convex polygon."""
+        # Check that we do not have a convex polygon for the plain exposure.
+        self.assertIsNone(self.exposureMiOnly.convex_polygon)
+
+        # Check that all the points in the bounding box are in the polygon
```
parejkoj, line 785: "in the padded bounding box"?

tests/test_exposure.py
```diff
@@ -777,6 +777,88 @@ def testGetCutout(self):
...
+        bbox = self.exposureMiWcs.getBBox()
+        # Grow by the default padding.
+        bbox.grow(10)
+        x, y = np.meshgrid(np.arange(bbox.getBeginX(), bbox.getEndX()),
+                           np.arange(bbox.getBeginY(), bbox.getEndY()))
+        wcs = self.exposureMiWcs.wcs
+        ra, dec = wcs.pixelToSkyArray(x.ravel().astype(np.float64),
```
parejkoj, line 792: Why astype here?

tests/test_exposure.py
```diff
@@ -777,6 +777,88 @@ def testGetCutout(self):
...
+                                      y.ravel().astype(np.float64))
+
+        poly = self.exposureMiWcs.convex_polygon
+        contains = poly.contains(ra, dec)
+        np.testing.assert_array_equal(contains, np.ones(len(contains), dtype=bool))
+
+        # Check that points outside of the bounding box are not in the polygon
```
parejkoj, line 799: "that points one pixel outside"?

tests/test_exposure.py
```diff
@@ -777,6 +777,88 @@ def testGetCutout(self):
...
+        bbox.grow(1)
+
+        ra, dec = wcs.pixelToSkyArray(
```
parejkoj, line 802: I'm a bit confused by all of these different versions. I think this is testing each side of the box? Maybe making a small closure would make the differences clearer? It might not help, though.

```
def check_outside(x, y):
    """Check that the x,y points are all outside the polygon."""
    ra, dec = wcs.pixelToSkyArray(x, y)
    contains = poly.contains(ra, dec)
    np.testing.assert_array_equal(contains, np.zeros(len(contains), dtype=bool))
```

tests/test_exposure.py
```diff
@@ -777,6 +777,88 @@ def testGetCutout(self):
...
+            np.full(100, bbox.getEndX()),
+            np.linspace(bbox.getBeginY(), bbox.getEndY(), 100))
+        contains = poly.contains(ra, dec)
+        np.testing.assert_array_equal(contains, np.zeros(len(contains), dtype=bool))
+
+    def testContainsSkyCoords(self):
+        """Test the sky coord containment code."""
+        self.assertRaises(ValueError,
```
parejkoj, line 828: Please use `RaisesRegex` here, so we can be sure that the exception is coming from where we think it is. (I've just had to deal with some tests that weren't getting the exception the author thought, even though it had the right type! So this is on my mind.)

parejkoj (commented):
Several more comments about the API: I think this will be really useful, but I do want us to think about angle units a bit here.

tests/test_exposure.py
erykoff: I suppose this should be in the `np.arange`.  Done.

tests/test_exposure.py
erykoff: So I spent about 2 minutes trying to come up with a cleverer version of getting the perimeter and decided that it would be more trouble than its worth for a simple test.

tests/test_exposure.py
```diff
@@ -825,10 +826,11 @@ def testGetConvexPolygon(self):
 
     def testContainsSkyCoords(self):
         """Test the sky coord containment code."""
-        self.assertRaises(ValueError,
-                          self.exposureMiOnly.containsSkyCoords,
-                          0.0,
-                          0.0)
+        self.assertRaisesRegex(ValueError,
```
parejkoj, line 829: Probably not worth changing for this simple case, but I always do assertRaises in a context manager so that the call signature is completely obvious:

```
with self.assertRaisesRegex(ValueError, msg):
    self.exposureMiOnly.containsSkyCoords(0, 0)
```

python/lsst/afw/image/exposure/exposureUtils.py
```diff
@@ -47,66 +49,53 @@ def bbox_to_convex_polygon(bbox, wcs, padding=10):
...
-        raise ValueError("Exposure does not have a valid wcs.")
-
-    if degrees:
-        _ra = np.deg2rad(np.atleast_1d(ra)).astype(np.float64)
-        _dec = np.deg2rad(np.atleast_1d(dec)).astype(np.float64)
-    else:
-        _ra = np.atleast_1d(ra).astype(np.float64)
-        _dec = np.atleast_1d(dec).astype(np.float64)
+    _ra = np.atleast_1d(ra.to(units.radian).value).astype(np.float64)
+    _dec = np.atleast_1d(dec.to(units.radian).value).astype(np.float64)
```
parejkoj, line 93: You can do `ra.to_value(units.radian)`, though it looks like that doesn't result in a measureable speedup.

parejkoj (approved):
Thanks Eli, this looks great, and I think the use of astropy angle units here makes for a nice clean interface.

PR #633, opened 2022-04-04
DM-33496: relax rtol to pass test on aarch64

mwittgen:


tests/test_function.py
```diff
@@ -585,16 +585,16 @@ def basic1DPoly(x, params):
                                     f(x), predVal, x, params,
                                     "params constructor")
                 self.assertFloatsAlmostEqual(
-                    f(x), predVal, msg=msg, atol=self.atol, rtol=None)
+                    f(x), predVal, msg=msg, atol=self.atol, rtol=1E-13)
```
erykoff, line 588: Should these be `self.rtol` as with the `self.atol` set in `setUp`?

erykoff (approved):

tests/test_function.py
mwittgen: One could set `self.rtol` and use it. Seems OK to me as this just overrides for these three tests.

PR #634, opened 2022-04-22
DM-31811: Add detection significance to footprint peaks

parejkoj:


include/lsst/afw/detection/Footprint.h
```diff
@@ -329,6 +329,25 @@ class Footprint : public afw::table::io::PersistableFacade<lsst::afw::detection:
...
+
+    /**
+     * Compute the significance of each peak, using the variance image.
+     *
+     * @param variance The variance plane of the image that this footprint is in.
+     * @param polarity +1 if this is detected as a positive footprint, -1 if negative.
+     *
+     * @throws pex::exceptions::NotFoundError Thrown if the the peaks do not have a significance field.
+     */
+    void computePeakSignificance(image::Image<float> const &variance, int polarity);
```
TallJimbo, line 349: Should document that these put the significance values in a peak field named "significance", which must already be part of the Footprint's peak schema.  I'd also be inclined to name them "updatePeakSignificance" instead of "computePeakSignificance", because I think that better captures the modify-in-place behavior, but that's really splitting hairs.

TallJimbo (approved):
Looks good.

include/lsst/afw/detection/Footprint.h
parejkoj: I wasn't sure about the name, and I like `updatePeakSignificance` better. Thanks.

parejkoj:
@TallJimbo : do you mind reviewing the new commit? I believe it fixes the ci_hsc breakage (I'm waiting on a jenkins run to confirm that, but I added a test that demonstrated the problem and ap_verify got past the breakage point in ip_diffim). I'm not sure this is the ideal approach, and I didn't add a check that the two peak schemas being merged are equal, I just took the first non-empty one.

include/lsst/afw/detection/FootprintSet.h
```diff
@@ -67,6 +68,21 @@ class FootprintSet {
...
+     *
+     * @param img Image to search for objects
+     * @param threshold threshold to find objects
+     * @param npixMin minimum number of pixels in an object
+     * @param setPeaks should I set the Peaks list?
+     * @param peakSchema Schema for peak records, even if we don't measure them here.
+     */
+    template <typename ImagePixelT>
+    FootprintSet(image::Image<ImagePixelT> const& img, Threshold const& threshold,
+                 table::Schema const& peakSchema, int const npixMin = 1, bool const setPeaks = true);
```
TallJimbo, line 84: It should also be possible to write this as a defaulting version of the existing similar overload by moving the `peakSchema` argument to the end:
```
template <typename ImagePixelT>
FootprintSet(image::Image<ImagePixelT> const& img, Threshold const& threshold,
             int const npixMin = 1, bool const setPeaks,
             table::Schema const& peakSchema = PeakTable::makeMinimalSchema())
```
In C++ that default value is effectively stored as a `static` variable, and it'll be copied into each call, so (unlike Python) it should be safe despite `Schema` being mutable.

src/detection/FootprintSet.cc
```diff
@@ -333,18 +333,28 @@ FootprintSet mergeFootprintSets(FootprintSet const &lhs,      // the FootprintSe
...
-                                                         * (e.g. if nbit is 2, idFinder IDs 0x5 and 0x6 both contain lhsId = 0) so we get duplicates
-                                                         * of peaks.  This is not too bad, but it's a bit of a pain to make the lists unique again,
-                                                         * and we avoid this by this two-step process.
-                                                         */
+    // Ensure fields added to the input peak schemas are preserved in the output.
+    table::Schema schema;
+    if (lhsFootprints.size() > 0) {
+        schema = lhsFootprints[0]->getPeaks().getSchema();
+    } else if (rhsFootprints.size() > 0) {
+        schema = rhsFootprints[0]->getPeaks().getSchema();
```
TallJimbo, line 341: It'd be pretty easy to add a check to see if the schemas are equal and throw an exception if they are not, and I think it's worth doing, given that not checking leaves the door open to somebody else basically repeating all the debugging you did today.

TallJimbo (approved):

PR #635, opened 2022-05-03
DM-34621: Use image caches to satisfy PSF bbox checks, when possible.

TallJimbo:


arunkannawadi (approved):
Looks good, Jim. 

This kind of ties the result of doComputeBBox as the bbox of the kernelImage. Given that it's a virtual method in this base class, I wonder how it'd hold up if a derived class (incorrectly) implements it as a bounding box of its image. This is of course a hypothetical situation for now.

I also wonder if there'd be any (dis)advantage to just caching the bounding boxes right beside the cached (k)images, instead of getting the cached image first and then getting its bounding box.

TallJimbo:
> Given that it's a virtual method in this base class, I wonder how it'd hold up if a derived class (incorrectly) implements it as a bounding box of its image. This is of course a hypothetical situation for now.

This is true, but I don't think it's avoidable; we just have to rely on derived classes getting it right when they override them.

> I also wonder if there'd be any (dis)advantage to just caching the bounding boxes right beside the cached (k)images, instead of getting the cached image first and then getting its bounding box.

I think this would only pay off if bounding box calculations were both pretty expensive on their own (true at some level for `CoaddPsf`, but I'm not sure how much) _and_ we're likely to have a lot of bounding box requests at a point before we get any PSF image requests at that point (I'm pretty confident this is not true).

PR #638, opened 2022-06-07
DM-35128: Valgrind error in lsst::afw::math::LeastSquares::getDiagnostic

PaulPrice:


src/math/LeastSquares.cc
```diff
@@ -360,6 +360,7 @@ LeastSquares::LeastSquares(Factorization factorization, int dimension) {
             break;
     }
     _impl->factorization = factorization;
+    _impl->whichDiagnostic = factorization;
```
TallJimbo, line 363: Mind instead adding these as arguments to the `Impl` (base class) constructor, and having the derived constructors pass down their own constant?  This is really something the constructor should guarantee.

TallJimbo (approved):

PR #641, opened 2022-06-23
DM-35281: Avoid duplicate peak ID's

fred3m:


tests/test_footprintMergeCatalog.py
```diff
@@ -102,35 +102,49 @@ def setUp(self):
...
-        self.catalog1 = afwTable.SourceCatalog(self.table)
+        # Use a different ID factory for each footprint so that the peak ID's
+        # overlap. This allows us to test ID collisions.
+        idFactory = afwTable.IdFactory.makeSimple()
+        table = afwTable.SourceTable.make(schema, idFactory)
+        self.catalog1 = afwTable.SourceCatalog(table)
         fp1.makeSources(self.catalog1)
 
+        idFactory = afwTable.IdFactory.makeSimple()
+        table = afwTable.SourceTable.make(schema, idFactory)
```
parejkoj, line 117: Since you're re-using the `idFactory` name every time here, it might be clearer to just put the `IdFactor.makeSimple()` inside the `SourceTable.make()` call in each of these cases?

tests/test_footprintMergeCatalog.py
```diff
@@ -102,35 +102,49 @@ def setUp(self):
...
-        del self.table
+
+    def assertUniqueIds(self, catalog):
+        """Ensure that all of the peak IDs in a single parent are unique
+        """
+        for src in catalog:
+            # The peak catalog may be non-contiguous,
+            # so add each peak individually
+            peaks = [peak.getId() for peak in src.getFootprint().peaks]
+            self.assertEqual(len(peaks), len(np.unique(peaks)))
```
parejkoj, line 142: How about `self.assertEqual(sorted(peaks), sorted(set(peaks))`? That would specify exactly which ones don't match.

tests/test_footprintMergeCatalog.py
```diff
@@ -102,35 +102,49 @@ def setUp(self):
...
         fp3.makeSources(self.catalog3)
 
     def tearDown(self):
         del self.catalog1
         del self.catalog2
         del self.catalog3
-        del self.table
+
+    def assertUniqueIds(self, catalog):
+        """Ensure that all of the peak IDs in a single parent are unique
```
parejkoj, line 136: Period at end of sentence.

parejkoj (commented):
A couple of cleanups on the tests, but otherwise this looks good. Good catch on the bug, and I like the tests that demonstrate the problem.

I would suggest merging the test with the C++ commit: they go together.

PR #647, opened 2022-07-29
DM-35457: Fix handling of FITS error messages passed to C++ exceptions 

mwittgen:


src/fits.cc
```diff
@@ -439,7 +439,12 @@ std::string makeErrorMessage(std::string const &fileName, int status, std::strin
     }
     os << "\ncfitsio error stack:\n";
     char cfitsioMsg[FLEN_ERRMSG];
+    // fits_read_errmsg can return a junk string with control characters
     while (fits_read_errmsg(cfitsioMsg) != 0) {
+        cfitsioMsg[FLEN_ERRMSG-1] = char(0); // ensure termination
+        std::size_t len=strlen(cfitsioMsg);
+        for(std::size_t i=0;i<len;i++)
+            if(cfitsioMsg[i] < char(32)) cfitsioMsg[i] = '.';
```
timj, line 447: (I think my previous comment got rebased out).

I think `isprint()` as the test would be more explicit.

mwittgen: Yes, ignore the previous version. isprint is clearly better.

python/lsst/afw/fits/_fits.cc
```diff
@@ -276,8 +276,9 @@ PYBIND11_MODULE(_fits, mod) {
     wrappers.addSignatureDependency("lsst.pex.exceptions");
     wrappers.addSignatureDependency("lsst.daf.base");
     // FIXME: after afw.image pybind wrappers are converted
-    //wrappers.addSignatureDependency("lsst.afw.image"); 
-    wrappers.wrapException<FitsError, lsst::pex::exceptions::IoError>("FitsError", "IoError");
+    //wrappers.addSignatureDependency("lsst.afw.image");
```
timj, line 279: Why do we think it thinks this line changed?

python/lsst/afw/fits/_fits.cc
```diff
@@ -276,8 +276,9 @@ PYBIND11_MODULE(_fits, mod) {
...
+    auto cls = wrappers.wrapException<FitsError, lsst::pex::exceptions::IoError>("FitsError", "IoError");
+    cls.def(py::init<std::string const &>());
```
timj, line 281: What does this do?

timj (approved):
I'm astonished that a control character in an error message can mess up python's exception handling but this fix seems fine.

python/lsst/afw/fits/_fits.cc
mwittgen: extra space previously at the end.

python/lsst/afw/fits/_fits.cc
mwittgen: Adding a ctor for the bound class, otherwise this happens

```
>>> from lsst.afw.fits import FitsError
>>> raise FitsError("test")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/wittgen/lsst/lsst_stack/stack/miniconda3-py38_4.9.2-4.1.0/Linux64/pex_exceptions/g48ccf36440+89c08d0516/python/lsst/pex/exceptions/wrappers.py", line 70, in __init__
    cpp = self.WrappedClass(message, *args, **kwds)
TypeError: lsst.afw.fits._fits.FitsError: No constructor defined!

```

PR #650, opened 2022-08-15
DM-35895: Use band for the dataId not filter

timj:
Gen3 uses band for coadds. If people want to retrieve by
physical filter it would likely need a change in API
or much more cleverness in trying to inspect the dataset type.

### DIFF
--- a/doc/lsst.afw.multiband/multiband.rst
+++ b/doc/lsst.afw.multiband/multiband.rst
@@ -480,17 +480,14 @@ to load an exposure from a file:
 .. code-block:: python
 
     from lsst.afw.image import MultibandExposure
-    from lsst.daf.persistence import Butler
+    from lsst.daf.butler import Butler
 
-    # This is an example dataset on lsstdev which may be out of date,
-    # replace with a local dataset
-    DATA_DIR = "/datasets/hsc/repo/rerun/RC/w_2018_22/DM-14547"
-    butler = Butler(inputs=DATA_DIR)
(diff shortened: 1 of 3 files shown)


doc/lsst.afw.multiband/multiband.rst
```diff
@@ -480,17 +480,14 @@ to load an exposure from a file:
...
 
     from lsst.afw.image import MultibandExposure
-    from lsst.daf.persistence import Butler
+    from lsst.daf.butler import Butler
 
-    # This is an example dataset on lsstdev which may be out of date,
-    # replace with a local dataset
-    DATA_DIR = "/datasets/hsc/repo/rerun/RC/w_2018_22/DM-14547"
-    butler = Butler(inputs=DATA_DIR)
+    DATA_DIR = "/repo/main"
```
erykoff, line 485: I don't think that `/repo/main` works anywhere ...

erykoff (approved):
Minor comment on the doc; otherwise if these aren't being used/tested, should we also deprecate them?

doc/lsst.afw.multiband/multiband.rst
timj: It's meant to be an example. Also, K-T was pondering defining `/repo/main` as a butler alias at SLAC.

timj: Hmm. Nothing is going to work without a collections parameter either here or in the `fromButler` method.

PR #657, opened 2022-09-13
DM-36207: Reset the Fits object status to 0 if we are throwing an error

weatherhead99:
Fix and test case added for https://jira.lsstcorp.org/browse/DM-36207.

### DIFF
--- a/include/lsst/afw/fits.h
+++ b/include/lsst/afw/fits.h
@@ -108,8 +108,12 @@ std::string makeLimitedFitsHeader(lsst::daf::base::PropertySet const& metadata,
 /**
  *  Throw a FitsError exception if the status of the given Fits object is nonzero.
  */
-#define LSST_FITS_CHECK_STATUS(fitsObj, ...) \
-    if ((fitsObj).status != 0) throw LSST_FITS_EXCEPT(lsst::afw::fits::FitsError, fitsObj, __VA_ARGS__)
+#define LSST_FITS_CHECK_STATUS(fitsObj, ...)                                              \
+    if ((fitsObj).status != 0) {                                                          \
+        auto except = LSST_FITS_EXCEPT(lsst::afw::fits::FitsError, fitsObj, __VA_ARGS__); \
+        (fitsObj).status = 0;                                                             \
+        throw except;                                                                     \
+    }
 
 /// Return the cfitsio integer BITPIX code for the given data type.
 template <typename T>
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -128,6 +128,20 @@ class FitsTestCase(lsst.utils.tests.TestCase):
         metadata = self.writeAndRead(header)
         self.assertEqual(metadata.getArray("FOO"), [None, None])
 
+    def test_ticket_dm_36207(self):
+        # test whether moving to invalid HDU and then moving back throws or not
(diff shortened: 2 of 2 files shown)


tests/test_fits.py
```diff
@@ -128,6 +128,24 @@ def testUndefinedVector(self):
...
+            fts.setHdu(5)
+
+        exc = None
+        try:
+            fts.setHdu(0)
+        except lsst.afw.fits.FitsError as err:
+            exc = err
+
+        # if the bug is still there, this will have raised and exc will not be None
+        self.assertEqual(exc, None)
```
timj, line 147: `self.assertIsNone(exc)` is preferred.

tests/test_fits.py
```diff
@@ -128,6 +128,24 @@ def testUndefinedVector(self):
...
+        testfile = os.path.join(testPath, "data", "parent.fits")
+        fts = lsst.afw.fits.Fits(testfile, "r")
+
+        # parent.fits has 2 HDUs.
+        with self.assertRaises(lsst.afw.fits.FitsError):
+            fts.setHdu(5)
+
+        exc = None
+        try:
+            fts.setHdu(0)
```
timj, line 142: I think rather than proving no exception was raised, I'd rather prove we did succeed in moving to the first HDU by reading it and proving that we got a header from the right HDU.

include/lsst/afw/fits.h
```diff
@@ -109,7 +109,8 @@ std::string makeLimitedFitsHeader(lsst::daf::base::PropertySet const& metadata,
  *  Throw a FitsError exception if the status of the given Fits object is nonzero.
  */
 #define LSST_FITS_CHECK_STATUS(fitsObj, ...) \
-    if ((fitsObj).status != 0) throw LSST_FITS_EXCEPT(lsst::afw::fits::FitsError, fitsObj, __VA_ARGS__)
+  if ((fitsObj).status != 0) { auto except =  LSST_FITS_EXCEPT(lsst::afw::fits::FitsError, fitsObj, __VA_ARGS__); \
```
timj, line 112: Is there a reason where there are two spaces after the `=` here and two spaces before the `=` below?

timj (approved):
Thanks for doing this. I have some minor comments. Please also make sure jenkins passes and also include ci_hsc_gen3 in the Jenkins build.

include/lsst/afw/fits.h
weatherhead99: the reason is that clang-format for some reason borked on this on my machine. Will fix manually

weatherhead99:
FYI @timj  I'm pretty sure for some reason I don't have permission to run it in Jenkins, not sure what the glitch is there...

erykoff:
Well, we have to figure out the jenkins thing.  You should also change the title of this PR to `DM-36207: Reset the Fits object status to 0 if we are throwing an error`.  See https://developer.lsst.io/work/flow.html#make-a-pull-request

PaulPrice:
On Slack you considered multiple approaches to fixing this. Could I encourage you to capture the logic from those considerations in the commit message so it's not (effectively) lost?

weatherhead99:
ok I believe all comments have been accounted for and I have squashed this to one commit with a hopefully informative message. Only thing remaining is the Jenkins run I think

PR #658, opened 2022-09-19
DM-36277: Check header is defined before using it for VisitInfo

timj:
Header existence checks are not sufficient because the C++ reader in PropertySet will refuse to return a string for an undefined value.

### DIFF
--- a/src/image/VisitInfo.cc
+++ b/src/image/VisitInfo.cc
@@ -70,11 +70,11 @@ auto const nan = std::numeric_limits<double>::quiet_NaN();
  * @param[in] metadata  metadata to get
  * @param[in] key  key name; the associated value must be of type double if the key exists
  * @returns value of metadata for the specified key, as a double,
- *   with a value of nan if the key is not present
+ *   with a value of nan if the key is not present or not defined.
  */
 double getDouble(daf::base::PropertySet const& metadata, std::string const& key) {
     try {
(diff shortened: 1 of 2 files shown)


src/image/VisitInfo.cc
```diff
@@ -70,11 +70,11 @@ auto const nan = std::numeric_limits<double>::quiet_NaN();
  * @param[in] metadata  metadata to get
  * @param[in] key  key name; the associated value must be of type double if the key exists
  * @returns value of metadata for the specified key, as a double,
- *   with a value of nan if the key is not present
+ *   with a value of nan if the key is not present or not defined.
  */
 double getDouble(daf::base::PropertySet const& metadata, std::string const& key) {
     try {
-        return metadata.exists(key) ? metadata.getAsDouble(key) : nan;
+        return metadata.exists(key) && !metadata.isUndefined(key) ? metadata.getAsDouble(key) : nan;
```
timj, line 77: I'm fine with removing this one since it's in a TypeError catch block anyhow. I added it for consistency with all the other checks.

parejkoj: I don't think it hurts anything to have it here.

parejkoj (approved):
Thanks for the quick fix and for the test to confirm it. Do you want write a ticket to convert VisitInfo to use HIERARCH for header persistence?

PR #659, opened 2022-09-27
tickets/DM-36230: Add new polygon contains interface

natelust:
Allow checking if a polygon contains various objects like vectors of points, and a point by passing an x, y
pair instead of a Point object. Expose this to python with an auto-vectorized interface to allow passing of numpy arrays of points.


erykoff:
Thank you for doing this!

czwa (approved):

PR #660, opened 2022-10-05
DM-27811: Remove deprecated getFilterLabel interfaces.

erykoff:


kfindeisen (approved):
Looks good, I don't see anything missing (or is that left behind?).

PR #664, opened 2022-10-27
DM-36762: Fix broken Spanset.asArray method

fred3m:


tests/test_spanSets.py
```diff
@@ -440,6 +440,24 @@ def testSpanIteration(self):
         self.assertEqual(len(span), len(points))
         self.assertEqual(points, [lsst.geom.Point2I(x, 4) for x in range(3, 9)])
 
+    def testAsArray(self):
+        spans = afwGeom.SpanSet(lsst.geom.Box2I(lsst.geom.Point2I(2, 3),
```
leeskelvin, line 444: Black assures me that this would make more sense written this way:
```
spans = afwGeom.SpanSet(
    lsst.geom.Box2I(lsst.geom.Point2I(2, 3), lsst.geom.Point2I(6, 7))
)
```
I think I agree, but I don't feel super strongly about it!

fred3m: Sure, I changed it. I don't like it written the way it was either, I just copied it from one of the tests above and didn't bother to reformat it.

PR #668, opened 2022-11-11
DM-36961: Increase test tolerance for osx-arm64.

erykoff:


ktlim:
Shouldn't this be "increase" tolerance? (from epsilon to 1e-15)

arunkannawadi (approved):

PR #669, opened 2022-11-23
DM-37085: Add copyWith method to VisitInfo

fred3m:


tests/test_visitInfo.py
```diff
@@ -228,6 +228,52 @@ def testValueConstructor_data2(self):
...
+            "hasSimulatedContent",
+        ]
+
+        kwargs1 = {k: getattr(visitInfo2, k) for k in updateFields1}
+        kwargs2 = {k: getattr(visitInfo2, k) for k in updateFields2}
+
+        newVisit1 = visitInfo1.copyWith(**kwargs1)
+        newVisit2 = visitInfo1.copyWith(**kwargs2)
+
+        print(newVisit1)
```
mfisherlevine, line 267: I assume this was for testing while writing and was supposed to go?

python/lsst/afw/image/_visitInfo.py
```diff
@@ -22,12 +22,105 @@
...
+            scienceProgram = self.scienceProgram
+        if observationReason is None:
+            observationReason = self.observationReason
+        if object is None:
+            object = self.object
+        if hasSimulatedContent is None:
+            hasSimulatedContent = self.hasSimulatedContent
+
+        return VisitInfo(
+            exposureId=0,
```
mfisherlevine, line 100: Shouldn't that be `exposureId=exposureId`? And doesn't that then need an accompanying block above like all the others? Or am I missing something about that being special somehow?

mfisherlevine: Oh, or are you saying that because it's a copy, it no longer corresponds to the original expId, and therefore must be zeroed out? If so, might be worth setting a more obviously suspicious value like `-1`, and either way, a comment to say that's why it's hard-coded to zero.

python/lsst/afw/image/_visitInfo.py
```diff
@@ -22,12 +22,105 @@
...
+from lsst.utils import continueClass
 from ._imageLib import VisitInfo
 
 
 __all__ = []
 
 
+@continueClass
+class VisitInfo:  # noqa: F811
+    def copyWith(
```
mfisherlevine, line 34: General comment: this doesn't quite seem to cover all the items. I think `[boresightHourAngle, boresightParAngle, localEra]` are missing. And then there's `isPersistable`, but I wonder where that ever comes from and if it makes sense to set... That feels like a question for @timj perhaps?

python/lsst/afw/image/_visitInfo.py
fred3m: Oh, this was me being lazy and just trying to make the tests work in time for the end of the day. The issue is that exposureId is deprecated (see https://github.com/lsst/afw/blob/main/python/lsst/afw/image/_visitInfo.py#L31-L35), so a newly generated exposureId is None. However the CPP API won't let a new `VisitInfo` be initialized with `None`, it must be zero. But I can put in a check for `self.exposureId is None` and set it to zero, that way this function will still be compatible with the deprecated API until it is removed.

python/lsst/afw/image/_visitInfo.py
fred3m: Oh, I guess I did miss those. I'll fix and update the PR.

fred3m: Those must all be derived quantities. I don't see any of them in the constructor for `VisitInfo`: https://github.com/lsst/afw/blob/main/include/lsst/afw/image/VisitInfo.h#L100-L130

python/lsst/afw/image/_visitInfo.py
mfisherlevine: Ah, gotcha OK, thanks.

python/lsst/afw/image/_visitInfo.py
mfisherlevine: Hmmm, I wonder how they get in there then, given that the object is immutable. Presumably as part of the init, so these will show up as intended after all?

fred3m: They aren't stored anywhere. Each time they are called they are calculated in C++:
https://github.com/lsst/afw/blob/main/src/image/VisitInfo.cc#L633
https://github.com/lsst/afw/blob/main/src/image/VisitInfo.cc#L635
https://github.com/lsst/afw/blob/main/src/image/VisitInfo.cc#L637-L649

They just look like attributes because they are wrapped as properties by pybind11: https://github.com/lsst/afw/blob/main/python/lsst/afw/image/_visitInfo.cc

mfisherlevine: Ah, interesting, thanks!

PR #670, opened 2022-12-19
DM-37412: move some exposure summary stats code from pipe_tasks

TallJimbo:


python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -20,76 +20,102 @@
...
-from typing import List
+import dataclasses
+from typing import TYPE_CHECKING
 import yaml
 import warnings
 
 from ..typehandling import Storable, StorableHelperFactory
 
+if TYPE_CHECKING:
+    from ..table import BaseRecord, Schema
```
laurenam, line 31: I have no doubt you've got this right, but I don't know what determines the state of `TYPE_CHECKING` and am wondering what happens if it's `False` and these imports don't happen?

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -141,7 +143,160 @@ def _read(bytes):
...
         return ExposureSummaryStats(**yamlDict)
+
+    @classmethod
+    def update_schema(cls, schema: Schema) -> None:
+        """Update an schema to includes for all summary statistic fields.
+
+        Parameters
+        -------
+        schema : `lsst.afw.table.Schema`
+            Schema to add which fields will be added.
```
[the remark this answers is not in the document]
TallJimbo: I'll drop the return value instead, as that's what I meant to do.

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -141,7 +143,160 @@ def _read(bytes):
...
+            doc="Scatter (via MAD) of size residual scaled by median size squared",
+        )
+        return schema
+
+    def update_record(self, record: BaseRecord) -> None:
+        """Write summary-statistic columns into a record.
+
+        Parameters
+        ----------
+        record : `BaseRecord`
```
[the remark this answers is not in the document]
TallJimbo: I think the mistake was in making the other schema one return something.  Yes, these could return the thing they're passed in, but I think that implies that the user should use the return value for something and that's not the case here.

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -141,7 +143,160 @@ def _read(bytes):
...
+            else:
+                record[field.name] = value
+
+    @classmethod
+    def from_record(cls, record: BaseRecord) -> None:
+        """Read summary-statistic columns from a record into ``self``.
+
+        Parameters
+        ----------
+        record : `BaseRecord`
```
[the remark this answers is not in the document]
TallJimbo: Oops, this one does have a return value; will add that doc section.  Got this comment and the one on `update_record` confused at first.

tests/test_exposure.py
```diff
@@ -1135,6 +1136,34 @@ def testExposureSummaryExtraComponents(self):
         self.assertEqual(summaryStats.ra, testDict['ra'])
         self.assertEqual(summaryStats.decl, testDict['decl'])
 
+    def testExposureSummarySchema(self):
+        """Test that we can make a schema for an exposure summary and populate
+        records with that schema.
+        """
+        schema = afwTable.Schema()
+        afwImage.ExposureSummaryStats.update_schema(schema)
+        self.maxDiff = None
```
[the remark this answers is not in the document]
TallJimbo: This makes `assertEquals` and the like actually show you what differs even when it's long.  It's something I added here while debugging a test failure originally, and I hadn't intended to leave it, but I now I wonder if I should, since I always find myself adding it when a test does fail and I need to investigate.

python/lsst/afw/image/_exposureSummaryStats.py
TallJimbo: This is only set when running mypy or something similar to do type-checking, so in this case (i.e. in afw) that amounts to "just inside some people's editors", and certainly never in production code; you'll note that these symbols are never used anywhere other than type annotations, and that means at runtime they aren't actually needed (and when they are, I've imported them inside function bodies).  That's all because afw subpackages are mess of circular dependencies and I wasn't at all sure that importing `lsst.afw.table` at module scope normally was safe here.

laurenam: Thanks for the explanation!

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -117,7 +143,164 @@ def _read(bytes):
...
+        ----------
+        record : `lsst.afw.table.BaseRecord`
+            Record to read from.  This is expected to frequently be an
+            `ExposureRecord` instance (with higher-level code adding other
+            columns and objects), but this method can work with any record
+            type, ignoring any attributes or columns it doesn't recognize.
+
+        Returns
+        -------
+        summary : `ExposureSummaryStats`
```
laurenam, line 294: `lsst.afw.image.ExposureSummaryStats`?

laurenam (approved):

python/lsst/afw/image/_exposureSummaryStats.py
TallJimbo: This one should be fine unqualified, as it's in the same module that defines this type.

PR #671, opened 2023-01-02
DM-37411: Add visit-level PSF model robustness metrics

laurenam:


python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -117,6 +117,12 @@ class ExposureSummaryStats(Storable):
     psfStarScaledDeltaSizeScatter: float = float('nan')
     """Psf stars MAD size scatter scaled by psfSize**2."""
 
+    psfTraceDelta: float = float('nan')
+    """Delta (max - min) of the model psf trace values evaluated on a grid of unmasked pixels (pixels)."""
```
TallJimbo, line 121: Let's say "trace radius" here and in the schema definition below instead of just "tract".  That's because "trace" refers to the trace of the `[[ixx, ixy], [ixy, iyy]]` matrix (`ixx + iyy`), while the trace radius is `\sqrt{(ixx + iyy)/2}` - i.e. derived from that, but not the same thing, and people do use the actual trace (often written as `T`, with units of `pixels^2` or `arcsec^2`) as a measure of PSF size.

TallJimbo (approved):

PR #678, opened 2023-01-27
DM-37534: Ensure that pex_exceptions is imported for afw.fits

timj:
Without this it is not possible to "import lsst.afw.fits" because the FitsError is not translated to python. Tests passed in afw because other parts of afw were importing pex_exceptions.

### DIFF
--- a/python/lsst/afw/fits/_fits.cc
+++ b/python/lsst/afw/fits/_fits.cc
@@ -283,7 +283,7 @@ void declareFitsModule(lsst::utils::python::WrapperCollection &wrappers) {
 }  // namespace
 PYBIND11_MODULE(_fits, mod) {
     lsst::utils::python::WrapperCollection wrappers(mod, "lsst.afw.fits");
-    wrappers.addSignatureDependency("lsst.pex.exceptions");
+    wrappers.addInheritanceDependency("lsst.pex.exceptions");
     wrappers.addSignatureDependency("lsst.daf.base");
     // FIXME: after afw.image pybind wrappers are converted
     //wrappers.addSignatureDependency("lsst.afw.image");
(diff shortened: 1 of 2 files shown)


timj:
This is the error without this change:

```
Traceback:
../../miniconda/envs/lsst-scipipe-5.0.0/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
python/lsst/obs/base/__init__.py:22: in <module>
    from ._fitsRawFormatterBase import *
python/lsst/obs/base/_fitsRawFormatterBase.py:27: in <module>
    import lsst.afw.fits
../../stack/lsst-scipipe-5.0.0/Darwin/afw/g87b8406481+bd7147d3c6/python/lsst/afw/fits/__init__.py:22: in <module>
    from ._fits import *
E   ImportError: generic_type: type "FitsError" referenced unknown base type "lsst::pex::exceptions::IoError"
```

@mwittgen I was under the impression that this line https://github.com/lsst/afw/blob/main/python/lsst/afw/fits/_fits.cc#L286 means that the python import was not needed. Seems I was wrong.

mwittgen (approved):

timj:
Looking at the code, we currently have explicit mentions of `lsst.pex.exceptions` in 5 places so we don't seem to attempt to always ensure that it's imported when we have defined a C++ exception. In `lsst.afw.fits` we have:

```c++
wrappers.addSignatureDependency("lsst.pex.exceptions");
```

and that doesn't seem to have the effect of importing pex_exceptions. In `meas_base` we have:

```c++
py::module::import("lsst.pex.exceptions");
```
which is explicit about it.

mwittgen:
https://github.com/lsst/afw/blob/87b8406481d008bd04dea2cdb41b6a1c746e3c47/python/lsst/afw/fits/_fits.cc#L286
This should be 
`wrappers.addInheritanceDependency("lsst.pex.exceptions"); `

The reason is
https://github.com/lsst/afw/blob/87b8406481d008bd04dea2cdb41b6a1c746e3c47/include/lsst/afw/fits.h#L36

the `fitsError` class, etc inheriting from ` lsst::pex::exceptions`  classes

timj:
@mwittgen changing does fix the problem. I also fixed afw.geom since that was also trying to export a wrapped exception.  Please take another look.

mwittgen (approved):
Looks good and this is correcting an earlier mistake of mine.

PR #682, opened 2023-03-28
DM-38457: Use skyToPixelArray et al in SIP test

timj:


tests/test_sipApproximation.py
```diff
@@ -561,8 +561,15 @@ def run(original, width, height, order, spacing=20, skyTol=1.0e-4, pixTol=1.0e-8
             # Evaluate performance
             points = [Point2D(xx, yy) for xx, yy in itertools.product(range(0, width, spacing),
                                                                       range(0, height, spacing))]
```
erykoff, line 563: Should this be something like
```
x_points, y_points = np.meshgrid(range(0, width, spacing), range(0, height, spacing))
x_points = x_points.ravel()
y_points = y_points.ravel()
```
And I guess we still need the spherepoint conversion because of the `getTangentPlaneOffset` call.

tests/test_sipApproximation.py
```diff
@@ -577,8 +584,8 @@ def run(original, width, height, order, spacing=20, skyTol=1.0e-4, pixTol=1.0e-8
             # Not comparing to absolute round-trip, but relative to the
             # original WCS round-trip. This is an important distinction: we're
             # recreating the original WCS, along with all its flaws.
-            roundTrip = np.array([wcs.skyToPixel(wcs.pixelToSky(pp)) for pp in points])
-            roundTrip -= np.array([original.skyToPixel(original.pixelToSky(pp)) for pp in points])
+            roundTrip = np.array(wcs.skyToPixelArray(*wcs_sky))
```
erykoff, line 587: Do we need to convert this to an array?  Oh, it's because we get an x, y tuple; I don't know if `vstack` would be clearer.

erykoff (approved):
Looks good; minor comments but not really necessary.  Shame about needing the SpherePoint.

tests/test_sipApproximation.py
timj: Right, we get a tuple of two ndarray back so I convert them again. I can change to vstack.

timj: I switched to vstack.

tests/test_sipApproximation.py
timj: I changed to meshgrid. The number of elements is the same but they come out in a different order -- that won't change the test.

python/lsst/afw/cameraGeom/testUtils.py
```diff
@@ -419,20 +419,29 @@ def makeTestRepositoryItems(self, isLsstLike=False):
...
             fromPoint = lsst.geom.Point2D(x, y)
-            res1 = func1(fromPoint)
-            res2 = func2(fromPoint)
-            self.assertPairsAlmostEqual(res1, res2)
+            points.append(fromPoint)
+
+    vres1 = func1(points)
+    vres2 = func2(points)
+    for res1, res2 in zip(vres1, vres2):
+        self.assertPairsAlmostEqual(res1, res2)
```
timj, line 444: @erykoff I tried using `self.assertPairListsAlmostEqual(vres1, vres2)` but it ends up slowing things down by half a second for each test. It seems odd that `assertPairsAlmostEqual` in a loop is faster than `assertPairListsAlmostEqual`.

python/lsst/afw/cameraGeom/testUtils.py
```diff
@@ -419,20 +419,29 @@ def makeTestRepositoryItems(self, isLsstLike=False):
 
 @inTestCase
 def compare2DFunctions(self, func1, func2, minVal=-10, maxVal=None, nVal=5):
```
timj, line 421: I considered leaving this function as it was and adding a new `compare2DFunctionsVector` but in lsst_distrib everything that calls this function is calling it with funcs that are vectorized so everything worked anyhow. @ktlim if you are worried about compatibility I can leave this unchanged and add the new method.

python/lsst/afw/cameraGeom/testUtils.py
erykoff: That is strange that it's slower, but looking at the code for `assertPairListsAlmostEqual` https://github.com/lsst/geom/blob/f0d0fc72e121e9654f877923b359f832be524fe5/python/lsst/geom/testUtils.py#L141 I can see why it isn't faster since it just loops over all the pairs anyway.

python/lsst/afw/cameraGeom/testUtils.py
erykoff: If it passes all the tests I'm okay with it.

erykoff (approved):

python/lsst/afw/cameraGeom/testUtils.py
timj: Maybe I should use the vectored one and hope that someday the vectored one becomes faster?

erykoff: It's not vectored ... and I wouldn't hold my breath.

timj: (I meant vectored in the sense that it takes everything in one call and so could be vectorized internally in the future whereas me looping over each point is guaranteed never to become vectorized)

PR #683, opened 2023-03-28
DM-38463: Skip DS9 show tests if XPA reports that no DS9 is available

timj:


tests/test_ds9.py
```diff
@@ -93,7 +94,13 @@ def testZoomPan(self):
     @unittest.skipUnless(ds9, "You must setup display.ds9 to run this test")
     def testStackingOrder(self):
         """ Un-iconise and raise the display to the top of the stacking order if appropriate"""
-        ds9.show()
+        try:
+            ds9.show()
+        except lsst.display.ds9.Ds9Error as e:
```
timj, line 99: I'd be happy to make this a generic `Exception` given that I'm looking for an explicit string.

PR #684, opened 2023-04-06
DM-38602: Include failing VisitInfo serialization version in error message

parejkoj:


src/image/VisitInfo.cc
```diff
@@ -320,7 +320,9 @@ class VisitInfoFactory : public table::io::PersistableFactory {
         int version = getVersion(record);
         if (version > SERIALIZATION_VERSION) {
             throw LSST_EXCEPT(pex::exceptions::TypeError, "Cannot read VisitInfo FITS version > " +
-                                                                  std::to_string(SERIALIZATION_VERSION));
+                                                                  std::to_string(SERIALIZATION_VERSION) +
+                                                                  ", found version " +
+                                                                  std::to_string(version) + " in file.");
```
arunkannawadi, line 325: Ending the log message in "in file. " is unnecessary and potentially confusing. This could lead a user to expect a path to appear after that, or think that the path did not appear for whatever reason.

tests/test_visitInfo.py
```diff
@@ -360,6 +360,11 @@ def testPersistenceVersions(self):
         self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-3.fits"), 3)
         self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-4.fits"), 4)
 
+        # Check that reading a newer file raises a useful exception.
+        with self.assertRaisesRegex(TypeError,
+                                    r"Cannot read VisitInfo FITS version .* found version 999999"):
```
arunkannawadi, line 365:
```suggestion
                                    r"Cannot read VisitInfo FITS version > .* found version 999999"):
```

arunkannawadi: It'd be good to include all parts of the string that don't vary in here, and leave only the variable version number for regex. I am not sure where the comma should go before the "found version", but I'll leave it to you whether to include it or not.

arunkannawadi (approved):

rearmstr (approved):

PR #687, opened 2023-05-05
DM-39053: Add WCS metadata stripping code that doesn't create a WCS.

isullivan:


src/geom/detail/frameSetUtils.cc
```diff
@@ -131,6 +131,51 @@ std::shared_ptr<ast::FrameSet> readFitsWcs(daf::base::PropertySet& metadata, boo
...
+
+    auto channel = getFitsChanFromPropertyList(metadata, excludeNames,
+                                               "Encoding=FITS-WCS, IWC=1, SipReplace=0, ReportLevel=3");
+    auto const initialNames = setFromVector(channel.getAllCardNames());
+
+    std::shared_ptr<ast::Object> obj;
+    try {
+        obj = channel.read();
+    } catch (std::runtime_error const&) {
+        throw LSST_EXCEPT(pex::exceptions::TypeError, "The metadata does not describe an AST object");
```
erykoff, line 158: I think this should just return and no throw.

tests/test_skyWcs.py
```diff
@@ -694,6 +695,10 @@ def testBasics(self):
         makeSkyWcs(self.metadata, strip=True)
         self.assertEqual(len(self.metadata.names(False)), 0)
 
+    def testBasicsStrip(self):
+        stripWcsMetadata(self.metadata)
+        self.assertEqual(len(self.metadata.names(False)), 0)
```
erykoff, line 700: We should add a test here to strip again and make sure it returns the same thing.

PR #689, opened 2023-05-05
DM-36403: Implement methd to extract flux from an image

fred3m:


python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+
+import numpy as np
+
+from lsst.utils import continueClass
+
+from lsst.geom import Point2I
+from ._detection import Footprint
+
+
+__all__ = []  # import this module only for its side effects
```
parejkoj, line 33: This goes above the imports:

https://developer.lsst.io/python/style.html#standard-code-order-should-be-followed

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@continueClass
+class Footprint:  # noqa: F811
+    def extractFluxFromArray(self, image: np.ndarray, xy0: Point2I) -> float:
+        """
+
```
parejkoj, line 40: no docstring?

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+        xy0:
+            The origin of the image array.
+
+        Returns
+        flux:
+            The flux from the image in pixels contained in the footprint
+        """
+        return self.spans.flatten(image, xy0).sum()
+
+    def extractFluxFromImage(self, image: "Image") -> float:
```
parejkoj, line 54: I don't know a lot about mypy, but I think this?

```suggestion
    def extractFluxFromImage(self, image: lsst.afw.image.Image) -> float:
```

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+
+@continueClass
+class Footprint:  # noqa: F811
+    def extractFluxFromArray(self, image: np.ndarray, xy0: Point2I) -> float:
+        """
+
+        Parameters
+        ----------
+        image:
+            The array containing the image pixels to extract
```
parejkoj, line 44: You need types on all these parameters and returns, spaces before the colon, periods at the end of sentences, and dimensions for the array.

I think we also generally don't say "The X", just "X", but I don't know that we have official guidance on this.

https://developer.lsst.io/python/numpydoc.html#array-types

```suggestion
        image : `numpy.ndarray`, (N, 2)
            Image array containing the pixels to sum over.
```

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+        xy0:
+            The origin of the image array.
+
+        Returns
+        flux:
+            The flux from the image in pixels contained in the footprint
```
parejkoj, line 50:
```suggestion
        Returns
        -------
        flux: `float`
            Flux from the image pixels contained in the footprint.
```

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Parameters
+        ----------
+        image:
+            The image to extract.
+
+        Returns
+        -------
+        flux:
+            The flux from the image in the pixels contained in the footprint.
```
parejkoj, line 66:
```suggestion
            The flux from the image pixels contained in the footprint.
```

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        return self.extractFluxFromArray(image.array, image.getBBox().getMin())
```
parejkoj, line 68: Instead of the bbox min, should this be `image.getXY0()`?

tests/test_footprint1.py
```diff
@@ -1016,6 +1016,27 @@ def testEdge(self):
...
+            np.arange(100, dtype=np.int32).reshape(10, 10),
+            xy0=lsst.geom.Point2I(23, 25),
+            dtype="I"
+        )
+        radius = 3
+        spans = afwGeom.SpanSet.fromShape(radius, afwGeom.Stencil.CIRCLE, offset=(27, 30))
+        footprint = afwDetect.Footprint(spans)
+
+        # The extracted footprint should be the same as the product of the
+        # spans and the overlapped bow with the image
```
parejkoj, line 1031: "overlapped bow"?

tests/test_footprint1.py
```diff
@@ -1016,6 +1016,27 @@ def testEdge(self):
...
+        truth = spans.asArray() * image.array[2:9, 1:8]
+        self.assertEqual(footprint.extractFluxFromImage(image), np.sum(truth))
+        # Test hard coded value
+        self.assertEqual(footprint.extractFluxFromImage(image), 1566)
```
parejkoj, line 1035: "hard coded value"? If you're just checking that `np.sum(truth)` is the number you expect it should be, I'd make that a separate assert above the actual tests, since it's a precondition on the test being sensible: `assert np.sum(truth) == 1566`

tests/test_footprint1.py
```diff
@@ -1016,6 +1016,27 @@ def testEdge(self):
...
+
+        # Test the array method with an offset
+        self.assertEqual(footprint.extractFluxFromArray(image.array-3, image.getBBox().getMin()), 1566-29*3)
```
parejkoj, line 1038: Why 29*3? Probably worth a comment, since it's not obvious to me why that's the right value for this offset.

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from lsst.afw.image import Image
```
parejkoj, line 23: Do you actually need the `if` here? Isn't `TYPE_CHECKING` always set at this point?

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+from lsst.geom import Point2I
+from ._detection import Footprint
+
+
+__all__ = []  # import this module only for its side effects
+
+
+@continueClass
+class Footprint:  # noqa: F811
+    def extractFluxFromArray(self, image: np.ndarray, xy0: Point2I) -> float:
```
parejkoj, line 38: I think these should be `computeFluxFromX`, not `extract`, since there's a `sum` here.

parejkoj (changes requested):
Sorry this took me so long. Several docs/formatting comments/questions, and some test tweaks.

python/lsst/afw/detection/_footprintContinued.py
fred3m: The latest version of the documenteer no longer requires us to put types on parameters if we use type hints. Which is preferred, since the type hints are checked while types in parameters more easily bitrot.

python/lsst/afw/detection/_footprintContinued.py
fred3m: This is a problem with a circular dependency that is introduced by type checking. See https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/ .

python/lsst/afw/detection/_footprintContinued.py
fred3m: See previous comment about circular import introduced by the type hints.

fred3m:
I think that I've either implemented or responded to all of the reviewer comments and suggestions.

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+@continueClass
+class Footprint:  # noqa: F811
+    def computeFluxFromArray(self, image: np.ndarray, xy0: Point2I) -> float:
+        """Calculate the total flux in the region of an image array
+        contained in this Footprint.
+
+        Parameters
+        ----------
+        image:
+            Array containing the pixels to extract.
```
parejkoj, line 44: You still say extract here, but this is now "compute". Maybe "2d array of pixels containing the footprint."?

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -0,0 +1,68 @@
...
+        xy0:
+            The origin of the image array.
+
+        Returns
+        flux:
```
parejkoj, line 49:
```suggestion
        Returns
        -------
        flux:
```

tests/test_footprint1.py
```diff
@@ -1016,6 +1016,29 @@ def testEdge(self):
...
+        image = afwImage.Image(
+            np.arange(100, dtype=np.int32).reshape(10, 10),
+            xy0=lsst.geom.Point2I(23, 25),
+            dtype="I"
+        )
+        radius = 3
+        spans = afwGeom.SpanSet.fromShape(radius, afwGeom.Stencil.CIRCLE, offset=(27, 30))
+        footprint = afwDetect.Footprint(spans)
+
+        # The extracted footprint should be the same as the product of the
```
parejkoj, line 1030: "extracted footprint"->"computed flux"?

parejkoj (approved):
Thanks, a handful of other suggestions.

PR #691, opened 2023-06-19
DM-39700:  Remove inheritance from unary/binary_function

mwittgen:


erykoff (changes requested):
There are more places that `std::unary_function` and `std::binary_function` are used in afw, particularly `lsstGil.h` and `integrate.h` that I can see with a search.  (But these bits of code may be unused, we should clean them up or remove them.)

erykoff:
Still doesn't work; there are uses of `binary_function` in at least `Pixel.h` that need to be cleaned up, not just the `unary_function`.

mwittgen:
Wasn't pushed. Pushed now

erykoff (approved):
This works on macos with 7.0.0dev and libcxx=16.

PR #694, opened 2023-06-26
DM-27275: Add enable if clauses to operator templates, clean up code

mwittgen:


include/lsst/afw/image/Pixel.h
```diff
@@ -570,24 +532,15 @@ ExprT1 operator+=(ExprT1& e1, ExprT2 e2) {
 //
 // The choice is made on the basis of std::is_arithmetic
```
kfindeisen, line 533: I don't think this comment is true anymore... `valid_operands` does not use `is_arithmetic`.

kfindeisen (approved):
Aside from some typos, no objections to the changes themselves. However, the commit blends together two or three distinct changes, making it hard to follow; please use atomic commits and explain what each does.

PR #697, opened 2023-07-05
DM-32138: Remove exposureId from VisitInfo

parejkoj:


src/image/ExposureInfo.cc
```diff
@@ -328,6 +328,7 @@ ExposureInfo::FitsWriteData ExposureInfo::_startWriteFits(lsst::geom::Point2I co
     }
 
     // So long as VisitInfo also writes exposureId (until DM-32138), let ours take precedence.
+    // TODO: what about this one?
```
kfindeisen, line 331: Remove the comments, and possibly move the block higher up? If the `setVisitInfoMetadata` call above no longer sets `EXPID`, then there's no conflict and no reason to care about the order.

src/image/VisitInfo.cc
```diff
@@ -247,71 +255,73 @@ class VisitInfoSchema {
...
-                      "observationReason", "reason this observation was taken, or its purpose", "", 0)),
-              object(schema.addField<std::string>("object", "object of interest or field name", "", 0)),
-              hasSimulatedContent(schema.addField<afw::table::Flag>(
-                      "hasSimulatedContent", "Was any part of this observation simulated?")) {}
+    VisitInfoSchema(int _version) : schema() {
+        if (_version < 5) {
+            // exposureId was removed in version 5, but we need to reserve space in the schema for old files.
+            // We don't save the field record, as we don't need to read into it.
+            schema.addField<table::RecordId>("exposureid", "exposure ID", "");
+        }
```
kfindeisen, line 263: This seems... not very portable if there are more schema changes in the future; the logic for exactly which fields belong in which version could get quite twisty. You also have to sacrifice the benefits of one-time initialization to do it this way. And then there's the potential confusion from `VisitInfoSchema` acting like a singleton but not actually being one (a dualton?).

Why not just have separate classes for versions 0-4 and 5+, and make `VisitInfoFactory` responsible for initializing the correct schema? Then you could extend to future incompatibilities without worrying about accidentally changing the old stuff.

tests/test_visitInfo.py
```diff
@@ -344,6 +345,7 @@ def testPersistenceVersions(self):
         self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-2.fits"), 2)
         self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-3.fits"), 3)
         self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-4.fits"), 4)
+        self._testFitsRead(self.data1, os.path.join(dataDir, "visitInfo-version-5.fits"), 5)
```
kfindeisen, line 348: I believe this test should go in the same commit that creates version 5. 🙂

kfindeisen (approved):
Thanks for taking care of this. Only substantial comment is that I think the support for reading old versions is going to be hard to maintain; obviously the commit cleanup will be hard to do if you're still aiming to merge today.

src/image/VisitInfo.cc
```diff
@@ -587,7 +554,7 @@ bool VisitInfo::operator==(VisitInfo const& other) const {
 
 std::size_t VisitInfo::hash_value() const noexcept {
     // Completely arbitrary seed
-    return utils::hashCombine(17, _exposureId, _exposureTime, _darkTime, _date, _ut1, _era, _boresightRaDec,
+    return utils::hashCombine(16, _exposureTime, _darkTime, _date, _ut1, _era, _boresightRaDec,
```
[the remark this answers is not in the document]
parejkoj: Oops! I thought that had something to do with the number of fields. I'll revert.

src/image/VisitInfo.cc
```diff
@@ -380,35 +378,13 @@ namespace detail {
 int stripVisitInfoKeywords(daf::base::PropertySet& metadata) {
     int nstripped = 0;
 
-    std::vector<std::string> keyList = {"EXPID",
```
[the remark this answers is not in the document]
kfindeisen: Or maybe not, if the field is shared with `ExposureInfo`... 😵‍💫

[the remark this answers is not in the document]
parejkoj: I think no, because that key is now handled by ExposureInfo. I think we should only be stripping metadata that is directly used by the given class.

src/image/VisitInfo.cc
parejkoj: I'd tried your latter approach, but ran into C++ compiler problems that I couldn't sort out. This approach was @TallJimbo 's suggest (on top of my initial attempt that forgot about the static members and had the "get which version" in the factory).

Line number no longer accurate, but this was the error (I had made a `VisitInfoSchemaWithExposureId` subclass):
```
src/image/VisitInfo.cc:322:39: error: member initializer 'schema' does not name a non-static data member or base class
    VisitInfoSchemaWithExposureId() : schema() {
```

kfindeisen: 🤦 Right, I've been working in Python for too long. But this version's also potentially unsafe re: its API (you're fine this time because you don't care about the expId key in either case), so maybe file a ticket to clean up the implementation later?

parejkoj: I'm game to try another approach here, if you have a suggestion (or a way around the error above, which I think is due to static-ness of `schema` shared between the parent and child classes). I think CRTP would be one way around it.

kfindeisen: I don't understand why `schema` would be `static` (it's not so in the current or previous versions?), but I don't think this is a problem we can solve in <1 hour. The current code at least gives the desired behavior for now.

parejkoj: https://jira.lsstcorp.org/browse/DM-39906

PR #705, opened 2023-07-28
DM-40032: Issue warning from caller code

timj:
Otherwise the warning comes from inside afw and the caller doesn't really know which line was causing the problem.

### DIFF
--- a/python/lsst/afw/image/_exposureSummaryStats.py
+++ b/python/lsst/afw/image/_exposureSummaryStats.py
@@ -155,7 +155,8 @@ class ExposureSummaryStats(Storable):
         if len(droppedFields) > 0:
             droppedFieldString = ", ".join([str(f) for f in droppedFields])
             warnings.warn(
-                f"Could not read summary fields [{droppedFieldString}]. Please use a newer stack.",
+                f"Summary fields [{droppedFieldString}] not recognized by this software version;"
+                " ignoring them.",
                 FutureWarning,
                 stacklevel=2,
             )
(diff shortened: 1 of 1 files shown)


python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -155,11 +155,9 @@ def _read(bytes):
         if len(droppedFields) > 0:
             droppedFieldString = ", ".join([str(f) for f in droppedFields])
             warnings.warn(
-                (
-                    f"Could not read summary fields [{droppedFieldString}]. "
-                    "Please use a newer stack."
-                ),
+                f"Could not read summary fields [{droppedFieldString}]. Please use a newer stack.",
```
timj, line 158: Thoughts on how to reword this warning would be welcomed. We get it in obs_base tests because we are reading an old file, so the version of the stack is not all that relevant.

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -155,11 +155,9 @@ def _read(bytes):
         if len(droppedFields) > 0:
             droppedFieldString = ", ".join([str(f) for f in droppedFields])
             warnings.warn(
```
timj, line 157: Should this be a `log.warning` instead?

TallJimbo: This one does feel more like a `warnings.warning` to me, just because it's something about backwards compatibility.

python/lsst/afw/image/_exposureSummaryStats.py
timj: In obs_base we should fix the file. Reading it in and writing it out again does fix the warning.

TallJimbo: Ah, if it's not trying to test backwards compatibility I agree.

TallJimbo (approved):

PR #708, opened 2023-09-11
DM-40739: Add column forwarding in exposure summary stats to allow decl to be renamed to dec on read.

erykoff:
This removes warnings from reading older calexps.

### DIFF
--- a/tests/test_exposure.py
+++ b/tests/test_exposure.py
@@ -26,6 +26,7 @@ Test lsst.afw.image.Exposure
 import dataclasses
 import os.path
 import unittest
+import warnings
 
 import numpy as np
 from numpy.testing import assert_allclose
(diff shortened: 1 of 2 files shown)


tests/test_exposure.py
```diff
@@ -1158,6 +1159,30 @@ def testExposureSummaryExtraComponents(self):
...
         self.assertEqual(summaryStats.dec, testDict['dec'])
 
+    def testExposureSummaryForwardComponents(self):
+        """Test that we can forward extra components (e.g. decl->dec).
+        """
+        testDict = {'ra': 10.0,
+                    'decl': 10.0}
+        bytes = yaml.dump(testDict, encoding='utf-8')
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
```
ktlim, line 1169: Maybe add a comment here that forwarding fields must _not_ result in a warning.

ktlim (approved):
Maybe add a comment in the testthat forwarding

PR #709, opened 2023-09-13
DM-40451: Update MultibandExposure.computePsfImage

fred3m:
This updates the `MultibandExposure.computePsf<Kernel>Image` methods to return a `MultibandImage`. It also adds the partial PSF to the `IncompleteDataError` so that users who try...except the error can obtain the PSF in all of the bands that were able to construct a PSF image.


cmsaunders (approved):
Just a few small comments, though I will also have some comments tied to this in the `meas_extensions_scarlet` review.

PR #711, opened 2023-10-02
DM-41012: Check for empty spanset prior to trying to transform it

erykoff:


src/geom/SpanSet.cc
```diff
@@ -848,6 +848,11 @@ std::shared_ptr<SpanSet> SpanSet::transformedBy(lsst::geom::AffineTransform cons
 }
 
 std::shared_ptr<SpanSet> SpanSet::transformedBy(TransformPoint2ToPoint2 const& t) const {
+    // Check if it's an empty bounding box
```
TallJimbo, line 851:
```suggestion
    // If the SpanSet is empty, its bounding box corners are not meaningful and cannot be transformed.
```
(but line-wrapped)

TallJimbo: As an aside, I think we should make `Box2I.getCorners` raise an exception if the box is empty.  Not suggesting it go on this ticket.

TallJimbo (approved):

PR #712, opened 2023-10-18
DM-30535-v24: Update multi-band APIs and conversion to arrays 

mwittgen:


fred3m (commented):
Everything looks fine except that this only includes a subset of the changes required for DM-40451 to work properly. I'm not sure about the current back-porting procedure, but I worry that only backporting one ticket at a time will lead to breakages along the way. Wouldn't it be easier to have a single ticket to backport the entire chain of tickets required to implement DM-40451 (and the bug fixes that followed)?

mwittgen:
I back ported all tickets accumulatively, stacked MRs, and did run the compounded in CI

PR #718, opened 2023-10-27
DM-41345: handle non-finite positions when calculating covariance

parejkoj:


src/table/wcsUtils.cc
```diff
@@ -91,35 +92,35 @@ void updateRefCentroids(geom::SkyWcs const &wcs, ReferenceCollection &refList) {
...
-    
     auto localGnomonicWcs = geom::makeSkyWcs(center, skyCenter, cdMatrix);
-    auto measurementToLocalGnomonic = wcs.getTransform()->then(
-        *localGnomonicWcs->getTransform()->inverted()
-    );
-    
+    auto measurementToLocalGnomonic = wcs.getTransform()->then(*localGnomonicWcs->getTransform()->inverted());
+
     Eigen::Matrix2d localMatrix = measurementToLocalGnomonic->getJacobian(center);
-    Eigen::Matrix2f d = localMatrix.cast < float > () * scale * (lsst::geom::PI / 180.0);
```
cmsaunders, line 110: I don't know how those spaces got merged! Oops.

src/table/wcsUtils.cc
```diff
@@ -91,35 +92,35 @@ void updateRefCentroids(geom::SkyWcs const &wcs, ReferenceCollection &refList) {
     }
 }
 
-Eigen::Matrix2f calculateCoordCovariance(geom::SkyWcs const& wcs, lsst::geom::Point2D center, Eigen::Matrix2f err) {
+Eigen::Matrix2f calculateCoordCovariance(geom::SkyWcs const &wcs, lsst::geom::Point2D center,
+                                         Eigen::Matrix2f err) {
+    if (!std::isfinite(center.getX()) || !std::isfinite(center.getY())) {
```
cmsaunders, line 97: I think this would be the end result anyway, but I agree it's better to just skip going forward in this situation.

cmsaunders (approved):
Looks good to me. Thanks for cleaning up the rest of the function--I obviously forgot to run anything to check formatting on DM-15180!

src/table/wcsUtils.cc
parejkoj: No, if the center is non-finite, the computed `skyCenter` will be NaN and the call to `geom::makeSkyWcs` will raise because it's not a valid WCS.

PR #719, opened 2023-12-05
DM-42012: Add a unittest for MaskedImageFitsReader

arunkannawadi:


timj (approved):
Looks okay to me. The obs_base test is the important one for butler.

PR #721, opened 2024-01-02
DM-40043: Remove value_type definition from CatalogT to work with latest pybind11

mwittgen:
…aut detrction

### DIFF
--- a/include/lsst/afw/table/Catalog.h
+++ b/include/lsst/afw/table/Catalog.h
@@ -103,7 +103,6 @@ public:
     using Table = typename Record::Table;
     using ColumnView = typename Record::ColumnView;
 
-    using value_type = RecordT;
     using reference = RecordT &;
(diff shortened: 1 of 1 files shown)


parejkoj (approved):
Looks fine. My only question is if there are any possible side effects of removing this. Looking at the docs for `value_type` doesn't really clarify that for me.

mwittgen:
A possible side effect would be to use the class with STL containers,
algorithms or nested containers. Defining `value_type` created a
side affect with pybind's new nested container check.
There are more complex ways to tell pybind11 not to check the
`value_type` traits recursively.
`CatalogT` does not implement a conforming STL-like container class in the first place, so I think
removing the `value_type` definition should be fine.

parejkoj:
Oh, please fix the commit message: it ends with a typo: "traut detrction". Preferably, keep the first message under 50 characters, too.

PR #723, opened 2024-01-09
DM-6391: Add centroids plotting for catalogs

parejkoj:


python/lsst/afw/display/interface.py
```diff
@@ -644,6 +644,26 @@ def erase(self):
         """
         self._impl._erase()
 
+    def centroids(self, catalog, symbol=None, *args, **kwargs):
```
PaulPrice, line 647: Why can't you do `symbol="o"`? It's not modifiable, like a list, right?

python/lsst/afw/display/interface.py
```diff
@@ -644,6 +644,26 @@ def erase(self):
...
+        """Draw the sources from a catalog at their pixel centroid positions
+        as given by ``getX()`` and ``getY()``.
+
+        See `dot` for explanation of ``symb`` and available kwargs.
```
PaulPrice, line 651: It's spelled `symbol` in the function def.
Make it more explicit that `args, kwargs` are passed to `dot`.

PaulPrice (approved):
Wonderful, thanks!

python/lsst/afw/display/interface.py
parejkoj: Because I was dumb. Good call!

PR #724, opened 2024-01-12
DM-42445: afw display cleanups and new image() method

parejkoj:


python/lsst/afw/display/interface.py
```diff
@@ -635,12 +632,12 @@ def Buffering(self):
         return self._Buffering(self._impl)
 
     def flush(self):
-        """Flush the buffers
+        """Flush the buffers.
```
gpdf, line 635: "Flush any buffering that may be provided by the backend.", maybe?

python/lsst/afw/display/interface.py
```diff
@@ -816,26 +805,27 @@ def zoom(self, zoomfac=None, colc=None, rowc=None, origin=afwImage.PARENT):
...
         See also
         --------
         Display.zoom
         """
-
         self.zoom(None, colc, rowc, origin)
 
     def interact(self):
-        """Enter an interactive loop, listening for key presses in display and firing callbacks.
+        """Enter an interactive loop, listening for key presses in display and
```
gpdf, line 824: "listening for key presses or equivalent UI actions in the display"

python/lsst/afw/display/interface.py
```diff
@@ -816,26 +805,27 @@ def zoom(self, zoomfac=None, colc=None, rowc=None, origin=afwImage.PARENT):
...
+        firing callbacks.
 
-        Exit with ``q``, ``CR``, ``ESC``, or any other callback function that returns a `True` value.
+        Exit with ``q``, ``CR``, ``ESC``, or any other callback function that
```
gpdf, line 827: Exit with ``q``, ``CR``, ``ESC``, or any equivalent UI action provided in the display.  The loop may  also be exited by returning `True` from a user-provided callback function.

python/lsst/afw/display/interface.py
```diff
@@ -855,15 +845,16 @@ def interact(self):
                         "Display._callbacks['{0}']({0},{1},{2}) failed: {3}".format(k, x, y, e))
 
     def setCallback(self, k, func=None, noRaise=False):
-        """Set the callback for a key
+        """Set the callback for a key.
```
gpdf, line 848: Add "Backend displays may provide an equivalent graphical UI action, but must make the associated key letter visible in the UI in some way."

gpdf (changes requested):
I made a few comments about docstrings.  This looks like a good first step to modernizing and strengthening the interface.

gpdf (approved):
Looks great now.

PR #725, opened 2024-01-17
DM-42467: add properties for ExposureCatalog objects

parejkoj:


python/lsst/afw/table/_exposure.cc
```diff
@@ -68,9 +68,13 @@ PyExposureRecord declareExposureRecord(WrapperCollection &wrappers) {
     return wrappers.wrapType(PyExposureRecord(wrappers.module, "ExposureRecord"), [](auto &mod, auto &cls) {
         cls.def("getId", &ExposureRecord::getId);
         cls.def("setId", &ExposureRecord::setId, "id"_a);
+        // Don't allow id to be set with property, to avoid accidental id changes.
+        cls.def_property_readonly("id", &ExposureRecord::getId);
```
TallJimbo, line 72: I don't actually think a property setter is any more dangerous than `setId`.

TallJimbo (approved):

PR #726, opened 2024-01-27
DM-42080: Add effTime to computeExposureSummaryStats

kadrlica:


python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -127,6 +127,20 @@ class ExposureSummaryStats(Storable):
     (pixels).
     """
 
+    effTime: float = float('nan')
+    """Effective exposure time calculated from psfSigma, skyBg, and
+    zeroPoint (seconds).
+    """
+
+    effTimeScalePsfSigma: float = float('nan')
```
erykoff, line 135: I think maybe this should be `effTimePsfSigmaScale` (and similarly below).

python/lsst/afw/image/_exposureSummaryStats.py
```diff
@@ -285,6 +299,27 @@ def update_schema(cls, schema: Schema) -> None:
             type="F",
             doc="Maximum distance of an unmasked pixel to its nearest model psf star (pixel).",
         )
+        schema.addField(
+            "effTime",
+            type="F",
+            doc="Effective exposure time calculated from psfSigma, skyBg, and "
+            "zeroPoint (seconds)."
```
erykoff, line 306: It's possible/preferred to add a unit to the schema as well.  (This was an oversight with the original version).

erykoff (changes requested):
Looks good; some minor changes requested.

python/lsst/afw/image/_exposureSummaryStats.py
kadrlica: Ok, I'll add units for all quantities that have them (also keeping units in the docstring).

erykoff (approved):

PR #730, opened 2024-03-20
DM-42920: Add evaluateArray interface to ChebyshevBoundedField for compatibility.

erykoff:


python/lsst/afw/math/_chebyshevBoundedField.py
```diff
@@ -72,3 +72,29 @@ def approximate(cls, boundedField,
...
+        x : `np.ndarray`
+            Array of x values.
+        y : `np.ndarray`
+            Array of y values.
+
+        Returns
+        -------
+        values : `np.ndarray`
+            Array of evaluated values.
+        """
```
TallJimbo, line 93: This should go on the base `BoundedField` class instead, I think.  That will make it work on any bounded field (including chebyshev), while still letting `CoaddBoundedField` override it.

TallJimbo (approved):
I think this should move to the base class, but that should be trivial.

erykoff:
Upon further review, I can overload the evaluate in the coadd bounded field by redefining the pybind11 call there.  So this PR is not necessary.

PR #741, opened 2024-07-30
DM-45486: Use cpputils rather than deprecated utils in C++ code

timj:


mwittgen (approved):
Browsed through the changes. Looks good.

python/lsst/afw/cameraGeom/_camera.cc
```diff
@@ -22,11 +22,11 @@
  */
 
 #include "pybind11/pybind11.h"
-#include <lsst/utils/python.h>
+#include <lsst/cpputils/python.h>
```
timj, line 25: @mwittgen I'm guessing this line should be removed because it's repeated a couple of lines below but without the `<>`.

PR #743, opened 2024-09-05
DM-44352: Prevent overwriting existing detector mapping

cmsaunders:
The existing afw code assumes a simple model for the pixels->focal plane mapping and automatically sets that for each detector. This change adds a check to see whether these mappings have already been defined, and if so, skips applying the default simple model.


PFLeget (approved):
Sounds good to me, but does that need to be covered with some unit test?

cmsaunders:
Yes, that's always a good idea. I added a test that builds a camera with a custom detector transform and checks that it is not the same as the one that gets added by default, which uses the detector orientation and pixel size.

PR #744, opened 2024-09-13
DM-46331: Fix test amplifier prescan geometry

czwa:
Fix amplifier geometry:

1. Fix inconsistent column ordering so future work is easier to do.
2. Fix prescan bboxes, by making them have vertical extent matching the raw_data bbox, and located directly next to the raw_data bbox at smaller x-position (on the opposite side of the data region from the horizontal overscan bbox).
3. Manually tweak the two-amplifier detectors so that the prescan is at larger x-position than the data region to again ensure that the prescan is opposite the overscan box.


Alex-Broughton (approved):
Looks good to me; all bboxes/readoutcorners seem reasonable now

PR #745, opened 2024-09-26
DM-45808: Fix pointer bug in SourceCatalog alias map

parejkoj:


TallJimbo (approved):
Looks good.  It's in much better shape now than I thought it would be when we put in our first attempt at a fix.

I'm ambivalent about whether hooking up the alias map to SimpleTable and ExposureTable is good or bad; it seems a little weird to do it for those two but not `PeakTable`, but it really doesn't matter either way.

parejkoj:
Well, if nothing else, `PeakTable.make` is quite a bit more complicated than the others. Ambivalence sounds like approval to me!

PR #748, opened 2024-10-16
DM-46720: Add uncalibrateImage method to afw.image.PhotoCalib

parejkoj:


include/lsst/afw/image/PhotoCalib.h
```diff
@@ -350,6 +350,20 @@ class PhotoCalib final : public table::io::PersistableFacade<PhotoCalib>, public
...
                                       bool includeScaleUncertainty = true) const;
 
+    /**
+     * Return a un-calibrated image, with pixel values in ADU (or whatever the original input to
+     * this photoCalib was).
+     *
+     * Mask pixels are propagated directly from the input image.
+     *
+     * @param maskedImage The masked image with pixel units of nJy to uncalibrate.
+     * @param includeScaleUncertainty Include the uncertainty on the calibration in the resulting variance?
```
TallJimbo, line 360: I think what this really does (or ought to do) is more along the lines of "Remove uncertainty from this calibration that was previously propagated into the variance plane.  Should have the same value as the parameter of the same name when `calibrateImage` was called."

src/image/PhotoCalib.cc
```diff
@@ -94,6 +96,31 @@ void toNanojanskyVariance(ndarray::Array<float const, 2, 1> const &instFlux,
...
+ */
+void fromNanojanskyVariance(ndarray::Array<float const, 2, 1> const &flux,
+                            ndarray::Array<float const, 2, 1> const &fluxVar, float scaleErr,
+                            ndarray::Array<float const, 2, 1> const &instFlux,
+                            ndarray::Array<float, 2, 1> out) {
+    auto eigenFlux = ndarray::asEigen<Eigen::ArrayXpr>(flux);
+    auto eigenFluxVar = ndarray::asEigen<Eigen::ArrayXpr>(fluxVar);
+    auto eigenInstFlux = ndarray::asEigen<Eigen::ArrayXpr>(instFlux);
+    auto eigenOut = ndarray::asEigen<Eigen::ArrayXpr>(out);
+    eigenOut = (eigenFluxVar / eigenFlux.square() - (scaleErr / eigenFlux * eigenInstFlux).square()) *
```
TallJimbo, line 120:
```suggestion
    eigenOut = (eigenFluxVar / eigenFlux.square() - (scaleErr / (eigenFlux * eigenInstFlux).square())) *
```
I don't remember the C++ precedence rules well enough to know if what you had was already correct, but this ensures it.

tests/test_photoCalib.py
```diff
@@ -594,11 +594,17 @@ def testCalibrateImageConstant(self):
         photoCalib = lsst.afw.image.PhotoCalib(self.calibration, self.calibrationErr)
         result = photoCalib.calibrateImage(maskedImage)
         self.assertMaskedImagesAlmostEqual(expect, result)
+        uncalibrated = photoCalib.uncalibrateImage(result)
+        # High rtol because our specific choice of calib values results in a
+        # very-close-to-but-not-1 value, which results in poor round-tripping.
+        self.assertMaskedImagesAlmostEqual(maskedImage, uncalibrated, rtol=4e-2)
```
TallJimbo, line 600: I'm hoping those parentheses in the calculation were important and you can shrink this after you add them, because that is indeed an extremely high tolerance.

TallJimbo (approved):

src/image/PhotoCalib.cc
parejkoj: Uh oh: the above is the exact expression used in the `ADU->nJy` version:

```
    eigenOut = eigenFlux.square() *
               (eigenInstFluxVar / eigenInstFlux.square() + (scaleErr / eigenFlux * eigenInstFlux).square());
```

Have we been doing that incorrectly the whole time? Because your change here does switch the precedence. I've never been sure whether that calibration error is a variance or not. In jointcal just copied it over from PhotoCalTask: https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/photoCal.py#L418

TallJimbo: Oh, wait, I read that incorrectly.  Your original version was correct (but not obviously so), and I was trying to write it as
```
    eigenOut = (eigenFluxVar / eigenFlux.square() - (scaleErr / (eigenFlux / eigenInstFlux).square())) *
```
(note the divide rather than multiply inside the new parentheses).

So the formula you had looks correct (though maybe not obviously so), and we've got a strange test tolerance to explain.

TallJimbo: Oh, and I apparently put the closing parentheses in the wrong place in my original suggestion to (now fixed in the post above).  Here's a fully corrected suggestion:
```suggestion
    eigenOut = (eigenFluxVar / eigenFlux.square() - (scaleErr / (eigenFlux / eigenInstFlux)).square()) *
```

parejkoj: Hmmm... I believe the original expression evaluates to `(scaleErr / eigenFlux) * eigenInstFlux`, because `/` and `*` have the same precedence, and are evaluated left-to-right, not the version that you wrote above.

parejkoj: But unfortunately, changing to make that `flux/instFlux` division explicit didn't fix the numerical problem.

TallJimbo: The way I think about it - which may be worth writing out more explicitly via another temporary variable for `scale` - is
```
scale = flux / instFlux
instFluxVar = instFlux**2 * (fluxVar / flux**2) - (scaleVar / scale**2)
```

TallJimbo: Would it be possible to modify the numbers going into the test make sure we're not adding/subtracting a big number to/from a small number here?

parejkoj: As we discussed on slack, I think we both now agree that the `PhotoCalib.calibrationErr` is supposed to be `sigma^2` (that's what comes out of `PhotoCalTask`), although my docstring is not really specific enough, sadly.

http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/classlsst_1_1afw_1_1image_1_1_photo_calib.html#adb7c4cc78845e8d64666b3751cf5a51b

TallJimbo (approved):

TallJimbo:
Looks good!

PR #751, opened 2024-11-02
DM-47344: Update plotting in cameraGeom to add integer detector ID label

laurenam:


python/lsst/afw/cameraGeom/utils.py
```diff
@@ -1022,7 +1022,8 @@ def showCamera(camera, imageSource=FakeImageDataSource(), imageFactory=afwImage.
                     displayUtils.drawBBox(
                         bbox, borderWidth=0.5, ctype=ctype, display=display)
                     dims = bbox.getDimensions()
-                    display.dot(ccd.getName(), bbox.getMinX() + dims.getX()/2, bbox.getMinY() + dims.getY()/2,
+                    ccdLabel = "{}\n[{}]".format(ccd.getName(), ccd.getId())
```
suberlak, line 1025: I guess it doesn't matter much, but was it on purpose that `showCcd` has `{} [{}]` and  `showCamera` has `{}\n[{}]` ( the detId  below rather than next to detName )?

suberlak (approved):
Thanks for the update!

python/lsst/afw/cameraGeom/utils.py
laurenam: Indeed, it was.  The one above is a plot "title", so just keeping it in one line seems preferable.  For the focal plane labels, we don't have a lot of horizontal real estate, hence the newline.

PR #752, opened 2024-11-08
Set the truthiness value of BackgroundList

arunkannawadi:
BackgroundList is expected to behave like a Python list. It is therefore natural to expect initializing an instance with no arguments to evaluate to False.

### DIFF
--- a/tests/test_background.py
+++ b/tests/test_background.py
@@ -673,6 +673,8 @@ class BackgroundTestCase(lsst.utils.tests.TestCase):
         approxWeighting = False
 
         backgroundList = afwMath.BackgroundList()
+        # An empty BackgroundList should eval to False, like an empty list.
+        self.assertFalse(backgroundList)
 
         for i in range(2):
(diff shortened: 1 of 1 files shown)


python/lsst/afw/math/_backgroundList.py
```diff
@@ -79,6 +79,9 @@ def __getitem__(self, *args):
     def __len__(self, *args):
         return self._backgrounds.__len__(*args)
 
+    def __bool__(self):
+        return bool(self._backgrounds)
+
```
TallJimbo, line 84: I'm actually a little surprised this is necessary - I thought defining `__len__` would do it.  But I see you have a test.

TallJimbo (approved):

python/lsst/afw/math/_backgroundList.py
arunkannawadi: You're right. I can skip the `__bool__` but would like to keep the tests.

PR #753, opened 2024-11-08
DM-38120: Add VIGNETTED bit mask

parejkoj:


include/lsst/afw/image/Mask.h
```diff
@@ -72,6 +72,10 @@ using MaskPlaneDict = std::map<std::string, int>;
  * - `INTRP` This pixel has been interpolated over @note should be called `INTERPOLATED`
  * - `SAT` This pixel is saturated and has bloomed @note should be called `SATURATED`
  * - `SUSPECT` This pixel is untrustworthy, and you may wish to discard any Source containing it
+ * - `NO_DATA` This pixel could not have received flux, e.g. fully vignetted, no available flux from coadd
```
TallJimbo, line 75:
```suggestion
 * - `NO_DATA` This pixel received little or no flux, e.g. fully vignetted, no available flux from coadd
```
It's usually useful to start setting this mask plane before the flux level drops all the way to zero.

include/lsst/afw/image/Mask.h
```diff
@@ -72,6 +72,10 @@ using MaskPlaneDict = std::map<std::string, int>;
...
+ * input exposures.
+ * - `VIGNETTED` This pixel is in a vignetted region, and has received less than 75% of the expected
```
TallJimbo, line 77: I do think it'd be better to leave the actual number to ISR configuration.

TallJimbo (approved):

PR #754, opened 2024-11-12
DM-46388: Add STREAK to default mask planes

parejkoj:


include/lsst/afw/image/Mask.h
```diff
@@ -76,6 +76,7 @@ using MaskPlaneDict = std::map<std::string, int>;
  * input exposures.
  * - `VIGNETTED` This pixel is in a vignetted region; exactly how much vignetting is set by the downstream
  * processing code.
+ * - `STREAK` This pixel is part of a long linear feature, possibly a satellite or fast-moving asteroid.
```
mrawls, line 79: Not bad! However, I'd prefer not to explicitly call out "satellite," to avoid people who may be particular about such things. Perhaps "This pixel is part of a long linear and likely non-astrophysical feature."?

mrawls (approved):

include/lsst/afw/image/Mask.h
parejkoj: Since the streak isn't going to be in our released AP images, and it won't matter whether it is in the DRP images (since they're after the 80 hour embargo), I don't think that's a problem? Do you really think we shouldn't mention satellites in our docs at all?

mrawls: Others have bent over backwards to not say "satellite" so I am mostly suggesting we follow that unwritten precedent. It's not a big deal to me either way.

PR #763, opened 2025-01-24
DM-45545: Update afw to use Numpy 2 newbyteorder type

roceb:


tests/test_methods.py
```diff
@@ -302,7 +302,7 @@ def testUnsignedImages(self):
 
         # arrays differ by a maximum of 2
         errMsg1 = imagesDiffer(image0, image1)
-        self.assertIn("maximum absolute error: |0 - 2| = 2 at position (0, 1).", errMsg1)
+        self.assertRegex(errMsg1, r"maximum absolute error: \|0 - 2\| = 2 at position \(.*0.*,.*1\)" )
```
timj, line 305: @roceb now that I look at this patch I think the better fix is to patch the imagesDiffer function itself that generates the error message and put `int` calls into the f-strings.

@TallJimbo the issue here is that numpy changed the way repr works so you get an explicit `np.int64(1)` in the error message rather than `1`.

tests/test_imageIo2.py
```diff
@@ -58,7 +58,7 @@ def doRoundTrip(self, image, compression=None, scaling=None):
             with astropy.io.fits.open(filename) as hduList:
                 hdu = hduList[1 if isCompressed else 0]
                 if hdu.data.dtype.byteorder != '=':
-                    hdu.data = hdu.data.byteswap().newbyteorder()
+                    hdu.data = hdu.data.byteswap().dtype.newbyteorder()
```
timj, line 61: Does this also work with numpy 1.x or do we have to have code that checks numpy version?

roceb: No sorry, this was supposed to be a draft. Working on it but wanted people to see that there was a PR out there.

timj: I fixed it. I followed the instructions in the numpy 2 migration guide.

tests/test_methods.py
timj: I moved the fix to where message creation happens.

timj:
Confirmed that these changes also work on rubin-env 9 so we can merge them before we switch over.

TallJimbo (approved):

PR #768, opened 2025-02-26
DM-49210: Allow FootprintMergeList to take a pre-initialized schema

parejkoj:


src/detection/FootprintMerge.cc
```diff
@@ -227,9 +227,11 @@ void FootprintMergeList::_initialize(afw::table::Schema &sourceSchema,
...
     for (auto const &iter : filterList) {
         KeyTuple &keys = _filterMap[iter];
-        keys.footprint = sourceSchema.addField<afw::table::Flag>(
-                "merge_footprint_" + iter,
-                "Detection footprint overlapped with a detection from filter " + iter);
+        keys.footprint = sourceSchema.addField(
+                table::Field<afw::table::Flag>(
+                        "merge_footprint_" + iter,
+                        "Detection footprint overlapped with a detection from filter " + iter),
+                /*doReplace=*/true);
```
nevencaplar, line 234: why are we adding doReplace=True here which was not there needed before?

parejkoj: This means that if the schema already has these fields (as it will in the new diffim code), it won't raise an exception.

nevencaplar (approved):

PR #773, opened 2025-03-18
Change 'filters' to 'bands' for multiband images and exposures

fred3m:


python/lsst/afw/multiband.py
```diff
@@ -83,10 +83,10 @@ def clone(self, deep=True):
         pass
 
     @property
-    def filters(self):
-        """List of filter names for the single band objects
+    def bands(self):
+        """List of band names for the single band objects
         """
-        return self._filters
```
arunkannawadi, line 89: I'd suggest keeping this for now, but deprecating it while adding `bands` as a new property.

arunkannawadi (approved):
This is a breaking API change but smoothly transitioning this in constructor methods is definitely cumbersome. But let's atleast continue to support accessor methods like `filters` with a proper deprecation warning.

python/lsst/afw/multiband.py
arunkannawadi: This appeared a few rows below I intended. I meant to keep `filters` as a property, not `_filters` attribute.

python/lsst/afw/multiband.py
```diff
@@ -82,6 +83,15 @@ def clone(self, deep=True):
         """
         pass
 
+    @property
+    def filters(self):
+        """List of filter names for the single band objects (deprecated)
+
+        Use `bands` instead.
+        """
+        warnings.warn("Use `bands` instead of `filters`", DeprecationWarning)
```
arunkannawadi, line 92: You could use the same presciption to deprecate the `staticmethod` so it appears in the documentation as well.
https://developer.lsst.io/stack/deprecating-interfaces.html#python-deprecation

fred3m: I initially tried that but unfortunately it doesn't work for properties.

arunkannawadi: Hmm, I tried it locally before commenting this. The order matters - property on top and deprecated below that.

fred3m: Oh... it's because I forgot to import deprecated. :(

PR #777, opened 2025-03-25
DM-49695: Allow hasattr to work for table for dunder properties if non-contiguous

timj:
Other property requests will fail.

### DIFF
--- a/python/lsst/afw/table/_base.py
+++ b/python/lsst/afw/table/_base.py
@@ -339,7 +339,17 @@ class Catalog(metaclass=TemplateMeta):
         try:
             return getattr(self.table, name)
         except AttributeError:
-            return getattr(self.columns, name)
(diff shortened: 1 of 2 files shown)


python/lsst/afw/table/_base.py
```diff
@@ -339,7 +339,17 @@ def __getattr__(self, name):
...
+            # Special case __ properties as they are never going to be column
+            # names.
+            if name.startswith("__"):
+                raise
+        # This can fail if the table is non-contiguous
+        try:
+            attr = getattr(self.columns, name)
+        except Exception as e:
+            e.add_note(f"Error retrieving column attribute '{name}' from {type(self)}")
+            raise
```
timj, line 351: I assume reraising this as an AttributeError would break things?

TallJimbo (approved):

PR #781, opened 2025-04-21
DM-45569: Add access to color value.

PFLeget:


include/lsst/afw/image/Color.h
```diff
@@ -8,11 +8,13 @@
 
 #include <cmath>
 #include <limits>
+#include <string>
 
 namespace lsst {
 namespace afw {
 namespace image {
 
+// TO DO: Change description here.
```
arunkannawadi, line 17: I think we have finalized the design enough to edit the description here.

include/lsst/afw/image/Color.h
```diff
@@ -24,51 +26,56 @@ namespace image {
...
+        if (_indeterminate != other._indeterminate) {
+            return false;
+        }
+        return _color_value == other._color_value && _color_type == other._color_type;
     }
-    bool operator!=(Color const &other) const noexcept { return !operator==(other); }
-    //@}
-
-    /// Return a hash of this object.
-    std::size_t hash_value() const noexcept { return isIndeterminate() ? 42 : std::hash<double>()(_g_r); }
```
arunkannawadi, line 56: Why was the hash removed altogether? It should hash the type and value together.

include/lsst/afw/image/Color.h
```diff
@@ -24,51 +26,56 @@ namespace image {
...
-namespace std {
-template <>
-struct hash<lsst::afw::image::Color> {
-    using argument_type = lsst::afw::image::Color;
-    using result_type = size_t;
-    size_t operator()(argument_type const &obj) const noexcept { return obj.hash_value(); }
-};
-}  // namespace std
-
-#endif
```
arunkannawadi, line 74: This should be restored _if_ hash_value is restored.

python/lsst/afw/image/_color.cc
```diff
@@ -37,25 +38,31 @@ namespace image {
...
+void wrapColor(lsst::cpputils::python::WrapperCollection & wrappers) {
+    PyColor cls(wrappers.module, "Color");
+
+    /* Constructors */
+    cls
+        // default ctor → indeterminate color
+        .def(py::init<>())
+        // fully-specified ctor: both ColorValue and ColorType required
+        .def(py::init<double, std::string>(),
+            "colorValue"_a, "colorType"_a);
```
arunkannawadi, line 50: I didn't know this style was supported by pybind11!

include/lsst/afw/image/Color.h
PFLeget: I don't understand the point of hashing the value here. Was maybe done when the place holder was doing nothing, but did not think it was useful at all.

PR #783, opened 2025-05-09
DM-50735: support code for visit-level background fitting

TallJimbo:


include/lsst/afw/math/ChebyshevBoundedField.h
```diff
@@ -156,6 +156,30 @@ class ChebyshevBoundedField : public table::io::PersistableFacade<ChebyshevBound
...
 
+    /**
+     *  Construct a matrix that can be used to fit or evaluate a Chebyshev
+     *  polynomial at a set of points.
+     *
+     *  @param[in]  bbox     Integer bounding box of the resulting approximation.  All
+     *                       given points must lie within lsst::geom::Box2D(bbox).
+     *  @param[in]  x        Array of x coordinate values.
+     *  @param[in]  y        Array of y coordinate values (same size as `x`).
+     *                       noise, w = 1/sigma.
```
erykoff, line 167: I don't see where `noise` enters into this.

tests/test_chebyshevBoundedField.py
```diff
@@ -428,6 +428,18 @@ def testArrayFit(self):
...
                     outField2.getCoefficients(), coefficients, rtol=1E-7, atol=1E-7)
+                # If we make a matrix for the same points and use it to fit,
+                # we get coefficients in unspecified order, but there are
+                # (up to round-off error) the same as the nonzero entries of
+                # the 2-d coefficients arrays.
+                matrix = lsst.afw.math.ChebyshevBoundedField.makeFitMatrix(self.bbox, self.xFlat, self.yFlat,
+                                                                           ctrl)
+                coefficientsPacked, _, _, _ = np.linalg.lstsq(matrix, array)
+                coefficientsPacked.sort()
+                coefficientsFlat = coefficients.ravel()
```
erykoff, line 439: I think in this case you probably want `coefficients.flatten()` which makes a copy, while `coefficients.ravel()` makes a view and thus you will be changing them in the next line and while that's fine in the test it's probably not what you want.

erykoff (approved):
A couple of small comments.

include/lsst/afw/math/ChebyshevBoundedField.h
TallJimbo: Accidentally copied an extra line.

PR #785, opened 2025-06-03
DM-50875: Add align/lock method to public display_firefly API

parejkoj:


python/lsst/afw/display/interface.py
```diff
@@ -847,6 +847,28 @@ def interact(self):
...
+    def alignImages(self, match_type="Standard", lock_match=True):
+        """Align and optionally lock the orientation of the images being
+        displayed.
+
+        See the Firefly native docs for additional kwargs reference:
+        https://caltech-ipac.github.io/firefly_client/api/firefly_client.FireflyClient.html#firefly_client.FireflyClient.align_images
+
+        Parameters
+        ----------
+        match_type : `str`, optional
```
timj, line 859: Do you know what the DS9 names would be? Having afw base class point at Firefly docs seems odd. We should be defining what these mean. I have no idea what "target" means for aligning and assumed that WCS and pixel were the obvious ones. (and why isn't it called "wcs" instead of "Standard"?)

PS "prigins" typo.

parejkoj: I only added this to the base class to make for a slightly more useful error message if someone tries to use it from ds9. I'm tempted to remove it from here (there are already some other things only implemented in display_firefly), given that we don't even have effort to support firefly, let alone ds9.

timj: I'm trying to say that the base class docs are the most important. They describe how you expect each plugin to work and so if match type is part of the public interface for all of afwDisplay we need to define what those options mean. If we are saying that match_type values can change for each backend (which seems bad) then we have to say that (and then not specify the firefly ones).

PR #788, opened 2025-06-10
DM-51331: Allow SkyWcs to hold a FITS approximation to itself.

TallJimbo:


include/lsst/afw/geom/SkyWcs.h
```diff
@@ -358,10 +357,20 @@ class SkyWcs final : public table::io::PersistableFacade<SkyWcs>, public typehan
...
-     * This feature is not yet implemented, so hasFitsApproximation is always false.
+     * Returns an empty pointer if and only if hasFitsApproximation() returns false.
      */
-    bool hasFitsApproximation() const { return false; }
+    std::shared_ptr<SkyWcs> getFitsApproximation() const { return _fitsApproximation; }
+
+    /**
+     * Return a copy of this SkyWcs with the given FITS approximation.
+     */
+    std::shared_ptr<SkyWcs> withFitsApproximation(std::shared_ptr<SkyWcs> fitsApproximation) const;
```
cmsaunders, line 373: I assume there's some good reason not to do a `setFitsApproximation`...
Could this be called `copyWithFitsApproximation` or something like that? The name took me a while to process.

cmsaunders (approved):
Looks good. Just a couple minor comments.

include/lsst/afw/geom/SkyWcs.h
TallJimbo: Yes, a setter can't work here because `SkyWcs` is immutable.  But I'm happy to change the name; I think `with*` as a pattern for copy-and-modify factories reads well once you're used to it, but there's already precedent in `SkyWcs` for something starting with `copy*`.

PR #792, opened 2025-06-18
DM-51445: Allow FITS approximations to WCSs that are alreadyt valid FITS.

TallJimbo:
Whether a WCS is already valid FITS is a squirrely definition, because it always involves AST trying to approximate and/or simplify a non-FITS representation and that can get lucky even if the functional form is FITS-hostile.


jeffcarlin (approved):
Looks good!

PR #796, opened 2025-06-20
DM-20746: allow camera geometry to know about parity flip vs. sky

TallJimbo:


TallJimbo (commented):
> I think the logic does not totally work for me here, but let me lay out how the code seems to work, and you can tell me if I'm interpreting things correctly.

Yes, you've understood it all correctly.

> Maybe I'm missing some way it's useful here or in the other PRs, but if not, I think this optional argument should be removed.

Will do.  I left it in as an optimization for careful callers, because I was originally quite worried that computing the parity of the transform from its Jacobian determinant would be slow.  But I benchmarked it and it's totally negligible, so there's really not need for that argument.

The `CameraBuilder` behavior is very much intentional - I didn't want to have to wade deep into the existing camera definitions to work out how to add a parity flip to each of them.  I'll try to clarify this in the docs.

PR #798, opened 2025-07-17
Fix SourceCatalog.asAstropy copy to make all columns writeable

fred3m:
There is a problem that copying a numpy array does not update wheather or not the column is read only. This commit changes the way that copy is implemented by building the astropy table and then copying the table, which is a true deep copy.


fred3m:
No. But it didn't work the other way so it wasn't ever tested (or at least not rigorously).

erykoff (approved):

PR #800, opened 2025-09-04
DM-51668: Add footprint intersection and footprint to numpy from scarlet lite

fred3m:


python/lsst/afw/detection/utils.py
```diff
@@ -49,3 +54,51 @@ def writeFootprintAsDefects(fd, foot):
...
+        width, height = bbox.getDimensions()
+        shape = (height, width)
+        xy0 = (bbox.getMinX(), bbox.getMinY())
+
+    if xy0 is None:
+        offset = (0, 0)
+    else:
+        offset = (-xy0[0], -xy0[1])
+
+    result = np.zeros(shape, dtype=bool)
```
taranu, line 97: `result = np.zeros(shape, dtype=bool if asBool else int)` (or else I don't think this can work correctly if `asBool = False`)

python/lsst/afw/detection/utils.py
```diff
@@ -49,3 +54,51 @@ def writeFootprintAsDefects(fd, foot):
     y0:     %4d                         # Starting row
     height: %4d                         # number of rows
 }""" % (bbox.getMinX(), bbox.getWidth(), bbox.getMinY(), bbox.getHeight()), file=fd)
+
+
+def footprintsToNumpy(
```
taranu, line 59: Needs a unit test.

python/lsst/afw/detection/_footprintContinued.py
```diff
@@ -32,6 +32,41 @@
...
+def getFootprintIntersection(
+    footprint1: Footprint,
+    footprint2: Footprint,
+    copyPeaks: bool = True,
+) -> Footprint:
+    """Calculate the intersection of two Footprints.
+
+    Parameters
+    ----------
+    footprint1:
```
taranu, line 44: We are apparently still not on documenteer >= 2.0 in pipelines docs and are supposed to be putting type hints here too, sadly.

taranu (changes requested):
See comments.

python/lsst/afw/detection/utils.py
fred3m: I added tests for both the footprint ID and boolean footprints and will rerun Jenkins.

PR #836, opened 2026-07-03
DM-55357: Add support for lsst.images in afw display

timj:
We have to convert CellCoadd and VisitImage to `lsst.images.MaskedImage` and then convert to legacy form before we can display. There is too much code in afw display assuming things like afw masks so we can not retain native form.

### DIFF
--- a/python/lsst/afw/display/_images_compat.py
+++ b/python/lsst/afw/display/_images_compat.py
@@ -1,0 +1,92 @@
+# This file is part of afw.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ["normalize_lsst_images"]
+
+import sys
+import lsst.afw.image
+
+
+def normalize_lsst_images(data, wcs=None):
+    """Convert an `lsst.images` object to its afw equivalent for display.
+
+    Parameters
+    ----------
+    data : `~typing.Any`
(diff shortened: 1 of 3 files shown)


python/lsst/afw/display/_images_compat.py
```diff
@@ -0,0 +1,93 @@
...
+
+    if data.sky_projection is not None:
+        if wcs is not None:
+            raise RuntimeError(
+                "You may not specify a wcs with an lsst.images object that has its own sky_projection"
+            )
+        wcs = data.sky_projection.to_legacy()
+
+    if isinstance(data, images.MaskedImage):
+        data = lsst.afw.image.MaskedImage(
```
timj, line 82: It might be better to request a masked_image component from the `MaskedImage` subclass and then call to_legacy on that rather than reimplementing the legacy function.

python/lsst/afw/display/_images_compat.py
```diff
@@ -0,0 +1,93 @@
...
+
+import sys
+
+
+def normalize_lsst_images(data, wcs=None):
+    """Convert an `lsst.images` object to its afw equivalent for display.
+
+    Parameters
+    ----------
+    data : `object`
```
TallJimbo:
```suggestion
    data
```
Better to have no type than an unhelpful one.

python/lsst/afw/display/_images_compat.py
```diff
@@ -0,0 +1,93 @@
...
+    Only the image and mask planes are ever forwarded to display backends,
+    so PSFs, variance, and other components are never serialized for
+    display.  Mask schemas with more than 31 named planes cannot be
+    represented as an `lsst.afw.image.Mask` and fail conversion.
+    """
+    images = sys.modules.get("lsst.images")
+    if images is None or not isinstance(data, (images.MaskedImage, images.Image, images.Mask)):
+        return data, wcs
+
+    import lsst.afw.image
```
TallJimbo: This seems fine to put at module scope.

python/lsst/afw/display/_images_compat.py
TallJimbo: Yeah, if we don't have methods to do that, we should add then.  I'm not sure what to call them; `as_masked_image()` doesn't quite feel right because it's on an object that formally is a `MaskedImage` already.

TallJimbo (approved):

python/lsst/afw/display/_images_compat.py
timj: Would have been easier if VisitImage and CellCoadd had constructed an internal MaskedImage and proxied the underlying properties through it. At the moment we populate the bits of a MaskedImage by calling the parent method and then hang off extras. Maybe a property `MaskedImage.masked_image` that returns `self` if it is a `MaskedImage`, else makes a `MaskedImage` if it's a subclass. That still leads to issues regarding whether it's a view on the original pixels or a copy. Since you are going away and we are running out of time to discuss it's probably best if we punt for now and I add a comment to the catch-all ticket.

